PR #2, opened 2022-09-14
IA-1512: when importing gpkg in new iaso account, out are check across the instance not the project

olethanh:
See Original PR https://github.com/BLSQ/iaso/pull-private/1030 for comments
A `'NoneType' object has no attribute 'projects'` exception was raised when importing a GPKG in the following situation:

- several OUT already in the system similar (same name, shortname and depth) to the one being imported
- None of those OUT are assigned to the import's project

## Changes

Rewrote the `get_or_create_org_unit_type()` function of the gpkg import so:

- The exception reported by Elie doesn't happen anymore (+ regression test for it)
- The code path is simpler (less risks of similar edge cases), and more Djangoesque (use the stock get_or_create() method)
- Since a (per-account) `get_or_create_org_unit_type()` can be useful in other contexts than just the GPKG import, I moved it to models file, added a proper docstring and some tests.

## Notes

To deploy carefully since I'm still new to the codebase, and while the get_or_create logic now seems simpler and less error prone, I'm not sure how it will deal with the data in production. 

In particular, it will throw an exception if there are already two identical -same name same depth- OUT in a given account. IIUC this shouldn't happen, so in case that's the clean solution would be to take this as an opportunity to do a quick data cleanup before deployment -> tell me if I'm wrong.


niconoe:
Following a comment by @olethanh (that disappeared due to repo move) stating that the production data was probably messy and that therefore the simplifications I made to the OUT creation code were unfortunately not possible, I came back to the basics and fixed the bug in the less invasive way.

iaso/models/org_unit.py
```diff
@@ -15,6 +15,26 @@
...
+    # (maybe synced with other devices).
+    # This function has probably the potential to create more messy data, so I still suggest to refactor CAREFULLY.
+    out = OrgUnitType.objects.filter(projects=project, name=name).first()
+    if not out:
+        count = OrgUnitType.objects.filter(name=name, short_name=name[:4], depth=depth).count()
+        if count == 0:
+            out = OrgUnitType.objects.create(name=name, short_name=name[:4], depth=depth)
+            out.save()
+        elif count > 1:
+            out = OrgUnitType.objects.filter(name=name, short_name=name[:4], depth=depth, projects=project).first()
```
olethanh: I think that's where us the original bug the `projects=project`, if condition of line 22 is not true there is no way this one is true

olethanh (commented):
LGTM but we can't probably remove the useless condition in comment

iaso/models/org_unit.py
niconoe: Just removing the `elif count > 1`, wont work, because `get()` will be called in the else branch when there are multiple matching OUT. We can also adjust that else branch, but then we're back at simplifying and adjusting the logic. That's what I did initially, but I had to revert it because of the messy data in production... It feels a bit like we're stuck in a loop here :/

niconoe:
After another round of discussion with @olethanh we decided to attempt another way to refactor (simplify stuff, but deals with the fact that we sometimes have multiples OUT per account in prod).

I removed the release batch so we have a bit more time for careful review.

niconoe:
@olethanh : I made adjustments, for some reason I can't add you as a reviewer ? Don't hesitate to review anyway :D

PR #3, opened 2022-09-14
Polio 304 make the campaign api multi tenant (account aware)

olethanh:
Make the campaigns list API mulit-tenant by linking the account to the campaign, see https://bluesquare.atlassian.net/browse/POLIO-304?filter=-1

## Changes

- migration: existing campaigns are associated with an account
- at campaign creation, an account id must be provided
- logged in users only see campaigns listed to their accounts
- anonymous users see all campaigns per default, can request a specific account via the `account_id` parameter

## Notes

I think the basics are okay, but the implications to the whole system are not totally clear to me yet (for example: similar endpoints to also change? client code to update?)


olethanh (commented):
@niconoe  Pushed a modification so we use the name to identify the polio account in case there is multiple and not the id. And not to use an environment variable since it's a pain and flaky to set up in prod during the migrations.

Otherwhise there is a problem. we cannot create new campaign from the webfrontend anymore, because the backend expect the account to be passed. We should not expect (nor trust) the client on this but should derive the account from the user.
(I can't set as request change since I reopened  the PR)

niconoe (approved):

PR #4, opened 2022-09-14
CDN support

olethanh:
In a nutshell, if you run with an env-variable `STATIC_URL`, this will assume you have your already built assets hosted at that location.

What this means is that we're using S3 to store uploaded files, but we're using the CDN to serve the (already built) static assets.
    
If you don't supply a `STATIC_URL` it is assumed that if you run with `USE_S3` that the assets will also be found in that bucket.

### DIFF
--- a/hat/settings.py
+++ b/hat/settings.py
@@ -22,6 +22,7 @@ import hashlib
 import html
 import re
 import urllib.parse
+from urllib.parse import urlparse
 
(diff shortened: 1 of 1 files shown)


olethanh:
For me it seems to correct.

@madewulf Can you take a second looks to be sure?
From my understanding this is the configuration we are using for WFP so it's already working there.

PR #8, opened 2022-09-14
POLIO-516: get scope from round(LQAS-IM)

olethanh:
LQA/IM maps were still using campaign scoped maps

## Changes

- Add `get_round_districts` method to `Campaign` model
- Replace `get_all_districts` with `get_round_district` in `LQASStatsViewSet` and `IMStatsViewSet`

## How to test

- create a campaign with different scopes per round
- Go to LQAS
- show campaign
- The maps should display different scopes for each round (it will appear in grey if there's no LQAS data, so you can see it anyway)
- repeat with IM

## Print screen / video

Screenshot 2022-09-13 at 12 03 56

### DIFF
--- a/plugins/polio/api.py
+++ b/plugins/polio/api.py
@@ -644,17 +644,17 @@ class IMStatsViewSet(viewsets.ViewSet):
         if stats_types == "HH,OHH":
             im_request_type = ""
 
-        # cached_response = cache.get(
-        #     "{0}-{1}-IM{2}".format(request.user.id, request.query_params["country_id"], im_request_type),
-        #     version=CACHE_VERSION,
-        # )
+        cached_response = cache.get(
+            "{0}-{1}-IM{2}".format(request.user.id, request.query_params["country_id"], im_request_type),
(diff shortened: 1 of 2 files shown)


plugins/polio/api.py
```diff
@@ -644,17 +644,17 @@ def list(self, request):
...
-            response = json.loads(cached_response)
-            cached_date = make_aware(datetime.utcfromtimestamp(response["cache_creation_date"]))
+        # if not request.user.is_anonymous and cached_response and not no_cache:
+        #     response = json.loads(cached_response)
+        #     cached_date = make_aware(datetime.utcfromtimestamp(response["cache_creation_date"]))
 
-            if latest_campaign_update and cached_date > latest_campaign_update:
-                return JsonResponse(response)
+        #     if latest_campaign_update and cached_date > latest_campaign_update:
+        #         return JsonResponse(response)
```
olethanh: the cache response is commented?

quang-le: Ah geez 🤦‍♂️

PR #9, opened 2022-09-14
Ia 1517 allow to search on the content of submissions in the api

olethanh:
The instance/submission API (list) can now be filtered by passing a `json_content` parameters with some JsonLogic filters, for example:

```
{
  "or": [
     {"==": [{"var": "gender"}, "M"]},
     {"and": [{"==": [{"var": "gender"}, "F"]}, {"<": [{"var": "age"}, 20]}]},
   ],
}
```
if you want to filter for all males, plus females under 20.

## Changes

The JsonLogic syntax was chosen because it's an existing standard, and we've already found good front-end libraries that make use of if such as https://react-querybuilder.js.org/docs/api/export

## How to test

Unit tests should be pretty explanatory, the test functions in `test_instances.py` can be used as inspiration if you want to test manually (by passing `json_content` in the list URL)

## Notes

- The front-end should be adapted to make use of this new feature
- Currently, all data (numbers, ...) is stored as strings in the JSONField. That's something we want to change to be able to take full advantage of this search feature.


olethanh:
looks great as a first step to me, I'm merging it so we can continue on this. Ideally we should add a bit of doc on the url parameters

PR #11, opened 2022-09-14
IA-1539: iterate all perms to find landing page

olethanh:
Some profiles were not correctly redirected when they didn't have permission for the landing page

## Changes

- change `getFirstAllowedUrl` to keep looking for eligible route if the first permission found isn't linked to any routes

## How to test

- create a user without the permission to forms but with the untranslated permission `iaso_beneficiaries` and at least 1 other permission (except the polio ones)
- log in with that user

## Print screen / video

There's no point really, just login and see if you get a the 401 screen

## Notes

The bug happened because of a dangling permission after a migration. Removing that permission is part of another ticket.
Anyway it's good to have a more robust fallback method, because we've had permissions leading to nowhere before.


quang-le:
I don't get the failing test. It passes locally. @codeKameleon @beygorghor , any suggestion is welcome

beygorghor (approved):

PR #12, opened 2022-09-15
WC2-53: list dynamic fields

beygorghor:
Use entity type list field entries to populate columns and fields popup on beneficiaries list page only ONE entity type is selected

- [ ] Did I add translations

## Changes

- adding type selector
- remove unused fields  
- refactor use columns and popup component to use fields from new api
- repair detail infos fields broken due to api change
- allow api to receive multiple entity_type_ids
- add org unit (Health Center) to the hardcoded values in the api

## How to test

Go to entity types page and edit a type with entities.
Add or remove list field items.
Open beneficiaries list page.
Filter those on the types you just edited and check that the columns are correct.
Open map tab and click on a marker to open popup => check that same fields are present .

## Print screen / video

https://user-images.githubusercontent.com/12494624/190381153-1a10bf1b-af1c-4f9a-b53c-4b36cd499534.mov

## Notes

We have some issues with the api, ticket is [here](https://bluesquare.atlassian.net/browse/WC2-59)


beygorghor:
CSV export is broken for now

iaso/api/entity.py
```diff
@@ -142,32 +149,42 @@ def export_entity_as_csv(entities):
...
+                    if h in fields_list and h not in header:
+                        header.append(h)
+        i = 0
+        for h in header:
+            for k, v in res.items():
+                match = False
+                if k == h:
+                    benef_data.append(f"found at {k}")
+                if i < len(header):
+                    i += 1
```
SimonDus, line 185: The csv export doesnt work in this version as I have to finish it. So this part of the code is a WIP.

iaso/api/entity.py
```diff
@@ -56,17 +56,24 @@ class Meta:
...
     submitter = serializers.SerializerMethodField()
+    org_unit = serializers.SerializerMethodField()
 
     def get_attributes(self, entity: Entity):
         if entity.attributes:
             return entity.attributes.as_full_model()
         return None
 
+    def get_org_unit(self, entity: Entity):
+        if entity.attributes.org_unit:
```
niconoe: entity.attributes might be `None` since the model allows it. Not sure if that happens in the real world (I created such an entity using the Django admin), but something to check IMHO (and if the idea is to always have attributes, let's the model/database enforce it).

iaso/tests/test_entities.py
```diff
@@ -381,12 +382,12 @@ def test_export_entity(self):
         self.assertEquals(response.get("Content-Disposition"), "attachment; filename=NEW CLIENT_ENTITY.xlsx")
 
         # export specific entity type as xlsx
-        response = self.client.get(f"/api/entity/?entity_type_id={entity_type.pk}&xlsx=true/")
+        response = self.client.get(f"/api/entity/?entity_type_ids={entity_type.pk}&xlsx=true/")
```
niconoe, line 385: Do we also have tests that make sure it works if multiple values are passed in `entity_type_ids`?

iaso/api/entity.py
```diff
@@ -325,69 +350,76 @@ def list(self, request: Request, *args, **kwargs):
...
-                                            data_list = _v
-                                            is_list = True
-                                    columns_list.append(value_dict)
-                result = {}
-                result["id"] = entity.pk
-                result["uuid"] = entity.uuid
-                result["entity_type_name"] = entity.entity_type.name
-                result["created_at"] = entity.created_at
-                result["updated_at"] = entity.updated_at
+
```
niconoe, line 380: General comment: the `list` method is getting really long and doing many different things. Would it be possible to 1) add a bit more comments (not what we do, but why we do it and describing more the API itself), and maybe split it into more digestable functions?

niconoe (commented):
Tested on my machine and works fine, well done!

From a user experience perspective, I wonder if we shouldn't change the condition to show specific columns, so it depend of the entities returned and not of the selected type in the form (=if a search return all entities of the same type => we know which columns to display, regardless of what's selected in "form type").

I'm thinking for example of the case of users having all entities of the same type: they wouldn't use the form to filter on entity type, but would still expect to see the "correct" columns

hat/assets/js/apps/Iaso/domains/entities/beneficiaries/types/fields.ts
```diff
@@ -6,3 +6,9 @@ export type Field = {
     key?: string;
     type?: FieldType;
 };
+
+export type Column = {
```
quang-le: Since it's exported maybe give it a more specific name to avoid confusion with the `Column` type we use for Tables

hat/assets/js/apps/Iaso/domains/entities/beneficiaries/index.tsx
```diff
@@ -133,36 +156,35 @@ export const Beneficiaries: FunctionComponent<Props> = ({ params }) => {
                         {!isFetching && (
                             <ListMap
                                 locations={
-                                    data?.beneficiary?.map(beneficiary => ({
+                                    data?.result?.map(beneficiary => ({
```
quang-le, line 132: One day we should go through all APIs and send `result`everywhere i.o. "org_units" or "instances" or whatever. One day...

hat/assets/js/apps/Iaso/domains/entities/beneficiaries/index.tsx
```diff
@@ -66,9 +68,9 @@ export const Beneficiaries: FunctionComponent<Props> = ({ params }) => {
...
     const { formatMessage } = useSafeIntl();
     const dispatch = useDispatch();
-    const columns = useColumns();
-    const { url: apiUrl } = useGetBeneficiariesApiParams(params);
 
+    const { url: apiUrl } = useGetBeneficiariesApiParams(params);
+    const { data: types } = useGetBeneficiaryTypesDropdown();
     const { data, isFetching } = useGetBeneficiariesPaginated(params);
     const [tab, setTab] = useState(params.tab ?? 'list');
     // const { mutate: deleteEntity, isLoading: deleting } =
```
quang-le: This can be deleted 😇

hat/assets/js/apps/Iaso/domains/entities/beneficiaries/index.tsx
```diff
@@ -85,10 +87,31 @@ export const Beneficiaries: FunctionComponent<Props> = ({ params }) => {
         };
         dispatch(redirectTo(baseUrl, newParams));
     };
+    const entityTypeIds = useMemo(
+        () => params.entityTypeIds?.split(',') || [],
+        [params.entityTypeIds],
+    );
 
+    const {
```
quang-le, line 85: I don't understand why not using a `select` in the query for this?

hat/assets/js/apps/Iaso/domains/entities/beneficiaries/hooks/requests.ts
```diff
@@ -55,38 +61,23 @@ type GetAPiParams = {
...
+            params.dateFrom &&
+            moment(params.dateFrom, 'DD-MM-YYYY').format(apiDateFormat),
+        dateTo:
+            params.dateTo &&
+            moment(params.dateTo, 'DD-MM-YYYY').format(apiDateFormat),
+        created_by_id: params.submitterId,
+        created_by_team_id: params.submitterTeamId,
+        entity_type_ids: params.entityTypeIds,
     };
-    if (params.search) {
```
quang-le, line 61: 😀

hat/assets/js/apps/Iaso/domains/entities/beneficiaries/config.tsx
```diff
@@ -31,133 +25,129 @@ import MESSAGES from '../messages';
...
+            Cell: (settings): ReactElement => (
+                // TODO: limit to user permissions
+                <section>
+                    <IconButtonComponent
+                        url={`/${baseUrls.entityDetails}/entityId/${settings.row.original.id}`}
+                        icon="remove-red-eye"
+                        tooltipMessage={MESSAGES.see}
+                    />
+                    {/* <DeleteDialog
+                            keyName="entity"
```
quang-le: This can be deleted, no?

iaso/api/entity.py
beygorghor: > Tested on my machine and works fine, well done!
> 
> From a user experience perspective, I wonder if we shouldn't change the condition to show specific columns, so it depend of the entities returned and not of the selected type in the form (=if a search return all entities of the same type => we know which columns to display, regardless of what's selected in "form type").
> 
> I'm thinking for example of the case of users having all entities of the same type: they wouldn't use the form to filter on entity type, but would still expect to see the "correct" columns

Yes but not now, we will have a second round on this by providing the option to change those visible colmumns with the UI

iaso/tests/test_entities.py
beygorghor: nope, my bad, i have added this option

iaso/api/entity.py
beygorghor: > entity.attributes might be `None` since the model allows it. Not sure if that happens in the real world (I created such an entity using the Django admin), but something to check IMHO (and if the idea is to always have attributes, let's the model/database enforce it).

i'll fix that one now

hat/assets/js/apps/Iaso/domains/entities/beneficiaries/components/Filters.tsx
```diff
@@ -24,43 +24,38 @@ import { OrgUnitTreeviewModal } from '../../../orgUnits/components/TreeView/OrgU
...
 };
 
-const Filters: FunctionComponent<Props> = ({ params }) => {
+const Filters: FunctionComponent<Props> = ({ params, types }) => {
+    const getParams = useFiltersParams();
     const classes: Record<string, string> = useStyles();
     const { formatMessage } = useSafeIntl();
     const dispatch = useDispatch();
-    const [filters, setFilters] = useState({
+    const [filters, setFilters] = useState<FilterType>({
```
quang-le, line 51: Someday we should look into how to use `useReducer`, which is apparently better suited for dealing with bigger state objects

hat/assets/js/apps/Iaso/domains/entities/beneficiaries/config.tsx
```diff
@@ -31,133 +25,129 @@ import MESSAGES from '../messages';
...
-                                ?.program ?? '--'}
+                            {getValue(
+                                extraColumn.name,
+                                settings.row.original,
+                                extraColumn.type,
+                            )}
                         </>
                     );
                 },
-            },
```
quang-le, line 84: :+1:

hat/assets/js/apps/Iaso/domains/entities/beneficiaries/index.tsx
```diff
@@ -1,32 +1,33 @@
...
     // @ts-ignore
     commonStyles,
     // @ts-ignore
-    Table,
-    // @ts-ignore
     LoadingSpinner,
     // @ts-ignore
     useSafeIntl,
     // @ts-ignore
     // AddButton as AddButtonComponent,
```
quang-le: can be deleted

hat/assets/js/apps/Iaso/domains/entities/beneficiaries/index.tsx
```diff
@@ -1,32 +1,33 @@
...
+import { TableWithDeepLink } from '../../../components/tables/TableWithDeepLink';
 import TopBar from '../../../components/nav/TopBarComponent';
 import { Filters } from './components/Filters';
 import DownloadButtonsComponent from '../../../components/DownloadButtonsComponent';
 // import { Dialog } from './components/Dialog';
 import {
     useGetBeneficiariesPaginated,
     useGetBeneficiariesApiParams,
+    useGetBeneficiaryTypesDropdown,
     // useDeleteBeneficiary,
```
quang-le: can be deleted

hat/assets/js/apps/Iaso/domains/entities/beneficiaries/index.tsx
```diff
@@ -1,32 +1,33 @@
...
     // useSaveBeneficiary,
 } from './hooks/requests';
+// import { useGetPossibleFields } from '../entityTypes/hooks/useGetPossibleFields';
```
quang-le: can be ...

quang-le (approved):
Very neat. Nothing blocking, but a few suggestions and/or bugs that fall out of scope:

- I think it would be nicer to have the dynamic fields shown before the static fields in the table, as they're more likely to have relevant info
- I agree with @niconoe 's suggestion regarding "shared fields". If the search only returns 1 beneficiary or 1 beneficiary type, it would be nice to display the dynamic fields for its type. I think it's a nice to have that can be spun in another ticket if we decide to implement it
- Another nice to have would be the possibility for the user to order the dynamic fields as they want (vs: auto sorted at the moment).
- The map view gets hard to read when there are a lot (20+) beneficiaries for the same org unit. Not blocking for this PR, but worth considering
Screenshot 2022-09-23 at 15 10 35
- csv download is broken (out of scope): 
Screenshot 2022-09-23 at 15 17 24
- xlsx download ignores search, ie: always download everything

hat/assets/js/apps/Iaso/domains/entities/beneficiaries/index.tsx
beygorghor: because data can be `undefined` before fetching it

hat/assets/js/apps/Iaso/domains/entities/beneficiaries/components/Filters.tsx
beygorghor: yes good point

beygorghor:
> Very neat. Nothing blocking, but a few suggestions and/or bugs that fall out of scope:
> 
> * I think it would be nicer to have the dynamic fields shown before the static fields in the table, as they're more likely to have relevant info
> * I agree with @niconoe 's suggestion regarding "shared fields". If the search only returns 1 beneficiary or 1 beneficiary type, it would be nice to display the dynamic fields for its type. I think it's a nice to have that can be spun in another ticket if we decide to implement it
> * Another nice to have would be the possibility for the user to order the dynamic fields as they want (vs: auto sorted at the moment).
> * The map view gets hard to read when there are a lot (20+) beneficiaries for the same org unit. Not blocking for this PR, but worth considering
>   Screenshot 2022-09-23 at 15 10 35
> * csv download is broken (out of scope):
>   Screenshot 2022-09-23 at 15 17 24
> * xlsx download ignores search, ie: always download everything

like i said, the order and shared fields can come after indeed.

codeKameleon (approved):

PR #13, opened 2022-09-15
IA-1554 Web Remove the devices filter on instances page

olethanh:
Remove both the device and device filters on the instances page, Reasoing is that both query to populate the select box were  slow and taxing on the backend and it doesn't seems it was used.

## Changes

On InstancesFiltersComponent remove the query and the Select component.
I moved the Map limit on the second component otherwhise the page looked unbalenced. see screenshot

## How to test

Open the Instances page and do filter to check everything still works.

## Print screen / video
image


beygorghor (approved):
Yes less reducer ! @olethanh  Can you create a JIRA ticket for this, so we have a trace on the release 😉

quang-le (approved):

quang-le:
There's a conflict with main

olethanh:
Fixed the conflict

PR #15, opened 2022-09-16
POLIO-463: Verification score as a number

beygorghor:
Verification score should be blocked for number data entry only 

- [ ] Did I add translations

## Changes

adding type and input props to adapt TextInput as a number input

## How to test

Edit a campaign , go to risk tab, verification score input should accept only numbers

## Print screen / video

Screenshot 2022-09-16 at 13 31 03

### DIFF
--- a/plugins/polio/js/src/forms/RiskAssessmentForm.js
+++ b/plugins/polio/js/src/forms/RiskAssessmentForm.js
@@ -39,6 +39,10 @@ export const RiskAssessmentForm = () => {
                             name="verification_score"
                             component={TextInput}
                             className={classes.input}
+                            type="number"
+                            inputProps={{
+                                min: 0,
+                            }}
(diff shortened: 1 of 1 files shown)


codeKameleon:
@beygorghor shouldn't we add a limite to 20 or add a error validation if the user put a number above ?

quang-le (changes requested):
I can still input text and save it as such:


https://user-images.githubusercontent.com/38907762/191191055-374cff2c-f360-4089-8772-35322c6d644c.mov

beygorghor:
> I can still input text and save it as such:
> 
>  Screen.Recording.2022-09-19.at.17.33.35.mov

It's a firefox issue in fact, on chrome you can only enter 'e'.

quang-le (approved):

PR #16, opened 2022-09-16
IA-1371: Automatically run mypy using GitHub actions

niconoe:
This is a work in progress for IA-1371 where the idea is to:

- Fix the current mypy errors (one shot)
- Run mypy on GitHub actions, just like we do for Django tests and black, so we detect type errors as soon as we introduce them
- Gradually add annotations when working on the code

**currently, most of the first point is donen, the rest not yet**. My initial idea was to wait until I have totally completed this point, but @olethanh encouraged me to already submit incremental PR to limit the number of conflict since this branch touch many files.

**The execution of Iaso shouldn't be impacted at all.**


mypy.ini
```diff
@@ -0,0 +1,20 @@
+[mypy]
+python_version=3.8
```
olethanh: should be 3.6

niconoe: Indeed. I temporarily switched to 3.8 to test something and forgot to reset this.

PR #19, opened 2022-09-19
WIP Testing

qgerome:

- [ ] Did I add translations

## Changes

Explain the changes that were made. The idea is not to list exhaustively all the changes made (Github already provides a full diff), but to help the reviewers better understand:

## How to test

Explain how to test your PR.

## Print screen / video

## Notes


qgerome:
@cypress

PR #21, opened 2022-09-21
IA-278 consistant datetime field

olethanh:
In the backend previously the field `created_at` and `updated_at` on Instance and OrgUnit where behaving differently that on the other field as we sometime used the date from the mobile and sometime from the server which was confusing.
This fix it by adding two new field prepended by `source_` which represent the date send by the mobile client. We now use the field update_at and created_at like for the other models.

## Changes

- Added field `source_updated_at` and `source_created_at` on Instance
- Added field `source_updated_at` and `source_created_at` on OrgUnit
- SQL migration to fill those field
- Changed logic so that when the Mobile send the created_at and updated_at field we save them in the new fields

## How to test

Check that you can still create Submission from the mobile app and then upload them. Then check in the database that the field correspond to what's intended.
Idem when creating from the Web frontend.

## Print screen / video
N/A

## Notes

For more information and analysis of the root problem and solutions  see the jira ticket https://bluesquare.atlassian.net/browse/IA-278


madewulf:
This looks like a nice improvment to me, but we need to show the new dates in the instance details now.

PR #22, opened 2022-09-21
IA-1279: Edit instance location in admin is not working

niconoe:
An exception ("Z dimension is missing") was raised when editing an instance location in the Django admin (by adding/moving the point on the map).

This PR fixes the issue.

## How to test

Edit an instance location in Django admin, make sure it can save correctly and without any exception.

@ekhalilbsq: as far as I understood the same (or at least a similar) issue might still happen in other places in Iaso. When you encounter one, can you create a JIRA ticket and assign me? thanks!

### DIFF
--- a/iaso/admin.py
+++ b/iaso/admin.py
@@ -1,5 +1,3 @@
-import json
-
 from django.contrib.admin import widgets
 from django.contrib.gis import admin, forms
 from django.db import models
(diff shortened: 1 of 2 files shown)


olethanh:
didn't test it yet but  if we already had an altitude are we not loosing it?

beygorghor (approved):

PR #27, opened 2022-09-23
IA-1562 entity admin: the entity type is now more clearly displayed

niconoe:
When adding/editing an entity in the Django admin site, there is a <select> to allow choosing the entity type. Only the name of the entity type was shown there. It made the thing difficult to use, because it was not possible to know to which account the entity type was linked (it also happens that we have multiple entity types with the same name in different accounts)

It is now fixed, the entries in the <select> look like: "Children under 5 (Account: RDC)"

## How to test

1) Go to the Django admin site
2) Create a new entity (or edit an existing one)
3) Make sure the account name is clearly visible in the list
4) Make sure the admin page still works as intended

## Print screen / video

**Before the change:**

Screenshot_20220922_144913

**After the change:**

Screenshot_20220923_122348

### DIFF
--- a/iaso/admin.py
+++ b/iaso/admin.py
@@ -279,6 +279,14 @@ class SourceVersionAdmin(admin.ModelAdmin):
 
 
 class EntityAdmin(admin.ModelAdmin):
+    def get_form(self, request, obj=None, **kwargs):
+        # In the <select> for the entity type, we also want to indicate the account name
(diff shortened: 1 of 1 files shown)


olethanh:
wouldn't it make more sense to directly set it on the model str?

beygorghor (approved):

PR #30, opened 2022-09-27
Ia 1531 allow meta/deprecated id when flattening xml to json

niconoe:
Logs are polluted with entries such as 

`[Sun Sep 11 11:56:56.745312 2022] [:error] [pid 8699] skipped 1 paths while parsing instance 362140 {'flat_json': {<JSON CONTENT HERE>}, 'skipped_paths': ['meta/deprecatedID']}`

`meta/deprecatedID` should be an allowed path in XML data but were not because of a simple type: `https://github.com/BLSQ/iaso/blob/85f9e9a71dd9704af4e45b29f7cda82e6ffbd8cd/iaso/models/base.py#L930`

This PR solves the issue, and do some light/careful refactoring on this part of the code.

## How to test

Not that easy, check the logs have improved post-deployment on staging?


SimonDus (approved):
Code wise, everything seems ok to me. But I don't really know how to test it locally.

PR #34, opened 2022-09-27
IA-1567 mark Instance.accurracy as not mandatory

olethanh:
so we don't have to set a false value in admin when we edit

- [NA] Are there enough tests
## Changes

Set blank=True on the model. It was already null=True so it doesn't really make any db change, it was probaly in the admi

## How to test
Check you can still modify an instance in the admin


niconoe (approved):
Looks good to me!

PR #43, opened 2022-09-29
IA-1559: Push GPS from OrgUnit fix

hakifran:

- [ ] Did I add translations
- [x] Are there enough tests
## Changes
- The isGpsEqual condition was not well defined, the fix was if(!isGpsEqual) insteade of if(isGpsEqual)
- The payload for latitude, longitude and altitude was not well set so the backend was not getting those parameters, the payload in useEditLocationWithGpsAction is fixed

## How to test
- Find a instance which has location field set
- Make sure the instance location is different to the one in the orgUnit related to it
- And click to the location marker in the icons on the top right of the instance detail page
## Print screen / video
Peek 2022-09-29 18-46


beygorghor (approved):

quang-le:
My mistake when I refactored, thanks for fixing it @hakifran !

PR #47, opened 2022-09-30
WC2-78-Entities-api-crashes-if-an-entity-doesn't-have-an-attribute

SimonDus:

- [x] Are there enough tests
## Changes

Make entity list  handle entities without attributes

## How to test

Create an entity without attributes and access beneficiaries webpage or entity api

### DIFF
--- a/iaso/api/entity.py
+++ b/iaso/api/entity.py
@@ -355,29 +355,44 @@ class EntityViewSet(ModelViewSet):
 
         if entity_type_ids is None or (entity_type_ids is not None and len(entity_type_ids.split(",")) > 1):
             for entity in entities:
-                entity_serialized = EntitySerializer(entity, many=False)
-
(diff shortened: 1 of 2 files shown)


iaso/api/entity.py
```diff
@@ -355,29 +355,44 @@ def list(self, request: Request, *args, **kwargs):
...
-                entity_serialized = EntitySerializer(entity, many=False)
-
-                attributes = entity_serialized.data.get("attributes")
-                file_content = attributes.get("file_content")
+                attributes = entity.attributes
+                attributes_pk = None
+                attributes_ou = None
+                file_content = None
+                if attributes is not None:
+                    file_content = entity.attributes.get_and_save_json_of_xml().get("file_content", None)
```
olethanh, line 363: you can use `entity.attributes.json` directly here.

PR #50, opened 2022-10-03
IA-1574 optimize getting org unit hierarchy (and optimize fetching Instance under an orgunit)

olethanh:
Optimize the DB usages for OrgUnitQueryset.hierarchy. which is used to filter on all the descendant of one or more orgunits. Before this necessitate a separate subquery now it does everything in one via  a subquery.

This PR also optimize the query for getting all the Submissions below an instance and allow querying at arbritrary depth

- [x] Are there enough tests. (there was coverage for this api before)

## Changes

- Updated django-ltree version to the latest one (0.5.4)
-  vendored some code from django 4.   since we can't update the Django version for now (because we are still on python 3.6)
- Rewrote hierarchy to use a subquery
- Modified the query in Instance

## How to test

These are optimizations so everything should works  the same as before. Try to use all the feature using org unit and see if they still works.
Try using the website with an user whose OrgUnit access is limited.

### DIFF
--- a/iaso/api/logs.py
+++ b/iaso/api/logs.py
@@ -18,10 +18,10 @@ from iaso.models import OrgUnit, Instance, Form
 def has_access_to(user: User, obj: Union[OrgUnit, Instance, models.Model]):
     if isinstance(obj, OrgUnit):
         ous = OrgUnit.objects.filter_for_user_and_app_id(user, None)
-        return obj in ous
+        return ous.filter(id=obj.id).exists()
     if isinstance(obj, Instance):
         instances = Instance.objects.filter_for_user(user)
-        return user.has_perm("menupermissions.iaso_submissions") and obj in instances
+        return user.has_perm("menupermissions.iaso_submissions") and instances.filter(id=obj.id).exists()
     if isinstance(obj, Form):
         forms = Form.objects.filter_for_user_and_app_id(user)
         return obj in forms
--- a/iaso/models/base.py
+++ b/iaso/models/base.py
@@ -819,15 +819,18 @@ class InstanceQuerySet(models.QuerySet):
 
     def filter_for_user(self, user):
         profile = user.iaso_profile
+        # Do a relative import to avoid an import loop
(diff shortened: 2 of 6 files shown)


iaso/api/logs.py
```diff
@@ -18,10 +18,10 @@
 def has_access_to(user: User, obj: Union[OrgUnit, Instance, models.Model]):
```
niconoe, line 18: Changes look good at first look, but I think such an important would deserve its own extensive testing (instead of being quickly tested through the API), this issue would have been a good opportunity to add those. Since it's release day, may I suggest that we release like it is, but create another ticket for the test?

iaso/models/base.py
```diff
@@ -819,15 +813,18 @@ def for_org_unit_hierarchy(self, org_unit):
 
     def filter_for_user(self, user):
         profile = user.iaso_profile
+        # Do a relative import to avoid an import loop
         from .org_unit import OrgUnit
```
niconoe, line 817: Good for me this time, but I think if we end up with too many local imports because of loops, we might actually need to restructure (divide more) the models folder.

iaso/models/base.py
```diff
@@ -729,17 +729,11 @@ def for_filters(
             queryset = queryset.filter(org_unit_id=org_unit_id)
 
         if org_unit_parent_id:
```
niconoe, line 731: Is the use of this parameter actually tested? If not, that would deserve to be done sooner than later.

niconoe (approved):
Thanks a lot! I especially like the comments that give good context and explain the "why"!

Approved for the release, but as said in the comment I think it would also have been a great opportunity to add a few more tests to those features, can we plan that in another PR/Ticket?

iaso/models/base.py
olethanh: totally agree

iaso/api/logs.py
olethanh: Son will open the tickets for it

PR #52, opened 2022-10-03
IA-1532 return 409 on fail enketo export

olethanh:
Made the code change base on the ticket description but haven't test if it actually works so I'm opening the PR as to not lost track but leaving as a draft for now. Anyone feel free to continue working on this.

### DIFF
--- a/iaso/api/enketo.py
+++ b/iaso/api/enketo.py
@@ -347,6 +347,7 @@ class EnketoSubmissionAPIView(APIView):
                     return Response(
                         {
                             "result": "error",
+                            "step": "export",
                             "message": error.message,
                             "description": error.descriptions,
                         },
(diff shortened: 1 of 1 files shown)


olethanh:
@mestachs @madewulf  Tested it, it change the error message displayed to the user but I didn't detect any other change in the behaviour
Before:
image

image


A bit beyond the scope of this PR but there is two complementary issue.

- Despite the error message it it still saved in Iaso, just not exported. This may confuse the error.
- There is not link to go back to. For the user return to the initial platform (this problem occure when we use `/enketo/public_create_url/` so from an external website.

I tried changing the Error message displayed in Enketo but it seems there is no way without modifying their source code.
List of possible error message in Enketo according to the status code it receive
image

olethanh:
Actually found a way to show a message in Enketo, will make a separate  jira ticket / PR for it as it a behaviour change and we will want to  discuss which message to send etc...

PR #59, opened 2022-10-05
IA-1564: Orgunit Detail - sort on children tab redirects to infos tab

beygorghor:
Removing useless useEffect

- [ ] Did I add translations
- [ ] Are there enough tests
## Changes
Removed a useEffect causing the problem, not sure that it was very usefull...

## How to test

Open details of an org unit, 
navigate lo links tab, 
change sort on a column, 
go to children tab,
change sort on a column, 
 => tab should still be children, not links or infos

## Print screen / video

https://user-images.githubusercontent.com/12494624/194027444-3425faae-8b84-4962-ba87-8dd308129f15.mov

## Notes

Please test also navigaion between other tabs, i d'ont remember why i put this useEffect

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/orgUnits/details.js
+++ b/hat/assets/js/apps/Iaso/domains/orgUnits/details.js
@@ -364,11 +364,6 @@ const OrgUnitDetail = ({ params, router }) => {
         sourcesSelected,
     ]);
 
-    useEffect(() => {
-        if (tab !== params.tab && params.tab) {
-            setTab(params.tab);
-        }
-    }, [tab, params.tab]);
     return (
         <section className={classes.root}>
             <TopBar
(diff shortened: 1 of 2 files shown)


codeKameleon (approved):

hat/assets/js/apps/Iaso/domains/orgUnits/details.js
```diff
@@ -364,11 +364,6 @@ const OrgUnitDetail = ({ params, router }) => {
         sourcesSelected,
     ]);
 
-    useEffect(() => {
```
quang-le, line 367: What was that useEffect for? It suspiciouslsy like something that was added to squash an older bug

beygorghor: duno ...

PR #69, opened 2022-10-10
Polio 626 polio 627 vaccine tab improvements

quang-le:
There was a wrong translation and a defauklt value missing in the vaccine tab

- [X] Did I add translations

## Changes

- Added a `useEffect` to add default value for doses per vial. 
- corrected french translation of PO number

## How to test

- go to a campaign with a clear vaccine tab. 
- add a vaccine -> the doses per vial field should auto fill
- check the translation for PO number. Should be " Numéro de PO"

## Print screen / video

https://user-images.githubusercontent.com/38907762/194870088-1ed9ed61-d159-42cd-bce6-9d4ece531efc.mov

Screenshot 2022-10-10 at 14 46 00

### DIFF
--- a/plugins/polio/js/src/forms/RoundVaccinesForm.tsx
+++ b/plugins/polio/js/src/forms/RoundVaccinesForm.tsx
@@ -27,6 +27,8 @@ const DEFAULT_WASTAGE_RATIOS = {
     bOPV2: 1.18,
 };
 
+const DEFAULT_DOSES_PER_VIAL = 50;
+
 export const RoundVaccinesForm: FunctionComponent<Props> = ({
     round,
     roundIndex,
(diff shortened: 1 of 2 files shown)


hakifran (commented):
I have one question. When the doses per vial has been saved and want to edit it, when selected vacine is changed the doses per vial still the saved one and not the default. Is it the way it should work? A part from that it seems good to me.
Peek 2022-10-10 16-24

plugins/polio/js/src/forms/RoundVaccinesForm.tsx
```diff
@@ -150,6 +152,34 @@ export const RoundVaccinesForm: FunctionComponent<Props> = ({
         vaccines.length,
     ]);
 
+    // Fill in number of doses with default value when adding vaccine
+    useEffect(() => {
+        vaccines.forEach((vaccine, index) => {
+            const dosesPerVial =
+                rounds[roundIndex].vaccines[index].doses_per_vial;
+            // I'm sorry for this, I really had to null check this value
```
beygorghor, line 160: 😄

beygorghor (approved):

quang-le:
> I have one question. When the doses per vial has been saved and want to edit it, when selected vacine is changed the doses per vial still the saved one and not the default. Is it the way it should work? A part from that it seems good to me. Peek 2022-10-10 16-24 [ Peek 2022-10-10 16-24 ](https://user-images.githubusercontent.com/19631540/194888931-cb9f61ac-5be8-4fe9-9dea-d4340d8c1929.gif) [ ](https://user-images.githubusercontent.com/19631540/194888931-cb9f61ac-5be8-4fe9-9dea-d4340d8c1929.gif)

Nicely spotted I will look into it.

Edit: on second thought, I think it's fine this way, otherwise if the user sets the doses per vial, then realizes they input the wrong vaccine, it will revert and they will have to change it again. So I think it's OK to give priority to the value set by the user

plugins/polio/js/src/forms/RoundVaccinesForm.tsx
quang-le: This is the most meme-able line of js I ever wrote

PR #72, opened 2022-10-17
IA-1590 org units perfs

beygorghor:
We have too much renders on the org unit search page.
To reproduce, go locally on orgunits search page, set Geographical data filter to only points.
Search and try to select a row in the multi select, it’s very slow, on Axel Computer it can take a few seconds !!

- [ ] Did I add translations
- [ ] Are there enough tests

## Changes
- split the map into another component to avoid re-render of the map on each user action
- implement useMemo to avoid rows calculation
- split cell into separate components.

## How to test

Explain how to test your PR.
Reproduce same search


hat/assets/js/apps/Iaso/domains/orgUnits/index.tsx
```diff
@@ -164,34 +146,22 @@ export const OrgUnits: FunctionComponent<Props> = ({ params }) => {
...
-            setSelection(newSelection);
+    const getSearchColor = useCallback(
+        currentSearchIndex => {
+            const currentSearch = searches[currentSearchIndex];
+            let currentColor;
+            if (currentSearch) {
+                currentColor = currentSearch.color;
+            }
+            if (!currentColor) {
+                currentColor = getChipColors(0);
```
olethanh: shouldn't it be getChipColors(currentSearchIndex) instead of 0?

olethanh (approved):

olethanh:
Haven't reviewed the whole code but tested it in the frontend and it seems to works well, except sometimes it lost the source parametter but apparently that's a known bug

codeKameleon:
still need to go through the whole code but works well ! I get huge perf improvements locally

hat/assets/js/apps/Iaso/domains/orgUnits/index.tsx
```diff
@@ -200,7 +170,7 @@ export const OrgUnits: FunctionComponent<Props> = ({ params }) => {
 
     const onSearch = useCallback(
         newParams => {
-            handleTableSelection('reset');
+            // handleTableSelection('reset');
```
codeKameleon, line 173: remove the comment ?

codeKameleon (approved):

PR #82, opened 2022-10-19
POLIO-580: reduce grouped campaigned dialog width

codeKameleon:

- [ ] Did I add translations
- [ ] Are there enough tests
## Changes

- change the value of prop maxWidth from 'md' to 'sm'

## How to test

Explain how to test your PR.

## Print screen / video

Capture d’écran du 2022-10-19 14-42-39


beygorghor (approved):

beygorghor (commented):
We still have white space to the left and the right
Input should use 100% of the width, you can also go for a "xs" not a "sm" value

PR #89, opened 2022-10-20
POLIO-650 fix listing possible state in filter

olethanh:
This fix listing the state in the dropdown filter when there was no list of campaign returned

- [] Are there enough tests.
## Changes

- Add an new endpoint to list possibles states. /polio/workflow/current/
- Use the new endpoint in the front-end

## How to test

General configuration is the same as for the budget workflow feature.
Check if the state are properly listed

## Notes

Filtering on campaign without budget is still broken, this is a known bug.

### DIFF
--- a/plugins/polio/js/src/pages/Budget/hooks/api/useGetBudget.ts
+++ b/plugins/polio/js/src/pages/Budget/hooks/api/useGetBudget.ts
@@ -1,11 +1,15 @@
 /* eslint-disable camelcase */
 import { useMemo } from 'react';
 import { UseQueryResult } from 'react-query';
+import { groupBy } from 'lodash';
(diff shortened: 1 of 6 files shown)


plugins/polio/js/src/pages/Budget/hooks/api/useGetBudget.ts
```diff
@@ -76,3 +80,29 @@ export const useGetBudgetForCampaign = (
         options: { enabled: Boolean(id) },
     });
 };
+
+const getBudgetWorkflow = () => {
+    return getRequest(`/api/polio/workflow/current/`);
```
quang-le, line 85: There's an API that returns the workflow? That's awesome, I've got an idea that needed such an endpoint 😀

quang-le (approved):

PR #91, opened 2022-10-20
show the current user on the top right corner and display menu when cl…

olethanh:
…icking on it

Explain what problem this PR is resolving : Lot of pains

## How to test

No config required. check the video

## Print screen / video

https://user-images.githubusercontent.com/82500/196963304-cab17935-95bb-47bf-b116-53333f9d6e7a.mp4

### DIFF
--- a/hat/assets/js/apps/Iaso/components/nav/TopBarComponent.js
+++ b/hat/assets/js/apps/Iaso/components/nav/TopBarComponent.js
@@ -13,6 +13,7 @@ import PropTypes from 'prop-types';
 
 import { toggleSidebarMenu } from '../../redux/sidebarMenuReducer';
 import { ThemeConfigContext } from '../../domains/app/contexts/ThemeConfigContext.tsx';
+import { useCurrentUser } from '../../utils/usersUtils';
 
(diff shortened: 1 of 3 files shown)


beygorghor (changes requested):
Ok i got the value of it, but for me clicking on the username should not open the side menu.
Also having a double navigation is a bit confusing.
If this is fr dev purpose only => we can merge
If this is for prod, i suggest we find a better solution:
- remove the back arrow and use always the burger menu
- keep both (arrow back and burger menu), it's a bit ugly
- go for double navigation but put the burger menu instead of the username to be more consistent with our users.
I'm curious of what @bmonjoie is thinking about this.

bmonjoie:
I completely agree that having the menu twice is confusing.
I would move the entries from the menu about the user (language settings, logout, etc.) il a popup that opens (on top of the username) when the user clicks on their username and leave the left menu as is.
I also would make the username pop out a little bit more:
* Add an avatar image next to it (even if it cannot be changed)
* Add a rounded border with some padding
* Make the inside of the border a little bit darker
* Make the font size of the name bigger.

PR #93, opened 2022-10-20
Wc2 96: search parameter doesn't work for /api/entity (list/export/details)

niconoe:
Due to a broken import, the "search" paramter/filter of the list /api/entity endpoint triggered a 500 error.

This fixes the issue and add some tests for this paramter. 

- [x] Are there enough tests

## Changes

Just added an import :)

## How to test

Go to an URL such as `/api/entity/?search=entity_name` and make sure:

- no error 500 is triggered
- the results match the search

## Notes

I added tests for the code I've touched, but IMHO this API is currently poorly tested and many more test functions should also be added if we want to ensure some robustness.

### DIFF
--- a/iaso/api/entity.py
+++ b/iaso/api/entity.py
@@ -6,7 +6,7 @@ from typing import List, Any, Union
 
 import xlsxwriter  # type: ignore
 from django.core.paginator import Paginator
-from django.db.models import Max
+from django.db.models import Max, Q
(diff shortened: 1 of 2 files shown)


beygorghor:
Looks good to me, just a remark, you can perform a search on json while type is not selected and json columns are not visible. User can be lost with results ...

Screenshot 2022-10-24 at 10 12 42

Screenshot 2022-10-24 at 10 12 50

beygorghor (approved):

PR #97, opened 2022-10-21
Polio 671 step node deletion

olethanh:
Make the api not crash when it cannot found a node or a step. (When they get  removed or deleted)

- [ ] Are there enough tests

## Changes

When we cannot found the transition or node to show the label we return the key itself as a fallback value

## How to test

Use the budget workflow. Then in the workflow definition  the current state of a campaign  and the last step

### DIFF
--- a/plugins/polio/budget/serializers.py
+++ b/plugins/polio/budget/serializers.py
@@ -251,10 +251,9 @@ class BudgetStepSerializer(serializers.ModelSerializer):
     created_by = UserSerializer()
 
     @swagger_serializer_method(serializer_or_field=serializers.CharField)
-    def get_transition_label(self, budget_step: BudgetStep):
+    def get_transition_label(self, budget_step: BudgetStep) -> str:
         workflow = get_workflow()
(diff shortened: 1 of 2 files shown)


plugins/polio/budget/serializers.py
```diff
@@ -77,11 +77,18 @@ class Meta:
...
         workflow = get_workflow()
-        node = workflow.get_node_by_key(campaign.budget_current_state_key)
-        return {
-            "key": campaign.budget_current_state_key,
-            "label": node.label,
-        }
+        try:
+            node = workflow.get_node_by_key(campaign.budget_current_state_key)
+        except:
+            return {
```
niconoe, line 83: Can we specify which specific exception(s) we expect here? thanks!

niconoe (changes requested):
Thanks @olethanh!

quang-le (approved):
Works on my machine!

plugins/polio/budget/serializers.py
olethanh: Will do it in a later pr. need to strenghten the whole system more in any case

PR #104, opened 2022-10-21
Polio 676 fix preparedness import chart

olethanh:
Some preparedness Spreadsheet were not importing, because some  of their worksheet were Special Graph and it was making the gspread library crash. We now skip these sheets

Also we no ignore hidden WorkSheet

- [x ] Did I use eslint and black formatters

## Changes
Skip worksheet that are not of GRID type 
Skip worksheet hidden one. This won't take effect retroactively for already imported sheet since we didn't save this meta data at the time


SimonDus (approved):
Code Wise, it looks ok.

PR #106, opened 2022-10-25
IA-1579 Use flexible site name in the reset  and invitations emails

olethanh:
Instead of hardcoding Iaso in the e-mail Subject and Body use the name defined in database so we can easily configure it for each environment

- [x] Did I add translations
- [x] Are there enough tests
## Changes

* More Consistantly use the Site settings. See Django Sites' Famework  for more information https://docs.djangoproject.com/en/4.1/ref/contrib/sites/
* Changed the text of the Invitation email and reset-email as well.
* The old budget stuff wasn't touched because we remove it in another PR and we already use the new style in the new budget engine.
* Remove some use of DNS_DOMAIN but not all yet. (There was some inconsistance of usage in it, some part of the code were expecting the protocol other not it seems)
* Updated documentation for Django translation handling.

## How to test

* Make sure the Site framework is configured correctly in the Admin, by going to `/admin/sites/site/1/change/`
* Try sending an invitation e-mail and check the result
* Idem for a Reset password e-mail

By default the content of the e-mail will be displayed in the Django terminal log.

### DIFF
--- a/README.md
+++ b/README.md
@@ -620,6 +620,18 @@ them reflected in the interface:
 
 ```manage.py compilemessages```
 
+You do not need to manage local for English as it is the default language
+
+If you get an error about `/opt/app` or cannot accessing docker:
+Change in settings.py LOCALE_PATHS to
+```python
+LOCALE_PATHS = [ "hat/locale/"]
+```
+
+And specify --ignore
+```sh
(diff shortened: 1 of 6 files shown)


niconoe (approved):
Thanks a lot @olethanh. Code looks nice, and I was able to successfully test in local (mail content looks good).

PR #109, opened 2022-10-25
Polio-620: search districts selected and not selected

hakifran:
On the scope tab there is a list(table) of selected district and the user can't be able to search in the selected districts nor in the unselected(not include) ones

- [x] Did I add translations
- [ ] Are there enough tests
## Changes
-  Add a search input and a filter button
- Add a check box to specify if the search is in all districts(country) or in scope(already selected) only
- Display districts in table according to the result of search
- Add actions(select or unselect both for district and region) to each district in the result
- The map won't change with the search result

## How to test
- Edit campaign
- Scope tab
- Enter a string in the search input and click on filter button
- In the result try to do action like select or unselect

## Print screen / video

Peek 2022-10-31 14-24

@cypress


beygorghor (changes requested):
I'm a bit lost on this.

The checkbox is acting on the result of the search, so the search button should be enabled again if you toggle the checkbox.
Also this is not clear that the checkbox will affect the results.

May i suggest a different approach here:
- move checkbox over the results
- apply changes directly on the results while clicking on this checkbox without having to click on search again

hakifran:
> 

The refactoring is done based on the above comments.thank you

plugins/polio/js/src/components/Inputs/ScopeInput.tsx
```diff
@@ -53,6 +54,10 @@ import {
     unselectedPathOptions,
 } from '../../styles/constants';
 import { useStyles } from '../../styles/theme';
+import { csvPreview } from '../../../../../../hat/assets/js/apps/Iaso/domains/dataSources/requests';
```
quang-le: defined but never used

plugins/polio/js/src/components/Inputs/ScopeInput.tsx
```diff
@@ -215,14 +223,52 @@ export const ScopeInput: FunctionComponent<FieldProps<Scope[], Values>> = ({
...
+                    }
+
+                    return dist;
+                });
+                setFilteredDistricts(newFilteredDistricts);
+            }
+
+            setScopes(newScopes);
+        },
+        [scopes, setScopes, filteredDistricts, setFilteredDistricts],
```
quang-le: `searchValue` is missing from the dependencies array.

plugins/polio/js/src/components/Inputs/ScopeInput.tsx
```diff
@@ -95,6 +100,7 @@ type Values = {
 export const ScopeInput: FunctionComponent<FieldProps<Scope[], Values>> = ({
     field,
     form: { values },
+    searchDistricts,
```
quang-le: `searchDistricts` does not exist in props type

plugins/polio/js/src/components/Inputs/ScopeInput.tsx
```diff
@@ -128,6 +134,8 @@ export const ScopeInput: FunctionComponent<FieldProps<Scope[], Values>> = ({
     const { data: districtShapes, isFetching: isFetchingDistricts } =
         useGetGeoJson(parentCountryId, 'DISTRICT');
 
+    const districts = districtShapes
```
quang-le: missing `;` 

I don't understand the new `const` here. It seems to me you can use districtShapes directly since you don't mutate it anywhere

plugins/polio/js/src/components/Inputs/ScopeInput.tsx
```diff
@@ -215,14 +223,52 @@ export const ScopeInput: FunctionComponent<FieldProps<Scope[], Values>> = ({
         }
         setScopes(newScopes);
     };
+
     const [selectedVaccine, setSelectedVaccine] = useState('mOPV2');
 
+    const addDistrictInVaccineScope = useCallback(
+        (district, selectedVaccine) => {
```
quang-le: `selectedVaccine` is already declared line 227, please change the name to avoid confusion

plugins/polio/js/src/components/Inputs/ScopeInput.tsx
```diff
@@ -234,6 +280,16 @@ export const ScopeInput: FunctionComponent<FieldProps<Scope[], Values>> = ({
             }
             // Remove org unit from selection if it's part of the scope
             if (scope.group.org_units.includes(district.id)) {
+                if (searchValue !== "") {
+                    const newFilteredDistricts = filteredDistricts.map(dist => {
+                        if (dist.id === district.id) {
+                            dist.vaccineName = "";
```
quang-le: You have type errors here

plugins/polio/js/src/components/Inputs/ScopeInput.tsx
```diff
@@ -284,9 +348,46 @@ export const ScopeInput: FunctionComponent<FieldProps<Scope[], Values>> = ({
                     OrgUnitId => !OrgUnitIdsToRemove.includes(OrgUnitId),
                 );
             });
+            if (searchValue !== "") {
+                const newFilteredDistricts = filteredDistricts.map(dict => {
```
quang-le: you have a type error here

plugins/polio/js/src/components/Inputs/ScopeInput.tsx
```diff
@@ -284,9 +348,46 @@ export const ScopeInput: FunctionComponent<FieldProps<Scope[], Values>> = ({
...
+            }
             setScopes(newScopes);
         },
-        [districtShapes, scopes, setScopes],
+        [districtShapes, scopes, setScopes, filteredDistricts, searchValue],
+    );
+
+    // Add all district in the same region as this district
+    const addRegionToTable = useCallback(
+        (shape: ShapeRow, selectedVaccine) => {
```
quang-le: `selectedVaccine` cf previous comment

plugins/polio/js/src/components/Inputs/ScopeInput.tsx
```diff
@@ -284,9 +348,46 @@ export const ScopeInput: FunctionComponent<FieldProps<Scope[], Values>> = ({
...
+                        dict.vaccineName = selectedVaccine;
+                    }
+                    return dict;
+                });
+                setFilteredDistricts(newFilteredDistricts);
+            }
+
+            setScopes(newScopes);
+        },
+        [districtShapes, scopes, setScopes, filteredDistricts, setFilteredDistricts],
```
quang-le: Missing deps in the array. This can lead to nasty bugs

plugins/polio/js/src/components/Inputs/ScopeInput.tsx
```diff
@@ -355,9 +456,51 @@ export const ScopeInput: FunctionComponent<FieldProps<Scope[], Values>> = ({
             </Tooltip>
         );
     };
+    // Launch the searchdistrictByname from parent component
+    useEffect(() => {
+        searchDistricts.current = searchDistrictByName;
+
+    }, [districts, filteredDistricts, searchValue, searchScope])
```
quang-le: missing deps in the array ⚠️

plugins/polio/js/src/components/Inputs/ScopeInput.tsx
```diff
@@ -355,9 +456,51 @@ export const ScopeInput: FunctionComponent<FieldProps<Scope[], Values>> = ({
             </Tooltip>
         );
     };
+    // Launch the searchdistrictByname from parent component
+    useEffect(() => {
+        searchDistricts.current = searchDistrictByName;
```
quang-le: You're mutating a prop directly, which is not recommended. See the [es-lint rule](https://eslint.org/docs/latest/rules/no-param-reassign)

plugins/polio/js/src/components/Inputs/ScopeInput.tsx
```diff
@@ -355,9 +456,51 @@ export const ScopeInput: FunctionComponent<FieldProps<Scope[], Values>> = ({
...
+
+    }, [districts, filteredDistricts, searchValue, searchScope])
+    // research when the checkbox(Search districts in scope) is checked on unchecked
+    useEffect(() => {
+        searchDistrictByName(searchValue);
+    }, [searchScope]);
```
quang-le: missing deps in the array ⚠️

quang-le (changes requested):
I think you have a problem with your linter. I listed the one most likely to cause bugs and unexpected behaviour in my comments. 
Please check your IDE to make sure it enforces the rules defined in the iaso repo

quang-le (changes requested):
I think you have a problem with your linter. I listed the one most likely to cause bugs and unexpected behaviour in my comments. 
Please check your IDE to make sure it enforces the rules defined in the iaso repo

About the feature itself, I would suggest some changes so that the behaviour matches other search fields in the app:
- Replace the `Filter` button with a `Seach`button (like in Org units)
- Disable the Filter/Search button unless the search term is changed
- Enable the search when clearing the field. eg: I searched for "x", now I empty the search field, I should be able the click the button and see all results
- Show/hide out of scope districts, even when no search is active (same behaviour as "show deleted" elsewhere).

hakifran:
> I think you have a problem with your linter. I listed the one most likely to cause bugs and unexpected behaviour in my comments. Please check your IDE to make sure it enforces the rules defined in the iaso repo
> 
> About the feature itself, I would suggest some changes so that the behaviour matches other search fields in the app:
> 
> * Replace the `Filter` button with a `Seach`button (like in Org units)
> * Disable the Filter/Search button unless the search term is changed
> * Enable the search when clearing the field. eg: I searched for "x", now I empty the search field, I should be able the click the button and see all results
> * Show/hide out of scope districts, even when no search is active (same behaviour as "show deleted" elsewhere).

I did the refactoring(moving some code from ScopeInput to ScopeForm

hakifran:
@cypress

plugins/polio/js/src/components/Inputs/ScopeInput.tsx
```diff
@@ -215,14 +251,60 @@ export const ScopeInput: FunctionComponent<FieldProps<Scope[], Values>> = ({
...
+
+    const [selectedVaccine, setSelectedVaccine] = useState<string>('mOPV2');
+
+    const addDistrictInVaccineScope = useCallback(
+        district => {
+            const newScopes: Scope[] = cloneDeep(scopes);
+            let scope: Scope | undefined = newScopes.find(
+                s => s.vaccine === selectedVaccine,
+            );
+            if (scope === undefined) {
```
beygorghor: `!scope`

plugins/polio/js/src/components/Inputs/ScopeInput.tsx
```diff
@@ -215,14 +251,60 @@ export const ScopeInput: FunctionComponent<FieldProps<Scope[], Values>> = ({
...
+            // Add org unit in selection if it's not part of the scope
+            if (!scope.group.org_units.includes(district.id)) {
+                scope.group.org_units.push(district.id);
+            }
+            // when the add is done on searched result and update the filteredDistricts state
+            if (searchLaunched || searchScopeChecked) {
+                const newFilteredDistricts = filteredDistricts.map(dist => {
+                    if (dist.id === district.id) {
+                        // eslint-disable-next-line no-param-reassign
+                        dist.vaccineName = selectedVaccine;
```
beygorghor: `const newFilteredDistricts = [...filteredDistricts];
filteredDistricts.forEach((dist, index) => {
  if (dist.id === district.id) {
    newFilteredDistricts[index].vaccineName = selectedVaccine;
  }
}`

plugins/polio/js/src/components/Inputs/ScopeInput.tsx
```diff
@@ -234,6 +316,17 @@ export const ScopeInput: FunctionComponent<FieldProps<Scope[], Values>> = ({
...
             if (scope.group.org_units.includes(district.id)) {
+                if (searchLaunched || searchScopeChecked) {
+                    const newFilteredDistricts = filteredDistricts.map(dist => {
+                        if (dist.id === district.id) {
+                            // eslint-disable-next-line no-param-reassign
+                            dist.vaccineName = '';
+                        }
+
+                        return dist;
+                    });
```
beygorghor: same here

plugins/polio/js/src/components/Inputs/ScopeInput.tsx
```diff
@@ -284,9 +392,72 @@ export const ScopeInput: FunctionComponent<FieldProps<Scope[], Values>> = ({
                     OrgUnitId => !OrgUnitIdsToRemove.includes(OrgUnitId),
                 );
             });
+
+            if (searchLaunched || searchScopeChecked) {
+                const newFilteredDistricts = filteredDistricts.map(dict => {
+                    if (OrgUnitIdsToRemove.includes(dict.id)) {
+                        // eslint-disable-next-line no-param-reassign
```
beygorghor: same as above, better use another variable

plugins/polio/js/src/components/Inputs/ScopeInput.tsx
```diff
@@ -284,9 +392,72 @@ export const ScopeInput: FunctionComponent<FieldProps<Scope[], Values>> = ({
...
+        (shape: ShapeRow) => {
+            const OrgUnitIdsToAdd = districtShapes
+                .filter(s => s.parent_id === shape.parent_id)
+                .map(s => s.id);
+
+            const newScopes: Scope[] = cloneDeep(scopes);
+            newScopes
+                .filter(sc => sc.vaccine === selectedVaccine)
+                .forEach(scope => {
+                    // eslint-disable-next-line no-param-reassign
```
beygorghor: i'm really not a huge fan of this

plugins/polio/js/src/components/Inputs/ScopeInput.tsx
```diff
@@ -284,9 +392,72 @@ export const ScopeInput: FunctionComponent<FieldProps<Scope[], Values>> = ({
...
+                    scope.group.org_units =
+                        scope.group.org_units.concat(OrgUnitIdsToAdd);
+                });
+
+            if (searchLaunched || searchScopeChecked) {
+                const newFilteredDistricts = filteredDistricts.map(dict => {
+                    if (OrgUnitIdsToAdd.includes(dict.id)) {
+                        // eslint-disable-next-line no-param-reassign
```
beygorghor: ...

plugins/polio/js/src/components/Inputs/ScopeInput.tsx
```diff
@@ -461,6 +637,13 @@ export const ScopeInput: FunctionComponent<FieldProps<Scope[], Values>> = ({
                 )}
             </Grid>
             <Grid xs={5} item>
+                <InputComponent
+                    type="checkbox"
+                    withMarginTop={false}
+                    onChange={() => onChangeSearchScope()}
```
beygorghor: why not directly `onChange={onChangeSearchScopeFunction}`?

hakifran:
@cypress

plugins/polio/js/src/forms/ScopeForm.tsx
```diff
@@ -33,19 +86,179 @@ export const ScopeForm: FunctionComponent = () => {
...
+            <Grid container spacing={4} justify="space-between">
+                <Grid xs={12} md={6} item>
+                    <Field
+                        name="separate_scopes_per_round"
+                        component={BooleanInput}
+                        label={formatMessage(MESSAGES.scope_per_round)}
+                    />
+                </Grid>
+
+                <Grid container spacing={3} xs={12} md={5}>
```
beygorghor: you need to add item in order to use xs prop

plugins/polio/js/src/forms/ScopeForm.tsx
```diff
@@ -33,19 +86,179 @@ export const ScopeForm: FunctionComponent = () => {
...
+    };
+
     return (
         <>
-            <Field
-                name="separate_scopes_per_round"
-                component={BooleanInput}
-                label={formatMessage(MESSAGES.scope_per_round)}
-            />
+            <Grid container spacing={4} justify="space-between">
```
beygorghor: `justify` => `justifyContent`

plugins/polio/js/src/components/Inputs/ScopeInput.tsx
beygorghor: also `keyValue` is required

beygorghor (changes requested):
This is not working for me unfortunately:

Screenshot 2022-11-10 at 15 07 37

hakifran:
@cypress

hakifran:
@cypress

plugins/polio/js/src/components/Inputs/ScopeInput.tsx
```diff
@@ -281,12 +439,81 @@ export const ScopeInput: FunctionComponent<FieldProps<Scope[], Values>> = ({
             newScopes.forEach(scope => {
                 // eslint-disable-next-line no-param-reassign
```
quang-le: Please do not disable this linter rule, it is here for a good reason: [https://eslint.org/docs/latest/rules/no-param-reassign](https://eslint.org/docs/latest/rules/no-param-reassign)

plugins/polio/js/src/components/Inputs/ScopeInput.tsx
```diff
@@ -281,12 +439,81 @@ export const ScopeInput: FunctionComponent<FieldProps<Scope[], Values>> = ({
...
+        (shape: ShapeRow) => {
+            const OrgUnitIdsToAdd = districtShapes
+                .filter(s => s.parent_id === shape.parent_id)
+                .map(s => s.id);
+
+            const newScopes: Scope[] = cloneDeep(scopes);
+            newScopes
+                .filter(sc => sc.vaccine === selectedVaccine)
+                .forEach(scope => {
+                    // eslint-disable-next-line no-param-reassign
```
quang-le: cf supra. It is a blocking issue for this PR

plugins/polio/js/src/components/Inputs/ScopeInput.tsx
```diff
@@ -92,9 +97,35 @@ type Values = {
...
+    vaccineName: string;
+    region: any;
+    id: number;
+    parent_id: number;
+};
+
+type Props = {
+    field;
+    form: FieldProps<Scope[], Values>;
+    values: values;
```
quang-le: `values` isn't a type. The TS linter should tell you this

plugins/polio/js/src/components/Inputs/ScopeInput.tsx
```diff
@@ -38,9 +39,13 @@ import sortBy from 'lodash/sortBy';
 
 import { FieldProps } from 'formik/dist/Field';
 
+import CheckIcon from '@material-ui/icons/Check';
+import SelectAllIcon from '@material-ui/icons/SelectAll';
+import InputComponent from 'Iaso/components/forms/InputComponent';
+import { values } from 'cypress/types/lodash';
```
quang-le: unused import

plugins/polio/js/src/components/Inputs/ScopeInput.tsx
```diff
@@ -38,9 +39,13 @@ import sortBy from 'lodash/sortBy';
 
 import { FieldProps } from 'formik/dist/Field';
```
quang-le: I get a warning for this import  
Screenshot 2022-11-15 at 09 36 58

quang-le (changes requested):
I get a lot of linter errors again, some of which are not trivial and indicative of errors in the code, like this one:

Screenshot 2022-11-15 at 09 16 45

Also, there are still  issues:


https://user-images.githubusercontent.com/38907762/201869856-7bec8e11-a949-47bb-8e18-718c2d5e28d5.mov


https://user-images.githubusercontent.com/38907762/201869906-c0626711-64d9-4c75-95b2-521a0286b346.mov

hakifran:
@beygorghor , here is what I didn't manage to do: 

-  When the scope is loaded with not "Separate scope per round" not selected it works fine and the search works
- But when the "Separate scope per round" is checked I didn't manage to update the scope and therefore filteredDistricts state is not update end the search don't work and the amount of districts displayed in the table is not the same as in the map's legend.
- What I would like is to be able to update scope accordingly to the "Separate scope per round" selected or not.

beygorghor:
@quang-le @hakifran @codeKameleon 
I did a refactor of the code, split scopeInput  into smaller components.
Moved ap to the right.
And review a bit the logic to make it work with scope per round or not.
@hakifran honestly it was a very complex task... 


https://user-images.githubusercontent.com/12494624/202727885-e3e57b8e-9cc0-4a4d-ba5e-914d14aec0c8.mov

beygorghor (approved):

plugins/polio/js/src/forms/ScopeForm.tsx
```diff
@@ -33,25 +48,157 @@ export const ScopeForm: FunctionComponent = () => {
...
+        scopes,
+        search,
+        searchScope,
+    ]);
+
+    useEffect(() => {
+        if (districtShapes) {
+            searchDistrictByName();
+        }
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
quang-le: please fill in all deps or leave a comment explaining why it's ok not to list them all

plugins/polio/js/src/forms/ScopeForm.tsx
```diff
@@ -33,25 +48,157 @@ export const ScopeForm: FunctionComponent = () => {
...
+                searchScopeChecked={searchScopeChecked}
+                addNewScopeId={(id, vacciName) => addNewScopeId(id, vacciName)}
+                isFetchingDistricts={isFetchingDistricts || !filteredDistricts}
+                isFetchingRegions={isFetchingRegions || !regionShapes}
+                districtShapes={filteredDistricts}
+                regionShapes={regionShapes}
+                searchComponent={
+                    <ScopeSearch
+                        search={search}
+                        setSearch={setSearch}
```
quang-le: I'm not sure it's always safe to pass a reference to `setSearch`. I remember having to switch to useReducer to be able to pass a reference to the `set_` function of `useState` as prop. 

Just mentioning it in case it comes up in debugging

plugins/polio/js/src/forms/ScopeForm.tsx
```diff
@@ -33,25 +48,157 @@ export const ScopeForm: FunctionComponent = () => {
...
+                });
+            }
+
+            if (searchScope) {
+                filtreds = filtreds.filter(d =>
+                    scopes.some(scope => scope.group.org_units.includes(d.id)),
+                );
+            }
+
+            if (search !== '') {
```
quang-le: We should discuss this. In my understanding, the search feature was a way to control what's shown in the table to facilitate selecting the scope. The map on the other hand should always show the current scope

plugins/polio/js/src/components/Inputs/ScopeInput.tsx
```diff
@@ -207,6 +148,32 @@ export const ScopeInput: FunctionComponent<FieldProps<Scope[], Values>> = ({
...
+                    let addedDistr: FilteredDistricts | undefined;
+                    if (OrgUnitsIdInSameRegion.includes(dist.id)) {
+                        addedDistr = dist;
+                        if (addedDistr) {
+                            addedDistr.vaccineName = selectedVaccine;
+                            addNewScopeId(dist.id, selectedVaccine);
+                        }
+                    }
+                    return addedDistr;
+                });
```
quang-le: That's a strange use of `.filter`. It's supposed to return a boolean but it returns `FilteredDistricts | undefined`. Maybe you can chain `.filter` to filter out unwanted values with `.map` to change the shape of the `dist` objects?

plugins/polio/js/src/components/Inputs/ScopeInput.tsx
```diff
@@ -207,6 +148,32 @@ export const ScopeInput: FunctionComponent<FieldProps<Scope[], Values>> = ({
...
+
+            if ((searchLaunched || searchScopeChecked) && districtShapes) {
+                let newListAfterAdding = [...filteredDistricts];
+                const addedDistricts = districtShapes.filter(dist => {
+                    let addedDistr: FilteredDistricts | undefined;
+                    if (OrgUnitsIdInSameRegion.includes(dist.id)) {
+                        addedDistr = dist;
+                        if (addedDistr) {
+                            addedDistr.vaccineName = selectedVaccine;
+                            addNewScopeId(dist.id, selectedVaccine);
```
quang-le: It would be better to avoid side-effects in the `filter` method, if just for easier later debugging.
If possible, I would spin out the side-effect in a `forEach`or something, like:

```
const addedDistricts = districtShapes.filter(filterFn);
addedDist.forEach(dist=>{addNewScopeId(dist.id, selectedVaccine})
```

plugins/polio/js/src/components/Inputs/ScopeInput.tsx
```diff
@@ -196,9 +120,26 @@ export const ScopeInput: FunctionComponent<FieldProps<Scope[], Values>> = ({
...
+                let orgIdToRemove: number | undefined;
+                if (!OrgUnitsIdInSameRegion.includes(OrgUnitId)) {
+                    orgIdToRemove = OrgUnitId;
+                    addNewScopeId(OrgUnitId, '');
+                }
+                return orgIdToRemove;
+            });
         } else {
             // Remove the OrgUnits from all the scopes
             newScopes.forEach(s => {
```
quang-le: I feel this can be the cause of one of the bugs I recorded

plugins/polio/js/src/components/Scopes/DistrictScopeTable.tsx
```diff
@@ -0,0 +1,442 @@
...
+    const { formatMessage } = useSafeIntl();
+    const { value: scopes = [] } = field;
+    const { setValue: setScopes } = helpers;
+    const [rowsPerPage, setRowsPerPage] = useState(10);
+    const [orderBy, setOrderBy] = useState('asc');
+    const [sortFocus, setSortFocus] = useState('DISTRICT');
+
+    const addDistrictInVaccineScope = useCallback(
+        district => {
+            const newScopes: Scope[] = cloneDeep(scopes);
```
quang-le: 😀

quang-le (changes requested):
Big work, and nice cleanup of the code. I noted some hiccups with the feature:

- When playing with the add/remove region button in the table, the entire scope in unselected when re-adding a region that was removed (excpet for said region):


https://user-images.githubusercontent.com/38907762/203033605-1743cc30-e6ec-4c0e-b062-dca6bf2db2ff.mov

- When toggling the "Show districts in Scope" check box, the districts' boundary colors change, and the selection changes as well:


https://user-images.githubusercontent.com/38907762/203033920-ee5541ea-3925-43dc-829c-ef8e570855ae.mov

- When making a search including out of scope districts, if the search get 1 hit, the map will focus (change the fit to bounds) to that single district. It's confusing. Also, checking back the "Show districts in scope" check box, will then show an empty map:


https://user-images.githubusercontent.com/38907762/203034393-285da3b6-7de7-4207-bc73-80b775362045.mov

I think there's some clarification needed for that feature, as my understanding was that the search should just facilitate finding some districts in the table, and not impact the map, which should always show the current selected scope.

quang-le:
It's been confirmed that the search should only change the results displayed by the table, and not the map. I updated the [ticket](https://bluesquare.atlassian.net/browse/POLIO-620)

plugins/polio/js/src/forms/ScopeForm.tsx
```diff
@@ -1,30 +1,45 @@
...
+    const [searchScope, setSearchScope] = useState(true);
+    const [currentTab, setCurrentTab] = useState('1');
+    const [search, setSearch] = useState('');
+
+    const [debouncedSearch] = useDebounce(search, 500);
+
+    const sortedRounds: Round[] = useMemo(
         () =>
             [...rounds]
                 .map((round, roundIndex) => {
```
quang-le, line 46: Line 46, I think `round` should be `cloneDeep` instead of of spread (complex nested object)

plugins/polio/js/src/forms/ScopeForm.tsx
```diff
@@ -1,30 +1,45 @@
...
+    const [page, setPage] = useState(0);
+    const [searchScope, setSearchScope] = useState(true);
+    const [currentTab, setCurrentTab] = useState('1');
+    const [search, setSearch] = useState('');
+
+    const [debouncedSearch] = useDebounce(search, 500);
+
+    const sortedRounds: Round[] = useMemo(
         () =>
             [...rounds]
```
quang-le: As I recently learned at my expense: you should use lodash's cloneDeep here. Round is a nested object.
I would even think that you can skip copying rounds altogether, since .map returns a new array without changing the original

plugins/polio/js/src/forms/ScopeForm.tsx
```diff
@@ -33,25 +48,115 @@ export const ScopeForm: FunctionComponent = () => {
...
     };
-    return (
-        <>
+
+    const scopes = useMemo(() => {
+        if (!scopePerRound) {
+            return values.scopes;
+        }
+        if (rounds) {
+            const currentRound = sortedRounds[parseInt(currentTab, 10) - 1];
```
quang-le: I think this where it breaks with round 0

plugins/polio/js/src/forms/ScopeForm.tsx
```diff
@@ -33,25 +48,115 @@ export const ScopeForm: FunctionComponent = () => {
...
+            if (currentRound) {
+                return currentRound.scopes;
+            }
+        }
+        return [];
+    }, [currentTab, rounds, scopePerRound, sortedRounds, values.scopes]);
+
+    const filteredDistricts = useMemo(() => {
+        if (districtShapes) {
+            let filtreds: FilteredDistricts[] = [...districtShapes].map(
```
quang-le: filtered

quang-le: Also, I think you could use `const filtered` and use an early return i.o reassigning.

plugins/polio/js/src/forms/ScopeForm.tsx
```diff
@@ -33,25 +48,115 @@ export const ScopeForm: FunctionComponent = () => {
...
+                        });
+                    });
+                });
+            }
+
+            if (searchScope) {
+                filtreds = filtreds.filter(d =>
+                    scopes.some(scope => scope.group.org_units.includes(d.id)),
+                );
+            }
```
quang-le: You could early return here (see previous comment)

plugins/polio/js/src/forms/ScopeForm.tsx
```diff
@@ -33,25 +48,115 @@ export const ScopeForm: FunctionComponent = () => {
...
+
+            if (debouncedSearch !== '') {
+                filtreds = filtreds.filter(d =>
+                    d.name
+                        .toLowerCase()
+                        .includes(debouncedSearch.toLowerCase()),
+                );
+            }
```
quang-le: and here (idem)

plugins/polio/js/src/forms/ScopeForm.tsx
quang-le: Also `[...districtShapes]` is not necessary since you `.map` it

plugins/polio/js/src/forms/ScopeForm.tsx
```diff
@@ -33,25 +48,115 @@ export const ScopeForm: FunctionComponent = () => {
...
+        }
+        return [];
+    }, [currentTab, rounds, scopePerRound, sortedRounds, values.scopes]);
+
+    const filteredDistricts = useMemo(() => {
+        if (districtShapes) {
+            let filtreds: FilteredDistricts[] = [...districtShapes].map(
+                district => {
+                    return {
+                        ...district,
```
quang-le: I think shapes are nested objects, so `cloeDeep` i.o spread would be warranted

plugins/polio/js/src/forms/ScopeForm.tsx
```diff
@@ -33,25 +48,115 @@ export const ScopeForm: FunctionComponent = () => {
...
+                isFetchingDistricts={isFetchingDistricts || !filteredDistricts}
+                isFetchingRegions={isFetchingRegions || !regionShapes}
+                districtShapes={districtShapes}
+                regionShapes={regionShapes}
+                searchComponent={
+                    <ScopeSearch search={search} setSearch={setSearch} />
+                }
+                page={page}
+                setPage={setPage}
             />
```
quang-le: I would move that out in a separate component

plugins/polio/js/src/forms/ScopeForm.tsx
```diff
@@ -62,20 +167,21 @@ export const ScopeForm: FunctionComponent = () => {
                             />
                         ))}
                     </TabList>
-                    {sortedRounds.map(round => (
+                )}
+                {!scopePerRound && renderField('scopes')}
```
quang-le: moving `renderField` in a separate component would have the advantage that it would appear here as JSX

plugins/polio/js/src/components/Scopes/DistrictScopeTable.tsx
```diff
@@ -0,0 +1,370 @@
...
+    toggleDistrictInVaccineScope,
+    page,
+    setPage,
+    isFetching,
+}) => {
+    const [, , helpers] = useField(field.name);
+    const classes: Record<string, string> = useStyles();
+    const { formatMessage } = useSafeIntl();
+    const { value: scopes = [] } = field;
+    const { setValue: setScopes } = helpers;
```
quang-le: wouldn't it be quicker to do `const setScopes = useField(field.name)[3].setValue`? 
Maybe it's not as null safe? I don't know.

plugins/polio/js/src/components/Scopes/DistrictScopeTable.tsx
```diff
@@ -0,0 +1,370 @@
...
+    const addRegionToTable = useCallback(
+        (shape: ShapeRow) => {
+            if (filteredDistricts && districtShapes) {
+                const OrgUnitIdsToAdd = districtShapes
+                    .filter(s => s.parent_id === shape.parent_id)
+                    .map(s => s.id);
+
+                const newScopes: Scope[] = cloneDeep(scopes);
+                newScopes
+                    .filter(sc => sc.vaccine === selectedVaccine)
```
quang-le: I think this is causing the bug when adding a region though the table

plugins/polio/js/src/components/Scopes/DistrictScopeTable.tsx
```diff
@@ -0,0 +1,370 @@
...
+    isFetching: boolean;
+};
+
+export const DistrictScopeTable: FunctionComponent<Props> = ({
+    field,
+    selectedVaccine,
+    regionShapes,
+    filteredDistricts,
+    districtShapes,
+    toggleDistrictInVaccineScope,
```
quang-le, line 66: why `toggleDistrictInVaccineScope`? It's only used to remove

plugins/polio/js/src/components/Scopes/DistrictScopeTable.tsx
```diff
@@ -0,0 +1,370 @@
...
+        sortFocus,
+        orderBy,
+        page,
+        rowsPerPage,
+        scopes,
+    ]);
+    const displayPlacHolder = useMemo(
+        () => isFetching || (!isFetching && filteredDistricts?.length === 0),
+        [filteredDistricts?.length, isFetching],
+    );
```
quang-le: No need for `useMemo`here since you're returning a boolean

plugins/polio/js/src/components/Inputs/ScopeInput.tsx
```diff
@@ -2,170 +2,78 @@
...
+    page,
+    setPage,
 }) => {
-    // eslint-disable-next-line no-unused-vars
-    const [_field, _meta, helpers] = useField(field.name);
-    const classes: Record<string, string> = useStyles();
-    const { formatMessage } = useSafeIntl();
     const [selectRegion, setSelectRegion] = useState(false);
+    const [selectedVaccine, setSelectedVaccine] = useState<string>('mOPV2');
+    const [, , helpers] = useField(field.name);
```
quang-le, line 66: cf infra about the declaration

plugins/polio/js/src/components/Inputs/ScopeInput.tsx
```diff
@@ -2,170 +2,78 @@
...
+    searchComponent,
+    page,
+    setPage,
 }) => {
-    // eslint-disable-next-line no-unused-vars
-    const [_field, _meta, helpers] = useField(field.name);
-    const classes: Record<string, string> = useStyles();
-    const { formatMessage } = useSafeIntl();
     const [selectRegion, setSelectRegion] = useState(false);
+    const [selectedVaccine, setSelectedVaccine] = useState<string>('mOPV2');
```
quang-le, line 65: This typing can be stronger. I think I introduce stronger typing for viruses and vaccines in the PR about selecting round in the calendar, FYI

plugins/polio/js/src/components/Inputs/ScopeInput.tsx
```diff
@@ -2,170 +2,78 @@
...
-
-    const getBackgroundLayerStyle = () => {
-        return {
-            color: 'grey',
-            opacity: '1',
-            fillColor: 'transparent',
-        };
-    };
-
     const toggleRegion = (
```
quang-le: useCallback

plugins/polio/js/src/components/Inputs/ScopeInput.tsx
```diff
@@ -254,192 +168,52 @@ export const ScopeInput: FunctionComponent<FieldProps<Scope[], Values>> = ({
         },
         [scopes, setScopes],
     );
-    //
+
     const onSelectOrgUnit = shape => {
```
quang-le: useCallback

quang-le (changes requested):
Great rewrite, the code is much more readable and easy to navigate Some changes here and there in the code. I tried to point out potential sources of bugs I found while testing the feature.

- Selecting a region from the table unselects all other regions. Also, if the vaccine has been changed, the region selection ignores it



https://user-images.githubusercontent.com/38907762/204263129-47fb6feb-9fb2-4e39-89ff-ffe41bb7609a.mov



https://user-images.githubusercontent.com/38907762/204263263-be84386e-f2df-4d22-a190-f20619bf3dce.mov


- Adding a round/re-adding a round after removing crashes the page:


https://user-images.githubusercontent.com/38907762/204263523-14a3c31e-16ae-4c80-a703-2bfa6bf9b4cd.mov

- Selecting `Separate Scopes per round` before saving the rounds crashes the app (probably same bug as above):


https://user-images.githubusercontent.com/38907762/204263916-a4ba59c7-dd01-4874-95b1-7086dedf30bc.mov

- The table doesn't show the scope for Round 0: 

Screenshot 2022-11-28 at 10 48 37


- There's an error in the console:

Screenshot 2022-11-28 at 10 51 56

quang-le (approved):

PR #118, opened 2022-10-28
WC2-97: iaso_storages not translated

beygorghor:
Adding translation in the messages for iaso_storages
Screenshot 2022-10-28 at 09 58 56


olethanh:
Missing translation in French

olethanh:
nervermin it was reusing another trad

PR #121, opened 2022-10-28
Budget Put the attachment required message on top

olethanh:
Open this PR for discussion, I'm proposing that we put the Attachement required message directly above the Files inputs, as to make the flow of the dialog easier to read. Especially since we added the Recipients teams

Current:
image

Proposal
image


quang-le:
I can see that the text at the bottom of the modal is not ideal, but I'm not a fan of putting it up there like this, since all errors/helper texts are under the relevant fields. 
Also putting a general/multi field error is consistent with what we do elsewhere in polio, eg for missing round dates

beygorghor:
Maybe a common field delimitation  (border) can do the trick... also remove the divider then

PR #122, opened 2022-10-28
Documentation Enketo

olethanh:
Document the Enketo workflow more. Especially the public_url
Added a Workflow diagram

@mestachs Can you check if it looks correct to you?
The flow is not visible directly in the PR but is readable here
https://github.com/BLSQ/iaso/blob/4b131857878f544176203c977c9bc09f06c5496b/iaso/enketo/README.md#flow-for-public_create_url


mestachs:
note that there's a flow where 
 * the invoice invoice app (which looks like what you documented)
    *  this one implies some CORS settings in dhis2 ([and I think in iaso too)]
    *  the parameters are "in the dhis2 world" because (we don't know much about iaso db details)
       * form_id=CS_QLTE_RDC003-Gestion_finance (the form_id not the technical id of the db)
       * period=2022Q3 (in dhis2 format)
       * external_user_id= dhis2 user id (an user/profile is created automatically)
       * external_org_unit_id=DWmKNyyDCWv   (source_ref of the iaso orgunit)
    *  it assume that iaso won't create "duplicate" instances for "single submission per period forms" (so it creates a submission or edit the existing one)
    * for the export 
       * for the moment we want the export to be synchronous
       * the user expect to see the changes in the data when redirected to the invoice app
       * if the export fails the user is "stuck" on the enketo submission
       * note that for the user is probably doesn't know that iaso exist
 * where "edit in enketo" : iaso calls that api too but redirects to iaso after edition ?

olethanh:
I commented from what I saw in the code and what I had discussed with Fleury at that time. I'm not familiar with Invoice the Invoice App. but we can add these information to the README.

It would be really helpful to have a list of which external service use theses Iaso  API endpoint so we don't accidentally break them.

There is also an url `/fill/<form_uuid>/<org_unit_id>/<period>` and I'm not sure who and what is using it (or even if it is used)

For the `Edit in Enketo`  it use `/enketo/create/` instead of `/enketo/public_create_url/` which behave nearly in the same way but not 100% exactly. It's restricted to currently logged in user and embed the current user in the form definition. There are other difference it take the internal iaso form_id and not the one from the XML, idem for Org Unit etc..

I think everything from  Step "4"  to Step "14" in the flow diagram would be the same however. And it use a different mechanism for Export

We should definitely document this more.

mestachs:
for "/fill" we should ask @madewulf , I think it's another usecase (sending links to people so they can encode things anonymously ?)

I think one call is missing it's a HEAD to know the "max content size"
https://github.com/BLSQ/iaso/blob/main/iaso/api/enketo.py#L284

https://docs.getodk.org/openrosa-form-submission/#content

olethanh:
Added the Head  call. Martin said the /fill url is not used at the moment

PR #131, opened 2022-11-03
WC-98: Logs per device endpoint: added "performed_at" filter

niconoe:
Added a performed_at filter to the "list logs per device" endpoint, as requested by @beygorghor in https://bluesquare.atlassian.net/browse/WC2-98

- [x] Are there enough tests

## How to test

Create a new StorageDevice called superbe_id of type NFC, and add it a few log entries (from Django Admin)

Go to `/api/storage/NFC/superbe_id/logs` and check that you can filter the logs with the performed_at filter, for example: `/api/storage/NFC/superbe_id/logs?performed_at=2022-11-03`

### DIFF
--- a/iaso/api/storage.py
+++ b/iaso/api/storage.py
@@ -1,5 +1,5 @@
 # TODO: need better type annotations in this file
-
+import datetime
 from typing import Tuple
 
 from django.db.models import Prefetch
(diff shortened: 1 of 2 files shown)


beygorghor:
You can test it directly with the dashboard too. Workng for me

beygorghor (approved):

SimonDus (approved):
Tested and approved.

Good work

PR #133, opened 2022-11-04
 IA-1618 Submission content filter add support for number in API

olethanh:
When querying for Submission content it now works with number. It was not working before because we are storing the number as string in the JSON on the DB. Now we cast them.

We still need to re enable it in the frontend.

- [ ] Did I add translations
- [x ] Are there enough tests

## Changes

If we receive a number to compare to we force cast the json field value to a   number for comparison

I had to do use arcane django stuff, glaned from their source code, but everything should be sufficiently documented in the comments.

## How to test

We still need to reenable it in the frontend. You can test via the API for now. See the call in the test.

### DIFF
--- a/iaso/tests/api/test_instances.py
+++ b/iaso/tests/api/test_instances.py
@@ -486,7 +486,7 @@ class InstancesAPITestCase(APITestCase):
             period="202001",
             org_unit=self.jedi_council_corruscant,
             project=self.project,
-            json={"name": "a", "age": 18, "gender": "M"},
+            json={"name": "a", "age": "18", "gender": "M"},
         )
 
         b = self.create_form_instance(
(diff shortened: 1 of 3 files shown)


niconoe (changes requested):
Thanks @olethanh, that's a great addition while we wait for the ideal solution (store things more "cleanly" when saving the JSON in the database).

My only tiny remark is that we could be a bit more specific/helpful in the different `ValueErrors`: `raise ValueError(f"Unsuported JsonLogic (unknown operator): {jsonlogic}")` or `raise ValueError(f"Unsuported JsonLogic (exactly 2 operands required): {jsonlogic}")` for example

olethanh:
I put in nicer error and fixed the remaining tests

niconoe (approved):
Great job, thanks @olethanh

PR #136, opened 2022-11-04
Polio 670 destruction update

quang-le:
There was only 1 destruction of vaccines per rounds, but we need many

- [X] Did I add translations

## Changes

- Created a `Destruction` model
- Updated `RoundSerializer` to use the new model
- Created `DestructionForm` and `DestructionsForm` to handle adding and removing destructions
- Updated yup validation to reflect changes
- fixed some translation typos
- removed validator error on number fields with value 0
- Add fix for `TextArea` infinite loop bug

## How to test

Go to campaigns, open one, go Vaccines tab and try adding and removing destructions

## Print screen / video

https://user-images.githubusercontent.com/38907762/200021794-18db2189-fd96-4a3d-a847-603ab2d2f880.mov

## Notes

There's room for improvement as most code (both front and back) is pretty much copy pasted from Shipments


plugins/polio/js/src/constants/translations/en.json
```diff
@@ -1,4 +1,5 @@
 {
+    "Date of the destruction report": "Date of the destruction report",
```
hakifran, line 2: This shouldn't be there.

plugins/polio/js/src/constants/translations/fr.json
```diff
@@ -103,6 +103,7 @@
     "iaso.polio.import_line_list": "Importer Line List",
     "iaso.polio.import_line_list.successMessage": "Import réussi pour {amount} campagne(s)",
     "iaso.polio.label.add": "Ajouter",
+    "iaso.polio.label.addDestruction": "Ajouter destruction",
```
hakifran: ce ne serait pas mieux si on disait "Ajouter une destruction"?

plugins/polio/js/src/constants/translations/fr.json
```diff
@@ -286,6 +289,7 @@
     "iaso.polio.label.rejectBudgetForCampaign": "Rejeter le budget pour la campagne: {campaign}",
     "iaso.polio.label.rejected": "Rejeté",
     "iaso.polio.label.Religious_leader": "Leader religieux",
+    "iaso.polio.label.removeLastDestruction": "Effacer dernière destruction",
```
hakifran: "Effacer la dernière destruction"

plugins/polio/js/src/forms/DestructionsForm.tsx
```diff
@@ -0,0 +1,108 @@
...
+}) => {
+    const { formatMessage } = useSafeIntl();
+    const { setFieldValue } = useFormikContext();
+    const { destructions = [] } = round ?? {};
+    const [enableRemoveButton, setEnableRemoveButton] =
+        useState<boolean>(false);
+
+    const lastIndex = destructions.length >= 1 && destructions.length - 1;
+
+    const handleAddShipment = () => {
```
hakifran: This method should be renamed to "handleAddDestruction"

plugins/polio/js/src/forms/DestructionsForm.tsx
```diff
@@ -0,0 +1,108 @@
...
+        const newShipments = [...destructions, {}];
+        setFieldValue(`${accessor}.destructions`, newShipments);
+    };
+
+    const handleRemoveLastShipment = useCallback(() => {
```
hakifran: This method too "handleRemoveLastDestruction"

plugins/polio/js/src/forms/DestructionsForm.tsx
```diff
@@ -0,0 +1,108 @@
...
+        const newShipments = [...destructions];
```
hakifran: Here too newDestructions

plugins/polio/js/src/forms/VaccineManagementForm.tsx
```diff
@@ -192,11 +193,16 @@ export const VaccineManangementForm: FunctionComponent<Props> = () => {
                         </Typography>
                     </Box>
                 </Grid>
-                <Grid container direction="row" spacing={2} item xs={12}>
-                    {/* Not sure yet if this should not be a map */}
+                <DestructionsForm
+                    accessor={accessor}
+                    round={rounds[roundIndex]}
+                />
+                {/* <Grid container direction="row" spacing={2} item xs={12}>
```
hakifran: This commented code be removed to make the code more clean. thank you

hakifran (commented):
Apart from some translation and methods name to change, it seems working for me and the remaining code seems to me correct. thank you @quang-le

plugins/polio/js/src/forms/DestructionsForm.tsx
quang-le: Me copy/pasting my own code?😅

beygorghor (changes requested):
Screenshot 2022-11-07 at 12 07 32
Screenshot 2022-11-07 at 12 07 09

I have an error on textArea while i shouldn't
Screenshot 2022-11-07 at 12 07 22
Also adding some more space before the divider if possible

quang-le:
> 

It seems validation for Shipments broke somewhere on the way

codeKameleon:
@quang-le I think it would be more clear to the user if we add required asterisk to the inputs that need to be filled before saving

plugins/polio/js/src/forms/DestructionsForm.tsx
```diff
@@ -0,0 +1,108 @@
...
+        useState<boolean>(false);
+
+    const lastIndex = destructions.length >= 1 && destructions.length - 1;
+
+    const handleAddDestruction = () => {
+        const newDestructions = [...destructions, {}];
+        setFieldValue(`${accessor}.destructions`, newDestructions);
+    };
+
+    const handleRemoveLastDestruction = useCallback(() => {
```
codeKameleon, line 36: @quang-le Is there a use case where the user would like to delete a older destruction ? For the moment the feature only allow to remove the last one

codeKameleon (changes requested):

plugins/polio/js/src/forms/DestructionsForm.tsx
quang-le: Not really. They might want to edit it, but once the vaccines are destroyed, they are destroyed. Same with Shipments

quang-le:
> @quang-le I think it would be more clear to the user if we add required asterisk to the inputs that need to be filled before saving

The thing is that it's conditional: all destruction inputs should be filled for a line (except comment) only if one has been filled. This as per client's request. Same with Shipments

quang-le:
> > 
> 
> It seems validation for Shipments broke somewhere on the way

It's not that, it's just that for new campaigns, the whole form will error when there's no obr_name. It works fine otherwise

codeKameleon:
> > @quang-le I think it would be more clear to the user if we add required asterisk to the inputs that need to be filled before saving
> 
> The thing is that it's conditional: all destruction inputs should be filled for a line (except comment) only if one has been filled. This as per client's request. Same with Shipments

Ok i see, so maybe in a future PR, show an info message in the Destruction section if the user put a value on one of the inputs, we could show "These fiellds need to be filled ..."

olethanh:
EDIT : Apparently I cannot read. You already did the AuditDestructionSerializer, you just need to add it to the AuditRoundSerializer

You need to add the new Models to the Audit so the change modifications get kept.

See following PR for example on how to do this
https://github.com/BLSQ/iaso/pull/60/files

But basically you need to ad an AuditSerializer and then reference it in the

hakifran (approved):
Seems good to me. thank you @quang-le

PR #137, opened 2022-11-07
IA-1613 expose the creator of an org unit in the api and show it in the org unit details

SimonDus:
- [x] Did I add translations
## Changes

Expose OU creator in OU List and ou Details

## How to test

Create an org unit, go to ou List or Ou details. 

## Print screen / video

image

image

### DIFF
--- a/iaso/api/serializers.py
+++ b/iaso/api/serializers.py
@@ -83,7 +83,13 @@ class OrgUnitSerializer(TimestampSerializerMixin, serializers.ModelSerializer):
         return org_unit.location.z if org_unit.location else None
 
     def get_creator(self, org_unit):
-        return None if org_unit.creator is None else org_unit.creator.username
+        creator = None
+        if org_unit.creator is not None:
+            if org_unit.creator.first_name is not None and org_unit.creator.last_name is not None:
+                creator = f"{org_unit.creator.username} ( {org_unit.creator.first_name} {org_unit.creator.last_name} )"
+            else:
+                creator = org_unit.creator.username
+        return creator
 
     class Meta:
         model = OrgUnit
@@ -112,9 +118,6 @@ class OrgUnitSerializer(TimestampSerializerMixin, serializers.ModelSerializer):
 
 
 class OrgUnitSmallSearchSerializer(OrgUnitSerializer):
-    def get_creator(self, org_unit):
-        return None if org_unit.creator is None else org_unit.creator.username
-
     class Meta:
         model = OrgUnit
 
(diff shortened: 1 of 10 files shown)


beygorghor (approved):
- "en" translation was missing for "iaso.orgUnits.createdBy"
- i removed width on id and group cell to make the columns fit automatically
- Also display a placeholder instead of nothing if no creator
I did those changes ;-)

quang-le (approved):

iaso/api/serializers.py
```diff
@@ -104,10 +107,14 @@ class Meta:
...
             "created_at",
             "org_unit_type_id",
+            "creator",
         ]
 
 
 class OrgUnitSmallSearchSerializer(OrgUnitSerializer):
+    def get_creator(self, org_unit):
+        return None if org_unit.creator is None else org_unit.creator.username
+
```
niconoe: Is it necessary to redefine here if the same code already appears in the parent class?

iaso/api/serializers.py
```diff
@@ -143,6 +152,15 @@ def get_instances_count(self, org_unit):
         else:
             return org_unit.instance_set.filter(~Q(file="") & ~Q(device__test_device=True) & ~Q(deleted=True)).count()
 
+    def get_creator(self, org_unit):
```
niconoe: I'm not familiar with the code, but I'm not sure why we redefine here an improved(?) version of get_creator() here. Would the improved version work everywhere (in that case I suggest moving it to the parent class and letting others inherit it?). Or is there a good reason for a different behavior in `OrgUnitSearchSerializer`? If so, can we add a comment explaining the "why"? thanks!

iaso/tests/api/test_serializers.py
```diff
@@ -251,10 +254,12 @@ def test_small_search(self):
         self.assertDictEqual(
             res,
             {
+                "creator": None,
                 "id": orgunit.id,
                 "name": "Endor Jedi Squad 1",
                 "org_unit_type_name": "Jedi Squad",
                 "parent": {
+                    "creator": None,
```
niconoe, line 267: Maybe we can improve the tests (or add a new one), to make sure the creator field has the expected value in case it's not None? thanks!

niconoe (changes requested):
Thanks a lot for the good work, I just let a few small comments.

iaso/api/serializers.py
SimonDus: Indeed, I forgot to remove it as the "OrgUnitSmallSearchSerializer" is used for the details and I did it first.

iaso/tests/api/test_serializers.py
SimonDus: Yep, done

olethanh:
I would say it need a prefetch_related to not create to much query. But I'm not sure which endpoint get impacted by this change. 

Also not sure why we are returning a string and not an embeded object like we do in other parts ?

SimonDus:
> I would say it need a prefetch_related to not create to much query. But I'm not sure which endpoint get impacted by this change.
> 
> Also not sure why we are returning a string and not an embeded object like we do in other parts ?



> I would say it need a prefetch_related to not create to much query. But I'm not sure which endpoint get impacted by this change.
> 
> Also not sure why we are returning a string and not an embeded object like we do in other parts ?

Because Son told me that on other part of Iaso we use " Username ( first_name last_name ) " I think in budget. So I made something similar. But for me it can return only the username... It's kinda the same to me

olethanh:
I think he mean the function getDisplayName in userUtils.ts but that  is in the frontend. I'm merging we can do the rest later

PR #143, opened 2022-11-08
POLIO-522: allow confirm if name and children campaigns

codeKameleon:

- [ ] Did I add translations
- [ ] Are there enough tests
## Changes

- update the `allowConfirm` value to make sure `campaignsToLink` select has a value 

## How to test

Go To Grouped campaigns > Create > Fill the inputs
You won't be able to save if either the name or a campaign to link input values is empty.

## Print screen / video

[recording-2022-11-08-11-38-44.webm](https://user-images.githubusercontent.com/25134301/200543289-172196d9-827a-4199-b365-82d729c4c1f5.webm)


plugins/polio/js/src/pages/GroupedCampaigns/GroupedCampaignDialog.tsx
```diff
@@ -9,6 +9,7 @@ import { Box, Divider, Grid } from '@material-ui/core';
 // @ts-ignore
 import ConfirmCancelDialogComponent from 'Iaso/components/dialogs/ConfirmCancelDialogComponent';
 import { useSafeIntl } from 'bluesquare-components';
+import { StringSchema } from 'yup';
```
beygorghor: this is not used, you can remove it

plugins/polio/js/src/pages/GroupedCampaigns/GroupedCampaignDialog.tsx
```diff
@@ -91,7 +92,9 @@ export const GroupedCampaignDialog: FunctionComponent<Props> = ({
         },
         [reset],
     );
-    const allowConfirm = Boolean(groupedCampaignName);
+    const allowConfirm =
+        Boolean(groupedCampaignName) && typeof campaignsToLink === 'string';
```
beygorghor: `Boolean (campaignsToLink )` is not working ?

beygorghor (approved):
Just remove unused import

plugins/polio/js/src/pages/GroupedCampaigns/GroupedCampaignDialog.tsx
quang-le: `Boolean(campaignsToLink)` would indeed be slightly better. We're safe here because the `Select` puts constraints on the value, but `typeof '' === 'string'` is `true` whereas `Boolean("")` is `false`

quang-le (changes requested):

plugins/polio/js/src/pages/GroupedCampaigns/GroupedCampaignDialog.tsx
codeKameleon: @quang-le @beygorghor The problem with` Boolean(camapignsToLink)` when you open the dialog and fill the name field, the allowConfirm will be set to `true` because `campaignsToLink` is an empty array and `Boolean(campaingsToLInk)` will be evaluated to `true` but we only want to be set to true if a campaign is selected and the value will be a string.

[recording-2022-11-14-14-18-46.webm](https://user-images.githubusercontent.com/25134301/201670292-b60b3813-fe38-4013-83ae-5c47a5342d38.webm)

quang-le: The mistake was initially mine when we shipped the feature. We left a workaround that was never fixed. I'll fix it now. basically, you need to ensure that the type of `campaignsToLink` remains constant, in this case an array

quang-le (commented):
I fixed the old ugly workaround, so it's really @beygorghor  who should review

beygorghor (approved):
Ok for me, i have added a loading on the campaign select

PR #147, opened 2022-11-09
IA-1622 Fix crash when creating campaign with vaccine data

olethanh:

- [x] Are there enough tests

## Changes

Fix crash by using the Round Serializer in campaign creation too. 
Added two test for creating and updating data on the vaccine tab.

## How to test

Test campaign creation and update on the Campaign

### DIFF
--- a/plugins/polio/serializers.py
+++ b/plugins/polio/serializers.py
@@ -337,7 +337,7 @@ class RoundSerializer(serializers.ModelSerializer):
         for shipment in shipments:
             Shipment.objects.create(round=round, **shipment)
         for destruction in destructions:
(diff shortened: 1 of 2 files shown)


plugins/polio/serializers.py
```diff
@@ -337,7 +337,7 @@ def create(self, validated_data):
         for shipment in shipments:
             Shipment.objects.create(round=round, **shipment)
         for destruction in destructions:
-            Shipment.objects.create(round=round, **destruction)
+            Destruction.objects.create(round=round, **destruction)
```
quang-le, line 340: shame on me

quang-le (approved):

PR #150, opened 2022-11-09
IA-1552: account id in urls

beygorghor:
The idea is to add account_id param everywhere in the dashboard in order to debug, share easily links

- [ ] Did I add translations
- [ ] Are there enough tests
## Changes

- refactor o `protectedRoute`, main `index` and `App`
- adding accountId in routes, plugin routes, sidebar menu, redirections
- fix tests for new urls

## How to test

Unfortunately you have to pass a bit everywhere.

## Notes

This is working also for anonymous plugins page like embedded calendar


beygorghor:
@cypress

beygorghor:
@cypress

beygorghor:
@cypress

olethanh:
`${siteBaseUrl}/dashboard/orgunits/detail/orgUnitId/${orgUnit.id}/tab/history/accountId/1`

Wouldn't it be more stable and future proof to have the accountId at the start ? Like this ?

`${siteBaseUrl}/dashboard/accountId/1/orgunits/detail/orgUnitId/${orgUnit.id}/tab/history`

would seems less prone to future conflict for me and less chance to mix it with deep linking.

Also wonder if it would be nice to have the account name instead

`${siteBaseUrl}/dashboard/account/pbf_rca/orgunits/detail/orgUnitId/${orgUnit.id}/tab/history`

beygorghor:
> `${siteBaseUrl}/dashboard/orgunits/detail/orgUnitId/${orgUnit.id}/tab/history/accountId/1`
> 
> Wouldn't it be more stable and future proof to have the accountId at the start ? Like this ?
> 
> `${siteBaseUrl}/dashboard/accountId/1/orgunits/detail/orgUnitId/${orgUnit.id}/tab/history`
> 
> would seems less prone to future conflict for me and less chance to mix it with deep linking.
> 
> Also wonder if it would be nice to have the account name instead
> 
> `${siteBaseUrl}/dashboard/account/pbf_rca/orgunits/detail/orgUnitId/${orgUnit.id}/tab/history`

@madewulf  ?

olethanh:
Also concerning putting it in front, I'm not certain but from my understanding of ReactRouter, if you  put it in front you could just wrap all the current in a parent <Route> component.

I'm seeing they are doing it like this for nesting in their documentation.

```
<Router>
    <Route path="/" component={App}>
      <Route path="about" component={About} />
      <Route path="inbox" component={Inbox}>
        <Route path="messages/:id" component={Message} />
      </Route>
    </Route>
  </Router>
```

https://github.com/remix-run/react-router/blob/v3.2.6/docs/guides/RouteConfiguration.md

hat/assets/js/apps/Iaso/constants/routes.js
```diff
@@ -815,4 +815,11 @@ export const routeConfigs = [
...
     storageDetailPath,
-];
+].map(conf => {
+    const newConf = { ...conf };
+    newConf.params.push({
+        isRequired: false,
+        key: 'accountId',
+    });
+    return newConf;
+});
```
quang-le: It would save us a loop through the conf to define it directly in each conf, the way we do with `paginationParams`.

hat/assets/js/apps/Iaso/domains/app/components/MenuItemComponent.js
```diff
@@ -73,13 +72,19 @@ function MenuItem(props) {
         <>
             <Link
                 className={classes.linkButton}
-                to={!hasSubMenu ? fullPath : ''}
-                target={urlLink ? "_blank" : ""}
+                to={
+                    !hasSubMenu
+                        ? `${fullPath}/accountId/${currentUser.account.id}`
```
quang-le: We could add the account directly when defining `fullPath`

hat/assets/js/apps/Iaso/domains/users/components/ProtectedRoute.js
```diff
@@ -1,109 +1,92 @@
...
-            </>
-        );
+    }, [allRoutes, currentUser, dispatch, isRootUrl, permissions]);
+
+    let isAuthorized =
+        permissions.length > 0
+            ? userHasOneOfPermissions(permissions, currentUser)
+            : true;
+    if (featureFlag && !hasFeatureFlag(currentUser, featureFlag)) {
+        isAuthorized = false;
```
quang-le: Why not define that earlier and use it in the useEffect above?
Also, line 65, I'm not sure I'm conmfortable with `isAuthorized`  being et to `true` by default. I guess it might be a byproduct of transposing `componentDidUpdate` to `useEffect`. If so, maybe it can be avoided by using `useSkipEffectOnMount` (I don't know, maybe you tried and it didn't work, I'm just thinking out loud really)

quang-le: Ok, I just saw that we set it to `true` in the old code as well. No idea why though

hat/assets/js/apps/Iaso/domains/users/components/ProtectedRoute.js
```diff
@@ -1,109 +1,92 @@
...
-        {
-            fetchCurrentUser: fetchCurrentUserAction,
-            redirectTo: redirectToAction,
-        },
-        dispatch,
-    ),
-    dispatch,
-});
-
-export default connect(MapStateToProps, MapDispatchToProps)(ProtectedRoute);
```
quang-le, line 109: 🍾

hat/assets/js/apps/Iaso/index.js
```diff
@@ -29,39 +29,6 @@ export default function iasoApp(
...
-        />
-    ));
-
-    const overrideLandingRoutes = plugins
-        .filter(plugin => plugin.overrideLanding)
-        .map(plugin => plugin.overrideLanding);
-    // using the last plugin override (arbitrary choice)
-    const overrideLanding =
-        overrideLandingRoutes.length > 0
-            ? overrideLandingRoutes[overrideLandingRoutes.length - 1]
```
quang-le, line 62: Out of curiosity: what was gained by moving this code to `App`?

hat/assets/js/apps/Iaso/routing/redirections.js
```diff
@@ -14,73 +14,114 @@ import Page404 from '../components/errors/Page404';
...
+                '(/withLocation/:withLocation)(/deviceId/:deviceId)(/deviceOwnershipId/:deviceOwnershipId)(/tab/:tab)(/columns/:columns)(/search/:search)(/showDeleted/:showDeleted)',
+        },
+        {
+            path: '/*',
+            component: ({ location }) => <Page404 location={location} />,
+        },
+    ];
 };
+const getRoutes = (baseRoutes, overrideLanding) =>
+    [...baseRoutes].concat(
```
quang-le: cloneDeep

hat/assets/js/apps/Iaso/utils/usersUtils.ts
```diff
@@ -5,6 +5,7 @@ export type Profile = {
     id: string;
     first_name: string;
     user_name: string;
```
quang-le: this one should be optional as well, no?

hat/assets/js/apps/Iaso/utils/usersUtils.ts
```diff
@@ -16,7 +17,7 @@ export type User = {
     first_name: string;
     last_name: string;
     user_name: string;
```
quang-le: cf supra

hat/assets/js/apps/Iaso/utils/usersUtils.ts
```diff
@@ -62,7 +63,7 @@ export const getDisplayName = (
     // Some endpoint have user_name and some username (without the _, fun)
     const userName = user.user_name ?? user?.username;
```
quang-le, line 64: shouldn't it be `user?.user_name` to avoid `undefined.user-name` ?

hat/assets/js/apps/Iaso/utils/usersUtils.ts
```diff
@@ -62,7 +63,7 @@ export const getDisplayName = (
...
     if (!user.first_name && !user.last_name) {
-        return userName;
+        return userName || '';
     }
     return `${userName} (${user.first_name ? `${user.first_name}` : ''}${
```
quang-le, line 68: if `userName === ''`,  this will return `(firstName, lastName)`. We should probably add a check to remove the partheses when `!userName`

quang-le (commented):
Haven't tested the feature, but I'll post already my comments about the code. 
If possible without breaking everything, I'll second @olethanh 's proposal to put the accountId at the beginning of the url rather than the end.

Nice job with `ProtectedRoute`😀!

hat/assets/js/apps/Iaso/routing/redirections.js
```diff
@@ -14,73 +14,114 @@ import Page404 from '../components/errors/Page404';
...
+const getRedirections = overrideLanding => {
+    const getPaginationParams = (order = 'id', pageSize = 20) =>
+        `/order/${order}/pageSize/${pageSize}/page/1`;
+    return [
+        {
+            path: '/',
+            to: overrideLanding ?? baseUrls.forms,
+        },
+        {
+            path: `${baseUrls.orgUnits}/accountId/:accountId`,
```
quang-le: I think the way we redirect now messes up with `goBack` in the router: I get a bug where clicking the back arrow in the `TopBar` brings me back to the url without the accountId then redirected immediately to the same url with the accounId. So I'm basically stuck on the page.

Also, when redirecting without history (deep linking) we need to pass the accountId in the params, otherwise we get a 404 (see Org Unit details for example)

quang-le (changes requested):
Works seamlessly overall with the exception of the back button in the `TopBar`:

- When deep linking, user gets a 404

https://user-images.githubusercontent.com/38907762/204509938-62686e5e-c1a2-4689-bf9f-431568c05fb1.mov


- When router has history, users gets stuck on the page. It seems the history remembers the url without the `accountId`, sends the user there, and then redirects again to the page they're trying to leave

https://user-images.githubusercontent.com/38907762/204510923-688211c7-c989-4a71-a952-1391fadffc3a.mov

- Some `GET`requests pass the `accountId` in the params though it's not necessary

I made a branch with a fix for the `GET` requests, a tentative fix for using the back button after deep linking (in OU details), and various minor bug fixes I stumbled upon while reviewing: https://github.com/BLSQ/iaso/tree/remove_account_id_from_requests

All Cypress tests passed locally

hat/assets/js/apps/Iaso/index.js
beygorghor: so i could use current user reducer

hat/assets/js/apps/Iaso/utils/usersUtils.ts
beygorghor: `userName ` cannot be ''

hat/assets/js/apps/Iaso/utils/usersUtils.ts
beygorghor: `user` cannot be null

beygorghor:
According to our discussion, accound id should be present at the beginning of the url.
We have to update tests and fix @quang-le issues

quang-le:
I added retries to the cypress tests. the tests will take even more time this way, but it will prevent some flaky tests from failing so it should be an overall win

quang-le (approved):
I approve my own changes. @beygorghor  I can' set you up as your own reviewer, so I tag you here

codeKameleon (changes requested):
Storages is broken (apart from this page, everything is works well!)
Capture d’écran du 2022-12-13 16-02-15

codeKameleon (approved):

niconoe (approved):
I've tried, without success, to generate issues by clicking everywhere for one hour, good job!

Given the nature of the changes, it would still be good if others could do the same!

PR #153, opened 2022-11-09
IA-1625: allow run multiple workflow at the same time

codeKameleon:

- [ ] Did I add translations
- [ ] Are there enough tests
## Changes

Explain the changes that were made. The idea is not to list exhaustively all the changes made (Github already provides a full diff), but to help the reviewers better understand:

## How to test

Explain how to test your PR.

## Print screen / video

## Notes

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -6,7 +6,7 @@ on:
 
 concurrency: 
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: false
 
 env:
   REGISTRY: ghcr.io


.github/workflows/cypress.yml
```diff
@@ -6,7 +6,7 @@ on:
 
 concurrency: 
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: false
```
quang-le, line 9: I think this reverts a change @qgerome  made to give us some performance gain

quang-le (commented):
Pls check your change with @qgerome

.github/workflows/cypress.yml
qgerome: This change in particular was not for performance but to cancel already running workflows when you trigger a new one. But there is an error in the `group` and it cancels all runs of this workflow instead. We can safely merge this.

qgerome (approved):

codeKameleon:
@cypress

codeKameleon:
@cypress

PR #165, opened 2022-11-14
IA-1628: [Form field filter] New version of form creates issues when searching in fields of old version of form

beygorghor:
Example from PEVCameroun on Iaso prod, form [Collecte des données de la FOSA](https://iaso.bluesquare.org/dashboard/forms/detail/formId/735):

In the version 2022110801 of the form there is a question “test_instance” allowing the user to indicate if this form is filled out for test reasons.

In the version 2022111001 of the form, this question was deleted by request of the client.

Now, when we try to find all forms that have test_instance = ‘yes’, we cannot select possible responses anymore, probably because the system takes the options from the lastest version of the form:

{"and":[{"==":[{"var":"test_instance"},"yes"]}]}  works however.

image-20221110-135211

- [ ] Did I add translations
- [ ] Are there enough tests

## Changes

Removed versionId param from api call to form to get form descriptor

## How to test

Create a new version of an existing form by removing a field (select with options) in the new one.

Launch the subissions search and try to search by the field you removed with the query builder, you should be able to select an option.

## Print screen / video

Screenshot 2022-11-14 at 11 33 33

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/instances/compare/hooks/useGetInstanceLogs.ts
+++ b/hat/assets/js/apps/Iaso/domains/instances/compare/hooks/useGetInstanceLogs.ts
@@ -66,23 +66,19 @@ export const useGetInstanceLogDetail = (
 };
 
 const getVersion = (
-    versionId: string | undefined,
-    formId: number | undefined,
+    formId: string | undefined,
 ): Promise<Record<string, any>> => {
-    return getRequest(
-        `/api/formversions/?version_id=${versionId}&form_id=${formId}&fields=descriptor`,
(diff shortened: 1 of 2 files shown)


codeKameleon (approved):

quang-le (approved):
Works fine for me. Maybe we should have a place to keep forms that can be used for testing and such stuff. A Google Drive or something. I'll check it out

hakifran:
@beygorghor  I'm getting this error when trying to test this PR:
Screenshot from 2022-11-17 19-24-53 . Any Idea of what happenned?

PR #166, opened 2022-11-14
POLIO-682  / IA-1558: Version info and hide infos on mobile

beygorghor:
POLIO-682: Username was visible hovering the search icon on the budget page
IA-1558: Adding version number on the same place

- [ ] Did I add translations
- [ ] Are there enough tests
## Changes

- hide username on mobile devices, also hides search icon on desktop
- display `IASO_VERSION` in the topbar
- remove username from the siderbar

## How to test

Simply log in and check the topbar, especially on budget page

## Print screen / video
Screenshot 2022-11-14 at 11 52 50
Screenshot 2022-11-14 at 11 52 36

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/app/components/SidebarMenuComponent.js
+++ b/hat/assets/js/apps/Iaso/domains/app/components/SidebarMenuComponent.js
@@ -12,6 +12,8 @@ import {
     Divider,
     Typography,
     Tooltip,
+    useMediaQuery,
+    useTheme,
 } from '@material-ui/core';
 
 import ArrowBackIcon from '@material-ui/icons/ArrowBack';
@@ -115,6 +117,8 @@ const SidebarMenu = ({
     const { plugins } = useContext(PluginsContext);
     const defaultSourceVersion = getDefaultSourceVersion(currentUser);
     const menuItems = getMenuItems(currentUser, plugins, defaultSourceVersion);
+    const theme = useTheme();
+    const isMobileLayout = useMediaQuery(theme.breakpoints.down('md'));
 
     return (
         <Drawer anchor="left" open={isOpen} onClose={toggleSidebar}>
@@ -141,6 +145,16 @@ const SidebarMenu = ({
             </List>
             <Box className={classes.user}>
                 <LanguageSwitch />
+                {isMobileLayout && (
+                    <Typography
+                        variant="body2"
+                        color="textSecondary"
+                        className={classes.userName}
+                    >
+                        {currentUser.user_name}
+                    </Typography>
+                )}
+
                 {currentUser.account && (
                     <Typography
                         variant="body2"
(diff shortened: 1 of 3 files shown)


hakifran (approved):
Good!

hakifran:
But just one suggestion, would it possible when on mobile to display the user on sidebarmenu and when it's on a desktop to display the username on the topbar.

codeKameleon (approved):

plugins/polio/js/src/pages/Budget/index.tsx
```diff
@@ -95,17 +95,19 @@ export const BudgetList: FunctionComponent<Props> = ({ router }) => {
...
-                    top={theme.spacing(2)}
-                    right={theme.spacing(2)}
-                >
-                    <SearchIcon
-                        onClick={() => {
-                            setExpand(value => !value);
-                        }}
-                    />
-                </Box>
+                {isMobileLayout && (
```
quang-le, line 98: This surprises me, I thought it was already conditional to the layout type

hat/assets/js/apps/Iaso/domains/app/components/SidebarMenuComponent.js
```diff
@@ -141,13 +141,6 @@ const SidebarMenu = ({
             </List>
             <Box className={classes.user}>
                 <LanguageSwitch />
-                <Typography
```
quang-le, line 144: I understand we don't really need it there anymore since it's in the top bar, but then when we switch to mobile layout (eg budget) we don't display the username anywhere.

plugins/polio/js/src/pages/Budget/index.tsx
beygorghor: yes don't get it, it was displayed on dekstop...

beygorghor:
> But just one suggestion, would it possible when on mobile to display the user on sidebarmenu and when it's on a desktop to display the username on the topbar.

Yes and no, only budget pages are used in mobile, all the rest of the app is not working at all on mobile

olethanh:
it's not a bad idea but a real version number  in prod look like this `v1.174b-142-ge762ac0f5`. I fear users will panic a bit if they see it displayed so prominently.

olethanh (changes requested):
discussed IRL with Christophe, It's too long, we need to find another way

hat/assets/js/apps/Iaso/domains/app/components/SidebarMenuComponent.js
beygorghor: Username is displayed on mobile in the sidebar

beygorghor:
> it's not a bad idea but a real version number in prod look like this `v1.174b-142-ge762ac0f5`. I fear users will panic a bit if they see it displayed so prominently.
Screenshot 2022-11-18 at 10 47 01
 Better now @olethanh  ?

olethanh:
yes better

quang-le (approved):

PR #167, opened 2022-11-14
POLIO-681 Mark node as completed when there is an alternative path

olethanh:
When we use a diamond to modelize that  a double approval is neeeded, add a way to mark the 'skipped' path as completed
 We do this by adding a mark_nodes_as_completed property on Node

- [] Did I add translations
- [x]  Are there enough tests

## Changes

- Add a `mark_nodes_as_completed` property in the workflow definition. When we pass by a node that has this property mark all the node listed (in the same category) as completed at the same time
- Re active the widget
- Fix crash in it.
- Fix calculation for Active and completed Category

## How to test

See wiki page on wikijs

The workflow definition in the fixture has been updated (for GPEI consolidation) so if you update your database with it you should have a correct workflow.

### DIFF
--- a/plugins/polio/budget/workflow.py
+++ b/plugins/polio/budget/workflow.py
@@ -36,6 +36,9 @@ class Transition:
 class Node:
     label: str
     key: str
+    # Hack for the widget. So we can mark node as completed when passing via other field
+    # must be in the same category
+    mark_nodes_as_completed: List[str] = field(default_factory=list)
     category_key: str = ""
     order: int = 0
     mandatory: bool = False
--- a/plugins/polio/js/src/pages/Budget/BudgetDetails/BudgetTimeline.tsx
+++ b/plugins/polio/js/src/pages/Budget/BudgetDetails/BudgetTimeline.tsx
@@ -1,12 +1,12 @@
 import { Box, makeStyles, Divider } from '@material-ui/core';
-import React, { FunctionComponent, useState, useEffect } from 'react';
+import React, { FunctionComponent, useState, useEffect, useMemo } from 'react';
 import classnames from 'classnames';
 import Stepper from '@material-ui/core/Stepper';
 import Step from '@material-ui/core/Step';
 import StepLabel from '@material-ui/core/StepLabel';
 import { CheckCircleOutline } from '@material-ui/icons';
 import moment from 'moment';
-import { findLast } from 'lodash';
+import { findLastIndex } from 'lodash';
 import { Categories } from '../types';
 
 type Props = {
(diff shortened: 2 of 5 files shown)


plugins/polio/js/src/pages/Budget/BudgetDetails/BudgetTimeline.tsx
```diff
@@ -85,21 +86,16 @@ const getColor = category => {
...
                 category => category.completed === false,
             );
-
-            const lastStepCompletedIndex =
-                categories.indexOf(lastStepCompleted);
-            setActiveStep(lastStepCompletedIndex);
         }
-    }, [categories, activeStep]);
+        return 0;
+    }, [categories]);
```
quang-le: nice !

plugins/polio/js/src/pages/Budget/BudgetDetails/BudgetTimeline.tsx
```diff
@@ -114,8 +110,9 @@ export const BudgetTimeline: FunctionComponent<Props> = ({
                             classes.step,
                         )}
                         key={category.key}
-                        completed={category.completed}
-                        active={category.active}
+                        // The widget automatically recalculate it so not needed
+                        // completed={category.completed}
+                        // active={category.active}
```
quang-le, line 113: I think we can just keep a comment and delete the commented props

quang-le: nitpick

plugins/polio/budget/workflow.py
```diff
@@ -32,6 +32,9 @@ class Transition:
 class Node:
     label: str
     key: str
+    # Hack for the widget. So we can mark node as completed when passing via other field
+    # must be in the same category
```
quang-le, line 40: This will not work for final approval I think

quang-le (changes requested):
I'm afraid limiting the scope of `mark_nodes-as_completed` the the nodes in the same category will be too limiting. For example, final  approval for polio budgets is in another category as the steps/nodes that lead to it.

Also, when a feedback is requested on the last step of a category, I think the category should not appear as completed, eg: RRT asks feedback to GPEI coordinator after `GPEI consolidation` -> the `Submit to RRT` step has be redone, so it the previous category should not appear as complete

Other than that it's very neat. It means also that we must be careful with the config or the widget may display wrong information, but I guess it would have been true of any solution that decouples the widget state from the workflow state

olethanh:
As discussed in meeting. We have decided on modifying the definition to resolve the last issues raised. We will move "Approval" out of it's own category. We will also move the feedback loop to new nodes which will be in the same category  as for approval.

quang-le:
> As discussed in meeting. We have decided on modifying the definition to resolve the last issues raised. We will move "Approval" out of it's own category. We will also move the feedback loop to new nodes which will be in the same category as for approval.

Which can all be handled by modifying the json config

quang-le (approved):

PR #169, opened 2022-11-15
POLIO-646 & POLIO-647 Attach file to email, use permanent URL

olethanh:
The URL generated for file that we used for quick transition and in e-mail where expiring so  they were unusable after 15 minutes. We now use a endpoint in iaso to act as a mediator that will redirect to the destination file and that we can share.

Attach file directly to the e-mails if they are less than 500k (the limit was set arbitrary for now on a guess on what server would always accept). Show a message in the template if some file were not attached.

Compared to the initial demand  we do not handle rejected e-mail !

- [] Did I add translations
- [NA ] My migrations file are included
- [] Are there enough tests. Could use more

## Changes

- Add `/file` endpoint on budget step
- add a get_absolute_url on model and a `permanent_url` on API from BugetStepFile to that url.
- Use that url in quickTransition
- Use that url

## How to test
- See wiki page on Budget. Properly set up your mail. Check using SMTPBucket
- Test making a step with one or more files attached.
- Check that there is a mail with the file
- Try the quick transition button

Check again with a file > 1 mb and a smaller file.

### DIFF
--- a/hat/assets/js/apps/Iaso/libs/utils.tsx
+++ b/hat/assets/js/apps/Iaso/libs/utils.tsx
@@ -2,12 +2,12 @@
  * Get request with params passed as query params
  * Remove undefined params
  * @param {string} url
- * @param {{[p: string]: T}} params
+ * @param {{[p: string]: T}} urlParams
  */
 // url should include closing slash
 export const makeUrlWithParams = (
     url: string,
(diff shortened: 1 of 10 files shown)


quang-le (approved):
Works like a charm on my machine. This tool is getting better and better

PR #171, opened 2022-11-15
IA-A634: improved completeness stats table

niconoe:
Improvement to make the completeness stats table more usable:

- show "%" character next to ration
- fixed a bug in ratio calculation
- display the total number of instances to fill foreach row

Screenshot_20221115_153956


quang-le (approved):
Given we're working in quick iterations and there's no client impact, I will approve this PR but we should:
- Clarify whether we can have org units with forms attached but no forms expected, even by mistake on the client's side
- If so decide how to display that info knowing that: we can't divide by 0 in the backend and we can't show 1 if the actual value is 0 in the front end

Otherwise nothing to complain about in the code, well done

PR #177, opened 2022-11-16
Fix for completeness stats api

madewulf:
Fixing the behaviour of the completeness stat api. 

- [ x] Is my code clear enough and well documented
- [ ] Are there enough tests


niconoe (approved):
Since it's a net improvement over the existing and we're working incrementally, I approve this so it can be deployed on staging. Once feature stabilised, I'll write tests to ensure we don't break the logic later.

PR #180, opened 2022-11-17
Ia 1639 fix cypress ci

quang-le:
Cypress binary was missing when running cypress manually in the CI

- [NA] Did I use eslint and black formatters
- [NA] Did I add translations
- [NA] Are there enough tests
## Changes

- Replaced the caching of node_modules as per [guidelines](https://docs.cypress.io/guides/continuous-integration/introduction#Caching)
- Cached cypress binary and `.npm` folder as in [recommended example](https://github.com/cypress-io/cypress-example-kitchensink/blob/master/.github/workflows/single.yml)

## How to test
Run Iaso manual cypress testing in Github actions

## Print screen / video

NA

## Notes

I just fixed the missing binary issue. the tests still take a long time (20min+), but this can probably be optimized further

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -6,8 +6,8 @@ on:
   workflow_dispatch: {}
   
 concurrency: 
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: false
+  group: ${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
 
 env:
   REGISTRY: ghcr.io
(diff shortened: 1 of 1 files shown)


.github/workflows/cypress.yml
```diff
@@ -119,22 +119,31 @@ jobs:
           node-version: '14'
           cache: npm
 
-      - name: Cache dependencies
```
qgerome: You need to keep this as by default the node_modules are not cached. You also need to keep the `id: cache` to be able to check it in the `npm ci` step.

.github/workflows/cypress.yml
```diff
@@ -119,22 +119,31 @@ jobs:
...
-        with:
-          path: ./node_modules
-          key: node_modules-${{ hashFiles('package-lock.json') }}
-
       - name: upgrade npm
         run: |
           npm install --global npm@8.5
+              # Restore the previous NPM modules and Cypress binary archives.
+
+      # source: https://github.com/cypress-io/cypress-example-kitchensink/blob/master/.github/workflows/single.yml
```
qgerome: This source is quite old, a lot has changed since then :p 
using `cache: npm` on the `setup/node` step will already cache the `npm cache`.

qgerome (changes requested):

codeKameleon:
@cypress

qgerome:
@quang-le I've pushed the correct fix. It worked here: https://github.com/qgerome/iaso/actions/runs/3513836648

quang-le:
> @quang-le I've pushed the correct fix. It worked here: https://github.com/qgerome/iaso/actions/runs/3513836648

I'm curious as to why it didn't work. It worked when I tested it on Friday

beygorghor:
@cypress

qgerome:
It has to be merged to work.Quentin Gérôme Let’s Build It SRLLe 21 nov. 2022 à 13:23, github-actions[bot] ***@***.***> a écrit :﻿
Test failed: https://github.com/BLSQ/iaso/actions/runs/3514318230

quang-le:
It works when you trigger the workflow manually (as did the config I pushed on Friday), so I'm merging this

PR #183, opened 2022-11-18
IA-1640: Allow multiple OU types selection in the completeness_stats table

niconoe:
When using the new completeness table, selecting multiple OU types (which was doable in the FE) triggered an error. 

## How to test

- Go to the completeness table
- Make sure you can filter by OU types (either no selection, one OU type or multiple OU types)


quang-le:
> When using the new completeness table, selecting multiple OU types (which was doable in the FE) triggered an error.
> ## Self proof reading checklist
> 
>     * [x]  Did I use eslint and black formatters
> 
>     * [x]  Is my code clear enough and well documented
> 
> 
> ## How to test
> 
>     * Go to the completeness table
> 
>     * Make sure you can filter by OU types (either no selection, one OU type or multiple OU types)

Please put the Jira ticket number in the PR title, it makes it easier to find

quang-le (changes requested):
When I use the OU types filter locally, I never get a response from the backend. Other filters seem to work fine

niconoe:
@quang-le: I've had a look at the performances issues and noted that the response is indeed very slow if we request an orgUnit type at the bottom of the tree (=> many results). This is because the pagination is implemented fully in Python and not at the database.  I've created a new issue for this (IA-1655) since this is not directly related to this PR, but only revealed by this. 

Therefore, I suggest approving/merging this PR (if all the rest looks ok to you) and continue the performance work in IA-1655. Does that makes sense?

quang-le:
> @quang-le: I've had a look at the performances issues and noted that the response is indeed very slow if we request an orgUnit type at the bottom of the tree (=> many results). This is because the pagination is implemented fully in Python and not at the database. I've created a new issue for this (IA-1655) since this is not directly related to this PR, but only revealed by this.
> 
> Therefore, I suggest approving/merging this PR (if all the rest looks ok to you) and continue the performance work in IA-1655. Does that makes sense?

It depends on how dependent the next steps of the completeness feature depend on this. It's really unusable in the UI. We should then at least hide it in the meantime, because if it's slow on my machine, it may just as well end up crashing in prod

mathvdh (approved):
Je suis pas trop sûr de comprendre les edge cases mais ça me semble bien aussinon ...

PR #188, opened 2022-11-22
IA-1649: update npm version dockerfile

codeKameleon:

the docker build config didn't install the right version of npm

- [ ] Did I add translations
- [ ] Are there enough tests
## Changes

- install npm 8.5 instead of npm 7
- remove node modules and forcing cache clean

### DIFF
--- a/docker/bundle/Dockerfile
+++ b/docker/bundle/Dockerfile
@@ -18,6 +18,10 @@ COPY package-lock.json .
 
 
 RUN npm install -g npm@8.5
+
+RUN rm -rf node_modules
+RUN npm cache clean --force
+RUN npm cache verify
 # ci install the exact version of package, compared to install which might a bump minor version
 RUN npm ci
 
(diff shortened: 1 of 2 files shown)


docker/bundle/Dockerfile
```diff
@@ -18,6 +18,10 @@ COPY package-lock.json .
 
 
 RUN npm install -g npm@8.5
+
+RUN rm -rf node_modules
```
quang-le, line 22: I don't understand why we do this. `npm ci` deletes the node_modules anyway. cf [https://docs.npmjs.com/cli/v6/commands/npm-ci](https://docs.npmjs.com/cli/v6/commands/npm-ci)

docker/bundle/Dockerfile
```diff
@@ -18,6 +18,10 @@ COPY package-lock.json .
...
+RUN npm cache clean --force
+RUN npm cache verify
```
quang-le, line 24: It seems to me these commands are mutually exclusive: if we clean the cache there's  nothing to verify, and if we want to "heal" the cache with `verify` there's no point in deleting it

quang-le (commented):
Not sure I want to block over this, but it may save use some build time to remove useless commands

docker/bundle/Dockerfile
codeKameleon: Indeed, i think we can remove it @beygorghor  ?

docker/bundle/Dockerfile
codeKameleon: yes ok, we remove the command `npm cache verify` ?

olethanh:
err for the modification on the bundle/Docker that feels like a code Smell, can someone justify theses modification and why they are here ?

PR #194, opened 2022-11-23
Wc2 107 auto blacklist old nfc cards

niconoe:
New feature for NFC card management:

Given the backend receives a new “WRITE_PROFILE” event
When the entity associated with that event has other cards associated to them
Then those other cards are automatically blacklisted with : 
 - the reason "OTHER" 
 - and the comment "Profile was written on {NEW NFC CARD ID}"

- [x] Are there enough tests

## How to test

Mobile-related endpoint, so difficult to test, two solutions come to mind:

- check the newly added tests in `test_storage.py` make sense and work
- for @bmonjoie: test the interactions directly from the mobile app

### DIFF
--- a/iaso/api/storage.py
+++ b/iaso/api/storage.py
@@ -309,7 +309,9 @@ class StorageLogViewSet(CreateModelMixin, viewsets.GenericViewSet):
 
                 performed_at = timestamp_to_utc_datetime(int(log_data["performed_at"]))
 
-                concerned_instances = Instance.objects.filter(uuid__in=log_data["instances"])
+                concerned_instances = Instance.objects.none()
+                if "instances" in log_data:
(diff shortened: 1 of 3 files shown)


bmonjoie (approved):

iaso/api/storage.py
```diff
@@ -309,7 +309,9 @@ def create(self, request):
 
                 performed_at = timestamp_to_utc_datetime(int(log_data["performed_at"]))
 
-                concerned_instances = Instance.objects.filter(uuid__in=log_data["instances"])
+                concerned_instances = Instance.objects.none()
```
madewulf, line 312: Not very important, but I think concerned is not the exact word (relevant could work) . Concerned is probably a translation of "concerné"

niconoe: You're absolutely right!

PR #203, opened 2022-11-25
Seed command, delete tenant, copy s3 content for an account, local hosting s3 assets

mestachs:
Super seed this PR : https://github.com/BLSQ/iaso/pull/79 and allows 
  - hosting static assets (js/..) in a minio (allowing to override the s3 url endpoint) via `AWS_S3_ENDPOINT_URL` env variable
  - put sql logs at debug with env variable DEBUG_SQL=true  
       - this allowed me to understand queries run by workers
       - fine tune sql to prevent loading the whole db in memory `where id (select id from ...)` vs `where id in (1,  2 ,3,... 1000000)`

See information shared on slack on how to produce a local hosting (db dump and s3 zip file) and restore it on a machine provisioned by https://github.com/BLSQ/ops-local-hosting

# Notes 

There's risk that I broke asset deployment for production settings (s3) 
if someone can deploy this on staging for testing before merging that would be great.


olethanh:
Haven't read the whole code yet as it quite a beast, tested by 
1. creating a new database,
2. Seeding it with ` docker-compose run --rm iaso manage seed_test_data --mode=seed --dhis2version=2.38.2.1`
3. Creating a new Account (which got the id `2`)
4. Running the command with: `docker-compose run --rm iaso manage delete_accounts --account-to-keep 2`

I still have User, Profile, Account on the wrong account and the external credential. I don't see any polio related stuff but from what I understand it is out of scope. 
Haven't tested the media migration logic yet. The config part looks ok.
Had to add a try catch because I didn't have the table `users_profile` on a fresh db. Made  a commit and pushed it

olethanh:
following discussion it was probably a manipulation error on my part. 
Device are not removed but we can adress it later.
Settings seems fine to me.

Added a commit to remove a warning in docker. Merged with main

PR #207, opened 2022-11-28
IA-1629 login  user name is not case senstive

SimonDus:

- [x] Are there enough tests
## Changes

When creating a new user, to check if the unsername was already existing the check was made on the exact match of the username. So if there was a capital letter or Uppercase etc, the profile API was accepting the username despite the fact that for django backend it would be a duplicate username. So I changed the check and used iexact to ensure the username doesn't exists in a variation ( case insensitive ).

https://user-images.githubusercontent.com/45785235/204291377-8a8cddbd-06c2-4528-9898-f5849285c25f.mov

### DIFF
--- a/iaso/api/profiles.py
+++ b/iaso/api/profiles.py
@@ -1,7 +1,7 @@
 from django.contrib.auth import update_session_auth_hash
 from django.contrib.sites.shortcuts import get_current_site
 
-from rest_framework import viewsets, permissions
+from rest_framework import viewsets, permissions, serializers
 from rest_framework.response import Response
 from django.shortcuts import get_object_or_404
 from django.core.paginator import Paginator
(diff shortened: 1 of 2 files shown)


mathvdh (commented):
Est-ce qu'on a vraiment besoin d'une regex ?

on peut pas utiliser `username__iexact` ou `username__icontains` ??

SimonDus:
> 

icontains marche pas c'est pourquoi j'ai utilisé un regex. Mais effectivement iexact fonctionne aussi, je n'y avais pas pensé.
Les deux fonctionnent.

mathvdh (approved):
Je suis toujours pour eviter une regex quand c'est possible ;)

niconoe (approved):
Thanks a lot @SimonDus, great work!

PR #210, opened 2022-11-29
IA-1681: Completeness stats API: parameter renamed

niconoe:
For the completeness stats table, there was a badly named filter parameter: `parent_id` was used to select the "current" OrgUnit, while the name suggested that the returned rows would actually be the children of the OU.

Since we'll need to implement another feature for the "real" parent_id soon, it's important to rename the existing one to avoid adding more confusion to the confusion.

- [x] Are there enough tests

## How to test

1) Go to the completeness stats table
2) Make sure the "OrgUnit" filter still works


olethanh (commented):
LGTM backend-wise

quang-le (approved):

PR #217, opened 2022-11-30
IA-1686: fix deep linking

quang-le:
Using deep linking with OU type filter n completeness stats crashed the page

- [ ] Did I add translations
- [ ] Are there enough tests
## Changes

- Add a null check on `ouType` to avoid crash when using deep linking

## How to test

Go to completeness stats, make a search with all 3 filters, copy the url, open it in a new. It should not crash and the filters should have the correct values

## Print screen / video

https://user-images.githubusercontent.com/38907762/204823822-ebccf19c-5be0-4ea5-a4ec-88e1ccc18642.mov


codeKameleon (approved):

niconoe (approved):
Thanks a lot @quang-le, works great on my machine!

PR #218, opened 2022-11-30
IA-1644/IA-1688: frontend grid and responsive behavior refactor

codeKameleon:
There was a lack of consistency in the design rules. Sometime the search button is under the filters, sometimes aside the filters

- [ ] Did I add translations
- [ ] Are there enough tests
## Changes

- used xs={12} if on mobile/tablet
- if there are <= 3 columns in filters, the search button is aligned on the same level 
- if there  are more then 3 columns, the search button is aligned below the filters
- used `useTheme` and `useMediaQuery`  to improve/fix design issues on mobile

## How to test

Go through the entire app, see if there are no design issues on desktop, tablet and mobile

## Print screen / video
Capture d’écran du 2022-12-01 17-36-58
Capture d’écran du 2022-12-01 17-36-15
Capture d’écran du 2022-12-01 17-36-03


codeKameleon:
@cypress

hat/assets/js/apps/Iaso/domains/users/components/Filters.js
```diff
@@ -115,44 +125,25 @@ const Filters = ({ baseUrl, params }) => {
...
-                        onChange={handleChange}
-                        value={filters.orgUnitTypes}
-                        type="select"
-                        options={orgUnitTypeDropdown}
-                        label={MESSAGES.orgUnitTypesDropdown}
-                        loading={isFetchingOuTypes}
-                        onEnterPressed={handleSearchPerms}
-                        clearable
-                    />
-                </Grid>
```
beygorghor: this one should not be removed !

hat/assets/js/apps/Iaso/components/tables/TableFilters.js
```diff
@@ -61,18 +62,21 @@ const Filters = ({
         }
         onSearch(tempParams);
     };
+
+    const runsBaseUrl = 'orgunits/sources/links/runs';
```
beygorghor: please use `baseUrls.algos` ! Also i would recommend  to use an array of urls were to use 3 columns.

beygorghor: `TableFilters` is used by `SingleTable` a bit everywhere, it could ba a good solution also to drill down columns number through SingleTable and use 4 or 3 by default.

beygorghor (changes requested):
Nice job!  it's much more coherent between pages now.
Be careful you removed orgunit type filter on users page !
Also have a look in table filters, i would go for the second option !

beygorghor (approved):

PR #229, opened 2022-12-05
IA-1669: Edit org unit: filter groups for ou source 

beygorghor:
In org unit details: When showing an org unit from a given source, proposed groups should be from the same source

- [ ] Did I add translations
- [ ] Are there enough tests
## Changes
Use current org unit source id to filter groups while editing an org unit.

## How to test

Edit an org unit and check that the list of groups are exclusively from the same source as the org unit.

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/orgUnits/hooks.js
+++ b/hat/assets/js/apps/Iaso/domains/orgUnits/hooks.js
@@ -1,4 +1,4 @@
-import { useState, useEffect } from 'react';
+import { useState, useEffect, useMemo } from 'react';
 import {
     useSnackQuery,
     useSnackMutation,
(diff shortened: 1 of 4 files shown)


codeKameleon (approved):

hat/assets/js/apps/Iaso/domains/orgUnits/hooks.js
```diff
@@ -42,12 +52,15 @@ export const useOrgUnitDetailData = (
             queryFn: () =>
                 getRequest(
                     `/api/groups/${
-                        isNewOrgunit ? '?&defaultVersion=true' : ''
+                        isNewOrgunit
+                            ? '?&defaultVersion=true'
+                            : `?&dataSource=${originalOrgUnit?.source_id}`
```
quang-le: if source_id is null or undefined you might send an incorrect query

quang-le (changes requested):
just a small edge case

quang-le (approved):

PR #240, opened 2022-12-12
Ia 1741 missing translation

quang-le:
There was no label for the org unit type filter at `/dashboard/orgunits/sources/links/list`

- [ ] Did I add translations
- [ ] Are there enough tests
## Changes

- Passed a `labelString` to the filter since the `label` prop didn't seem to function as expected

## How to test

Go to the page, see if the label is there

## Print screen / video

Screenshot 2022-12-12 at 16 47 47

## Notes

If the PR depends on a PR in bluesquare-components, or merges into another PR (i.o. main), it should also be mentioned he

### DIFF
--- a/hat/assets/js/apps/Iaso/constants/filters.js
+++ b/hat/assets/js/apps/Iaso/constants/filters.js
@@ -638,7 +638,6 @@ export const linksFilters = props => {
         fetchingAlgorithms,
         fetchingSources,
     } = props;
-    console.log('OUTYPES', orgUnitTypes);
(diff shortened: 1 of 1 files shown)


codeKameleon (approved):

beygorghor (changes requested):
It looks like we have the same issue on Orgunit details / children page, can you have a look please ?
Screenshot 2022-12-14 at 09 04 05

beygorghor (approved):

PR #243, opened 2022-12-13
Add more test on api for enketo

olethanh:
Add a test for formList and formDownload. Check that it work with duplicate instance. Use a real xml form so we can check that the id are correctly replaced I also added the source xlsform in the fixture so it can be used when testing manually


iaso/tests/api/test_enketo.py
```diff
@@ -387,3 +388,153 @@ def test_public_create_url_non_single_create_2(self):
         # self.assertEqual(responses.calls[0].request.url, 1)
         self.assertTrue(responses.assert_call_count("https://enketo_url.host.test/api_v2/survey/single", 1))
         self.assertEqual(old_count + 1, Instance.objects.count())
+
+    def test_form_list_work_with_duplicate_instance(self):
+        "Check form list work when there is two instance with the same uuid"
```
niconoe: Typos: `... when there are two instances with the same UUID`

niconoe (changes requested):
Great, thanks, I just left a minor comment for typos in a docstring

PR #249, opened 2022-12-15
IA-1756: Completeness stats table: make the org unit type filter a real filter

niconoe:
A real filter=something that filters out rows rather than change the row selection logic.

Hopefully that would make the system simpler, (IA-1748) the interface clearer and would circumvent the current performance issues.

- [x] Are there enough tests

## How to test

1) go to the completeness table
2) filter: select a parent org unit (example: country) 
3) filter: select the org unit type directly below country (example: region), make sure all regions below this country are still returned
4) change org_unit_type filter to a much lower level such as health center => there should now have ) results since health centers are not direct children of the country (**previous/incorrect behavior was: load all health centers that are indirect children of the country and create performance issues**)


quang-le (commented):
Seems OK to me. You might want to merge main for easier testing since the limitation on OU type filter has been removed.

Try asking @olethanh for a review if @madewulf is out of time

niconoe:
Thanks @quang-le. Main is indeed already merged for the reason you mention!

PR #263, opened 2022-12-22
IA-1780: fix columns order

codeKameleon:
Columns are not well ordered. Org unit should be the first column


beygorghor (changes requested):
Like we discussed, what is the default order of this table, can we sort it by default on orgunit please, or at least highlight the default sort

codeKameleon:
@hakifran  @beygorghor Nicolas suggested to do this in  a separate PR : https://bluesquare.atlassian.net/browse/IA-1780?focusedCommentId=57158

hakifran (approved):

PR #269, opened 2022-12-23
POLIO-766 connect override api to front end

quang-le:
Create front-end to use budget override API

- [X] Did I add translations
- [ ] Are there enough tests Ha ha, no
## Changes

- Add api hook to the new endpoint
- Add `CreateOverrideStep` based on `CreateBudgetStep`. Made all eisting fields optional and added mandatory `new_state_key`
- Add validation for `CreateOverrideStep` based on the existing one fro `CreateBudgetStep` 
- Merged main and fixed migration issue
- Fixed TS typing of Api.ts

## How to test

In a Db [with a budget workflow](https://wiki.bluesquare.org/en/Dev-team/product-management/iaso/polio_budget_workflow)
Log as a super user
Go to a budget
Use the Override Button

## Print screen / video

https://user-images.githubusercontent.com/38907762/209355707-cd4d7423-0fd5-4816-a79c-14e58511a749.mov

## Notes

This is a 0.1 version. We still need an API endpoint to be able to display all the possible states in a dropdown list. When this is done, the text Input for `new_state_key`  should be replaced with a single select component.

There's also some clarification needed as to which emails should be sent to which teams. This will impact the front-end, since the `recipient_teams` remain empty for now


olethanh:
actually we forgot but we already had a endpoint to list the possible state, it 's the one we use for filtering :D Will update the pr

PR #270, opened 2022-12-28
IA-1749: adding helperText prop to TextArea component

codeKameleon:
Adding a helperText prop to display a message below the textarea if it's disabled because required inputs need to be filled first

- [ ] Did I add translations
- [ ] Are there enough tests
## Changes

- add a helperText prop to TextArea and in MultilineText where TexttArea is used.


beygorghor (changes requested):
Can you add a error prop like it's done in `PeriodPicker` please, then we can put the helper Text in red when in error state

beygorghor (approved):

PR #281, opened 2023-01-09
POLIO-630 Fix timeout on calendar map merged shape- precalculate shape at update

olethanh:
Precalculate the geojson for each campaign at update time instead of when calling the API

- [NA] Did I add translations
- [?] Are there enough tests

## Changes

- Added `geojson` GeoJson field  on Campaign with it's shape (well a list of)
- Also rigged the admin so it's recalculate when updating the model from there
- and added a Admin action to recalculate shapes
- Added a migration

## How to test

With polio setup
Try to change the scope a campaign, multiple time
See if it is still displayed in the Calendar
Also in Anonymous/embedded mode


plugins/polio/api.py
```diff
@@ -471,6 +471,7 @@ def return_cached_response_if_valid(cache_key, update_dates):
         url_path="v2/merged_shapes.geojson",
     )
     def shapes_v2(self, request):
+        "Deprecated, should return the same format as shapes v3, kept for comarison"
```
quang-le: typo 😁

plugins/polio/serializers.py
```diff
@@ -685,7 +689,10 @@ def update(self, instance: Campaign, validated_data):
 
     class Meta:
         model = Campaign
-        fields = "__all__"
+        # TODO in the future specicy the fields we actually want
```
quang-le: typo lol

quang-le (commented):
Seems roughly 2x faster on my machine 😀

olethanh:
fixed the typois

beygorghor:
@olethanh you have a migration conflict here

olethanh:
I dont' understand how that happened since the other file was not even in my branch, seems like a strange github cache problem. Anyway fixed it my merging main and merging migrations

mathvdh (approved):
Il n'y a rien qui me choque mais je ne connais pas bien geojson ou les campaigns .. ?

niconoe (approved):
Same for me, not easy to understand all the global implications but the code looks fine IMHO.

PR #297, opened 2023-01-12
IA-1824  Fix Table Cell fixing the whole page

beygorghor:
Add an error boundary which show the raw value and an error in the console.

It helped me recently fix a bug I had in the Tasks table, which was a rendering bug in the launcher column (cf IA-1247)

- [ ] Did I add translations
- [ ] Are there enough tests
## Changes

using new version uf bluesquare-components

## How to test

Return null in one column for a cell

## Print screen / video

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5996,7 +5996,7 @@
         },
         "node_modules/bluesquare-components": {
             "version": "0.2.0",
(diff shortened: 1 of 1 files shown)


beygorghor:
@cypress

PR #300, opened 2023-01-13
Ia-1828: completeness stats api refactoring

niconoe:
**branch based on [IA-1781](https://github.com/BLSQ/iaso/pull/298) and [IA-1788](https://github.com/BLSQ/iaso/pull/291) so the code will be smaller/easier to review after those two are merged.**

Some refactoring and documentation additions for the completeness stats API. Ne change of behavior expected.

- [x] Are there enough tests

## How to test

No change of behavior expected: play with the completeness table and check if it still works as before. (there's good test coverage for this API)

## Notes

It's quite difficult to work with the test of this API without having a good overview of the test data, which is complex and spread in two different places. I therefore added a .png file with a diagram. Unfortunately I don't think the generation of this file can be easily automated.


quang-le:
Note: you can change the base of your branches so they are safer to merge. Also, the diff would be based on the base branch, so it wouldn't impede/slow down review.

PR #303, opened 2023-01-13
Aggregate handle dhis 2.38 exceptions

mestachs:
Dhis2 2.38 changed the http code in case of imports with warnings

## Changes

The good news is that the current code was already a bit considering this as an exception
I just had to adapt a bit the parsing of the conflicts field because the payload changed a bit too.

## How to test

- seed on a 2.38 version : 
`docker-compose run --rm iaso manage seed_test_data --mode=seed --dhis2version=2.38`
- test the export of a single submission of the quantity form
- then change in dhis2 the data element type from Number to Integer (https://play.dhis2.org/2.38.2.1/dhis-web-maintenance/index.html#/edit/dataElementSection/dataElement/nymNRxmnj4z) 
- modify the submission json via django admin to have decimals (something like 
  via something like http://localhost:8081/admin/iaso/instance/56359/change/
image
- test the export of the modified, verify the error messages looks like this
image


olethanh (approved):
test and works for me.
fixed a warning in the frontend that was already there

PR #307, opened 2023-01-16
IA-1782: account id warning

beygorghor:
Use account id that is now in the url to warn user logged into one account that he’s trying to access another account. 

Example: if you are logged into account with id 1 and that you access this url: http://localhost:8081/dashboard/settings/tasks/accountId/2 , you should see a popup telling you:

You are connected to the wrong account, would you like to log out and connect in the correct account to see this link?

- [x] Did I add translations
- [ ] Are there enough tests

## Changes

Create a dialog allowing to logout if account id in the url is not the same as current user account id.

## How to test

Open the dashboard, change accountId in the url to something else, a dialog should appear

## Print screen / video

https://user-images.githubusercontent.com/12494624/212679529-1c1cd330-3659-4413-a576-1fd9ea123474.mov


hat/assets/js/apps/Iaso/domains/app/translations/en.json
```diff
@@ -547,6 +547,8 @@
     "iaso.label.workflow": "Workflow",
     "iaso.label.workflows": "Workflows",
     "iaso.label.workflowVersion": "Workflow version",
+    "iaso.label.wrongAccount": "You are connected to the wrong account, would you like to log out and connect in the correct account to see this link?",
```
quang-le: You are connected to the wrong account, would you like to log out and connect TO the correct account to see this link?

I wonder if "with" wouldn't be even better

hat/assets/js/apps/Iaso/domains/app/translations/fr.json
```diff
@@ -546,6 +546,8 @@
     "iaso.label.workflow": "Workflow",
     "iaso.label.workflows": "Workflows",
     "iaso.label.workflowVersion": "Version du workflow",
+    "iaso.label.wrongAccount": "Vous êtes connecté au mauvais compte, voulez-vous vous déconnecter pour vous connecter avec le bon compte?",
```
quang-le, line 549: Here as well do we connect to an account or with an account?

hat/assets/js/apps/Iaso/domains/users/components/ProtectedRoute.js
```diff
@@ -24,6 +25,15 @@ const ProtectedRoute = ({
...
 
+    const isWrongAccount = useMemo(
+        () =>
+            Boolean(
+                params.accountId &&
+                    params.accountId !== `${currentUser.account.id}`,
+            ),
+        [currentUser.account.id, params.accountId],
+    );
+
```
quang-le: I don't think you need to memoize booleans

hat/assets/js/apps/Iaso/domains/users/components/WrongAccountModal.tsx
```diff
@@ -0,0 +1,100 @@
...
+    content: {
+        overflow: 'visible',
+        paddingBottom: theme.spacing(2),
+    },
+    action: {
+        paddingBottom: theme.spacing(2),
+        paddingRight: theme.spacing(2),
+    },
+}));
+const WrongAccountModal: FunctionComponent<Props> = ({ isOpen }) => {
```
quang-le, line 48: Why not use/edit the modal from bluesquare-components?

quang-le (changes requested):
it works when I mnually change the account, but when I paste a link with a different account id, the app makes the correction the account id of the current user:

https://user-images.githubusercontent.com/38907762/213430091-a74a5a58-4242-426a-acc6-25d698fe5685.mov

Changing this behaviour might increase the scope of the PR more than previously thought

hat/assets/js/apps/Iaso/domains/app/translations/fr.json
codeKameleon: i would say "with" too

codeKameleon (changes requested):
Mainly agree with Son for the changes. 
Just concerning this comment :https://github.com/BLSQ/iaso/pull/307#pullrequestreview-1261391099

I don't have the issue personnaly : 

[recording-2023-01-23-12-35-44.webm](https://user-images.githubusercontent.com/25134301/214030548-d7807e0f-220a-432c-919a-688b04ef9700.webm)

hat/assets/js/apps/Iaso/domains/users/components/WrongAccountModal.tsx
beygorghor: because it's impossible to hide the button

beygorghor:
@quang-le indeed i cannot reproduce it:

https://user-images.githubusercontent.com/12494624/214067897-06ee986e-4e18-4d6e-a25d-08485723d486.mov

on left = chrome with user account 1
on right = firefox with user account 3

codeKameleon (approved):

quang-le (approved):
I can't repro anymore either. Mystery bug...

PR #314, opened 2023-01-19
WC-118: fix back arrow not working on storage details

codeKameleon:
When you changed filters on storage details and then clicked on the back arrow, the navigation dint work well because you were redirect to last url withouth the filters params though the back arrow should redirect to the storages list

- [ ] Did I add translations
- [ ] Are there enough tests
## Changes

- removed router from params
- use `redirectToReplace` instead to force redirect to storage list

## How to test

Go to Storages > choose a element from the list and see detail 
Then change the filters, apply them 
Use the back arrow 
It should lead to the storages list page

## Print screen / video

[recording-2023-01-19-13-07-19.webm](https://user-images.githubusercontent.com/25134301/213439971-495e3295-4e5e-4e78-a907-2ef0bc12e094.webm)

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/storages/components/LogsFilters.tsx
+++ b/hat/assets/js/apps/Iaso/domains/storages/components/LogsFilters.tsx
@@ -28,7 +28,7 @@ type Props = {
 const baseUrl = baseUrls.storageDetail;
 export const LogsFilters: FunctionComponent<Props> = ({ params }) => {
     const { filters, handleSearch, handleChange, filtersUpdated } =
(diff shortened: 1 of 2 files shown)


quang-le (changes requested):
The `prevPath` is supposed to make sure you back to the page you came from. We use that pattern so much that there's a `useGoBack`hook for that now by the way (just FYI). 

This  fix will not work if we link to the details page from somewhere else. 

The problem is that there's a `redirectTo` somewhere that should be a `redirectToReplace`, but it's somewhere else

quang-le (approved):
All good now 👍

PR #316, opened 2023-01-19
Ia-1826: completeness stats table counts are > 100% when multiple submissions to the same ou

niconoe:
This PR improves the counting logic in the completeness stats table.

 if a orgunit has two instances of a given form attached:

- Previously, it was counted as "2 (forms filled ) / 1 (forms to fill)  - 200%".
- Now, it is counted as "1/1 - 100%". A new flag (`has_multiple_direct_submissions`) has been added to the API response to make it clear that there is more than one submission for the OU/form combination (a front-end ticket to display the status of this flag will be created soon)

- [x] Are there enough tests

## How to test

- Make sure the completeness table still works
- Check that the new way to calculate is applied: 2 instance for a given fom/ou combination is reported as 1/1


quang-le (commented):
Looks OK front end side

PR #327, opened 2023-01-23
Polio-741 Make vaccine and forma api more resilient

SimonDus:
/api/polio/vaccines was crashing due to the connexion to WHO servers failing for a couple of countries.  A 403 error was returned resulting into the crash of the request.  A notification email can be automatically send  in case of login error by creating a config into polio under the name **vaccines_emails**.  The content must be an array of emails. 

After checking thousands of responses we can't filter on edited forms as the edited field in the response is all the time set to False. Therefore I filter the request on the last 24 months as discussed. 

Related JIRA tickets : POLIO-741

- [x] Are there enough tests

## How to test

go to /api/polio/vaccines and you should have a response without 500 or 403.


plugins/polio/api.py
```diff
@@ -1095,8 +1096,16 @@ def handle_ona_request_with_key(request, key, country_id=None):
         if country_id is not None and config.get("country_id", None) != cid:
             continue
         forms = get_url_content(
-            url=config["url"], login=config["login"], password=config["password"], minutes=config.get("minutes", 60)
+            url=config["url"] + "?date_created__month__lte=24",
+            login=config["login"],
+            password=config["password"],
+            minutes=config.get("minutes", 60),
         )
+        if not forms:
```
olethanh: I would put the try except at this level so it doesn't affect the other place we could use get_url_content  and more info about the error could be included in the e-mail

plugins/polio/helpers.py
```diff
@@ -14,15 +15,20 @@ def get_url_content(url, login, password, minutes=60):
...
+            cached_response.content = response.text
+            cached_response.save()
+            j = response.json()
+
+        else:
+            j = json.loads(cached_response.content)
+        return j
+    except HTTPError:
+        # Return false in case the WHO server returns an error.
+        return False
```
olethanh: `logging.exception(<exception>)` should be called so it's properly logged in sentry

plugins/polio/api.py
```diff
@@ -1095,8 +1096,16 @@ def handle_ona_request_with_key(request, key, country_id=None):
         if country_id is not None and config.get("country_id", None) != cid:
             continue
         forms = get_url_content(
-            url=config["url"], login=config["login"], password=config["password"], minutes=config.get("minutes", 60)
+            url=config["url"] + "?date_created__month__lte=24",
```
olethanh: Not too sure about the hard coding of this parameter, is this what you want @madewulf  ?

Also is the parameter actually correct? Month smaller than 24 seems strange

olethanh (commented):
@madewulf Can you confirm this is actually the behaviour you expect?

plugins/polio/api.py
SimonDus: Well, I can add the params in the config. It would be more flexible. 
The last 24 months is set as we discussed with Martin

SimonDus: https://api.ona.io/static/docs/data.html#query-submitted-data-of-a-specific-form-using-date-modified

olethanh:
I have updated the branch with main  and changed the behavior a bit (removed the filter on the months as it is not relevant anymore), @mathvdh ,  can you take a look?

mathvdh (approved):
Seems quite straightforward to me ... 
Only question is config["login"] contains the country name then ?
I approved but I'm curious ..

olethanh:
the user name for the country on the who server it think

PR #330, opened 2023-01-24
POLIO-789: fix csv download

quang-le:
download url was incorrect

Related JIRA tickets : POLIO-789

- [ ] Did I add translations
- [ ] Are there enough tests

## Changes

- Changed `useGetCampaigns` so that the `fieldset` parameter is only added to the `query` (and not to `exportToCsv`)
- Changed `exportToCsv` so that it returns the url i.o. handling the download
- Created `PageActionWithLink` which handles the  `exportToCsv`url. This allows the displaying of the url on hover

## How to test

Go to polio campaigns and play with CSV button

## Print screen / video

https://user-images.githubusercontent.com/38907762/214286955-4713e34e-b37c-45b0-88b4-881b9423f866.mov

### DIFF
--- a/plugins/polio/js/src/hooks/useGetCampaigns.js
+++ b/plugins/polio/js/src/hooks/useGetCampaigns.js
@@ -27,14 +27,12 @@ export const useGetCampaigns = (
             // Ugly fix to prevent the full list of campaigns showing when waiting for the value of countries
             enabled: options.enabled ?? true,
             last_budget_event__status: options.last_budget_event__status,
-            fieldset: options.fieldset ?? undefined,
         }),
         [
             options.campaignGroups,
             options.campaignType,
             options.countries,
             options.enabled,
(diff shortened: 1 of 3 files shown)


beygorghor (approved):

plugins/polio/js/src/hooks/useGetCampaigns.js
```diff
@@ -69,16 +67,21 @@ export const useGetCampaigns = (
...
-                ...params,
-                limit: undefined,
-                page: undefined,
-                format: 'csv',
-            })}`),
+        exportToCSV: `${getURL({
+            ...params,
+            limit: undefined,
+            page: undefined,
+            format: 'csv',
```
olethanh, line 74: wouldn't it be more consistent to just override the fieldset here like we do for the page and limit?

PR #336, opened 2023-01-26
IA-1592 Restrict user to set default_source_version to version linked to his …

SimonDus:
When doing a PUT request at   /api/accounts/{the account id}/ with  {"default_version": the_version_id_from_other_account}
it was possible to set a version from another account. It is not longer possible. 

Related JIRA tickets : IA-1592

- [x] Are there enough tests

## How to test

Create two account

Create a Data source and a version inside one of the account.

Log on the other account, make an API call to  PUT /api/accounts/{the account id}/ with 

{"default_version": the_version_id_from_other_account}
What happen

Request fail and the account is not modified. The error received from the API should contain the field and the error.

### DIFF
--- a/iaso/api/accounts.py
+++ b/iaso/api/accounts.py
@@ -2,7 +2,7 @@
 from rest_framework.request import Request
 
 from .common import ModelViewSet, HasPermission
-from iaso.models import Account, SourceVersion
+from iaso.models import Account, SourceVersion, Project
 from rest_framework import serializers, permissions
 from rest_framework.generics import get_object_or_404
 
@@ -55,3 +55,11 @@ class AccountViewSet(ModelViewSet):
     queryset = Account.objects.all()
     # FIXME: USe a PATCH in the future, it make more sense regarding HTTP method semantic
     http_method_names = ["put"]
+
+    def update(self, request: Request, *args, **kwargs):
+        default_version = request.data["default_version"]
+        version = get_object_or_404(SourceVersion, pk=default_version)
+        project = Project.objects.get(account=request.user.iaso_profile.account)
+        if project not in version.data_source.projects.all():
+            raise serializers.ValidationError({"Error": "Account not allowed to access this default_source"})
+        return super().update(request, args, kwargs)
--- a/iaso/api/accounts.py
+++ b/iaso/api/accounts.py
@@ -59,7 +59,7 @@ class AccountViewSet(ModelViewSet):
     def update(self, request: Request, *args, **kwargs):
         default_version = request.data["default_version"]
         version = get_object_or_404(SourceVersion, pk=default_version)
-        project = Project.objects.get(account=request.user.iaso_profile.account)
+        project = Project.objects.get(account=request.user.iaso_profile.account)  # type: ignore
         if project not in version.data_source.projects.all():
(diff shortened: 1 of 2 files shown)


mathvdh (approved):
Seems Good

iaso/tests/api/test_account.py
```diff
@@ -79,8 +81,8 @@ def test_account_set_default_fail_wrong_account(self):
         # invert on the other account/user to be sure
         self.client.force_authenticate(self.john)
         response = self.client.put(f"/api/accounts/{self.ghi.pk}/", {"default_version": self.ghi_version.pk})
-        j = self.assertJSONResponse(response, 403)
-        self.assertEqual(j, {"detail": "You do not have permission to perform this action."})
+        j = self.assertJSONResponse(response, 400)
+        self.assertEqual(j, {"Error": "Account not allowed to access this default_source"})
```
olethanh: this test was checking that an user couldn't modify an account from another client. this is not the case anymore,  and the new check hide the previous error message. 
A new separate test should be used to check the new condition.

iaso/api/accounts.py
```diff
@@ -55,3 +55,11 @@ class AccountViewSet(ModelViewSet):
...
     # FIXME: USe a PATCH in the future, it make more sense regarding HTTP method semantic
     http_method_names = ["put"]
+
+    def update(self, request: Request, *args, **kwargs):
+        default_version = request.data["default_version"]
+        version = get_object_or_404(SourceVersion, pk=default_version)
+        project = Project.objects.get(account=request.user.iaso_profile.account)  # type: ignore
+        if project not in version.data_source.projects.all():
+            raise serializers.ValidationError({"Error": "Account not allowed to access this default_source"})
+        return super().update(request, args, kwargs)
```
olethanh: this should be in the serializer (which already has an update method)

olethanh (changes requested):

iaso/tests/api/test_account.py
SimonDus: Done

iaso/api/accounts.py
SimonDus: Done

iaso/api/accounts.py
```diff
@@ -18,12 +18,16 @@ class Meta:
...
     def update(self, account, validated_data):
         default_version = validated_data.pop("default_version", None)
+        user = self.context["request"].user
         if default_version is not None:
             sourceVersion = get_object_or_404(
                 SourceVersion,
                 id=default_version.id,
                 number=default_version.number,
             )
+            project = Project.objects.get(account=user.iaso_profile.account)  # type: ignore
```
olethanh: an account might be linked to multiple project so this query will probably fail in production.

it might be easier to filter directly in the get_object_or_404,  something like `SourceVersion.filter(data_source__project__account_user=user)` not sure of the exact syntax

olethanh (changes requested):

iaso/api/accounts.py
SimonDus: I couldn't make it work this way as projects is a many to many so I did it differently and adapted the tests to take into consideration this case

PR #344, opened 2023-01-31
Fix cypress

beygorghor:
Cypress tests were failing on detailChildrenTab, submissions/list, detailLinksTab and detailLinksTab.
Especially with the call to groups that has been changed

- [ ] Did I add translations
- [ ] Are there enough tests

## Changes

- rectify intercept calls for groups
- change test of an input text with selectall option

## How to test

run cypress tests


beygorghor:
@cypress

PR #355, opened 2023-02-03
IA-1759 json widget for the django admin

mathvdh:
Replace the default JSON Widget in Django Admin by a much better one with colored syntax, and error detection.

Related JIRA tickets : IA-1759

## Changes

Add `django-json-widget` to requirements and added it to the loaded apps in `/hat/settings.py`
Modified `/iaso/admin.py` and `plugins/polio/admin.py` to override the default JSON widget when necessary
In `plugins/polio/admin.py` > WorkflowAdmin removed the pretty_json read only field as it's not necessary anymore
Added `/hat/templates/admin/base.html` in order to add a custom css file to customize the look of the widget.
Added `/iaso/static/css/django-json-widget.css` with custom css for dark mode.

## How to test

Go to django admin and open for example a form or a workflowversion or any of the modified ModelAdmins

## Print screen / video

Example for a form "possible fields"

Screenshot 2023-02-03 at 11 36 28

## Notes

For the frontend people you can still customize the CSS more in `/iaso/static/css/django-json-widget.css` if you'd like this is not totally perfect style-wise


beygorghor (approved):

quang-le (commented):
Did you commit the css file? I don't see it anywhere. 

My admin looks like this locally (see pic). Still love it though. There's even an `undo` button!

Screenshot 2023-02-07 at 11 40 03

mathvdh:
@quang-le Apparently I didn't commit the CSS ... I will do it

mathvdh:
Apparently everything in `/iaso/static` is in `.gitignore` (line 123 of .gitignore)
I don't understand why ?
Moreover all the images in `/iaso/static/img `are committed even thou they are in gitignore

hat/templates/admin/base.html
```diff
@@ -0,0 +1,12 @@
+{% extends 'admin/base.html' %}
+{% load static %}
+
+
+
+{% block extrastyle %}
+  <link rel="stylesheet" href="{% static 'css/admin-json-widget.css' %}">
```
olethanh: Not really import but FYI  you can set the  CSS via the  JSON_EDITOR_CSS settings (or in the inheritance of the widget, if you want to add a supplementary file to the builtin one)

cf 

https://github.com/jmrivas86/django-json-widget/blob/master/django_json_widget/widgets.py#L15

https://docs.djangoproject.com/en/3.2/topics/forms/media/

olethanh (approved):

hat/templates/admin/base.html
mathvdh: Well the ting is that I'm not replacing the default JSON_EDITOR_CSS, I'm adding to it .. ?

mathvdh:
@quang-le  it should work now with the added CSS

niconoe:
Works quite well for me, a few comments still:

- I don't encounter the CSS issue, so I guess the file is eventually included now?
- Did you consider using a base class or a mixin to avoid repetition with `formfield_overrides = {models.JSONField: {"widget": IasoJSONEditorWidget}}`?
- When using the widget and trying to save incorrect JSON ("save and continue editing"): no error is clearly reported, and the JSON is reverted to the previous state (changes removed). I don't think that is consistent with standard fields in the Admin site.
- Maybe it's just the errors I introduced were too complicated, but the "repair JSON" feature (wrench icon) doesn't seem to do anything.

mathvdh:
> Did you consider using a base class or a mixin to avoid repetition with formfield_overrides = {models.JSONField: {"widget": IasoJSONEditorWidget}}?

I actually considered it but it required creating a base class for all django admin "pages" and then changing it everywhere also, or monkey patching the origin django ModelAdmin, I prefered to have this explicit line.

> Maybe it's just the errors I introduced were too complicated, but the "repair JSON" feature (wrench icon) doesn't seem to do anything.

Didn't really try this feature honestly ..

> When using the widget and trying to save incorrect JSON ("save and continue editing"): no error is clearly reported, and the JSON is reverted to the previous state (changes removed). I don't think that is consistent with standard fields in the Admin site.

This could be more of an issue .. this need to be tested more !

hat/templates/admin/base.html
mathvdh: I'll have a look into it because I think it would be better if we dont need the `/hat/templates/admin/base.html` file

mathvdh:
@olethanh I used your solution with the Media subclass, it's better than overriding the base admin template I think ..

@niconoe, I tested the "repair json" and it works for simple cases, I guess it cannot fix complex cases ...

Regarding the "doesn't save" when the json is invalid, I'm not sure what is the "good approach", we don't want to save a broken JSON I guess ... What do you think @olethanh  ?

olethanh:
@mathvdh if it's a json postgresql field the DB won't allow saving it anyway but yeah an error should be shown, it's strange if it doesn't?

mathvdh:
@olethanh I checked the widget's github and it's not supported at the moment .. I'm wondering if we shouldn't switch to another widget or fork this one ... As the author doesn't seem to be very active anymore

quang-le:
> Apparently everything in `/iaso/static` is in `.gitignore` (line 123 of .gitignore) I don't understand why ? Moreover all the images in `/iaso/static/img `are committed even thou they are in gitignore

As far as I'm concerned we can merge it as is and figure out the styling later. It's already usable and a huge improvement as is

mathvdh:
It was corrected for the style.
The issue now is what happens when the JSON is "bad", as it let's you save the model in django admin but doesn't actually save the changes in the JSON

mathvdh:
Only way to solve this would be to fork the django-admin-widget project and customize it. As it doensn't seem to be very active anymore. I did that in my private github account but didn't find time yet to work on it ...

quang-le (approved):

olethanh:
@mathvdh Forking seems a sensible idea.

actually we could check in the github fork if someone hasn't solved the issue already

mathvdh:
@olethanh I looked but didn't see much progress in the forks

PR #356, opened 2023-02-03
Polio 798 timeline widget backend

quang-le:
Backend of POLIO budget override step needs to tell front-end if a step has been skipped or cancelled due to an admin performing an override

Related JIRA tickets :  POLIO-798

- [ ] Is my code clear enough and well documented: I guess??
- [ ] Did I add translations
- [ ] Are there enough tests -> I'll make another ticket for that

## Changes

- Updated `CampaignBudgetSerializer`'s `get_timeline` method:
    - get all nodes from workflow to match the start and destination of the override step to an actual node
    - make a list of override steps
    - Once all steps have been listed: for each override step: 
        - if the step has been created before the override step and is between the override step's start end destination nodes: add either a `skipped` or `cancelled` entry to the step's data
    - Once all skipped and cancelled steps have been marked, determine which sections of the progress bar have been completed  
    - Only "filler steps" (the greyed out steps in the UI) are skipped. If we skip an actual step, it has to have been cancelled first by a previous override (logically)
    - Only actual steps are cancelled. Previously skipped steps revert to a neutral state, ie. they will be displayed as greyed out steps in the UI. The idea is that cancelling a placeholder makes no sense and would clutter the UI

## How to test

- Go to Budgets, 
- Select a budget
- Perform an override that goes forward in the workflow: in the API response, all steps that have been skipped should have  `skipped: true`
- Perform an override that goes backwards in the worflow: in the API response, all steps between start end end point should have `cancelled:true`
 

## Notes

Merges into POLIO-768, to take advantage of a fix that prevents a 500 when posting override step from front end


plugins/polio/budget/serializers.py
```diff
@@ -383,22 +475,31 @@ def save(self, **kwargs):
...
             for link_data in data.get("links", []):
                 link_serializer = BudgetLinkSerializer(data=link_data)
                 link_serializer.is_valid(raise_exception=True)
                 link_serializer.save(step=step)
 
+            # campaign.budget_current_state_key = node_key
+            # for file in data.get("files", []):
+            #     step.files.create(file=file, filename=file.name)
+            # campaign.budget_current_state_label = to_node.label
+            # campaign.save()
```
olethanh, line 490: ?

olethanh (approved):
wfm

PR #385, opened 2023-02-16
IA-1916 entity mandatory field

quang-le:
In entity types create/edit modal, the `List fields` and `Details info fields` need to be made mandatory to prevent bugs with the mobile app

Related JIRA tickets : IA-1916

- [ ] Did I add translations
- [ ] Are there enough tests

## Changes

- Added validator for both fields
- Marked them as required
- Changed the default value in `intialState` fromm `[]` to `undefined` to enable the validation check
- Filtered out forms without XLS version to avoid users having no fields to choose from after having selected a reference form when creating an entity type
- Fixed `usePossibleFields` so it fetches the fields when user changes selection in UI
- Fixed Dialog so it would reset `onClose` to avoid keeping the values after a `Create`
- **UPDATE 2023/02/17**: add `useGetPossibleFieldsForEntityTypes` to remove glitch on close (see below in comments thtread and code review).

## How to test

Go to Beneficiary Types
Create a Beneficiary: you shouldn't be able to save unless all fields have a value

## Print screen / video

https://user-images.githubusercontent.com/38907762/219353290-04a94906-bd96-4984-ba74-029d2fec036f.mov

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/entities/entityTypes/components/EntityTypesDialog.tsx
+++ b/hat/assets/js/apps/Iaso/domains/entities/entityTypes/components/EntityTypesDialog.tsx
@@ -20,7 +20,7 @@ import { baseUrls } from '../../../../constants/urls';
 
 import { useGetForms } from '../hooks/requests/forms';
 import { useTranslatedErrors } from '../../../../libs/validation';
-import { useGetPossibleFields } from '../../../forms/hooks/useGetPossibleFields';
+import { useGetPossibleFieldsForEntityTypes } from '../../../forms/hooks/useGetPossibleFields';
 import MESSAGES from '../messages';
 
 type RenderTriggerProps = {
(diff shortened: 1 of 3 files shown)


quang-le:
cypress

codeKameleon (approved):

beygorghor (changes requested):
While clicking on cancel, the fields are displaying an error state, not very sure how to handle this
https://user-images.githubusercontent.com/12494624/219588271-ac39fd02-c6fa-42ea-a7b9-894bd4f2727a.mov

quang-le:
> While clicking on cancel, the fields are displaying an error state, not very sure how to handle this https://user-images.githubusercontent.com/12494624/219588271-ac39fd02-c6fa-42ea-a7b9-894bd4f2727a.mov

That's probably because we reset the state values but the formik context hasn't reset

quang-le:
cypress

PR #389, opened 2023-02-16
WC2-167 search feature on entity api

SimonDus:
Mobile specific API to serve entity types /api/mobile/entitytypes

You can retrieve entities from entity types by /api/mobile/entitypes/{entity_type_id}/entities
filter on date by ?limit_date
json logic by ?json_content

Related JIRA tickets : WC2-167

- [x] Are there enough tests

## How to test

Create entity types, entities from different accoun, play with the api and check the data


iaso/api/mobile/entity.py
```diff
@@ -95,6 +96,7 @@ def get_queryset(self):
         orgunits = OrgUnit.objects.hierarchy(profile.org_units.all())
         queryset = Entity.objects.filter(account=profile.account)
         # we give all entities having an instance linked to the one of the org units allowed for the current user
+        # For now, it can be filtered on OU without any access restriction
```
bmonjoie: This comment is wrong and should be deleted

iaso/api/mobile/entity_type.py
```diff
@@ -0,0 +1,99 @@
...
+)
+from iaso.api.mobile.entity import MobileEntitySerializer
+from iaso.models import Entity, Instance, EntityType
+from iaso.utils.jsonlogic import jsonlogic_to_q
+
+
+class LargeResultsSetPagination(PageNumberPagination):
+    page_size = 1000
+    page_size_query_param = "page_size"
+    max_page_size = 1000
```
bmonjoie: I think it's probably a good idea to reuse the same class as the one in `entity.py` rather than copy/pasting it here

iaso/api/mobile/entity_type.py
```diff
@@ -0,0 +1,99 @@
...
+    class Meta:
+        model = EntityType
+        fields = [
+            "name",
+            "created_at",
+            "updated_at",
+            "reference_form",
+            "account",
+            "is_active",
+        ]
```
bmonjoie: The list of fields returned by the `MobileEntityType` serializer should be the same as the one returned by the `EntityTypeSerializer`:

- id
- name
- created_at
- updated_at
- reference_form
-  entities_count
- account
- fields_detail_info_view
- fields_list_view

iaso/api/mobile/entity_type.py
```diff
@@ -0,0 +1,99 @@
...
+                q = jsonlogic_to_q(jsonlogic=json_content, field_prefix="json__")  # type: ignore
+                queryset = queryset.filter(q)
+            except ValidationError:
+                return Response("Error in json logic format", status.HTTP_400_BAD_REQUEST)
+
+        page = self.paginate_queryset(queryset)
+
+        serializer = MobileEntitySerializer(page, many=True)
+
+        return self.get_paginated_response(serializer.data)
```
bmonjoie: A lot of this code seems like it has been copy/pasted from MobileEntityViewSet, it would probably be a good idea to make a base class to share the shared logic between the two.

iaso/tests/api/test_entity_types.py
```diff
@@ -184,3 +185,70 @@ def test_entity_types_are_multitenancy(self):
...
+
+        # different account
+        EntityType.objects.create(name="beneficiary", reference_form=self.form_1, account=self.the_gang)
+
+        response = self.client.get("/api/mobile/entitytypes/")
+
+        entity_type_count = EntityType.objects.filter(account=self.yoda.iaso_profile.account).count()
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.json()["count"], entity_type_count)
```
bmonjoie: Testing on "count" is often useless; you have no idea what has been returned.
Also, testing on "count" with only one value is not testing anything. If your code fails the limit and has a limit of 1, you won't see it at all!
Your test doesn't even check if the only entity type returned is the one from your account. What guarantees that the entity type returned is not from the other account?

Also, Since they have the same name and reference_form values, you can't even test the value returned to ensure that.

iaso/tests/api/test_entity_types.py
```diff
@@ -184,3 +185,70 @@ def test_entity_types_are_multitenancy(self):
...
+
+        response = self.client.get(f"/api/mobile/entitytype/{entity_type}/entities")
+
+        self.assertEqual(response.json()["count"], 2)
+
+        Entity.objects.create(
+            name="New Client",
+            account=self.yoda.iaso_profile.account,
+            entity_type=second_entity_type,
+        )
```
bmonjoie: This should be moved above the first call to entity types.

iaso/tests/api/test_entity_types.py
```diff
@@ -184,3 +185,70 @@ def test_entity_types_are_multitenancy(self):
...
+
+        response = self.client.get(f"/api/mobile/entitytype/{second_entity_type}/entities")
+
+        self.assertEqual(response.json()["count"], 1)
```
bmonjoie: Same remark as before, testing on count is useless

iaso/tests/api/test_entity_types.py
```diff
@@ -184,3 +185,70 @@ def test_entity_types_are_multitenancy(self):
...
+        self.client.force_authenticate(self.yoda)
+
+        EntityType.objects.create(name="beneficiary", reference_form=self.form_1, account=self.the_gang)
+
+        response = self.client.get("/api/mobile/entitytypes/")
+
+        entity_type_count = EntityType.objects.filter(account=self.yoda.iaso_profile.account).count()
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.json()["count"], entity_type_count)
```
bmonjoie: You are testing with only one entity type, you should test with more and you should test the values, testing on count is not a safe test

bmonjoie (changes requested):

iaso/api/mobile/entity.py
```diff
@@ -18,6 +21,44 @@ class LargeResultsSetPagination(PageNumberPagination):
...
+
+    p = Prefetch(
+        "instances",
+        queryset=Instance.objects.filter(deleted=False).exclude(file=""),
+        to_attr="non_deleted_instances",
+    )
+
+    queryset = queryset.prefetch_related(p).prefetch_related("non_deleted_instances__form")
+
+    return queryset
```
bmonjoie: As we talked about, this should be rewritten to be cleaner.
Also, it introduces a bug in `/api/mobile/entitytypes/id/entities` as it now filters on the accessible org units when it should not.

bmonjoie (changes requested):

iaso/api/mobile/entity.py
```diff
@@ -59,7 +99,9 @@ class Meta:
 
     @staticmethod
     def get_instances(entity: Entity):
-        return list(map(lambda instance: MobileEntityAttributesSerializer(instance).data, entity.non_deleted_instances))  # type: ignore
+        return list(
+            map(lambda instance: MobileEntityAttributesSerializer(instance).data, entity.non_deleted_instances)  # type: ignore
+        )
```
bmonjoie: @olethanh told me this should be written like this:

```suggestion
        return [ MobileEntityAttributesSerializer(instance).data for instance in entity.non_deleted_instances ]
```

olethanh: actually you would write `return MobileEntityAttributesSerializer(entity.non_deleted_instances, many=True).data`

olethanh: actually rreading the whole code you don't even need a function

`instances = MobileEntityAttributesSerializer(many=True, source="non_deleted_instances")` should suffice I think

iaso/api/mobile/entity.py
```diff
@@ -92,19 +122,11 @@ def get_serializer_class(self):
 
     def get_queryset(self):
         profile = self.request.user.iaso_profile
-        orgunits = OrgUnit.objects.hierarchy(profile.org_units.all())
-        queryset = Entity.objects.filter(account=profile.account)
+        queryset = filter_queryset_for_mobile_entity(Entity.objects.filter(account=profile.account), self.request)
```
olethanh, line 125: Maybe add a comment or alter the name to signify that the function alter the queryset by adding the `non_deleted_instances` fields ?

iaso/api/mobile/entity.py
```diff
@@ -18,6 +21,32 @@ class LargeResultsSetPagination(PageNumberPagination):
...
+
+    json_content = request.query_params.get("json_content", None)
+    if json_content:
+        try:
+            q = jsonlogic_to_q(jsonlogic=json_content, field_prefix="json__")  # type: ignore
+            queryset = queryset.filter(q)
+        except ValidationError:
+            raise Http404("Invalid Json Content")
+
+    p = Prefetch(
```
olethanh, line 40: didn't know prefetch, nice find :+1:

olethanh (approved):
a few nitpick on style but LGTM. Not sure if Benjamin had anything else that was blocking?

I'm still doubtful on the whole aspect of access and lack of perm but that's another discussion

PR #396, opened 2023-02-20
Fix docker iaso output

olethanh:
Fix the iaso docker not outputing any log in local dev.

Check if it works for you when you launch it. (also tell me if it was actually  working before for you)


olethanh:
@quang-le fixed

PR #399, opened 2023-02-21
WC2-170: Duplicate search fields configuration

bmonjoie:
To be able to configure which fields serve as a comparison for duplicate search, the backend must expose a list of fields to the mobile.
The web should also allow modification of those fields along with the list and details fields.
The duplicate search fields are not required; an empty list indicates we don’t want the duplicate search feature enabled.

Related JIRA tickets : WC2-170

- [x] Did I add translations
- [x] Are there enough tests

## Changes

Add a new `fields_duplicate_search` property to `EntityType`.
Expose this new field in the corresponding APIs.
Fix `/api/mobile/entitytype` was not pointing on the correct `ViewSet`. 
Make the mobile entity type endpoints expose the fields as non-nullable. 
Modify EntityType's configuration interface to add the duplicate search fields configuration.

## How to test

Edit an existing EntityType (or create a new one) and add one or more fields to the new "Duplicate search fields".


iaso/urls.py
```diff
@@ -125,7 +125,7 @@
 router.register(r"entitytype", EntityTypeViewSet, basename="entitytype")
 # At the moment we use the same view set but separate it for the future for when we want to be able to
 # change the format in the future
-router.register(r"mobile/entitytype", EntityTypeViewSet, basename="entitytype")
+router.register(r"mobile/entitytype", MobileEntityTypesViewSet, basename="entitytype")
```
SimonDus, line 128: Dont we use plurals name now ? 
I added mobile/entitytypes", MobileEntityTypesViewSet, basename="entitytypes at line 124 in urls.py in my previous PR about mobile entity types. Indeed on this branch we have entitytype and entitypes working. So I think maybe something went wrong during your merge of main ? And/Or I forgot to remove the old url

SimonDus (changes requested):

iaso/urls.py
bmonjoie: The old URLs are still there for backward compatibility reasons

SimonDus (approved):
Seems good to me.

quang-le:
cypress

quang-le (approved):
Front-end is ok to me

PR #403, opened 2023-02-22
Fix search by json logic not working

bmonjoie:
Also add support for the "in" operator for strings.

Related JIRA tickets : WC2-167

- [ ] Did I add translations
- [X] Are there enough tests

## Changes

Add import of `jsonlogic_to_q`.
Make sure that the value passed to `jsonlogic_to_q` is a JSON object.
Fix the prefix field on which the query was applied.
Add a test to ensure it is working properly.
Slight change in `jsonlogic_to_q` to include the [`in`](https://jsonlogic.com/operations.html#in) operator

## How to test

Call `/api/mobile/entitytypes/{id}/entities` with a valid `json_content` query param.

### DIFF
--- a/iaso/tests/fixtures/test_entity_data2.xml
+++ b/iaso/tests/fixtures/test_entity_data2.xml
@@ -2,11 +2,11 @@
           <start>2022-09-20T16:50:40.889+02:00</start>
           <end>2022-09-20T16:51:17.352+02:00</end>
           <register>
-            <name>Don Diego de la Vega</name>
-            <father_name>Don Alejandro de la Vega</father_name>
+            <name>Prince of Euphor</name>
+            <father_name>Professor Procyon</father_name>
             <age_type>0</age_type>
             
-            <birth_date>1919-08-09</birth_date>
+            <birth_date>1978-07-03</birth_date>
             <child_details>
               <gender>male</gender>
               <caretaker_name/>
@@ -15,6 +15,6 @@
             </child_details>
           </register>
           <meta>
-            <instanceID>uuid:abe3d0c2-0940-4e9c-8e3d-c0985bb57177</instanceID>
+            <instanceID>uuid:4901dff4-30af-49e2-afd1-42970bb8f03e</instanceID>
           </meta>
         </data>
--- a/iaso/api/mobile/entity.py
+++ b/iaso/api/mobile/entity.py
@@ -1,11 +1,12 @@
+import json
+
 from django.core.exceptions import ValidationError
 from django.db.models import Prefetch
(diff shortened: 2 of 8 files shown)


iaso/tests/fixtures/test_entity_data.xml
```diff
@@ -2,11 +2,11 @@
           <start>2022-09-20T16:50:40.889+02:00</start>
           <end>2022-09-20T16:51:17.352+02:00</end>
           <register>
-            <name>Zoro Marlboro</name>
-            <father_name>Gold Marlboro</father_name>
+            <name>Don Diego de la Vega</name>
+            <father_name>Don Alejandro de la Vega</father_name>
```
quang-le, line 6: lol

iaso/tests/fixtures/test_entity_data2.xml
```diff
@@ -0,0 +1,20 @@
+<data xmlns:jr="http://openrosa.org/javarosa" xmlns:orx="http://openrosa.org/xforms" id="wfp_reg_ok" version="2022081601" iasoInstance="18">
+          <start>2022-09-20T16:50:40.889+02:00</start>
+          <end>2022-09-20T16:51:17.352+02:00</end>
+          <register>
+            <name>Prince of Euphor</name>
+            <father_name>Professor Procyon</father_name>
```
quang-le, line 6: double lol

quang-le (commented):
I can't really comment on the backend code other than it seems to make sense, but I think since you enabled a new operator, the change should be reflected in the web front-end as well (either in the PR or in a separate ticket)

bmonjoie:
@quang-le I have indeed added a new operator, but we should test it with the other kind of data (arrays, etc.), as I have only tested with text, to add it in the web front-end. This could be added in a future PR if the need for this operator in the web front-end arises.

SimonDus (approved):
Python wise everything make sense. Good job.  My bad on this one !

mathvdh (approved):
RAS c'est très bien pour moi.
Ce que je voulais dire par rapport à convertir le JSON login en Q .. 

C'est pourquoi on a pas utilisé un truc du genre [json-logic-py](https://github.com/nadirizr/json-logic-py) et puis populé les vars ... 

mais j'ai pas suivi le truc dans tous ses details donc je raconte p-e n'importe quoi

bmonjoie:
@mathvdh Je pense que la raison pour laquelle on n'utilise pas JsonLogic-py c'est parce qu'on veut faire les requêtes en DB et que JsonLogic-py ne fait que du traitement de données en mémoire.

PR #405, opened 2023-02-23
Add an API to retrieve storage passwords

bmonjoie:
To write data on storage devices like NFC cards in a secure way, we need to share a password with the authenticated users of our application.

Related JIRA tickets : WC2-48

- [ ] Did I add translations
- [X] Are there enough tests

## Changes

I have added a new model to store the passwords.
I have exposed the modification through the admin.
I made a new mobile endpoint to retrieve the passwords.

## How to test

Add a password or more to some projects.
Retrieve them through `/api/mobile/storage/passwords`


iaso/admin.py
```diff
@@ -500,6 +501,19 @@ class StorageDeviceAdmin(admin.ModelAdmin):
     ]
 
 
+class StoragePasswordAdmin(admin.ModelAdmin):
+    fields = (
```
mathvdh, line 505: Je ne crois pas que t'es obligé de specifier tous les fields @bmonjoie , par défaut il les prends tous ...

mathvdh (approved):
Except for my little comment above which is more informative than anything this is really clean .. I would say even cleaner that what the backends usually do themselves @bmonjoie !

iaso/admin.py
mathvdh: Sauf si tu veux changer l'ordre ou faire des groupes : https://docs.djangoproject.com/en/4.1/ref/contrib/admin/#django.contrib.admin.ModelAdmin.fields

bmonjoie: OK, merci, je ne savais pas :slightly_smiling_face:

SimonDus (approved):
I have to admit that Benjamin is raising the bar. Everything works locally, so Martin said no more " looks good to me " but everything looks good in this case

PR #413, opened 2023-03-02
IA-1790 open the form versions api in read access to any connected users

mathvdh:
The idea is that, at the moment, if a project has the boolean `needs_authentication` to False. Then any user can access in read only the Forms that are linked to this project. We decided to extend this to the formversions linked to the forms to the project. (Initially it was discussed to hide this behind a feature flag but after careful discussion and consideration, it doesn't make sense to give access to forms but not the formversions related)

Related JIRA tickets : IA-1790

- [x] Are there enough tests

## Changes

Creation of a new Permission class `HasFormVersionPermission` derived from `HasFormPermission` for the api.
Modification of the form_versions api to handle the case when a user is not authentified but instead a app_id is given.
Addidition of a new test file to test all these changes.

## How to test

You can check `iaso/tests/api/test_form_versions_readonly.py`

But basically the idea is that if you call GET `/api/formversions` with no auth but and app id, it will return the versions of the forms linked to the project reference by this app_id

And the same if you call GET `/api/formversions/FORM_ID/?app_id=PROJECT` it should return data even if you are not loggued in or you normally dont have the rights to access this form


mathvdh:
I forgot to mention that this also includes a little change to the script `runiasodev.sh` to allow passing arguments to it. At the moment it calls "runserver" only .. now you can give arguments to it like "migrate" or "test". (it doesn't change the behaviour if you don't give any argument to the script)

mathvdh:
@bmonjoie Can you please have a look ? I think the mobile app is the one that needs this change ... ?

iaso/tests/api/test_form_versions_readonly.py
```diff
@@ -0,0 +1,192 @@
...
+
+        response = self.client.get(f"{BASE_URL}?app_id={self.blue_project_no_need_auth.app_id}")
+        self.assertJSONResponse(response, 200)
+        form_versions_data = response.json()["form_versions"]
+
+        assert len(form_versions_data) == 1
+
+        for form_version_data in form_versions_data:
+            self.assertValidFormVersionData(form_version_data)
+            self.assertNotIn("descriptor", form_version_data)
```
bmonjoie, line 105: Why do you test against that?
I need the descriptors when requested, so I want to ensure they'll still be provided, even for anonymous users, when requested.

iaso/api/form_versions.py
```diff
@@ -132,6 +133,18 @@ def update(self, form_version, validated_data):
...
+class HasFormVersionPermission(HasFormPermission):
+    def has_object_permission(self, request, view, obj) -> bool:
+        if not self.has_permission(request, view):
+            return False
+
+        ok_forms = Form.objects_include_deleted.filter_for_user_and_app_id(
+            request.user, request.query_params.get("app_id")
+        )
+
+        return obj.form in ok_forms
```
olethanh: do not use `in` to check I made this mistake before somehow it make django fetch all the records. Use .filter().exists()

olethanh (changes requested):
> The idea is that, at the moment, if a project has the boolean needs_authentication to False. Then any user can access in read only the Forms that are linked to this project. We decided to extend this to the formversions linked to the forms to the project. (Initially it was discussed to hide this behind a feature flag but after careful discussion and consideration, it doesn't make sense to give access to forms but not the formversions related)

I'm a bit confused by this description and the code, is it only for project with the `need_authentification` or not? And where is that checked in the code?

More of a overall comment than this PR in particular but I'm getting the nagging feeling that we are crowbaring the whole user right management for the sake of some features in the mobile app without much oversight and global strategy and that we are going to regret it at some point

iaso/api/form_versions.py
mathvdh: I was actually thinking about it but the current code in [/api/forms.py](https://github.com/BLSQ/iaso/blob/e89da17f8e8cb2e0a522251a300ed51c14fa3726/iaso/api/forms.py#L30) has the same 

```python
return obj in Form.objects_include_deleted.filter_for_user_and_app_id(
    request.user, request.query_params.get("app_id")
)
```

So this should be corrected too then ?

mathvdh:
@olethanh for your general comment:  

yes it's only for Project with `needs_authentication = False` and it's checked on Line 141 on `/api/form_versions.py` :  `Form.objects_include_deleted.filter_for_user_and_app_id(..)` which calling the Queryset functions from `Form` model which itself is calling the queryset functions from `Project` model where the check actually happens (it's all the same as in `/api/forms.py`)

If we are overriding some user right management, it was there already in the Past, I'm just copying from Form and adding the same for FormVersions ... This is a more general discussion I guess

iaso/api/form_versions.py
olethanh: yep

olethanh:
ah yes, I see. didn't check one level higher my bad

iaso/api/form_versions.py
olethanh: I corrected both and pushed it

olethanh (approved):

PR #417, opened 2023-03-03
IA-1964 SQL /explore/ interface add a Lines Chart widget

olethanh:
Add a new visualization for the sql UI (at /explore/ )

This widget will automatically be automatically activated when the query
has the columns: line_label, line_quantity and line_category

Related JIRA tickets :  IA-1964

- [NA] Did I add translations
- [NA] Are there enough tests

## Changes

Add a new widget using the method  documented in the lib
https://django-sql-dashboard.datasette.io/en/stable/widgets.html#custom-widgets

## How to test

Go to /explore with a super user account and try a few query

As starting point you can use the query for cumulative submission per month
```
select line_label,
       line_category,
       sum(line_quantity) over (PARTITION BY line_category order by line_label) as line_quantity
from (
select TO_CHAR(date_trunc('month', iaso_instance.created_at), 'YYYY/MM') as line_label, count(*) as line_quantity,  iaso_project.name as line_category from iaso_instance inner join iaso_project on iaso_instance.project_id = iaso_project.id
 group by line_label, iaso_project.name order by line_label, line_quantity desc limit 200
) as data
```

or for generated data
```
select TO_CHAR(gen_date.generate_series, 'YYYY/MM') as line_label,
random() * 1000::int                         as line_quantity,
name                                         as line_category
from (select *
from generate_series('2008-03-01 08:00'::timestamp,
'2009-03-04 12:00'::timestamp, '1 month')) gen_date
cross join (VALUES ('foo'), ('bar'), ('baz')) as categories (name)
```

This is for inside advanced use so the ui is a bit more clonky

## Print screen / video

image

## Notes

When this is a bit more mature it would be nice to have this moved to the upstream lib


mathvdh (approved):
Pour moi ça a l'air bon ... On peut merger @olethanh

PR #424, opened 2023-03-07
POLIO-867 Fix preparedness data cron not refreshing

olethanh:
the else was passed as an argument and considered a campaign filter

Related JIRA tickets : POLIO-867


mathvdh (approved):
@olethanh c'est difficile à tester en local evidemment .. Je peux merger avant la release ?

PR #425, opened 2023-03-07
Allow pagination of mobile OrgUnit API

bmonjoie:
We received a complaint that some devices cannot synchronize OrgUnits when they are too many.
As we couldn't reproduce, the assumptions are:
 - The list is too big in memory
 - The filtering that is done to remove the `parent_id` value of children for which the parent hasn't been returned is too consuming.

This PR allows 2 new parameters: `page_size`and `page`, to process the OrgUnit in smaller chunks.
Since we won't be able to remove the `parent_id` of children without parents with only part of the data, the removal is now done in the backend.

Therefore, this addresses both assumptions while still being backward compatible.

Related JIRA tickets: IA-1914

- [X] Are there enough tests

## Changes

The `/api/mobile/orgunits` endpoint is now capable of returning a paginated list.
The `parent_id` value of OrgUnits with a parent that has a `REJECTED` status is now set to `None`.

## How to test

Call `/api/mobile/orgunits?app_id=...` **before** applying the changes.
Call `/api/mobile/orgunits?app_id=...` **after** applying the changes.
The results should be the same

Then, call `/api/mobile/orgunits?app_id=...` with a `page_size` and `page` parameters and ensure the right chunk of the data has been returned.

### DIFF
--- a/iaso/api/mobile/org_units.py
+++ b/iaso/api/mobile/org_units.py
@@ -3,12 +3,20 @@ import json
 from django.contrib.gis.geos import Point
 from django.core.cache import cache
 from rest_framework import viewsets, permissions
+from rest_framework.pagination import PageNumberPagination
 from rest_framework.response import Response
 
 from hat.api.export_utils import timestamp_to_utc_datetime
 from iaso.api.common import get_timestamp
 from iaso.api.common import safe_api_import
+from iaso.api.query_params import APP_ID, PAGE_SIZE, PAGE
 from iaso.models import OrgUnit, Project
+
+
+class MobileOrgUnitsSetPagination(PageNumberPagination):
+    page_size_query_param = PAGE_SIZE
+    page_query_param = PAGE
+    page_size = None  # None to disable pagination by default.
 
 
 class HasOrgUnitPermission(permissions.BasePermission):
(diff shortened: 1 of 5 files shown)


iaso/models/org_unit.py
```diff
@@ -371,6 +371,13 @@ def as_dict_for_mobile(self):
             "aliases": self.aliases,
         }
 
+    def get_valid_parent_id(self, default_to_none: bool = True):
+        return (
+            self.parent_id
+            if self.parent_id is None or self.parent.validation_status == OrgUnit.VALIDATION_VALID
+            else (self.parent.get_valid_parent_id(default_to_none) if default_to_none is not True else None)
+        )
+
```
bmonjoie: @olethanh Even though I have added `.prefetch_related("parent")` to the `queryset`, I'm afraid this could result in a lot of queries to the database.
Could you advise if this is OK and, if not, what would be a better way?

olethanh: you can do a .prefetch_related("parent").prefech_related("parent_parent").. etc if you do 4 level it should be enough

iaso/api/mobile/org_units.py
```diff
@@ -3,14 +3,22 @@
...
 from iaso.api.common import safe_api_import
+from iaso.api.query_params import APP_ID, PAGE_SIZE, PAGE
 from iaso.models import OrgUnit, Project
 
 
+class MobileOrgUnitsSetPagination(PageNumberPagination):
+    page_size_query_param = PAGE_SIZE
+    page_query_param = PAGE
+    page_size = None  # None to disable pagination by default.
+
```
madewulf, line 21: Everywhere else, the paginations parameter is limit instead of page_size, I think. It would be better to stick with that

iaso/api/mobile/org_units.py
```diff
@@ -32,52 +40,66 @@ def has_object_permission(self, request, view, obj):
...
-        if app_id:
-            cached_response = cache.get(app_id)
-        else:
-            return Response()
 
+        pagination = MobileOrgUnitsSetPagination()
+        page = pagination.paginate_queryset(queryset, request)
+        page_number = 1
+        page_size = None
+        if page:
```
madewulf: This works, and you probably found samples like these in the older APIs, but pagination can be fully handled by Django Rest Framework with way less code

madewulf: But maybe this is done in this way to stay backward compatible?

iaso/tests/api/test_mobile_orgunits.py
```diff
@@ -0,0 +1,127 @@
+from iaso.api.query_params import APP_ID, PAGE_SIZE, PAGE
```
madewulf: Was there no test for this mobile org unit api before?

madewulf (changes requested):

iaso/api/mobile/org_units.py
bmonjoie: I first put `limit` but I saw that you put `page_size` on entities so I used that to stay consistent:
image

iaso/api/mobile/org_units.py
bmonjoie: I know, and I tried, but for the Django Rest Framework to handle the pagination fully, we have to use a serializer. Since the serialization is done through an `as_dict_for_mobile()` method, I couldn't use a serializer without redoing it all.

iaso/tests/api/test_mobile_orgunits.py
bmonjoie: None that I could find

iaso/api/mobile/org_units.py
madewulf: That's a mistake. We should stick to limit. Do you use that parameter already on entities?

bmonjoie: Actually, the code on mobile passes `limit` :sweat_smile: , I'll change it in both places

madewulf: 👍

iaso/api/mobile/org_units.py
madewulf: I think that would be a very small serializer as it does not follow any foreign key
CleanShot 2023-03-09 at 11 09 35@2x
but if some older apps are expecting results without pagination, then, it's probably painful to get right in the DRF (that will paginate everything by default, which is probably a sane approach)

madewulf: All those as_dict() method were just me being ignorant of good practices of the DRF and ideally should disappear, FYI.

iaso/tests/api/test_mobile_orgunits.py
madewulf: Could you add a test for the non paginated version then? It should be fast with what you have written already. Thanks!

madewulf (changes requested):

iaso/tests/api/test_mobile_orgunits.py
bmonjoie: That's literally my first test :thinking:

iaso/api/mobile/org_units.py
```diff
@@ -1,16 +1,91 @@
...
 
 from hat.api.export_utils import timestamp_to_utc_datetime
-from iaso.api.common import get_timestamp
+from iaso.api.common import get_timestamp, TimestampField, ModelViewSet
 from iaso.api.common import safe_api_import
+from iaso.api.query_params import APP_ID, LIMIT, PAGE
 from iaso.models import OrgUnit, Project
 
 
+class MobileOrgUnitsSetPagination(PageNumberPagination):
```
mathvdh: Why are we not subclassing the Paginator from `/iaso/api/common.py` ? (Just a question ..)

bmonjoie: Because it's not the one that was used by @madewulf in [`/iaso/api/mobile/entity.py`](/BLSQ/iaso/pull/425/files#diff-5098f37ac9f52301b6841b2fe96a74bf556dcca106bcb9c0fdf4659985762bce). Should I change to the one in `/iaso/api/common.py`?

bmonjoie: I have changed to use the one in `/iaso/api/common.py`

olethanh (approved):

PR #431, opened 2023-03-08
WC2-175: Storages, beneficiaries, move export buttons up

beygorghor:
Move export buttons before page result on those two pages

Related JIRA tickets : WC2-175

- [ ] Are there enough tests

## Changes

Moved and wrap buttons component before table on both pages

## How to test

Explain how to test your PR.

Visit Beneficiaries and storage page, check export buttons

## Print screen / video
Screenshot 2023-03-08 at 14 28 40
Screenshot 2023-03-08 at 14 28 33

### DIFF
--- a/hat/assets/js/apps/Iaso/components/DownloadButtonsComponent.tsx
+++ b/hat/assets/js/apps/Iaso/components/DownloadButtonsComponent.tsx
@@ -21,12 +21,14 @@ type Props = {
     csvUrl?: string;
     xlsxUrl?: string;
     gpkgUrl?: string;
+    disabled?: boolean;
 };
 
 const DownloadButtonsComponent: FunctionComponent<Props> = ({
     csvUrl,
     xlsxUrl,
     gpkgUrl,
+    disabled = false,
 }) => {
     const classes = useDownloadButtonStyles();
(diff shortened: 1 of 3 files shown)


quang-le (commented):
works fine but I wonder if it wouldn't be better to put disable the buttons while `isFetching` and put the `LoadingSpinner` in the button, like with the create bulk user feature. That would prevent the pop up effect:

https://user-images.githubusercontent.com/38907762/223963212-5a395fa6-3313-4184-85c4-95d41cc79c70.mov

beygorghor:
> works fine but I wonder if it wouldn't be better to put disable the buttons while `isFetching` and put the `LoadingSpinner` in the button, like with the create bulk user feature. That would prevent the pop up effect:

Ok for the disable, not for the loadingspinner, it's a bit overkill i think

quang-le:
> > works fine but I wonder if it wouldn't be better to put disable the buttons while `isFetching` and put the `LoadingSpinner` in the button, like with the create bulk user feature. That would prevent the pop up effect:
> 
> Ok for the disable, not for the loadingspinner, it's a bit overkill i think

Fine by me

PR #434, opened 2023-03-08
IA-1905 preparedness google sheets alerts send 4 emails

mathvdh:
There are alerts by email when the preparedness Google sheets do not load properly. Today, we receive them 4 times per day, at the same time. This should be corrected before we configure with the WHO AFRO people’s emails

Also, a same loading issue is mentioned in the email every day. Ideally it should only show once to avoid dissolving the attention of the user

Related JIRA tickets : IA-1905

## Changes

It was decided to move the management command who does this to a task.
To create a `/launch_task/<task_name>/<user_name>/` url that will be called by EBS native cron job management, where : 
- `task_name` is a fully qualified task_name = module path + task fn name, example : 'plugins.polio.tasks.refresh_preparedness_data.refresh_data' 
- `user_name` is the username of the user that will be associated with the task (we decided on  `polio_cron_task_user` for this case)

The EBS cron job is defined by a `cron.yaml` in root of project.

## How to test

You can call the url  `/launch_task/<task_name>/<user_name>/ ` (if in local you need to have a tasks_worker running in the background)

You can then find a task in the admin or tasks page of the dashboard.

## Notes

⚠️ A user called `polio_cron_task_user` need to be created on polio and staging polio and need to have a profile with the good account in order for this to work. ⚠️


iaso/views.py
```diff
@@ -65,3 +69,58 @@ def health(request):
         res["error"] = "db_fail"
 
     return JsonResponse(res)
+
+
+@csrf_exempt
+def task_launcher(request, task_name, user_name):
```
olethanh: would it be possible to have this in the beanstalk_worker module instead?

iaso/views.py
```diff
@@ -65,3 +69,58 @@ def health(request):
...
+        return JsonResponse(
+            {"status": "fail", "error": f"Error while loading user {user_name}", "error_details": str(e)}
+        )
+
+    try:
+        import importlib
+
+        task_parts = task_name.split(".")
+        task_fn_str = task_parts.pop()
+        task_module_str = ".".join(task_parts)
```
olethanh: not important but in case you don't know, you can use rsplit for this
`task_module_str, task_fn_str = task_name.rsplit('.', 1)`

```python
In [2]: task_module, task_fn = "a.b.c".rsplit('.', 1)

In [3]: task_module, task_fn
Out[3]: ('a.b', 'c')
```

iaso/views.py
```diff
@@ -65,3 +69,58 @@ def health(request):
         res["error"] = "db_fail"
 
     return JsonResponse(res)
+
+
+@csrf_exempt
+def task_launcher(request, task_name, user_name):
+    try:
```
olethanh: I would add a special check for if the task is in polio plugin and early and quietly return in that case to not spam useless errors (since most env are not polio but all the cron are in polio ).

You can use for this `apps.is_installed("plugins.polio")`  or check the plugin settings, not sure which is better

iaso/views.py
```diff
@@ -65,3 +69,58 @@ def health(request):
...
+        import importlib
+
+        task_parts = task_name.split(".")
+        task_fn_str = task_parts.pop()
+        task_module_str = ".".join(task_parts)
+
+        task_module = importlib.import_module(task_module_str)
+        the_task_fn = getattr(task_module, task_fn_str)
+
+    except Exception as e:
```
olethanh: add a logging.exception here so we get an error in sentry otherwhise we will miss all the error as they are catched

iaso/views.py
```diff
@@ -65,3 +69,58 @@ def health(request):
...
+        return JsonResponse(
```
olethanh: nitpick but add a status_code (400 IIRC ?)

iaso/views.py
```diff
@@ -65,3 +69,58 @@ def health(request):
...
+            call_args = {**call_args, **body_json}
+
+        if len(request.GET) > 0:
+            call_args = {**call_args, **request.GET}
+
+        the_task = the_task_fn(**call_args)
+        return JsonResponse({"status": "OK", "task": the_task.as_dict()})
+
+    except Exception as e:
+        return JsonResponse(
```
olethanh: idem, logging.exception and status_code

.ebextensions/70_crontab.config
```diff
@@ -28,43 +28,10 @@ files:
...
       [Install]
         WantedBy=timers.target
 
 # according to Amazon doc these are run after the file are copied with root user
 container_commands:
   90enable_cron_weekly_email:
-    command: "systemctl enable polio_weekly_email.timer && systemctl start polio_weekly_email.timer"
-  91enable_cron_refresh_preparedness:
-    command: "systemctl enable polio_refresh_preparedness.timer  && systemctl start polio_refresh_preparedness.timer"
+    command: "systemctl enable polio_weekly_email.timer && systemctl start polio_weekly_email.timer"
```
olethanh, line 37: this command also need to be moved

olethanh:
a few comment, and what we discussed this morning. but otherwhise the code seems to be working quite well

mathvdh:
@olethanh I moved the URLS as discussed an added some tests ...
I also added a `IN_TESTS` var in `settings.py` to catch this specific case

mathvdh:
The changes in the imports of some files are just because I'm using the "Sort imports" on files now (using isort) ..

beanstalk_worker/views.py
```diff
@@ -22,3 +25,54 @@ def run_all(request):
...
+        return JsonResponse(
+            {"status": "fail", "error": f"Error while loading user {user_name}", "error_details": str(e)}
+        )
+
+    try:
+        import importlib
+
+        task_parts = task_name.split(".")
+        task_fn_str = task_parts.pop()
+        task_module_str = ".".join(task_parts)
```
olethanh: not important but in case you don't know, you can use rsplit for this
`task_module_str, task_fn_str = task_name.rsplit('.', 1)`

```python
In [2]: task_module, task_fn = "a.b.c".rsplit('.', 1)

In [3]: task_module, task_fn
Out[3]: ('a.b', 'c')
```

beanstalk_worker/views.py
```diff
@@ -22,3 +25,54 @@ def run_all(request):
...
+def task_launcher(request, task_name, user_name):
+    try:
+        the_user = User.objects.get(username=user_name)
+    except User.DoesNotExist:
+        return JsonResponse({"status": "fail", "error": f"User not found for {user_name}"})
+    except Exception as e:
+        return JsonResponse(
+            {"status": "fail", "error": f"Error while loading user {user_name}", "error_details": str(e)}
+        )
+
```
olethanh, line 53: I would add a special check for if the task is in polio plugin and early and quietly return in that case to not spam useless errors (since most env are not polio but all the cron are in polio ).

You can use for this `apps.is_installed("plugins.polio")`  or check the plugin settings, not sure which is better

beanstalk_worker/views.py
```diff
@@ -22,3 +25,54 @@ def run_all(request):
...
+
+@csrf_exempt
+def task_launcher(request, task_name, user_name):
+    try:
+        the_user = User.objects.get(username=user_name)
+    except User.DoesNotExist:
+        return JsonResponse({"status": "fail", "error": f"User not found for {user_name}"})
+    except Exception as e:
+        return JsonResponse(
```
olethanh, line 49: add a logging.exception here so we get an error in sentry otherwhise we will miss all the error as they are catched.

Add a status code as well, like 400?

beanstalk_worker/views.py
```diff
@@ -22,3 +25,54 @@ def run_all(request):
...
+
+        task_parts = task_name.split(".")
+        task_fn_str = task_parts.pop()
+        task_module_str = ".".join(task_parts)
+
+        task_module = importlib.import_module(task_module_str)
+        the_task_fn = getattr(task_module, task_fn_str)
+
+    except Exception as e:
+        return JsonResponse(
```
olethanh, line 75: idem, logging.exception and status_code

beanstalk_worker/views.py
```diff
@@ -22,3 +25,54 @@ def run_all(request):
...
+            call_args = {**call_args, **request.POST}
+
+        if len(request.GET) > 0:
+            call_args = {**call_args, **request.GET}
+
+        the_task = the_task_fn(**call_args)
+        return JsonResponse({"status": "success", "task": the_task.as_dict()})
+
+    except Exception as e:
+        return JsonResponse(
```
olethanh, line 97: idem, logging.exception and status_code

olethanh (approved):

PR #435, opened 2023-03-08
IA-1326 Fix issue with EMPTY GEOM

olethanh:
we had OrgUnit with empty geom in the database that could not be edited anymore because they had the issue `Column has Z dimension but geometry does not`

they were created via the GPKG import or directly in the database by inserting a POINT Z EMPTY, but they couldn't be edited anymore from django because we got a 2d empty point in return.

After discussion with the geo type we decided that it's better to replace empty point with `null` value in database.

So this patch fix the issue by:
replacing Empty geom with None during GPKG import
Allow editing existing org unit by fixing their value in the save()

Related JIRA tickets : IA-1326

- [NA] New translations have been added or updated if new string have been introduced in the frontend
- [x] Are there enough tests
- [NA] Documentation has been included (for new feature)

## How to test

Ask the geo team

## Notes
Don't know if we should bother doing a DB data migration for all the existing orgunit? It might help for bulk_update operation or other stuff that skip the save(), not sure if we have any. But on the other hand the migration might block because of the group trigger like the one which added source_updated_at

### DIFF
--- a/iaso/gpkg/import_gpkg.py
+++ b/iaso/gpkg/import_gpkg.py
@@ -1,3 +1,4 @@
+import math
 import sqlite3
 from copy import deepcopy
 from typing import Dict, List, Optional, Tuple, Union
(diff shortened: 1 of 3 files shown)


mlennert (commented):
Hi @olethanh, I cannot really test the changes, so just a theoretical response, and probably mostly for my own education: what are the reasons that you choose to solve the problem in convert_to_geography(), and not create_or_update_orgunit(), avoding the call to convert_to_geography() ? Why not enhance the if geometry test at line 105 to also detect the case of an empty geometry ?

olethanh:
I had answered Moritz IRL but to have a written down trace, it's in convert_to_geography so we can reuse this function in the future without having to worry what kind of geometry we call it with

PR #445, opened 2023-03-15
POLIO-870 Campaign Data Emails Sent Four Times

mathvdh:
Weekly Campaign email was sent multiple times, 

Related JIRA tickets : POLIO-870

## Changes

- Switch to the new cron system as done in #434 
- Transform the management command to a task
- Remove old management command and folders
- Remove Old crontab files from EBS config files

## How to test

visit the URL mentioned `cron.yaml` file :
`/tasks/launch_task/plugins.polio.tasks.weekly_email.send_email/<YOUR_USERNAME>/`

Should return some JSON looking like  : 

```JSON
{
    "status": "success",
    "task": {
        "id": 16,
        "created_at": 1678885064.050157,
        "started_at": null,
        "ended_at": null,
        "params": {
            "args": [],
            "kwargs": {},
            "module": "plugins.polio.tasks.weekly_email",
            "method": "send_email"
        },
        "result": null,
        "status": "QUEUED",
        "launcher": {
            "id": 1,
            "first_name": "",
            "user_name": "<YOUR_USERNAME>",
            "last_name": "",
            "email": "<YOUR_EMAIL>",
            "language": null,
            "user_id": 1
        },
        "progress_value": 0,
        "end_value": 0,
        "name": "refresh_data",
        "queue_answer": {
            "result": "recorded into DB"
        },
        "progress_message": null,
        "should_be_killed": false
    }
}
```


olethanh (approved):
thanks. had a crash because we were trying to set the pk for the campaigns inside the progress_value but one is a uuid and the other is a int. So I fixed it using an enumerate instead. (also added a counter for the number of email sent)

PR #446, opened 2023-03-15
POLIO-830: connect budget tool with budget tab

quang-le:
When users create steps in the budget tool, the data should populate the budget tab in the main campaign form

Related JIRA tickets : POLIO-830

- [ ] Are there enough tests

## Changes

- Added logic to the `TransitionToSerializer` so that the corresponding field in the budget tab is written with the creation date of step. Caveats:
    - The current implementation only writes if there is no preexisting data in the corresponding field in the `Campaign`model, i.e: the field is empty in the budget tab. That means that we only capture the first round of feedback/response when they happen
    - Repeat steps are not taken into account
    - Override steps: moving backwards will remove existing data in the budget tab for the steps that come after the override destination
    - Override steps: moving forward will leave a gap in data for the skipped steps. This may cause trouble in the dashboards
    - Parallel path (GPEI Consolidation, approval): this is handled by code that is tailor made for the current version of the workflow. It should be updated if the workflow changes or it will break.
- The `budget_status` of the campaign is also updated 
- The `budget_responsible` field has been removed (hidden) in the front-end as it is obsolete. There is a separate issue to improve the layout following this change: https://bluesquare.atlassian.net/browse/POLIO-898

## How to test
You need a config for budget: [wiki](https://wiki.bluesquare.org/en/Dev-team/product-management
/iaso/polio_budget_workflow)

The current budget config (json) can be found on iaso staging (polio account)

Create a step in a campaign, then check the budget tab for this campaign:
- It should be blocked for edition
- You should see the creation date of your step in the corresponding field

## Print screen / video

https://user-images.githubusercontent.com/38907762/225331567-7c9d9304-b95a-4263-9292-61e5ac118c96.mov

### UPDATE 17/03: save budget state

Screenshot 2023-03-17 at 17 32 57


quang-le:
Still needs to update budget_status (for calendar) and to auto-fill responsible

beygorghor (approved):

olethanh (approved):

PR #454, opened 2023-03-16
Hotfix pagination

olethanh:
hotfix


olethanh:
I'm mergin this, this is the same as the other already merged pr but like this we wont break history and the tags

PR #461, opened 2023-03-21
IA-1990 Fix Botswana preparedness google sheet

olethanh:
by enabling new title for the score box that we use to detect the google sheet

Related JIRA tickets : IA-1990

- [NO ] Are there enough tests


quang-le (approved):
Works on my machine 😀, though client should be aware that if they fill out data in the wrong column (REGION) they will have funny results

Screenshot 2023-03-21 at 11 36 16

PR #471, opened 2023-03-23
Polio-839 export budget list as csv

quang-le:
We need to export budget list as csv

Related JIRA tickets : POLIO-839

- [ ] Are there enough tests - Almost
- [ ] Documentation has been included (for new feature) - It will be

## Changes

- Back-end:
    - Add `CSVExportMixin` class
    - Add `ExportCampaignBudgetSerializer`, with `fields` to export and `labels` for the columns
    - Use the mixin and serializer to `BudgetCampaignViewSet`
    - set `export_filename` and `exporter_serializer_class` properties to `BudgetCampaignViewSet`
- Front-end:
    - Add `CsvButton`for desktop and mobile layouts
    - filter out pagination params from the csv query 

## How to test

Explain how to test your PR.
- Go to polio>budget
- play around with the filters and the csv button
- the data in the csv should match the filters applied
- Try as well in mobile layout

## Print screen / video
Desktop

Screenshot 2023-03-23 at 13 53 24

Mobile

Screenshot 2023-03-23 at 13 53 50

### DIFF
--- a/plugins/polio/js/src/pages/Budget/index.tsx
+++ b/plugins/polio/js/src/pages/Budget/index.tsx
@@ -45,15 +45,6 @@ import { CsvButton } from '../../../../../../hat/assets/js/apps/Iaso/components/
 type Props = {
     router: any;
 };
-const style = () => {
-    return {
-        pagination: {
-            '&.MuiPagination-root > .MuiPagination-ul': {
-                justifyContent: 'center',
-            },
-        },
-    };
-};
 
 const getCsvParams = (apiParams: Record<string, any>): string => {
     const {
@@ -72,7 +63,14 @@ const getCsvParams = (apiParams: Record<string, any>): string => {
     return new URLSearchParams(filteredParams).toString();
 };
 
-const usePaginationStyles = makeStyles(style);
+const usePaginationStyles = makeStyles({
+    pagination: {
+        '&.MuiPagination-root > .MuiPagination-ul': {
+            justifyContent: 'center',
+        },
+    },
+    alignRight: { textAlign: 'right' },
+});
 
 export const BudgetList: FunctionComponent<Props> = ({ router }) => {
     const { params } = router;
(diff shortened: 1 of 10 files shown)


plugins/polio/js/src/pages/Budget/index.tsx
```diff
@@ -53,6 +55,23 @@ const style = () => {
...
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        pageSize,
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        page,
+        ...paramsForCsv
+    } = apiParams;
+    const filteredParams: Record<string, any> = Object.fromEntries(
+        Object.entries(paramsForCsv).filter(
+            // eslint-disable-next-line no-unused-vars, @typescript-eslint/no-unused-vars
+            ([_key, value]) => value !== undefined,
```
beygorghor, line 60: didi you try only "_" instead of "_key" ? linter should not complain

plugins/polio/js/src/pages/Budget/index.tsx
```diff
@@ -125,6 +152,13 @@ export const BudgetList: FunctionComponent<Props> = ({ router }) => {
                             params={params}
                             statesList={possibleStates}
                         />
+                        <Grid container justifyContent="flex-end">
```
beygorghor: Grid is not mandatory here, you can use a Box

beygorghor (approved):
Nothing blocking

plugins/polio/js/src/pages/Budget/index.tsx
quang-le: I can't use `flex-end` on a `Box`

plugins/polio/js/src/pages/Budget/index.tsx
quang-le: Actually it should not complain with `_key` either, but somehow it does anyway

quang-le: The linter complains anyway. i think we have discrepancies between vanilla iaso and polio linters

plugins/polio/js/src/pages/Budget/index.tsx
quang-le: ok `textAlign:right` got it

olethanh (approved):

PR #472, opened 2023-03-23
IA-1983: registry

beygorghor:
New registry domain.

[Mock-ups](https://whimsical.com/microplanning-cmr-9o1A7M8MaV6H5Bm2guUD4t) (only two first screens)

An org unit can be defined with a reference submission. You can basically add a submission to an org unit.
Org unit is normally of a specific org unit type.
This org unit type has sub org unit types.
The idea is to show: 
-  an org unit with his reference submission 
- all the submissions made of a specific form on the children org units  (included in the sub org unit types)

For the reference submission i used this form:
[microplan_obc_2022112302.xlsx](https://github.com/BLSQ/iaso/files/11087553/microplan_obc_2022112302.xlsx)

For the reference form of the children i used this form:
[microplan_localite_2022112301.xlsx](https://github.com/BLSQ/iaso/files/11087557/microplan_localite_2022112301.xlsx)

Related JIRA tickets : IA-1983

- [x] New translations have been added or updated if new string have been introduced in the frontend
- [ ] Are there enough tests

## Changes

- new registry domain
- new `useEnketoUrl` hook
- new tools to calculate map Bounds in `hat/assets/js/apps/Iaso/utils/mapUtils.ts`

## How to test

- Create two new forms with the XLSX provided.
- Edit the second form in the admin and add a few `label_keys`
- make a submission for the first one and a few of the second
(you can by pass those creation and use existing submissions.)
- Select an org unit having a shape, children with geolocation
- Make sure the org unit type of this org unit include the sub org unit types of the previous children
- attach the main submission to the main org unit with the admin (edit the org unit and provide the instance id)
- attach the children submissions to the children org unit too
- Navigate to the Registry entry in the dashboard menu
Screenshot 2023-03-28 at 11 11 27

- Select the main org unit (here "Aketi Zone de Santé") in the tree and click on See registry 
Screenshot 2023-03-28 at 11 21 18

- Select the form used to create children submission in the form selector
Screenshot 2023-03-28 at 11 22 39

- You should be able to see instances list with pre selected columns (the `label_keys`)
- Play with the map, the table, the visible columns and with enketo if enabled.

PS: I have translations issue with the second form, there is a [JIRA](https://bluesquare.atlassian.net/browse/IA-2012) ticket for it


hat/assets/js/apps/Iaso/domains/instances/utils/index.tsx
```diff
@@ -161,9 +160,12 @@ const renderValue = (settings: Setting<Instance>, c: VisibleColumn) => {
 export const useGetInstancesColumns = (
     showDeleted = false,
     // eslint-disable-next-line no-unused-vars
+    getActionCell: (settings: any) => ReactElement = settings => (
```
quang-le: Maybe we should put in the style guide to create types for arrow functions. It would make declarations easier to read. I do it as well, but it's a bit confusing I think, compared to something like:
```javascript
type RenderCell = (settings:any) => ReactElement
// ...some code

getActionCell: RenderCell = settings => // ... rest
```

hat/assets/js/apps/Iaso/domains/instances/config.js
```diff
@@ -1,22 +1,7 @@
-import React from 'react';
 import orderBy from 'lodash/orderBy';
 
 import { INSTANCE_METAS_FIELDS } from './constants';
 import MESSAGES from './messages';
-import ActionTableColumnComponent from './components/ActionTableColumnComponent';
-
-export const actionTableColumn = (formatMessage = () => ({}), user) => {
```
quang-le, line 8: ❤️

hat/assets/js/apps/Iaso/domains/orgUnits/components/TreeView/requests.js
```diff
@@ -4,9 +4,9 @@ import { dispatch } from '../../../../redux/store';
 import { enqueueSnackbar } from '../../../../redux/snackBarsReducer';
 import { errorSnackBar } from '../../../../constants/snackBars';
 
-export const getChildrenData = id => {
+export const getChildrenData = (id, validationStatus = 'all') => {
```
quang-le, line 7: This is strange, I had the feeling I had done this change already in a previous PR 🤔🤷‍♂️

hat/assets/js/apps/Iaso/domains/registry/hooks/useGetInstances.tsx
```diff
@@ -0,0 +1,35 @@
...
+        showDeleted: false,
+    };
+    const url = makeUrlWithParams('/api/instances/', apiParams);
+    return useSnackQuery({
+        queryKey: ['registry-instances', params, orgUnitTypeId],
+        queryFn: () => getRequest(url),
+        options: {
+            enabled: Boolean(orgUnitTypeId && params.formIds),
+            staleTime: 1000 * 60 * 15, // in MS
+            cacheTime: 1000 * 60 * 5,
```
quang-le, line 32: ❤️

hat/assets/js/apps/Iaso/domains/registry/components/OrgUnitMap.tsx
```diff
@@ -0,0 +1,222 @@
...
+
+export const OrgUnitMap: FunctionComponent<Props> = ({
+    orgUnit,
+    subOrgUnitTypes,
+    childrenOrgUnits,
+}) => {
+    const map: any = useRef();
+    const { formatMessage } = useSafeIntl();
+    const theme = useTheme();
+    const [currentTile, setCurrentTile] = useState<Tile>(tiles.osm);
```
quang-le: `tiles`should be all caps, as its a constant

hat/assets/js/apps/Iaso/domains/registry/config.tsx
quang-le, line 1: This sooooooooo much more legible now 😀

hat/assets/js/apps/Iaso/domains/registry/details.tsx
```diff
@@ -0,0 +1,141 @@
...
+                                id="form-contents"
+                                className={classes.paper}
+                                title={formatMessage(MESSAGES.submission)}
+                                IconButton={IconButton}
+                                iconButtonProps={{
+                                    onClick: () => getEnketoUrl(),
+                                    overrideIcon: EnketoIcon,
+                                    color: 'secondary',
+                                    tooltipMessage: MESSAGES.editOnEnketo,
+                                }}
```
quang-le, line 128: That icon button should be hidden if the user does not have permission to edit submissions

hat/assets/js/apps/Iaso/domains/registry/details.tsx
```diff
@@ -0,0 +1,141 @@
...
+    );
+
+    const getEnketoUrl = useGetEnketoUrl(
+        window.location.href,
+        orgUnit?.reference_instance,
+    );
+    return (
+        <>
+            <TopBar
+                title={formatMessage(MESSAGES.title)}
```
quang-le, line 84: Maybe the title should mention the org unit name

hat/assets/js/apps/Iaso/domains/registry/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+                            preselected={selectedOrgUnitsIds}
+                            preexpanded={selectedOrgUnitParents}
+                            selectedData={selectedOrgUnits}
+                            onUpdate={onUpdate}
+                            multiselect={false}
+                            inputLabelObject={MESSAGES.search}
+                        />
+                    </Grid>
+                    <Grid item xs={12} md={6}>
+                        {selectedOrgUnits?.length > 0 && (
```
quang-le, line 106: I think it would look less empty if we showed the Paper with  placeholder values when there's no org unit selected yet. Or alternately have an animation/transition i.o the Paper just popping up, which doesn't give a great effect

hat/assets/js/apps/Iaso/utils/mapUtils.ts
quang-le, line 1: 🦾

hat/assets/js/apps/Iaso/utils/mapUtils.ts
```diff
@@ -206,3 +206,91 @@ export const polygonDrawOption = (
...
+    zoom: number;
+};
+type Map = {
+    leafletElement: {
+        // eslint-disable-next-line no-unused-vars
+        fitBounds: (bounds: Bounds, boundsOptions) => void;
+    };
+};
+
+export const getOrgUnitBounds = (orgUnit: OrgUnit): Bounds | undefined => {
```
quang-le, line 235: nitpick: `getSingleOrgUnitBounds` or `getOrgUnitsListBounds` for the other function would be better names in my opinion. Easier and quicker to tell the 2 func apart

hat/assets/js/apps/Iaso/utils/mapUtils.ts
```diff
@@ -206,3 +206,91 @@ export const polygonDrawOption = (
...
+    }
+};
+
+export const defaultViewport: DefaultViewPort = {
+    center: [1, 20],
+    zoom: 3.25,
+};
+export const boundsOptions: BoundsOptions = {
+    padding: [50, 50],
+};
```
quang-le: all caps name

quang-le (changes requested):
Impressive work, super clean code. Great to see so many components reused.
Mostly details.

Worth noting:
- The thing with the enketo button and the permission
- The feature needs the `iaso-org_units` permission to work. Not a problem per se, but worth mentioning. It will probably be of interest for @hakifran  and the user profiles feature

hat/assets/js/apps/Iaso/domains/registry/details.tsx
beygorghor: the name of the org unit is in the title of the map

hat/assets/js/apps/Iaso/domains/registry/index.tsx
beygorghor: Yes the paper is not on the mock-up, let's wait for customer / Martin feed-back first

hat/assets/js/apps/Iaso/domains/registry/details.tsx
beygorghor: I would ask @madewulf first

olethanh (approved):
It's quite nice I like it.

Didn't review the whole JS code.  but I didn't encounter any problem and the backend is ok for me.

It's minor but the center the map button don't seems to work for me, when I'm on Sierra Leone from the test data set from DHIS2/seed. I think it's because it doesn't have a shape but the districts have a shape so there is still something on the map.

I have a few nitpick, improvement suggestion for the UI but it is nothing blocking : 

Wouldn't it make more sense to not have two separate page between the first page with the  picker and the "registry" for the orgunit. We could probably merge the two.

image
Shouldn't this be "Search an orgunit" ?
Not sure it make sense to keep the grey background either. I know it's in the ou picker but it make it looks like a debug thing


image
not sure period make sense

image
We should probably rename it to something else than submission and hide the question key.

Also it would be nice on the list of child org. unit to have a button/link  to navigate to  the registry for that orgunit without having to go back

PR #479, opened 2023-03-28
Polio-918: fix budget pagination bug

quang-le:
In budegt list, filtering on status, then using table pagination would result in a 404
Related JIRA tickets : POLIO-918

- [ ] Are there enough tests

## Changes

- Updated polio config so that url params and api params are the same
- Updated `BudgetFilters`'s `keyValues to reflect the changes
- Removed `resetPageToOne`related code as it's deprecated and was actually causing a bug
- Removed dead code that was a legacy of the first version of the budget tool

## How to test

Go to Budget list, filter on status, then use table pagination
Do the same with filter on country

## Print screen / video

https://user-images.githubusercontent.com/38907762/228274056-73921f49-e6b1-49e8-81e0-7c5dd95a69b9.mov


codeKameleon (approved):

beygorghor (approved):
i'm still having a "-" in the url while selecting "no budget submitted"
Screenshot 2023-03-31 at 12 01 08

quang-le:
> i'm still having a "-" in the url while selecting "no budget submitted" Screenshot 2023-03-31 at 12 01 08

That's kinda normal. It's the default "key" for when the field is `None` in the backend

PR #490, opened 2023-03-31
POLIO-934 timeline widget bug

quang-le:
The timelinewidget never appeared to be completed even with every step completed

Related JIRA tickets : POLIO-934

- [ ] Are there enough tests

## Changes

- The problem is that there' a check on unique labels to verify that all steps have been completed (made a bit complicated by the existence of override steps). The list of unique labels is a set generated from a list of labels. The problem is that the workflow json config had steps that had teh same label. I update the config on campaigns, campaigns-staging and staging, then the related translations in the front-end for consistency

## How to test

- go the budget page and go through a workflow until your each the last step

## Print screen / video

Screenshot 2023-03-31 at 13 32 46

## Notes

The logic should be refactored to take the keys into account i.o the labels, but that will be part of [another ticket](https://bluesquare.atlassian.net/browse/POLIO-936).


beygorghor (approved):
Can you update the YAML as well please ?

PR #533, opened 2023-04-20
POLIO-973: add read-only endpoint for configs

quang-le:
Configs need to be accessed easily to (automate) check if they are up to date for upcoming campaigns

Related JIRA tickets : POLIO-973

- [X] Is my code clear enough and well documented - I think so
- [X] Are there enough tests

## Changes

- Add a users field to Config model
- Add an a ViewSet that will return the Configs in which the current user in part of its users
- Only allow GET requests
- Update Django admin to be able to add users to configs

## How to test

- Try to access `/api/polio/configs`. You should get a 403
- Login to polio
- Make sure your DB has some configs (if it's recent it should have some)
- Try to access `/api/polio/configs`. You should get an empty list
- Go to Django admin
- Add your user to `Users` in one or several configs
- Try to access `/api/polio/configs`. You should get a list with the configs you added your user to
- Try to access `/api/polio/configs/<key>`, where `key` is the `key` field of one of the configs you added your user to. You should get the config 

## Print screen / video

Well, I'm not going to post a screenshot with sensitive data on it am I?

## Notes

I implemented a check on the `User` but not on the permissions, or the accounts. I don't know if I should add it of if it would be redundant.


mathvdh (approved):
Rien à dire, @quang-le est prêt pour faire du backend ;)

PR #537, opened 2023-04-21
IA-2063: Register, add button to create a children

beygorghor:
Add a button to create a new org unit as the children of the current one.
Related JIRA tickets : IA-2063

## Self proofreading checklist

- [x] Did I use eslint and black formatters
- [x] Is my code clear enough and well documented
- [x] Are my typescript files well typed
- [ ] New translations have been added or updated if new strings have been introduced in the frontend
- [ ] My migrations file are included
- [ ] Are there enough tests
- [ ] Documentation has been included (for new feature)

## Changes

- accept levels as default param for parent while creating a org unit
- fetch parent info end set it as parent to the new org unit
- custom paper component for registry
- add link to create children org unit with current org unit as level in the url
- placeholder on map while loading list of children

## How to test

Go to register for an existing org unit.
You should see a + button nearby the name of the org unit, if you click on it you should arrive on the org unit creation page with the org unit selected in registry as parent 

## Print screen / video

https://user-images.githubusercontent.com/12494624/233610010-6f303b2b-9d17-4000-ad6a-3e0259e85455.mov

## Notes

⚠️ Current base is this [PR](https://github.com/BLSQ/iaso/pull/535)

PR #539, opened 2023-04-21
IA-2074: registry toggle types

beygorghor:
This should be interactive and allow to toggle on and off the org units shown on the map

Related JIRA tickets : IA-2074,

- [ ] Are there enough tests

## Changes

- change legend to accept active prop
- custom map legend component toggling active org unit
- show / hide active org units
- refactor fit to bounds to fit only active org units

## How to test

Open a registry for an org unit whit su b org units (with geo locations)
You should be able to show/hide those children types

## Print screen / video

https://user-images.githubusercontent.com/12494624/233645142-bac5d4a9-a8ec-47ae-9179-d9cb095e55f3.mov

## Notes

⚠️ Curent base is this [PR](https://github.com/BLSQ/iaso/pull/537)

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/registry/components/MapLegend.tsx
+++ b/hat/assets/js/apps/Iaso/domains/registry/components/MapLegend.tsx
@@ -1,0 +1,101 @@
+import React, {
+    FunctionComponent,
+    Dispatch,
+    SetStateAction,
+    useCallback,
+} from 'react';
+import { Paper, makeStyles, Box } from '@material-ui/core';
+import { cloneDeep } from 'lodash';
+import CheckIcon from '@material-ui/icons/Check';
+
+import { Legend } from '../hooks/useGetLegendOptions';
+
+const useStyles = makeStyles(theme => ({
+    root: {
+        position: 'absolute', // assuming you have a parent relative
+        zIndex: 500,
+        top: 'auto',
+        left: 'auto',
+        right: theme.spacing(1),
+        bottom: theme.spacing(3),
+        width: 'auto',
+    },
+    option: {
+        cursor: 'pointer',
+    },
+    activeOption: {
+        cursor: 'pointer',
+    },
+    roundColor: {
+        borderRadius: 20,
+        height: 20,
+        width: 20,
+        display: 'inline-block',
+        marginRight: theme.spacing(1),
+    },
+    mapLegendLabel: {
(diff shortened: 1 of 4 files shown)


hat/assets/js/apps/Iaso/domains/registry/components/MapLegend.tsx
```diff
@@ -0,0 +1,100 @@
...
+}) => {
+    const classes = useStyles();
+    const handleClick = useCallback(
+        (optionIndex: number): void => {
+            const newOptions = cloneDeep(options);
+            if (newOptions && newOptions[optionIndex]) {
+                newOptions[optionIndex].active =
+                    !newOptions[optionIndex].active;
+                setOptions(newOptions);
+            }
```
quang-le: from what I read while trying out the Immer library, we don't necessarily need to `cloneDeep`, we can spread the original object, and overwrite the fields we replace (which is less expensive than deep copying the full object).

So something like :
```typescript
const newOptions = [...options];
if(newOption?.[optionIndex]){
    newOptions[optionIndex].active = !newOptions[optionIndex].active;
    setOptions(newOptions);
}
```
 should work;
 https://immerjs.github.io/immer/#without-immer

hat/assets/js/apps/Iaso/domains/registry/hooks/useGetLegendOptions.ts
```diff
@@ -0,0 +1,39 @@
...
+        if (orgUnit) {
+            options.unshift({
+                value: `${orgUnit.id}`,
+                label: formatMessage(MESSAGES.selectedOrgUnit),
+                color: theme.palette.secondary.main,
+                active: true,
+            });
+        }
+        return options;
+    };
```
quang-le, line 37: shouldn't this be in a useCallback?

quang-le: Not really since it's only used to initialize state an then never called again

quang-le (approved):

hat/assets/js/apps/Iaso/domains/registry/hooks/useGetLegendOptions.ts
beygorghor: yep it was wrapped in a useCallback but as it is only call once

PR #559, opened 2023-04-27
POLIO-942: Implement filter by org unit groups on Campaigns, SIA calendar and Budget tool

hakifran:
- There was not orgUnit groups filter on campaigns, calendar and budgets pages
- This PR implements the filter by org unit groups on Campaigns, SIA calendar and Budget tool

Related JIRA tickets : [POLIO-942](https://bluesquare.atlassian.net/browse/POLIO-942)

- [ ] Are there enough tests

## Changes
- Add a to_display field on group model(to be able to select which groups to see in select input)
- Add a checkbox on the groupsdialog to be able to edit the to_display field
- filter groups based on the to_display field
- Add select box which contains orgUnit groups on campaign, calendar and budget pas
- Filter on orgUnit groups in backend campaigns

### Filter campaigns
- Go on campaign, calendar or budget page
- Select one or multiple orgUnit groups in the input select(group)
- Try to filter
### Edit a group to make it displayed
- Go to orgUnits -> Groups
- Try to create or edit a group
- Check group to be displayed checkbox
- Save

### Edit a group to make it displayed
Peek 2023-04-27 17-40

### Filter campaigns
Peek 2023-04-27 17-41

[POLIO-942]: https://bluesquare.atlassian.net/browse/POLIO-942?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hakifran:
@cypress

plugins/polio/budget/api.py
```diff
@@ -59,6 +59,11 @@ def get_queryset(self) -> QuerySet:
 
     def filter_queryset(self, queryset):
         queryset = super().filter_queryset(queryset)
+
+        org_unit_groups = self.request.query_params.get("orgUnitGroups")
+        if org_unit_groups:
+            queryset = queryset.filter(initial_org_unit__groups__in=org_unit_groups.split(","))
```
olethanh: I think this  should filter on Campaign.country, initial_org_unit is not guaranteed to be a country (the `country` field is automatically calculated from the initial_org_unit)

plugins/polio/api.py
```diff
@@ -185,6 +187,8 @@ def filter_queryset(self, queryset):
             campaigns = campaigns.filter(is_preventive=False).filter(is_test=False)
         if campaign_groups:
             campaigns = campaigns.filter(grouped_campaigns__in=campaign_groups.split(","))
+        if org_unit_groups:
+            campaigns = campaigns.filter(initial_org_unit__groups__in=org_unit_groups.split(","))
```
olethanh: idem filter on `country` not `initial_org_unit`

olethanh (changes requested):
The backend code seems sound but the  problem I see with this is that this is in the Iaso code/ frontend, even without the Polio plugin. Also it's not very clear what this field do in when you encounter it (maybe it should add some doc?)

Not sure what would be a better approach, would using a GroupSet (group of group) with a hard coded source_ref be a possible way?

plugins/polio/budget/api.py
hakifran: Fixed

plugins/polio/api.py
hakifran: fixed

olethanh (approved):

hat/assets/js/apps/Iaso/domains/orgUnits/hooks/requests/useGetGroups.ts
```diff
@@ -12,22 +11,31 @@ import { staleTime } from '../../config';
...
 
-const makeGroupsQueryParams = ({ dataSourceId, sourceVersionId }) => {
+const makeGroupsQueryParams = ({
+    dataSourceId,
+    sourceVersionId,
+    blockOfCountries,
+}) => {
     if (sourceVersionId) return `?version=${sourceVersionId}`;
     if (dataSourceId) return `?dataSource=${dataSourceId}`;
+    if (blockOfCountries) return `?blockOfCountries=${blockOfCountries}`;
```
quang-le, line 24: you could probably have the type of `blockOfCountries` set to `boolean` and rewrite the condition:
```javascript
if (blockOfCountries) return `?blockOfCountries=True`
```

That way you could pass `true` instead of `True`when you call your hook in the component. It's not really blocking, though

plugins/polio/js/src/components/campaignCalendar/Filters.js
```diff
@@ -78,14 +87,17 @@ const Filters = ({ router, disableDates, disableOnlyDeleted, showTest }) => {
...
+        campaignGroups,
+        orgUnitGroups,
         router,
         dispatch,
     ]);
     const { data, isFetching: isFetchingCountries } = useGetCountries();
     const { data: groupedCampaigns, isFetching: isFetchingGroupedGroups } =
         useGetGroupedCampaigns();
+    const { data: groupedOrgUnits, isFetching: isFetchingGroupedOrgUnits } =
+        useGetGroups({ blockOfCountries: 'True' });
```
quang-le, line 100: See previous comment

quang-le (changes requested):
There is a bug with the campaign page when we try to order the table. 
My guess is it has to do with your `isCalendar` props which changes the order of the filters. It probably changes also the order of the params in the url, and react-router doesn't like when we change the order of the params, so it kicks us to the 404 page

https://user-images.githubusercontent.com/38907762/235915003-a32fa776-dc57-4302-9e88-7834f7daf9c2.mov

quang-le (approved):

PR #565, opened 2023-05-02
IA-2115: Org uit details fi to bounds

beygorghor:
Fi to bounds feature on this page was not working well.

It was using, forms, org unit types and sources value to try to fit on the map, but those value where always empty (need leaflet upgrade or `useRef`)

Now the map should only fit once while you have a current org unit and / or a parent with a shape or a point

Related JIRA tickets : IA-2115

 

https://user-images.githubusercontent.com/12494624/235629563-d43cb58f-7c9f-4823-a7a3-505c4534a1ff.mp4

- [ ] Are there enough tests

## Changes

- remove fit to bounds from sources and org unit types
- remove forms, sources and org unit types form fi to bounds computation, was not working at all
- local state to fit only once when current org unit and parent shapes or point are loaded

## How to test

- create an org unit with a parent (having a shape)
-  zoom a bit on the map
-  add a location marker and save
- the map should not zoom automatically to the first point

## Print screen / video

Uploading Screen Recording 2023-05-02 at 11.26.33.mov…

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/orgUnits/components/orgUnitMap/OrgUnitMap/OrgUnitMap.tsx
+++ b/hat/assets/js/apps/Iaso/domains/orgUnits/components/orgUnitMap/OrgUnitMap/OrgUnitMap.tsx
@@ -99,6 +99,7 @@ export const OrgUnitMap: FunctionComponent<Props> = ({
     const didLocationInitialize = useRef(false);
     const didCatchmentInitialize = useRef(false);
     const [isCreatingMarker, setIsCreatingMarker] = useState<boolean>(false);
+    const [isMapFitted, setIsMapFitted] = useState<boolean>(false);
     // eslint-disable-next-line no-unused-vars, @typescript-eslint/no-unused-vars
     const [state, setStateField, _, setState] = useFormState(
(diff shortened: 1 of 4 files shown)


hat/assets/js/apps/Iaso/domains/orgUnits/components/orgUnitMap/OrgUnitMap/OrgUnitMap.tsx
```diff
@@ -109,7 +110,7 @@ export const OrgUnitMap: FunctionComponent<Props> = ({
         fetchInstanceDetail,
         fetchSubOrgUnitDetail,
     } = useRedux();
-
+    // console.log('state', state);
```
quang-le, line 113: This can be deleted

quang-le (changes requested):
It doesn't refit to bounds when creating the org unit, but it when coming back to the org unit, `fitToBounds` has a strange behaviour

https://user-images.githubusercontent.com/38907762/235640593-baa5c557-035d-4015-8bce-ef3c725a846d.mov

quang-le (approved):

PR #576, opened 2023-05-04
IA-2132 upgrade postgresql version used in CI

olethanh:
IA-2132 The old 10 version was used in CI while locally and in prod we use 12

10 is not supported anymore
Also we used a unofficial image, this now use the official postgis image


olethanh:
@cypress

mathvdh (approved):
Good to use the official image :)

PR #582, opened 2023-05-08
POLIO-977: show districts without location in scope

beygorghor:
Today districts need to have a shape or at least a geographical point in order to appear in the list of districts when selecting the scope in “Campaigns”.

This is blocking because when districts do not have a shape but have been added to the Organization Units module, they do not appear in the scope part, impeding having the “correct” scope. 

Related JIRA tickets : POLIO-977

- [ ] Are there enough tests

## Changes

- change API call to get all org units
- filter it for the map
- add a localisation column  to display if org unit has a location
- new `OrgUnitLocationIcon`

## How to test

Edit a campaign with a scope.
Create a new district (org unit) without shape or marker linked to the parent of the selected campaign.
Open scope tab for this campaign, you should be able to see the new org unit.
You should be able to sort org unit by their locations too

## Print screen / video

https://user-images.githubusercontent.com/12494624/236832864-c2f38d17-06bf-412d-a6ae-11b0c018a4ff.mov


codeKameleon (approved):

plugins/polio/js/src/components/Scopes/MapScope.tsx
```diff
@@ -83,12 +84,31 @@ export const MapScope: FunctionComponent<Props> = ({
         };
     };
 
+    const districts = useMemo(
+        () =>
+            districtShapes.filter(
+                ogrUnit =>
+                    ogrUnit.has_geo_json ||
```
quang-le, line 91: ogrUnit, lol

plugins/polio/js/src/components/Scopes/MapScope.tsx
```diff
@@ -83,12 +84,31 @@ export const MapScope: FunctionComponent<Props> = ({
...
+                    (ogrUnit.latitude && ogrUnit.longitude),
+            ),
+        [districtShapes],
+    );
+    const regions = useMemo(
+        () =>
+            regionShapes.filter(
+                ogrUnit =>
```
quang-le, line 99: du travai encore du travail

quang-le (approved):
You make it look easy

PR #589, opened 2023-05-12
POLIO-1011 fix group dropdown in campaigns page

olethanh:
Now we don't need the group permission to have the Group dropdown.
And it works in the anonymous calendar

Related JIRA tickets : POLIO-2011

- [x] Are there enough tests

## Changes

Added a new API for group that we can use later in dropdowns: `/api/groups/dropdown/`

## How to test

- Test the polio interface with an user without the group permission
- Test the embeddedCalendar in an private page.


hat/assets/js/apps/Iaso/domains/orgUnits/hooks/requests/useGetGroups.ts
```diff
@@ -55,3 +55,48 @@ export const useGetGroups = ({
...
+    const queryString = urlSearchParams.toString();
+    return useSnackQuery({
+        queryKey: ['groups', queryString],
+        queryFn: () => getRequest(`/api/groups/dropdown/?${queryString}`),
+        snackErrorMsg: MESSAGES.fetchGroupsError,
+        options: {
+            staleTime,
+            select: data => {
+                if (!data) return [];
+                return data.groups.map(group => {
```
hakifran: There is no groups in data, you should map directly data `TypeError: data.groups is undefined.`
Screenshot from 2023-05-15 10-04-54

hakifran (changes requested):
And also there should be this `{
        isRequired: false,
        key: "orgUnitGroups"
    }` to be added in the /polio/js/config.js on campaignsFilters so that it can be taken in account while filtering on orgUnitGroups.

quang-le (changes requested):
I get the response from the API, but I don't see anything in the dropdown list:
Screenshot 2023-05-15 at 18 27 40

quang-le (changes requested):
In the embedded calendar, the dropdown shows, but the filter is not applied in the video, applying the filter should only keep the Benin campaign. It works in the normal calendar though)


https://github.com/BLSQ/iaso/assets/38907762/b4acc7b8-a5dd-4d87-9144-590bfa93502a

quang-le (approved):

PR #592, opened 2023-05-15
IA-2162: add permissions to update org unit types

beygorghor:
A separate permission is required to update org unit types in the dashboard

Related JIRA tickets : IA-2162

- [ ] Are there enough tests

## Changes

- added permission + migrations
- change menu permission to `iaso_update_org_unit_types`
- translate permission

## How to test

Use a test user 
Give him `iaso_update_org_unit_types` permission, you should be able to see and update an org unit type
Remove the permission, give him `iaso_org_units`  permissions and visit the search org unit page, you should still see types in the drop box.

## Print screen / video
Screenshot 2023-05-15 at 10 26 34

## Notes

This is not preventing the user to update type using the API !


hat/menupermissions/migrations/0047_alter_custompermissionsupport_options.py
```diff
@@ -0,0 +1,54 @@
...
+                    ("iaso_entities", "Entities"),
+                    ("iaso_storages", "Storages"),
+                    ("iaso_completeness_stats", "Completeness stats"),
+                    ("iaso_workflows", "Workflows"),
+                    ("iaso_polio_budget_admin", "Budget Polio Admin"),
+                    ("iaso_entity_duplicates_read", "Read Entity duplicates"),
+                    ("iaso_entity_duplicates_write", "Write Entity duplicates"),
+                    ("iaso_datastore_read", "Read data store"),
+                    ("iaso_datastore_write", "Write data store"),
+                    ("iaso_update_org_unit_types", "Editer types d'unité d'organisation"),
```
quang-le, line 50: Should be in English

hat/menupermissions/models.py
```diff
@@ -54,6 +54,7 @@
 ENTITY_DUPLICATES_WRITE = _("Write Entity duplicates")
 DATASTORE_READ = _("Read data store")
 DATASTORE_WRITE = _("Write data store")
+UPDATE_ORG_UNIT_TYPES = _("Editer types d'unité d'organisation")
```
quang-le: Should be in English

hat/menupermissions/models.py
```diff
@@ -113,4 +114,5 @@ class Meta:
             ("iaso_entity_duplicates_write", ENTITY_DUPLICATES_WRITE),
             ("iaso_datastore_read", DATASTORE_READ),
             ("iaso_datastore_write", DATASTORE_WRITE),
+            ("iaso_update_org_unit_types", UPDATE_ORG_UNIT_TYPES),
```
quang-le: @olethanh is right, we probably should have a standard way of naming read/write permissions

quang-le (approved):
Just the name thing but not blocking for me.


I see we use  the org units permissions to access "groups" as well. Maybe we should do something similar to anticipate further problems?

olethanh:
As discussed I would just name the permission `iaso_org_unit_types`. As the update is  a special keyword we use for the editing (and not creating) of submission, so it's going to be confusing. If you really want to specify the permission is for writing you can use the `_write` suffix. But really since it give access to the Org Unit Type page I would just name it `org_unit_types` to stay consistent with the other permissions.

Also the description in backend should be in English

PR #603, opened 2023-05-17
Add unit test for Limit downloads to root

olethanh:
Add testing for the previous hotfix in https://github.com/BLSQ/iaso/pull/598


beygorghor (approved):
tres zoli

PR #606, opened 2023-05-19
IA-2157: Add attachments to forms

bmonjoie:
To embed external media/datasets into ODK forms, Iaso must be able to provide attachment to the Android application and the Enketo server.

Related JIRA tickets : IA-2157

- [X] Are there enough tests

## Changes

A new `FormAttachment` model has been added with its own suite of APIs.

## How to test

The unit tests are self explainatory IMHO

## Notes

This is the first part as I still need to allow the application to download the manifest with all the attachments


.pre-commit-config.yaml
```diff
@@ -2,7 +2,7 @@
 # see Black section in README
 repos:
 -   repo: https://github.com/ambv/black
-    rev: 21.5b1
+    rev: 23.3.0
```
olethanh: this will cause conflict since it's not the same verywhere

bmonjoie: I cannot run black without this because of click. Shouldn't we update black everywhere to fix this issue?

olethanh: do you have the version of click that is in the requirements-dev.txt?

We can change it everywhere if this version is compatible with python 3.8 but I think there was some settings change so it might need modification of a bunch of file

olethanh (approved):

PR #607, opened 2023-05-21
Wc2 232 entities return 502 on coda dev environment

madewulf:
Wc2 232 entities return 502 on coda dev environment

## Changes
there were multiple problems:
- the API was returning 970 entities while there were 29 entities in the database in my case (a missing distinct)
- after solving the previous problem, the api was still making 450+ sql requests to build the response:
     - I fixed that by adding a few prefetch_related
     - caching the possible form_version that have not been deleted to use that to filter the answers (instead of doing one sql request for any such test)

### DIFF
--- a/iaso/api/mobile/entity.py
+++ b/iaso/api/mobile/entity.py
@@ -137,12 +137,12 @@ class MobileEntityViewSet(ModelViewSet):
 
     """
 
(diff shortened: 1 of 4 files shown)


mathvdh (approved):
Ok for me

PR #609, opened 2023-05-23
IA-2153: Make the difference between allowed-to-create children types and children-type to show in Iaso

beygorghor:
Have a way on mobile to differentiate the org unit type we would like to allow people to suggest adding, and those that are fixed and are not in the scope of suggestions for additions (ie regions, districts)

Related JIRA tickets : IA-2153

- [ ] Are there enough tests

## Changes

- model and migration
- duplicate code for sub unit type creation
- adding custom tooltip

## How to test

Create or edit an org unit type.
You should be able to add sub unit types displayed and sub unit types allowed for the creation in the mobile app
I have added same check on circular hierarchy as the one sub unit types

## Print screen / video

https://github.com/BLSQ/iaso/assets/12494624/2aa2dcec-fe79-4bcb-bb1d-28077d5acf59


codeKameleon (approved):

hat/assets/js/apps/Iaso/domains/orgUnits/orgUnitTypes/components/OrgUnitsTypesDialog.tsx
```diff
@@ -297,18 +309,58 @@ export const OrgUnitsTypesDialog: FunctionComponent<Props> = ({
...
                 label={MESSAGES.depth}
             />
-
             <InputComponent
                 multi
                 clearable
                 keyValue="sub_unit_type_ids"
                 onChange={onChange}
-                value={formState.sub_unit_type_ids.value}
+                loading={!allOrgUnitTypes}
```
quang-le: I think it's safer to use the `isLoading`from the react-query hook, because otherwise there's a risk to be stuck in and endless loading state

hat/assets/js/apps/Iaso/domains/orgUnits/orgUnitTypes/components/OrgUnitsTypesDialog.tsx
```diff
@@ -297,18 +309,58 @@ export const OrgUnitsTypesDialog: FunctionComponent<Props> = ({
...
             />
+
+            <Grid container spacing={1}>
+                <Grid item xs={11}>
+                    <InputComponent
+                        multi
+                        clearable
+                        keyValue="allow_creating_sub_unit_type_ids"
+                        onChange={onChange}
+                        loading={!allOrgUnitTypes}
```
quang-le: same as above

hat/assets/js/apps/Iaso/domains/app/translations/en.json
```diff
@@ -675,6 +675,8 @@
     "iaso.orgUnits.bulkChangeCount": "You are about to change {count} Org units",
     "iaso.orgUnits.confirmMultiChange": "Confirm bulk changes ?",
     "iaso.orgUnits.createdBy": "Created By",
+    "iaso.orgUnits.createSubUnitTypes": "Sub org units types to create",
+    "iaso.orgUnits.createSubUnitTypesInfos": "Enable the creation of the above sub org unit types (mobile)",
```
quang-le: I would replace with "Allow the creation of these sub org unit types (mobile)"

The "above sub org unit types"  is not ideal, since the tooltip is actually next to the relevant input

hat/assets/js/apps/Iaso/domains/orgUnits/orgUnitTypes/components/OrgUnitsTypesDialog.tsx
```diff
@@ -297,18 +309,58 @@ export const OrgUnitsTypesDialog: FunctionComponent<Props> = ({
...
                 keyValue="sub_unit_type_ids"
                 onChange={onChange}
-                value={formState.sub_unit_type_ids.value}
+                loading={!allOrgUnitTypes}
+                value={allOrgUnitTypes && formState.sub_unit_type_ids.value}
                 errors={formState.sub_unit_type_ids.errors}
                 type="select"
                 options={subUnitTypes}
                 label={MESSAGES.subUnitTypes}
             />
```
quang-le: I would put a tooltip next to this input as well

quang-le (changes requested):
Beware of the migration conflict.

Just some comments about the tooltips. The blocking one is the one about the phrasing of the tooltip. I think it's confusing

iaso/api/org_unit_types/serializers.py
```diff
@@ -101,15 +111,35 @@ def get_sub_unit_types(self, obj: OrgUnitType):
...
         parents = get_parents(self.context["request"].data.get("id", None))
+        # validate sub org unit type
         sub_types_errors = []
         for sub_type in data.get("sub_unit_types", []):
             if sub_type.id in parents:
                 sub_types_errors.append(sub_type.name)
         if len(sub_types_errors) > 0:
             raise serializers.ValidationError({"sub_unit_type_ids": sub_types_errors})
+        # validate sub org unit type allowed to be created
+        create_sub_types_errors = []
```
olethanh, line 139: I'm not sure I understand, this is to check for loop in the ancestry?

iaso/api/org_unit_types/serializers.py
```diff
@@ -101,15 +111,35 @@ def get_sub_unit_types(self, obj: OrgUnitType):
             context=self.context,
         ).data
 
+    def get_allow_creating_sub_unit_types(self, obj: OrgUnitType):
+        unit_types = obj.allow_creating_sub_unit_types.all()
+        app_id = self.context["request"].query_params.get("app_id")
+        if app_id is not None:
+            unit_types = unit_types.filter(projects__app_id=app_id)
```
olethanh, line 120: this is to return only the sub type for the app_id so the mobile one only have the correct one, do I understand correctly ? Can we add a comment?

iaso/models/org_unit.py
```diff
@@ -105,6 +105,7 @@ class OrgUnitType(models.Model):
     updated_at = models.DateTimeField(auto_now=True)
     category = models.CharField(max_length=8, choices=CATEGORIES, null=True, blank=True)
     sub_unit_types = models.ManyToManyField("OrgUnitType", related_name="super_types", blank=True)
+    allow_creating_sub_unit_types = models.ManyToManyField("OrgUnitType", related_name="create_types", blank=True)
```
olethanh, line 109: Please add a help_text describing what it does and that it only apply to mobile.

olethanh (approved):
a few nitpick requesting more doc/comments. otherwhise for the backend it LGTM, haven't looked at the frontend or tested it. It would be nice to have some backend testing for this. but we can open a separate ticket for it

olethanh:
I have fixed the black issue, we have changed the black version in main and the format is not 100% the same.

iaso/api/org_unit_types/serializers.py
beygorghor: yes same has above

quang-le (approved):

PR #613, opened 2023-05-23
Fixing an issue with black in pre-commit hook

bmonjoie:
When running the current version of black in the pre-commit hook, it installs an incompatible version of "click".

https://stackoverflow.com/questions/71769191/

## Changes

I added some configuration to force an older version of "click" which is compatible with the version of black we are using.


olethanh:
we adopter the alternative method in the end and updated the black version

PR #615, opened 2023-05-24
Fix XML file is malformed

bmonjoie:
The URL on the server contains `&` and it's making the XML malformed.

Related JIRA tickets : IA-2157

- [X] Are there enough tests

## Changes

Make use of the [`escape`](https://wiki.python.org/moin/EscapingXml) function to fix the XML


madewulf (approved):

mathvdh (commented):
Ok for me

PR #633, opened 2023-06-05
IA-2169: org unit groups permissions

beygorghor:
Allow access to group page only to admin or those with correct permission

Related JIRA tickets : IA-2169

- [ ] Are there enough tests

## Changes

- added new permission + migration
- change menu permission of groups to `iaso_org_unit_groups`
- translate permission

## How to test

Use a test user that doesn't have the permission `iaso_org_unit_groups`
Check that he does't have access to groups page
check that he can see groups in dropdown on ohter pages

## Print screen / video

Screenshot 2023-06-05 at 14 20 13


codeKameleon (approved):

olethanh:
The users are going to lose access to the page, should we have some kind of migration or communication with the implementations  team on how to handle the transition?

quang-le (changes requested):
Same comment as @olethanh 

I foresee this might be problem with polio, where they use the group feature to create country blocs

We should have a definitive answer to what to do about making a migration to add the permission to some users before merging

beygorghor:
@quang-le 
Screenshot 2023-06-14 at 13 17 48

PR #640, opened 2023-06-07
Upgrade dependencie ipython

olethanh:
hopefully this should fix the crash in the shell in the live server when doing tab completion.
This is the last version supported by Python 3.8


beygorghor (approved):

mathvdh:
See IA-2255 for the ticket !

PR #643, opened 2023-06-07
IA-2184 Make Form Attachment works with Enketo

olethanh:
The form attachment feature was added for the mobi84le but didn't work with enketo, this implement it.

Related JIRA tickets : IA-2184

- [NA] New translations have been added or updated if new strings have been introduced in the frontend
- [no] Are there enough tests
- [no] Documentation has been included (for new feature)

## Changes

- Pass the manifest url via the formDownload endpoint so enketo know to retreive the attachment. (in mobile the app construct it)
- Had to make a hack in the manifest endpoint to always have absolute uri, with the enketo domain hack for docker.
- Modification in enketo settings so it would download the attachement properly without a security error. This should only be needed in local

## How to test

- Ask Benjamin for the proper file (the xlsform and the csv)
- Launch your docker with Enketo. See the doc in the readme if you don't know how.
- Create a new form with the file (Form version + attachment)
- Create submissions

### DIFF
--- a/docker/enketo/config/config.json
+++ b/docker/enketo/config/config.json
@@ -1,6 +1,13 @@
 {
     "app name": "Enketo for iaso dev",
     "encryption key": "Test encryption key do not use in prod",
+    "ip filtering": {
+        "comment": "modification necessary for attachment download",
+        "allowIPAddressList": [],
+        "allowMetaIPAddress": false,
+        "allowPrivateIPAddress": true,
+        "denyAddressList": []
+    },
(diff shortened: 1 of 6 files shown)


olethanh:
I tested with embedding image (Bird example  from https://xlsform.org/en/#media ) but they don't seems to display. Not sure if it's a Enketo bug  or what but it's a bit hard to debug so let's check if we put more time in that.
But at least the WFP Form work

beygorghor (approved):

PR #665, opened 2023-06-21
IA-2251 fix  form attachment not working in enketo prod

olethanh:
The problem was that the API responded as needing an authentification I think the issue wasn't visible in local since it was on the same domain name ? The behaviour was modified so Enketo now pass the app_id (we pass it in the app definition) like the mobile does, and which allow accessing without auth

Related JIRA tickets : IA-2251

- [NA] New translations have been added or updated if new strings have been introduced in the frontend
- [x] Are there enough tests
- [NA] Documentation has been included (for new feature)

## Changes

Modify the manifest url to include the app_id.

## How to test

use  a Form with an attachement. You can copy this one https://iaso-staging.bluesquare.org/api/forms/153/manifest/
Make a submission via Enketo

### DIFF
--- a/iaso/api/enketo.py
+++ b/iaso/api/enketo.py
@@ -272,7 +272,10 @@ def enketo_form_list(request):
     downloadurl = public_url_for_enketo(request, "/api/enketo/formDownload/?uuid=%s" % i.uuid)
     # Only add a manifest if we actually have attachment so it doesn't make more unecessary request
     if i.form.attachments.exists():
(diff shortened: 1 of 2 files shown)


beygorghor (approved):
This is working for the WFP form including a CSV file (https://iaso-staging.bluesquare.org/dashboard/forms/detail/accountId/1/formId/153)
But not for form including images, but that was already the case before i think (https://iaso-staging.bluesquare.org/dashboard/forms/detail/accountId/1/formId/154)

PR #671, opened 2023-06-26
IA-2259 Separate Version for orgunit type api with allow_creating_sub_unit_types

olethanh:
Confusingly the V2 is the current behavior with the new allow_creating_sub_unit_types field (separate from sub_unit_types)
the existing url is with the new behavior where we map sub_unit_types to allow_creating_sub_unit_types.
so the mobile application don't need any update to alter the url.

Add a Data migration  to fill allow_creating_sub_unit_types so there is no loss of functionality for mobile application users.

The current behavior will be that allow_creating_sub_unit_types is used on mobile to restrict which org unit can be created
and sub_unit_types is used on web to control which type are displayed

Related JIRA tickets : IA-2259

- [na] Are my typescript files well typed
- [na] New translations have been added or updated if new strings have been introduced in the frontend
- [x] Are there enough tests

## Changes

- See summary, but add /v2/ with current behaviour and modified existing url with new behaviour for mobile.
- Ensure there is tests for both version
- Modified the frontend to use the new url
- Added doc

## Notes

The web and external API user should from now on only use the /v2/ endpoint.

At some point we should restrict the action on the current url to only
what the mobile app support, to reduce the amount of code needed for
maintenance. but we need to ensure this won't be problematic for external users

### DIFF
--- a/iaso/api/org_unit_types/viewsets.py
+++ b/iaso/api/org_unit_types/viewsets.py
@@ -8,8 +8,12 @@ from ..common import ModelViewSet
 
 
 class OrgUnitTypeViewSet(ModelViewSet):
-    """Org unit types API
+    """Org unit types API (deprecated)
 
+    This endpoint it deprecated, Use /v2/orgunittypes/ instead, this is kept only  for compatiblity with the mobile
+    application
+
+    Confusingly in this version  `sub_unit_types` map to allow_creating_sub_unit_types.
     This API is open to anonymous users.
 
     GET /api/orgunittypes/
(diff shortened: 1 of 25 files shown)


iaso/api/org_unit_types/viewsets.py
```diff
@@ -3,20 +3,24 @@
...
 
 
 class OrgUnitTypeViewSet(ModelViewSet):
-    """Org unit types API
+    """Org unit types API (deprecated)
+
+    This endpoint it deprecated, Use /v2/orgunittypes/ instead, this is kept only  for compatiblity with the mobile
+    application
 
+    Confusingly in this version  `sub_unit_types` map to allow_creating_sub_unit_types.
```
bmonjoie, line 16: Nit-picking, you didn't put the back quotes around "allow_creating_sub_unit_types"

```suggestion
    Confusingly in this version  `sub_unit_types` map to `allow_creating_sub_unit_types`.
```

bmonjoie (approved):

beygorghor (approved):

PR #693, opened 2023-07-03
IA-2292 500 when using a csv with semicolon in batch user creation + add projects and user roles

SimonDus:

Related JIRA tickets : IA-2292 and IA-2289

- [X] Are there enough tests

## Changes

Field Updates : 

Mandatory columns  have been updated to make it less error prone. 

New column " user_roles " and " projects " have been added.

Mandatory columns are : ["username", "password", "email", "first_name", "last_name", "orgunit", "profile_language", "dhis2_id", "permissions", "user_roles", "projects"]

I added a check of csv delimiter to handle files separated by both coma and semicolon. For some reason sometimes ne Sniffer doesnt work and result into a error. So the code check first with Sniffer, if it fails it try a " manual " check by parsing the file. 

Also, more tests.

## How to test

Explain how to test your PR.
Test with a CSV created with Excel.
Upload a CSV file with semicolon separator, it should work. Also test the new fields. Use one of the working templates to test.


hakifran (changes requested):
While testing there is a bug 
Screenshot from 2023-07-05 11-14-21

iaso/api/bulk_create_users.py
```diff
@@ -88,17 +88,24 @@ def create(self, request, *args, **kwargs):
...
-            i = 0
             csv_indexes = []
             file_instance = BulkCreateUserCsvFile.objects.create(
                 file=user_csv, created_by=request.user, account=request.user.iaso_profile.account
             )
             file_instance.save()
-            for row in reader:
+            for i, row in enumerate(reader):
                 org_units_list = []
                 if i > 0:
```
madewulf, line 158: Wondering if we should not have an except for all exceptions and raising then an error saying "there is a problem with your file" ?

SimonDus:
@hakifran  I just did a test and I don't have the error. Can you describe me what you are doing ?

iaso/api/bulk_create_users.py
SimonDus: I added more check to the csv files/Fields etc. It should be more secure now

SimonDus:
I just found a bug that I have to fix.

hakifran (changes requested):
@SimonDus , when a user role it's created there is a prefix which is added on the user role name. The prefix is the account id and an underscore like '1_manager'. So in the csv file if the prefix is not added it won't find the user role.Could you fix it?
Screenshot from 2023-08-01 14-29-42

Screenshot from 2023-08-01 14-30-21

hakifran (approved):
On my point of view it's good

mathvdh (approved):
Seems good to me !

mathvdh:
Juste une question : pourquoi il m'affiche ce message bizarre tout en haut de cette page ? juste parce qu'il y  a un `;` dans le nom de la branche ?

Screenshot 2023-08-04 at 08 44 17

SimonDus:
@mathvdh Oui c'est juste à cause du " ; " c'est pas bien grave.

PR #701, opened 2023-07-06
Configure the server URL for local use with an environment variable

bmonjoie:
Allow to configure the server URL for local use with an environment variable

TODO: Update the wiki when merged

## Changes

The current configuration for local use requires the user to modify several files.
Also, it is not foolproof as new media won't be served correctly (like the form attachments).
This solution will allow to easily configure the local environment, without modifying code, and, therefore, without the risk of committing local configuration.

## How to test

Launch ngrok, copy the URL, and add it into your `.env` file like so:
```
SERVER_URL=<copied URL>
```

Launch docker and go to the form API, the URLs should be absolute instead of relative.

### DIFF
--- a/hat/settings.py
+++ b/hat/settings.py
@@ -355,6 +355,7 @@ AWS_S3_REGION_NAME = os.getenv("AWS_S3_REGION_NAME", "eu-central-1")
 AWS_ACCESS_KEY_ID = os.getenv("AWS_ACCESS_KEY_ID")
 AWS_SECRET_ACCESS_KEY = os.getenv("AWS_SECRET_ACCESS_KEY")
 
+MEDIA_URL_PREFIX = "/media/"
 if USE_S3:
     # https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html
     AWS_S3_OBJECT_PARAMETERS = {"CacheControl": "max-age=86400"}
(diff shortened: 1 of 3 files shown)


olethanh (approved):

bmonjoie:
FYI, this is the currently documented way of setting up Iaso for mobile use: https://github.com/BLSQ/iaso/wiki/Development-environment#1---setup-ngrok

> 1 - Setup Ngrok
Download and setup Ngrok on https://ngrok.com/. Once Ngrok installed and running you must add your ngrok server url in settings.py by adding the following line :
```
FILE_SERVER_URL = os.environ.get("FILE_SERVER_URL", "YOUR_NGROK_SERVER_URL")
```
> After this step you have to import settings.py and add FILE_SERVER_URL to forms.py in iaso/models/forms as shown on the following lines :
```
"file": settings.FILE_SERVER_URL + self.file.url,
"xls_file": settings.FILE_SERVER_URL + self.xls_file.url if self.xls_file else None
```

hat/settings.py
```diff
@@ -389,7 +390,8 @@ def is_superuser(u):
 
     DEFAULT_FILE_STORAGE = "storages.backends.s3boto3.S3Boto3Storage"
 else:
-    MEDIA_URL = "/media/"
+    SERVER_URL = os.environ.get("SERVER_URL", "")
```
mathvdh, line 393: On mettrait pas une valeur par défault?

mathvdh (approved):
Just a little question ...

hat/settings.py
bmonjoie: Laquelle?

PR #703, opened 2023-07-07
POLIO-1089:skip rejected OU,improve string format

quang-le:
lqas/im API could sometimes return the district id of a deleted org unit of there were 1 valid and 1 rejected version of the org unit, or not show the district at all (same conditions). It also couldn't match all parent alias names when trying to match region names

Related JIRA tickets :  POLIO-1089

- [ ] Are there enough tests

## Changes

- filter out NEW and REJECTED org units when getting queryset
- strip and convert parent aliases to lower case for better name matching

## How to test

- You need a DB with the same org unit ids as prod
- test name matching: replicate TOG-39DS-06-022 from polio prod: there should be no "district not found" (as opposed to "golfe" in current polio prod)
- rejected org units: I couldn't replicate locally the case where a district is returned with the id of its rejected version (my datasource is outdated). The other bug can be tested by replication locally the campaign ZMB-30DS-02-2022 and creating a rejected version of a district in scope, eg Luangwa. It should not appear in the response when you're on main, but should appear when testing this branch. 

IMPORTANT: don't forget to set the cache time to 1 or comment out caching code when testing.

## Print screen / video
Not visible in the UI, you should compare json from API responses from main and from the branch

### DIFF
--- a/plugins/polio/serializers.py
+++ b/plugins/polio/serializers.py
@@ -313,6 +313,7 @@ class RoundDateHistoryEntrySerializer(serializers.ModelSerializer):
         ]
 
     modified_by = UserSerializer(required=False, read_only=True)
+    round = serializers.PrimaryKeyRelatedField(read_only=True, many=False)
 
     def validate(self, data):
         if not data["reason"]:
--- a/plugins/polio/serializers.py
+++ b/plugins/polio/serializers.py
@@ -313,7 +313,7 @@ class RoundDateHistoryEntrySerializer(serializers.ModelSerializer):
         ]
 
     modified_by = UserSerializer(required=False, read_only=True)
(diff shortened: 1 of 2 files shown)


madewulf (approved):

plugins/polio/serializers.py
```diff
@@ -313,6 +313,7 @@ class Meta:
         ]
 
     modified_by = UserSerializer(required=False, read_only=True)
+    round: Field = serializers.PrimaryKeyRelatedField(read_only=True, many=False)
```
madewulf, line 317: Isn't this already in another commit/PR?

quang-le: Yep, this one, that wa smerged yesterday: https://github.com/BLSQ/iaso/pull/704

I think it has been deployed as hotfix already: that's the fix to avoid the crash in reasons for delay

PR #718, opened 2023-07-17
IA-1940: Improve "user management" permission

bmonjoie:
Split the user management permission into "admin user manager" and "user management"

Related JIRA tickets : IA-1940

- [x] Are there enough tests

## Changes

- Replace the hardcoded permissions with constants
- Implement permission in `/api/profiles/` as well as in `/api/tasks/create/profilesbulkupdate/`
- Slightly change the UI to take into account the new permission

## How to test
Assign the permission "Users (management)" to a user, optionally, assign an org unit to that user to limit its permission to users within that org unit.
Log in with that user and manage users within its health pyramid.


hat/menupermissions/models.py
```diff
@@ -17,46 +17,75 @@
 `/api/permissions/` endpoint
 """
 
-
 from django.db import models
 from django.utils.translation import gettext_lazy as _
 
-MODIFICATIONS = _("Modifications")
+_ASSIGNMENTS = "iaso_assignments"
```
madewulf, line 25: Didn't you remove translations here. _() is used to translate the presentation text of a permission, normally

hat/menupermissions/models.py
```diff
@@ -66,57 +95,57 @@ class CustomPermissionSupport(models.Model):
...
-        "iaso_sources",
-        "iaso_data_tasks",
-        "iaso_reports",
+        _FORMS,
+        _SUBMISSIONS,
+        _MAPPINGS,
+        _COMPLETENESS,
+        _ORG_UNITS,
+        _LINKS,
+        _USERS,
```
madewulf: Here, you could fix another existing ticket by including all the permissions for a new account_user

hat/menupermissions/models.py
bmonjoie: I have moved the translation to the list and used the constants for the keys.

hat/menupermissions/models.py
bmonjoie: Could you link the ticket here?

iaso/api/profiles.py
```diff
@@ -15,19 +16,50 @@
...
+    @staticmethod
+    def has_permission_over_user(request, view):
+        pk = view.kwargs.get("pk")
+        if not pk:
+            raise PermissionDenied(f"User with '{permission.USERS_MANAGED}' cannot create users.")
+
+        if pk == request.user.id:
+            raise PermissionDenied(f"User with '{permission.USERS_MANAGED}' cannot edit their own permissions.")
+
+        org_units = OrgUnit.objects.hierarchy(request.user.iaso_profile.org_units.all()).values_list("id", flat=True)
```
madewulf, line 53: I'm questioning this. A user that can modify the permissions of everybody without creating one could be an interesting option too and would maybe match better what we do with org units limitation (basically, you're more powerful if you have no org units assigned than if you have some)

madewulf:
Wondering how all this mixes with the user roles that were recently created and the bulk user update too that we just merged. I have the feeling that we would need to rebase this on main an take these questions into consideration

beygorghor (approved):

SimonDus (approved):
I have tested it on my local instance and it seems to work as expected. 
Good job !

PR #726, opened 2023-07-19
IA-2209 Add filter on ou account while creating user

SimonDus:

Related JIRA tickets : IA-X2209

- [x] Are there enough tests

## Changes

Small fix to restrict creating user with access to OU based on ou accessible by the requesting user.

## How to test

upload a csv /api/bulkcreateuser/ 
Use one of the valid test csv as template and try to create a user with a org unit you don't have access to. 
You should get a error

### DIFF
--- a/iaso/api/bulk_create_users.py
+++ b/iaso/api/bulk_create_users.py
@@ -99,7 +99,7 @@ class BulkCreateUserFromCsvViewSet(ModelViewSet):
             )
             file_instance.save()
 
-            orgunits_hierarchy = OrgUnit.objects.hierarchy(request.user.iaso_profile.org_units.all())
+            orgunits_hierarchy = OrgUnit.objects.hierarchy(user_access_ou)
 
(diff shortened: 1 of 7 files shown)


iaso/api/bulk_create_users.py
```diff
@@ -164,7 +164,9 @@ def create(self, request, *args, **kwargs):
                                         )
                             except ValueError:
                                 try:
-                                    org_unit = OrgUnit.objects.get(name=ou)
+                                    org_unit = OrgUnit.objects.get(
+                                        name=ou, pk__in=request.user.iaso_profile.org_units.all()
```
madewulf: I think this is not going to work like it should. We want to allow any org unit that are children (including themselves, as the awkard definition of children in computer science trees means it) of the ones in iaso_profile.org_units.all().  See calls to hiearchy() in some other places of the code base

madewulf (approved):

PR #727, opened 2023-07-19
Ia 2328 remove countries from dropdown

quang-le:
Dropdown options in LQAs/Im screens showed countries for which there was no config. 

Related JIRA tickets : IA-2328

- [ ] Are there enough tests

## Changes

- Add an endpoint that returns countries with configs for lqas or im, and formats the response to be directly usable by the front-end
- Modify front-end accordingly

## How to test

Go to LQAs then IM
In each case check the countries in the select box. The list shouldn't contain all countries in the datasource, only the countries for which there is a config in the DB

## Print screen / video

Screenshot 2023-07-19 at 16 33 38


plugins/polio/api.py
```diff
@@ -2018,6 +2019,31 @@ def list(self, request):
...
+    ordering_fields = ["name", "id"]
+    filter_backends = [
+        filters.OrderingFilter,
+        DjangoFilterBackend,
+    ]
+
+    def get_queryset(self):
+        category = self.request.query_params.get("category")
+        configs = Config.objects.filter(slug=f"{category}-config").first().content
+        country_ids = [config["country_id"] for config in configs]
```
hakifran: @quang-le I think here you should make another filter on countries which have datastore(JsonDataStore). Because the request is made on datastore when a country is selected.
Peek 2023-07-20 11-01

hakifran (changes requested):

beygorghor (approved):

plugins/polio/api.py
```diff
@@ -2021,6 +2022,34 @@ def list(self, request):
...
+        DjangoFilterBackend,
+    ]
+
+    def get_queryset(self):
+        category = self.request.query_params.get("category")
+        configs = Config.objects.filter(slug=f"{category}-config").first().content
+        country_ids = [
+            (config["country_id"] if JsonDataStore.objects.filter(slug=f"{category}_{config['country_id']}") else None)
+            for config in configs
+        ]
```
beygorghor: @hakifran we can have an array containing none value then. @mathvdh any proposal to fix this ?

beygorghor (changes requested):

plugins/polio/api.py
mathvdh: I have to look into it, as I don't know (yet) what's being done with array later ..

hakifran: @beygorghor I refactored the code and for now when the country has not datastore I do a continue in the loop.

beygorghor (approved):

plugins/polio/api.py
```diff
@@ -2021,6 +2022,36 @@ def list(self, request):
...
+        filters.OrderingFilter,
+        DjangoFilterBackend,
+    ]
+
+    def get_queryset(self):
+        category = self.request.query_params.get("category")
+        configs = Config.objects.filter(slug=f"{category}-config").first().content
+        country_ids = []
+        for config in configs:
+            if JsonDataStore.objects.filter(slug=f"{category}_{config['country_id']}"):
```
mathvdh: Shouldn't we add .`exists()` after `.filter(..)` ? It will stop as soon as it finds one matching instance and doesn't fetch all matching instances which can be more efficient (In this case I don't think it makes much of a difference but It's more readable ...)

mathvdh (approved):
Except for the little comment I put, I think this is OK (I approved anyway as the change resquested is not really important)

beygorghor (approved):

PR #729, opened 2023-07-20
IA-1870 Give all permissions to user created with api setupaccount

SimonDus:
User created with the API setupaccount are now given all permissions. 

Related JIRA tickets : IA-1870

- [x] Are there enough tests

## Changes

User is superuser at creation.

## How to test

Add a user with the endpoint /api/setupaccount. It must be superuser.


iaso/api/setup_account.py
```diff
@@ -44,6 +44,7 @@ def create(self, validated_data):
             password=validated_data["password"],
             first_name=validated_data.get("user_first_name", ""),
             last_name=validated_data.get("user_last_name", ""),
+            is_superuser=True,
```
madewulf: This is unfortunately not the solution: this gives access to all the accounts through the admin and should NOT be given to an account admin.

madewulf: This is unfortunately not the solution: this gives access to all the accounts through the admin and should NOT be given to an account admin.

iaso/api/setup_account.py
```diff
@@ -12,6 +12,56 @@
 logger = logging.getLogger(__name__)
 
 
+DEFAULT_PERMISSIONS_FOR_NEW_SETUP_ACCOUNT_USER = [
```
beygorghor: From @bmonjoie we should use constants here and i totally agree.
Also, this is not a superuser, we have to maintain this while adding new permissions.

beygorghor (changes requested):

iaso/api/setup_account.py
SimonDus: I reworked it so the list comes from the model file.

bmonjoie: @SimonDus What @beygorghor is saying is not going to be possible until my PR is merged #718

madewulf (approved):

hakifran (approved):

beygorghor (approved):

PR #734, opened 2023-07-25
IA-1787 Instances in multiple projects are counted multiple times in the submissions count

SimonDus:
The submission count was wrong has the instance count was somehow multiplied by the number of projects. This fix it and test it.

Related JIRA tickets : IA-1787

- [x] Are there enough tests

## Changes

Changes are straightforward. Added a distinct after the query filter as each time a filter was True the Count() is adding it to the total.

## How to test

Have a form with multiple instances from multiple projects. 
For example if you have 2 instances, in different projects you should have 2 submissions for the form. But the list will show 4.
With the fix you will have the correct count.

### DIFF
--- a/iaso/api/forms.py
+++ b/iaso/api/forms.py
@@ -246,9 +246,11 @@ class FormsViewSet(ModelViewSet):
                         & ~Q(instances__device__test_device=True)
                         & ~Q(instances__deleted=True)
                         & Q(instances__org_unit__in=orgunits)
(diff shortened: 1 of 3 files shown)


beygorghor (approved):

hakifran (changes requested):
The added test don't work as expected.

SimonDus:
I'll start it from a new branch... This is too old

PR #742, opened 2023-08-02
IA-1991 completeness map

beygorghor:
Displaying completeness on a map tab

Related JIRA tickets : IA-1991

- [ ] Are there enough tests

## Changes

Display tab and map if only ONE form is selected on completeness stats page

## How to test

Find an account with submissions linked to org unit.
Browse to completeness stats page, select the form with those submissions.
Try to navigate down and up through org units levels.

## Print screen / video

https://github.com/BLSQ/iaso/assets/12494624/6b0e5fa1-2925-4b53-ae2a-98c6346507f8


iaso/api/completeness_stats.py
```diff
@@ -332,60 +333,97 @@ def list(self, request: Request, *args, **kwargs) -> Response:
                 )
 
         def to_dict(row_ou: OrgUnitWithFormStat):
+            # TODO: improve this (check validation, account ,...)
+            children_count = OrgUnit.objects.all().filter(parent_id=row_ou.id).count()
```
beygorghor: @mathvdh can you have a look on this please, we need to annotate the query and make it more robust

iaso/api/completeness_stats.py
```diff
@@ -332,60 +333,97 @@ def list(self, request: Request, *args, **kwargs) -> Response:
...
-                "parent_org_unit": row_ou.parent.as_dict_for_completeness_stats() if row_ou.parent else None,
+                "parent_org_unit": row_ou.parent.as_dict_for_completeness_stats_with_parent()
+                if row_ou.parent
+                else None,
+                "has_children": True if children_count > 0 else False,
             }
 
+        def to_map(row_ou: OrgUnitWithFormStat):
+            # TODO: improve this (check validation, account ,...)
+            children_count = OrgUnit.objects.all().filter(parent_id=row_ou.id).count()
```
beygorghor: @mathvdh same here

beygorghor:
still need to fix tests and `has_children` property in api

iaso/api/completeness_stats.py
beygorghor: @SimonDus

iaso/api/completeness_stats.py
beygorghor: @SimonDus

iaso/api/completeness_stats.py
```diff
@@ -342,61 +343,98 @@ def list(self, request: Request, *args, **kwargs) -> Response:
                     ),
                 )
 
+        def has_children(row_ou):
```
quang-le: Wouldn't it be better to define these methods outside of `list()`?

iaso/tests/api/test_completeness_stats.py
quang-le, line 1: There's no test for the the "map" payload ( that uses the `to_map`method)?

hat/assets/js/apps/Iaso/utils/map/mapUtils.ts
```diff
@@ -185,7 +185,9 @@ export type Bounds = {
     isValid: () => boolean;
 };
 
-export const getOrgUnitBounds = (orgUnit: OrgUnit): Bounds | undefined => {
+export const getOrgUnitBounds = (
+    orgUnit: Record<string, any>,
```
quang-le: Shouldn't the type be a variation of `OrgUnit[]`?

hat/assets/js/apps/Iaso/domains/completenessStats/types.ts
```diff
@@ -26,13 +27,35 @@ export type CompletenessStats = {
...
+    has_geo_json: boolean;
+    id: number;
+    is_root?: boolean;
+    latitude?: number;
+    longitude?: number;
+    name: string;
+    org_unit_type?: NameAndId;
+    parent_org_unit?: ParentOrgUnit;
+    has_children: boolean;
+};
```
quang-le, line 53: Wouldn't that be the type to use in `getOrgUnitBounds`?

hat/assets/js/apps/Iaso/domains/completenessStats/components/MapLegend.tsx
```diff
@@ -0,0 +1,118 @@
...
+                )}
+
+                {!showDirectCompleteness && (
+                    <LegendThreshold scale={getLegend}>
+                        {labels =>
+                            labels.reverse().map(label => {
+                                return (
+                                    <LegendItem
+                                        key={`legend-direct-${label.value}`}
+                                        margin={theme.spacing(0, 0, 1, 0)}
```
quang-le, line 93: I think the `keys` are mixed up. This one is not the `legend-direct`

hat/assets/js/apps/Iaso/domains/completenessStats/components/MapLegend.tsx
quang-le, line 1: Unless I missed something, you can spare some lines by declaring the `key` (which is a string, so no render problem) and the `scale` (which is either `getLegend` or `getDirectLegend` which are hoisted, so no problem here either) with  a ternary:
```typescript
//...some code
const theme = useTheme()
const scale = showDirectCompleteness ? getDirectLegend : getLegend
const legendItemBaseKey = showDirectCompleteness ? "legend-direct" : "legend"
// more code
                    <LegendThreshold scale={scale}>
// and so on
```

hat/assets/js/apps/Iaso/domains/completenessStats/index.tsx
```diff
@@ -39,10 +54,14 @@ export const CompletenessStats: FunctionComponent<Props> = ({
...
 }) => {
     const classes: Record<string, string> = useStyles();
+
+    const [tab, setTab] = useState<'list' | 'map'>(params.tab ?? 'list');
     const dispatch = useDispatch();
     const { formatMessage } = useSafeIntl();
     const { data: completenessStats, isFetching } =
         useGetCompletenessStats(params);
+    const { data: completenessMapStats, isFetching: isFetchingMapStats } =
+        useGetCompletnessMapStats(params);
```
quang-le: maybe we can disable it when the map is not displayed to avoid useless fetching?

hat/assets/js/apps/Iaso/domains/completenessStats/index.tsx
```diff
@@ -89,24 +133,54 @@ export const CompletenessStats: FunctionComponent<Props> = ({
...
+                            count={completenessStats?.count ?? 0}
+                            baseUrl={baseUrl}
+                            countOnTop={false}
+                            params={params}
+                            extraProps={{ loading: isFetching }}
+                            onTableParamsChange={p => {
+                                dispatch(redirectTo(baseUrl, p));
+                            }}
+                            // @ts-ignore
+                            rowProps={getRowStyles}
```
quang-le, line 186: ❤️

hat/assets/js/apps/Iaso/domains/completenessStats/components/Map.tsx
```diff
@@ -0,0 +1,219 @@
...
+    const classes: Record<string, string> = useStyles();
+
+    const bounds: Bounds | undefined = useMemo(
+        () => locations && getOrgUnitsBounds(locations),
+        [locations],
+    );
+
+    const [currentTile, setCurrentTile] = useState<Tile>(tiles.osm);
+
+    const isLoading = isFetchingLocations;
```
quang-le: The prop could be called `isLoading` then

quang-le (changes requested):
Nice work. Some suggestions for improvement in the code. Nothing really blocking, but some worth checking out at least.

This screen is confusing to me:
- It's the view for haut Lomami (Province).
- If I click on a shape:
    - The button see parent takes me to the parent of Haut Lomami
    - The button see children takes me to the org unit of the shape
 I get that the view actually goes down or up a level in the pyramid, but I find the wording a bit confusing

Screenshot 2023-08-22 at 17 54 39

maybe we should have a fallback to go back up a level in the map when the selected org unit has no shape or marker displayed:
Screenshot 2023-08-22 at 18 00 30

Actually, it might make sense to have only the "See children" button in the pop up, and have a separate "See parent" button available some place else on the map

I don't think we have on other maps either, but the tile selection is not deep linked

iaso/api/completeness_stats.py
mathvdh: Yes I agree, it's not "bad" but its ugly and the list() method is already long enough. These method can even be put outside of the viewset class itself.

iaso/api/completeness_stats.py
```diff
@@ -342,61 +343,98 @@ def list(self, request: Request, *args, **kwargs) -> Response:
...
+                "latitude": row_ou.location.y if row_ou.location else None,
+                "longitude": row_ou.location.x if row_ou.location else None,
+                "altitude": row_ou.location.z if row_ou.location else None,
+                "org_unit_type": row_ou.org_unit_type.as_dict_for_completeness_stats() if row_ou.org_unit_type else {},
+                "parent_org_unit": row_ou.parent.as_dict_for_completeness_stats_with_parent()
+                if row_ou.parent
+                else None,
+                "has_children": has_children(row_ou),
+            }
+            temp_org_unit["geo_json"] = None
```
mathvdh: This is redundant with line 369 if I'm not mistaken

iaso/api/completeness_stats.py
mathvdh: Moreover has these functions have access to their parent context its not explicit or clear what params they are using ...

iaso/models/org_unit.py
```diff
@@ -482,6 +482,13 @@ def as_dict_for_csv(self):
             "org_unit_type": self.org_unit_type.name,
         }
 
+    def as_dict_for_completeness_stats_with_parent(self):
```
mathvdh, line 485: Just a remark : Didn't we say we are gonna stop using the `to_dict` and `as_dict` methods and use serializers in general ?

iaso/tests/api/test_completeness_stats.py
```diff
@@ -253,13 +256,14 @@ def test_base_row_listing(self):
...
                     "parent_org_unit": None,
+                    "has_children": False,
                 },
             ],
             "has_next": False,
             "has_previous": False,
             "page": 1,
             "pages": 1,
-            "limit": 10,
+            "limit": "10",
```
mathvdh: How is this a string ?

mathvdh (changes requested):
A few (minor) comments, otherwise nice work !

iaso/models/org_unit.py
beygorghor: yes but... i used same pattern as Olivier, and the org unit ai is not using the serializer correctly i think

iaso/tests/api/test_completeness_stats.py
beygorghor: the api was returning a string in fact ...

hat/assets/js/apps/Iaso/domains/completenessStats/components/MapLegend.tsx
beygorghor: Yes the label is different too

quang-le (approved):

iaso/api/completeness_stats.py
```diff
@@ -349,54 +355,86 @@ def to_dict(row_ou: OrgUnitWithFormStat):
...
                 "form_stats": row_ou.form_stats,
                 "org_unit_type": row_ou.org_unit_type.as_dict_for_completeness_stats() if row_ou.org_unit_type else {},
-                "parent_org_unit": row_ou.parent.as_dict_for_completeness_stats() if row_ou.parent else None,
+                "parent_org_unit": row_ou.parent.as_dict_for_completeness_stats_with_parent()
+                if row_ou.parent
+                else None,
+                "has_children": has_children(row_ou),
             }
 
+        def to_map(row_ou: OrgUnitWithFormStat):
```
mathvdh, line 364: It was maybe not clear but all these methods could have been moved out of the class ... It doesn't really matter and I'll approve it

mathvdh (approved):

PR #747, opened 2023-08-08
WC2-249 beneficiary types improvements

beygorghor:
Multiple tasks here to be done on the beneficiary type list:

- replace the target of the eye button to be the list of beneficiaries 
- add a different button with a different icon to target the reference form (which was the target of the eye)
- add a permission (like forms) that allows to see the list of beneficiaries and beneficiary types but with no access to edition of the beneficiary type
- move the beneficiary type entry in the menu above the beneficiary list

I also added beneficiary type entries in the side menu to make it easier

Related JIRA tickets WC2-249

- [x] Are there enough tests

## Changes

- move permissions messages in one file
- create ne permission `iaso_entity_type_write` to allow CRUD on entity types
- adapt api to allow only user with correct permission to CRUD 
- adapt front the same way
- get entity types and populate side menu with dynamic names
- change action buttons as requested
- wrote some unit test and cypress tests

## How to test

Create a user with only entities permission, check that he can only read infos
Play with the navigation between types and entities

## Print screen / video

https://github.com/BLSQ/iaso/assets/12494624/c78b23eb-8ec9-4485-ae4f-6789d2444fbc

## Print screen / video

@bmonjoie make sure this doesn't interfere with the mobile app (i don't think you can edit entity type with the app)
⚠️ to display beneficiaries by types in the menu you need to activate SHOW_BENEFICIARY_TYPES_IN_LIST_MENU on your account


hakifran (changes requested):
when I want to go to menu there is a bug.
[Screencast from 2023-08-21 10-50-05.webm](https://github.com/BLSQ/iaso/assets/19631540/9568d02d-150c-4a66-b2e4-660eff9db752)

beygorghor:
It looks like i have an issue when Polio plugin is on, also i need to refresh beneficiary types in the menu after a CRUD operation on it

beygorghor:
- fixed menu for Polio
- order types by name
- use account feature flag (SHOW_BENEFICIARY_TYPES_IN_LIST_MENU) to display beneficiaries as list or by types

hat/assets/js/apps/Iaso/constants/menu.tsx
```diff
@@ -0,0 +1,347 @@
...
+        basicItems.push({
+            label: formatMessage(MESSAGES.dhis2),
+            key: 'dhis2',
+            url: currentUser.account.default_version.data_source.url,
+            icon: props => <DHIS2Svg {...props} />,
+        });
+    }
+
+    // filter by user permissions
+    const items = [...basicItems, ...pluginsMenu, admin].filter(menuItem => {
```
hakifran, line 345: This three objects don't have the same structure on label key that is why the error persist when trying to access the menu on polio. For **basicItems** and **admin** labels are string but for **pluginsMenu** it's a translation object.

Screenshot from 2023-08-22 11-21-32

hakifran: @beygorghor sorry, my bad. When I pulled it didn't succeed that why I was still getting the bug.

hat/assets/js/cypress/integration/07 - entities/types.spec.js
```diff
@@ -393,6 +384,35 @@ describe('Entities types', () => {
         });
     });
 
+    it('User wihtout write access should not be able to create,edit or delete a type', () => {
```
hakifran: Typo "without". but it's not blocking

iaso/api/entity.py
```diff
@@ -127,6 +127,64 @@ def get_queryset(self):
...
+        """
+
+        if not request.user.has_perm(permission.ENTITY_TYPE_WRITE):
+            return Response(status=status.HTTP_403_FORBIDDEN)
+
+        name = request.data.get("name", None)
+        if name is None:
+            raise serializers.ValidationError({"name": "This field is required"})
+        try:
+            print(pk)
```
hakifran: If it's not needed anymore, could you remove this print. but not blocking

iaso/api/entity.py
```diff
@@ -127,6 +127,64 @@ def get_queryset(self):
...
+        """
+        DELETE /api/entitytypes/{id}/
+        Provides an API to delete the an entity type
+        Needs iaso_entity_type_write permission
+        """
+        if not request.user.has_perm(permission.ENTITY_TYPE_WRITE):
+            return Response(status=status.HTTP_403_FORBIDDEN)
+
+        try:
+            obj = EntityType.objects.get(pk=pk)
```
hakifran: I don't know if I'm wrong but I think this **get_object_or_404** method can be used here, as it will try to get the object and raise 404 status code when the object doesn't exist. But not blocking too

hakifran (approved):
It's working fine apart from some comments I made which on my point of view are not blocking

hat/assets/js/apps/Iaso/constants/menu.tsx
```diff
@@ -0,0 +1,347 @@
...
+        const permissionsList = listMenuPermission(menuItem);
+        return userHasOneOfPermissions(permissionsList, currentUser);
+    });
+    return items;
+};
+
+export default useMenuItems;
```
quang-le: why not `export const useMenuItems` like we do in most places?

hat/assets/js/apps/Iaso/domains/app/components/MenuItemComponent.js
```diff
@@ -9,20 +9,20 @@ import {
...
     Collapse,
     List,
+    Box,
 } from '@material-ui/core';
 import ExpandLess from '@material-ui/icons/ExpandLess';
 import ExpandMore from '@material-ui/icons/ExpandMore';
 
 import PropTypes from 'prop-types';
 import {
-    injectIntl,
```
quang-le, line 18: ❤️

hat/assets/js/apps/Iaso/domains/app/components/MenuItemComponent.js
```diff
@@ -42,17 +42,24 @@ function MenuItem(props) {
...
         classes,
-        intl,
         subMenuLevel,
         currentPath,
-        currentUser,
         url,
     } = props;
-
+    const currentUser = useCurrentUser();
     const urlLink = url;
```
quang-le, line 50: I think this is useless

hat/assets/js/apps/Iaso/domains/app/components/MenuItemComponent.js
```diff
@@ -42,17 +42,24 @@ function MenuItem(props) {
...
-    const path = urlLink ? `${currentPath}` : `${currentPath}/${menuItem.key}`;
+    const { formatMessage } = useSafeIntl();
+
+    const path = useMemo(() => {
+        if (urlLink || !menuItem.key) {
+            return `${currentPath}`;
+        }
+        return `${currentPath}/${menuItem.key}`;
+    }, [currentPath, menuItem.key, urlLink]);
```
quang-le: They are all string, no? So `useMemo` is not necessary?

hat/assets/js/apps/Iaso/domains/app/components/SidebarMenuComponent.js
```diff
@@ -138,13 +121,16 @@ const SidebarMenu = ({
             </div>
             <Divider />
             <List className={classes.list}>
-                {menuItems.map(menuItem => {
-                    const permissionsList = listMenuPermission(menuItem);
-                    if (userHasOneOfPermissions(permissionsList, currentUser)) {
```
quang-le, line 143: nice to move that check in the hook 👍

hat/assets/js/apps/Iaso/domains/entities/entityTypes/index.tsx
```diff
@@ -64,26 +68,27 @@ export const EntityTypes: FunctionComponent<Props> = ({ params }) => {
...
+                        Permission.ENTITY_TYPE_WRITE,
+                        currentUser,
+                    ) && (
+                        <EntityTypesDialog
+                            titleMessage={MESSAGES.create}
+                            renderTrigger={({ openDialog }) => (
+                                <AddButtonComponent
+                                    dataTestId="add-entity-button"
+                                    onClick={openDialog}
+                                />
```
quang-le, line 81: Booooh, old modal

hat/assets/js/apps/Iaso/domains/entities/entityTypes/index.tsx
```diff
@@ -64,26 +68,27 @@ export const EntityTypes: FunctionComponent<Props> = ({ params }) => {
...
                 </Grid>
                 <Table
                     data={data?.types ?? []}
                     pages={data?.pages ?? 1}
                     defaultSorted={[{ id: 'name', desc: false }]}
-                    columns={columns({
-                        formatMessage,
-                        deleteEntityType,
-                        saveEntityType,
-                    })}
```
quang-le, line 86: ❤️

hat/assets/js/apps/Iaso/domains/entities/index.tsx
```diff
@@ -99,11 +95,26 @@ export const Beneficiaries: FunctionComponent<Props> = ({ params }) => {
...
+        if (entityTypeIds.length === 1) {
+            const currentType = types?.find(
+                type => `${type.value}` === entityTypeIds[0],
+            );
+            if (currentType) {
+                return currentType.label;
+            }
+        }
+        return undefined;
+    }, [entityTypeIds, types]);
```
quang-le: `entityTypeName` is `string | undefined` so the `useMemo` is not necessary

hat/assets/js/apps/Iaso/domains/users/permissionsMessages.ts
quang-le, line 1: You should document this, so that we don't forget to put all permission related massages in this file

hat/assets/js/apps/Iaso/libs/apiHooks.ts
```diff
@@ -140,7 +141,13 @@ const useBaseSnackMutation = <
...
             }
             if (invalidateQueryKey) {
-                queryClient.invalidateQueries(invalidateQueryKey);
+                if (isArray(invalidateQueryKey)) {
+                    invalidateQueryKey.forEach(queryKey =>
+                        queryClient.invalidateQueries(queryKey),
+                    );
+                } else {
+                    queryClient.invalidateQueries(invalidateQueryKey);
+                }
```
quang-le, line 150: YES!

quang-le: is the lodash version better than the JS `Array.isArray`?

iaso/api/entity.py
```diff
@@ -127,6 +127,64 @@ def get_queryset(self):
...
+        Provides an API to edit an Entity Type
+        Needs iaso_entity_type_write permission
+        """
+
+        if not request.user.has_perm(permission.ENTITY_TYPE_WRITE):
+            return Response(status=status.HTTP_403_FORBIDDEN)
+
+        name = request.data.get("name", None)
+        if name is None:
+            raise serializers.ValidationError({"name": "This field is required"})
```
quang-le, line 142: I think that check should be in the Serializer. Not blocking for me

iaso/api/entity.py
```diff
@@ -127,6 +127,64 @@ def get_queryset(self):
...
+        try:
+            print(pk)
+            entity_type = EntityType.objects.get(pk=pk)
+            entity_type.name = name
+            entity_type.fields_duplicate_search = request.data.get("fields_duplicate_search", None)
+            entity_type.fields_list_view = request.data.get("fields_list_view", None)
+            entity_type.fields_detail_info_view = request.data.get("fields_detail_info_view", None)
```
quang-le, line 148: this should maybe be validated by the serializer.
@SimonDus  @mathvdh Is it OK like this or is it worth refactoring? (Or have I got it wrong?)

quang-le (approved):
Nothing blocking for me. It would be good to have some feedback from a backend to make sure your `PATCH` is well withing guidelines

hat/assets/js/apps/Iaso/domains/app/components/MenuItemComponent.js
beygorghor: `currentUser` is used on line 65

hat/assets/js/apps/Iaso/libs/apiHooks.ts
beygorghor: Duno

iaso/api/entity.py
beygorghor: ya for those two maybe @SimonDus or @mathvdh can give their opinion

PR #751, opened 2023-08-11
POLIO-1127 nopv2 vaccine authorization page

SimonDus:
New API for Vaccine Authorizations. Check the ticket for details.
Cron To alert users if vacc auth is about to be expired in 60 days or has expired.

Related JIRA tickets : POLIO-1128 , POLIO-1166, POLIO-1182

- [x] Are there enough tests

## Changes

new api at /api/polio/vaccineauthorizations. 
One extra action at /api/polio/vaccineauthorizations/get_most_recent_authorizations/

There is also the front end under Polio Tab, Vaccine.

## How to test

Use the api, create entries, with differents users from different account, test the filters, the ordering... Check if the behavior works as expected.

Documentation will be added in the wiki.


plugins/polio/models.py
```diff
@@ -874,3 +874,26 @@ class Meta:
...
         return str(self.event)
+
+
+class VaccineAuthorization(SoftDeletableModel):
+    STATUS = (
+        ("ongoing", "Ongoing"),
+        ("signature", "Sent for signature"),
+        ("validated", "Validated"),
+        ("expired", "Expired"),
+    )
```
quang-le: Apparently we're switching to using `models.TextChoices` for this. See `RoundDatesHistory` for an example

quang-le (changes requested):
Filters don't work when passign `get_most_recent` param

plugins/polio/models.py
SimonDus: What's the benefice of this ? it's just creating a class for the choices outside of the model class ?

quang-le (changes requested):
The filter on status is not working properly: In the capture below, Burkina Faso should be excluded from the results after the filter is applied.

https://github.com/BLSQ/iaso/assets/38907762/cb8f4434-4efb-4c49-8db9-f3472e945209

plugins/polio/models.py
mathvdh: I think the benefit is more readability, and a common way of doing stuff in the project. Otherwise I think both ways are officially supported by Django, see https://docs.djangoproject.com/en/3.1/ref/models/fields/#enumeration-types

plugins/polio/api.py
```diff
@@ -2082,6 +2091,230 @@ def get_queryset(self):
...
+
+class CountryField(serializers.PrimaryKeyRelatedField):
+    def to_representation(self, value):
+        country = OrgUnit.objects.get(pk=value.pk)
+        serializer = CountryForVaccineSerializer(country)
+        return serializer.data
+
+
+class VaccineAuthorizationSerializer(serializers.ModelSerializer):
+    country = CountryField(queryset=OrgUnit.objects.filter(org_unit_type__name="COUNTRY"))
```
mathvdh, line 2158: Neat

plugins/polio/api.py
```diff
@@ -2082,6 +2091,230 @@ def get_queryset(self):
...
+        expiration_date = TimestampField(read_only=True)
+
+    def create(self, validated_data):
+        user = self.context["request"].user
+        validated_data["account"] = user.iaso_profile.account
+
+        return super().create(validated_data)
+
+
+class HassVaccineAuthorizationsPermissions(permissions.BasePermission):
```
mathvdh, line 2184: I'm wondering if we shouldnt generalize this to a Read / Write generic Auth Permission class that takes to basis permissions (read and write) or if it's not already there somewhere ?

plugins/polio/api.py
```diff
@@ -2082,6 +2091,230 @@ def get_queryset(self):
...
+        user = self.request.user
+        user_access_ou = OrgUnit.objects.filter_for_user_and_app_id(user, None)
+        user_access_ou = user_access_ou.filter(org_unit_type__name="COUNTRY")
+        country_id = self.request.query_params.get("country", None)
+        queryset = VaccineAuthorization.objects.filter(account=user.iaso_profile.account, country__in=user_access_ou)
+        block_country = self.request.query_params.get("block_country", None)
+        search = self.request.query_params.get("search", None)
+        auth_status = self.request.query_params.get("auth_status", None)
+
+        if country_id:
```
mathvdh, line 2256: I much prefer to define or extend custom filters for this, block_country ... search .. auth_status .. See [iaso/api/deduplication/filters.py](https://github.com/BLSQ/iaso/blob/ccc24e3e2ca3edd351b923aaaefe43e13a0e418c/iaso/api/deduplication/filters.py) and `EntityDuplicateViewSet` in [iaso/api/deduplication/entity_duplicate.py](https://github.com/BLSQ/iaso/blob/fb93ea9bd51cf86bbde48a10c1a9a943249b96e7/iaso/api/deduplication/entity_duplicate.py) for some examples ... It might be personal preferences but it's much cleaner to my taste

plugins/polio/api.py
```diff
@@ -2082,6 +2091,230 @@ def get_queryset(self):
...
+        country = self.request.data.get("country")
+        if not self.request.user.is_superuser:
+            if country not in [
+                str(ou.id) for ou in OrgUnit.objects.filter_for_user_and_app_id(self.request.user, None)
+            ]:
+                raise serializers.ValidationError({"Error": "You don't have access to this org unit."})
+
+        return super().create(request)
+
+    @action(detail=False, methods=["POST", "GET", "PUT", "DELETE"])
```
mathvdh: Not sure why we are allowing all theses methods ? Shouldn't it be only GET or I'm missing something ?

mathvdh (commented):
I have only checked the backend part .. See some of my comments, nothing major or blocking I would say

plugins/polio/models.py
kemar: [Enumeration](https://docs.python.org/3/library/enum.html#module-enum) types are cool 😎

`models.TextChoices` gives you a small API useful in tests and in the code:

```python
Model.Status.choices
# [('A', 'Foo'), ('B', 'Bar'), ('C', 'Baz')]

dict(Model.Status.choices)

Model.Status.values
# ['A', 'B', 'C']

Model.Status.labels
# ['Foo', 'Bar', 'Baz']

Model.Status.A.value
# 'A'

Model.Status.A.label
# 'Foo'

str(Model.Status.A) == Model.Status.A.value
```

kemar (commented):
I've read it but don't feel confident enough yet to approve it because it's very big and I'm still a newbie on this project :)

mathvdh (approved):
Ok we can improve the auth thing in the future. It's more cosmetic than anything

quang-le (approved):

PR #753, opened 2023-08-16
 IA-2360: Validate Feature Flags when project is updated

bmonjoie:
Some Feature Flags won't work without the 'REQUIRE_AUTHENTICATION' one added alongside. 
Therefore, we mark those as requiring authentication and we validate on project creation/update.

Related JIRA tickets: IA-2360

- [X] Are there enough tests

## Changes

A new field was added on the model `feature_flag` called `requires_authentication`.
This allows us to check when the project is created/updated that the `REQUIRE_AUTHENTICATION` feature flag is also present or it raises a 400.

## How to test

1. Go into the project edition screen.
1. Try to create or update a project 
1. In the feature flag screen, add "Planning" only
1. Save and you should see an error
1. Add "REQUIRE_AUTHENTICATION" (might have to be added in the admin first, I don't think there is a migration for it)
1. Save and it should work

### DIFF
--- a/iaso/admin.py
+++ b/iaso/admin.py
@@ -268,7 +268,7 @@ class ProjectAdmin(admin.ModelAdmin):
 
 @admin_attr_decorator
 class FeatureFlagAdmin(admin.ModelAdmin):
-    list_display = ("code", "name")
+    list_display = ("code", "name", "requires_authentication")
 
 
 @admin_attr_decorator
--- a/iaso/api/apps/serializers.py
+++ b/iaso/api/apps/serializers.py
@@ -1,11 +1,14 @@
 import logging
 
 from rest_framework import serializers
+from rest_framework.exceptions import ValidationError
 
 from iaso.models import Project, FeatureFlag, Form
 from ..projects import ProjectSerializer
 
 logger = logging.getLogger(__name__)
+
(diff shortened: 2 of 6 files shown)


beygorghor (approved):
So we need to make something in the front too in order to display if it requires authentification no ?

bmonjoie:
@beygorghor No, we don't, the error message returned by the backend is enough and well displayed in the UI

quang-le:
> @beygorghor No, we don't, the error message returned by the backend is enough and well displayed in the UI

But is it translated?🤓

mathvdh (approved):
@bmonjoie just a question to be sure to understand : 

We have a feature flag "REQUIRES_AUTHENTICATION" that should be added if we are adding any feature flag that does require authentication ? 

Otherwise it seems good to me

bmonjoie:
@mathvdh Yep, that's correct.

If you are trying to enable a feature flag with the boolean `requires_authentication` to true, the other feature flag `REQUIRES_AUTHENTICATION` must be enabled simultaneously. Otherwise, it will raise an exception.

mathvdh:
@bmonjoie it seems that when I merge this branch the python tests on main fail ... it's because we have migration conflicts with the previous branch merged .. Can you please merge main and fix the migration conflict ? Or let me know if you want me to do it .. ?

mathvdh:
@bmonjoie I fixed it .. No worries

PR #755, opened 2023-08-21
HOTFIX: fix round history bug with round 0

quang-le:
Polio users get 500 when adding round 0

Related JIRA tickets :  POLIO-1139

- [ ] Are there enough tests

## Changes

- prevent users to create datelog when adding round 0
- backend: raise error when receiving datelogs on create

## How to test

Explain how to test your PR.

## Print screen / video

BEFORE
Screenshot 2023-08-21 at 14 53 29

AFTER 

Screenshot 2023-08-21 at 14 56 31

## Notes

CAN BE DEPLOYED AS HOTFIX

### DIFF
--- a/plugins/polio/js/src/components/Rounds/RoundDates/RoundDates.tsx
+++ b/plugins/polio/js/src/components/Rounds/RoundDates/RoundDates.tsx
@@ -14,6 +14,7 @@ import { RoundDatesHistoryModal } from '../RoundDatesHistory/RoundDatesHistoryMo
 
 type Props = {
     roundIndex: number;
+    roundNumber: number;
     setParentFieldValue: (
         // eslint-disable-next-line no-unused-vars
         field: string,
@@ -27,6 +28,7 @@ type Props = {
 
 export const RoundDates: FunctionComponent<Props> = ({
     roundIndex,
+    roundNumber,
     setParentFieldValue,
     parentFieldValue,
     // TODO pass parentFieldValue to allow spreading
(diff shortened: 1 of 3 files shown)


plugins/polio/serializers.py
```diff
@@ -356,6 +356,9 @@ def create(self, validated_data):
         destructions = validated_data.pop("destructions", [])
         started_at = validated_data.get("started_at", None)
         ended_at = validated_data.get("ended_at", None)
+        datelogs = validated_data.get("datelogs", None)
+        if datelogs:
+            raise serializers.ValidationError({"datelogs": "Cannot have modification history for new round"})
```
madewulf, line 361: How could this have happened previously? Why is this needed?

quang-le: It could happen previously because the UI would let users add a `datelog` when they would create a round 0 in a campaign that has already other rounds. 

the PR fixes the problem front-end side. I just added the check in the backend for good measure

plugins/polio/js/src/components/Rounds/RoundDates/RoundDates.tsx
```diff
@@ -36,11 +38,21 @@ export const RoundDates: FunctionComponent<Props> = ({
         values: { rounds = [] },
         initialValues,
     } = useFormikContext<Campaign>();
+    // const currentStartDate = rounds.find(
+    //     round => round.number === roundNumber,
+    // )?.started_at;
+    // const currentEndDate = rounds.find(
+    //     round => round.number === roundNumber,
+    // )?.ended_at;
```
beygorghor: 🗑️

plugins/polio/js/src/components/Rounds/RoundDates/RoundDates.tsx
```diff
@@ -36,11 +38,21 @@ export const RoundDates: FunctionComponent<Props> = ({
...
     const currentStartDate = rounds?.[roundIndex]?.started_at;
     const currentEndDate = rounds?.[roundIndex]?.ended_at;
+    // For initial data, wee need to perform a find, because if we're adding round 0
+    // We'll be addinga round at the start of the rounds array which will lead to index related errors
     const hasInitialData = Boolean(
-        initialValues.rounds?.[roundIndex]?.started_at &&
-            initialValues.rounds?.[roundIndex]?.ended_at,
+        initialValues.rounds.find(round => round.number === roundNumber)
```
beygorghor, line 46: i would prefer to store round initial values in another variable

beygorghor (approved):
it's working but can be improved a bit

plugins/polio/js/src/components/Rounds/RoundDates/RoundDates.tsx
quang-le: I'm not convinced it would be more legible. Since `initialValues` comes from formik, it's pretty transparent. Since `rounds` is already used, I would have to make `intialRounds`  or something, which is not much shorter and less obviously related to formik.

PR #758, opened 2023-08-22
POLIO-1133: add new reason for delay

quang-le:
Add "Campaign moved forward by MOH" to the list of reasons for delay of campaign rounds (polio)

Related JIRA tickets :  POLIO-1133

- [ ] Are there enough tests

## Changes
- Update list of reasons back-end side
- add migration
- update front-end code (add reason to list and type)
- Add translation

## How to test

Go to a campaign
Edit round dates
The option "Campaign moved forward by MOH" should be available.

## Print screen / video

Screenshot 2023-08-22 at 18 45 43


hakifran (approved):
Good to me. Only thing(not blocking) I noticed, I may be wrong or not relevant, there is no test to check if the given reason is in the scope of available reasons. A part from that every thing are good.

beygorghor (approved):

quang-le:
> Good to me. Only thing(not blocking) I noticed, I may be wrong or not relevant, there is no test to check if the given reason is in the scope of available reasons. A part from that every thing are good.

Good catch, I'll double check!

PR #760, opened 2023-08-23
IA-2310: Data source initial setup

hakifran:
- When setting an account it creates datasource, source version and account but not the project
- This ticket is about to setup an account with an initial project

Related JIRA tickets : [IA-2310](https://bluesquare.atlassian.net/browse/IA-2310)

- [x] Are there enough tests

## Changes
- Just add the project creation when setting up an account and link it to the initial data source

## How to test
- Setup an account with api/setup_account
- Login and go in the list of projet there will be an initial project

## Print screen / video
[Screencast from 2023-08-23 10-40-40.webm](https://github.com/BLSQ/iaso/assets/19631540/650c3c56-479f-406f-bad7-1ecab77a1afb)

[Screencast from 2023-08-23 10-41-20.webm](https://github.com/BLSQ/iaso/assets/19631540/d4881cca-59cf-4abf-9db7-3dfc0fa78fb7)

[IA-2310]: https://bluesquare.atlassian.net/browse/IA-2310?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


beygorghor (approved):

iaso/api/setup_account.py
```diff
@@ -52,6 +51,21 @@ def create(self, validated_data):
...
         )
 
+        # Create a setup_account project with an app_id represented by the account name
+        initial_project = Project.objects.create(
+            name=validated_data["account_name"] + " project", account=account, app_id=validated_data["account_name"]
+        )
+
+        # Create an initial orgUnit type and link it to project
+        initial_orgunit_type = OrgUnitType.objects.create(name="COUNTRY", short_name="country", depth=1)
+        initial_orgunit_type.projects.set([initial_project])
```
quang-le: I'm not sure about this. Will all projects have "COUNTRY" as org unit type? If not, will it not mess up when importing a data source?

iaso/api/setup_account.py
```diff
@@ -52,6 +51,21 @@ def create(self, validated_data):
             user_manual_path=validated_data.get("user_manual_path"),
         )
 
+        # Create a setup_account project with an app_id represented by the account name
+        initial_project = Project.objects.create(
+            name=validated_data["account_name"] + " project", account=account, app_id=validated_data["account_name"]
```
bmonjoie: I would probably make sure that the account name doesn't have spaces in its name for the app_id:
```suggestion
            name=validated_data["account_name"] + " project", account=account, app_id=validated_data["account_name"].replace(' ', '.')
```

iaso/api/setup_account.py
bmonjoie: Also, does it have to be shouting at the user? :sweat_smile: 

I think the name would be much better as "Country"

iaso/tests/api/test_setup_account.py
```diff
@@ -122,3 +122,17 @@ def test_setup_account_has_all_perms(self):
...
+        data = {
+            "account_name": "initial_project_account",
+            "user_username": "username",
+            "password": "password",
+        }
+
+        response = self.client.post("/api/setupaccount/", data=data, format="json")
+        self.assertEqual(response.status_code, 201)
+        project = m.Project.objects.filter(name="initial_project_account project")
+        self.assertEqual(len(project), 1)
```
bmonjoie: This is not testing that the project has the correct app_id, that it's linked to the correct account, or that it is linked to the proper data source.

iaso/api/setup_account.py
hakifran: > @quang-le The idea of the ticket is to create a complete account setup so that the user can directly create an orgUnit, the Country is the default orgUnit type. I added the default one because it's not possible for the user to create directly an orgunit without the orgunit type.

iaso/api/setup_account.py
```diff
@@ -52,6 +51,26 @@ def create(self, validated_data):
             user_manual_path=validated_data.get("user_manual_path"),
         )
 
+        # Create a setup_account project with an app_id represented by the account name
+        app_id = (
+            validated_data["account_name"]
+            if " " not in validated_data["account_name"]
+            else validated_data["account_name"].replace(" ", "-")
+        )
```
bmonjoie: Why not just:
```suggestion
        app_id = validated_data["account_name"].replace(" ", "-")
```

iaso/api/setup_account.py
```diff
@@ -52,6 +51,26 @@ def create(self, validated_data):
...
+        initial_project = Project.objects.create(
+            name=validated_data["account_name"] + " project", account=account, app_id=app_id
+        )
+
+        # Create an initial orgUnit type and link it to project
+        initial_orgunit_type = OrgUnitType.objects.create(name="Country", short_name="country", depth=1)
```
bmonjoie, line 62: I wonder if depth shouldn't be `0` here.
@madewulf, any idea on this?

hakifran: @bmonjoie I checked in the DB how the depth is normally used and I found that it's like orgUnit levels as for Country we have 1, Region there is 2  and district it's  3

bmonjoie (approved):

PR #761, opened 2023-08-23
IA-2377: Fix issue in users management

bmonjoie:
Super-admins have both user admin and user management rights.
Since the code was checking for user management rights, the restrictions were also applied to admins.

Related JIRA ticket: IA-2377

- [X] Are there enough tests

## Changes

The endpoints are protected with a check that the user has either the user admin or user management permission. Therefore, instead of checking if the user has user management permission before applying the restrictions, we check if the user doesn't have user admin permissions.

## How to test

Login with a super user and try to assign the user admin permission to a user

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/users/components/PermissionsSwitches.tsx
+++ b/hat/assets/js/apps/Iaso/domains/users/components/PermissionsSwitches.tsx
@@ -16,6 +16,7 @@ import { Permission } from '../../userRoles/types/userRoles';
 import { useGetUserRolesDropDown } from '../hooks/useGetUserRolesDropDown';
 import { userHasPermission } from '../utils';
 import * as Permissions from '../../../utils/permissions';
+import { useCurrentUser } from '../../../utils/usersUtils';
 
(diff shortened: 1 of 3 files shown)


beygorghor (approved):

mathvdh (approved):
This seems OK to me.
I don't really understand the replacement of :

`if user.has_perm(USERS_MANAGED) and permission_codename == _USERS_ADMIN:
`
by :

`if not user.has_perm(USERS_ADMIN) and permission_codename == _USERS_ADMIN:`

bmonjoie:
@mathvdh That's basically the bug I'm fixing.

In the case of a super admin, the following condition: 
`if user.has_perm(USERS_MANAGED) and permission_codename == _USERS_ADMIN: ` 
will be true and trigger the exception.

But the other condition:
`if not user.has_perm(USERS_ADMIN) and permission_codename == _USERS_ADMIN:`
will be false and therefore not trigger the exception.

Yet, both will be true for a user with "only" the `USERS_MANAGED` permission and therefore trigger the exception for those and not for `USERS_ADMIN` nor super admin.

PR #763, opened 2023-08-23
Polio 1141 refresh dashboard api

quang-le:
refreshing prepâredness data from the Ui doesn't update the endpoint used for the powerBi dashboards

Related JIRA tickets :  POLIO-1141

- [X] Are there enough tests

## Changes

- Add endpoint to launch `refresh_data` and pass it an `obr_name` (to avoid refreshing all campaigns, which takes about an hour)
- Update front-end code to post on new endpoint when clicking "refresh" button.

## How to test

- Start a worker (see README for the command)
- Go to polio
- Find or create a campaign wit at least 1 round in the future
- Select campaign go to preparedness tab
- Select the future round and click "generate spreadsheet"
- Open the sheet in a new tab. Fill in some values
- Go back to polio,  and click "Refresh preparedness data"
- When the data has finished refreshing, go to Tasks
- You should see a `refresh_data` task with a timestamp matchiing your refresh (see video)

You can also open the browser inspector and check that a call to `tasks/create/refreshpreparedness` is made when you click "Refresh preparedness data".

## Print screen / video

https://github.com/BLSQ/iaso/assets/38907762/b068481d-a429-401b-8217-155fd4f38d14

### DIFF
--- a/plugins/polio/tasks/api/create_refresh_preparedness_data.py
+++ b/plugins/polio/tasks/api/create_refresh_preparedness_data.py
@@ -1,4 +1,3 @@
-import logging
 from plugins.polio.models import Campaign
 from plugins.polio.tasks.refresh_preparedness_data import refresh_data
 
@@ -7,11 +6,7 @@ from rest_framework.response import Response
 
 from iaso.api.common import HasPermission
 from iaso.api.tasks import TaskSerializer
-from iaso.models import DataSource
-from iaso.tasks.dhis2_ou_importer import dhis2_ou_importer
 from hat.menupermissions import models as permission
-
-logger = logging.getLogger(__name__)
 
 
 class RefreshPreparednessLaucherSerializer(serializers.Serializer):
--- a/plugins/polio/tasks/api/create_refresh_preparedness_data.py
+++ b/plugins/polio/tasks/api/create_refresh_preparedness_data.py
@@ -1,0 +1,42 @@
+import logging
+from plugins.polio.models import Campaign
+from plugins.polio.tasks.refresh_preparedness_data import refresh_data
+
+from rest_framework import viewsets, permissions, serializers
+from rest_framework.response import Response
+
+from iaso.api.common import HasPermission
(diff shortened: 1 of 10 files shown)


beygorghor (approved):

plugins/polio/tasks/api/create_refresh_preparedness_data.py
```diff
@@ -0,0 +1,42 @@
...
+from rest_framework import viewsets, permissions, serializers
+from rest_framework.response import Response
+
+from iaso.api.common import HasPermission
+from iaso.api.tasks import TaskSerializer
+from iaso.models import DataSource
+from iaso.tasks.dhis2_ou_importer import dhis2_ou_importer
+from hat.menupermissions import models as permission
+
+logger = logging.getLogger(__name__)
```
mathvdh: You don't seem to be using it ...

mathvdh (changes requested):
Except for my minor comment, It's a great job!
Also, I was wondering if we shouldn't rate-limit this endpoint @quang-le as it's launching a task in the background. Or at least maybe check if an existing task is not running already for this obr_name ? (Not sure how long the task take in background..)

plugins/polio/tasks/api/create_refresh_preparedness_data.py
mathvdh: I mean the "logger"

quang-le:
> Except for my minor comment, It's a great job! Also, I was wondering if we shouldn't rate-limit this endpoint @quang-le as it's launching a task in the background. Or at least maybe check if an existing task is not running already for this obr_name ? (Not sure how long the task take in background..)

The task is very short when you only run it for one round:
Screenshot 2023-08-29 at 15 36 04

Also the button is deactivated while refreshing the data, so multiple clicks should not happen (the other query takes longer than the task):

https://github.com/BLSQ/iaso/assets/38907762/5a77b169-f8f9-410b-bfa8-e7bb3bf06527

mathvdh (approved):

PR #765, opened 2023-08-25
POLIO-1156 Proposed temporary fix

mathvdh:
We are seeing performance issues on campaigns installation (POLIO Outbreaks).
After analysis they seem to be related to the preparedness dashboard which implies a lot of computations and fetching from google sheets API.

Related JIRA tickets : POLIO-1156

## Changes

We propose a "quick" fix that should fix the performance issues. Basically we are not expiring our cache after 24h and we are gonna launch a manual task to cache the results of old campaigns too (until now only the campaigns rounds not older than 180 days where cached).

## Notes

Every night a task is launch by cron to refresh the data in cache (data of the last 180 days)
After that fix we are gonna launch a manual task which will put in cache all available data (not only 180 days). The task has been modified to allow for an optional parameter to override this 180 days

### DIFF
--- a/plugins/polio/preparedness/client.py
+++ b/plugins/polio/preparedness/client.py
@@ -23,7 +23,7 @@ class QuotaLimiter:
     """This class allow limiting the number of call to an API within the allowed quota.
 
     Just call the function, and then it will wait before returning if we are over the rate.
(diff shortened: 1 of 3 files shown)


mathvdh:
FYI this has already been released as a hotfix on polio prod (campaigns)

quang-le (approved):

PR #766, opened 2023-08-25
Polio-1129 vaccine auth front end

quang-le:

Related JIRA tickets : POLIO-1129

- [ ] Are there enough tests

## Changes

- Add pages for nOPV2 Authorisation: list, details + create/edit modals
- Modal has a validation that prevents setting a date in the future if status is EXPIRED
- There's an ugly redirection with `useEffect` to add the `order`, pageSize` and `page` params when they're missiing. Couldn't use the 'redirections' of vanilla iaso, so 🤷‍♂️

## How to test

- Go to polio > vaccine module > nOPV2 Authorisations
- Check the table and filters
- Click on a row to go to details view (per country)
- Try adding, editing and deleting an authorisation

**Note**: current expiration date is the most recent expiration date from authorisations with either VALIDATED or EXPIRED status. next expiration date is the most recent expiration date from authorisations with either ONGOIND or SIGNATRE status

⚠️ This PR merges into  https://github.com/BLSQ/iaso/pull/751, which will fix the block of countries filter, and the table sorting⚠️

## Print screen / video

https://github.com/BLSQ/iaso/assets/38907762/5419adc9-356d-41c4-bb06-754f2a407407


hat/assets/js/apps/Iaso/utils/requests.js
```diff
@@ -392,3 +392,25 @@ const lockInstance = dispatch => instance =>
...
+    const copy = cleanupParams(params);
+    if (params.pageSize) {
+        copy.limit = params.pageSize;
+        delete copy.pageSize;
+    }
+    if (params.accountId) {
+        delete copy.accountId;
+    }
+    return copy;
+};
```
beygorghor, line 416: ❤️

plugins/polio/api.py
```diff
@@ -2121,7 +2121,12 @@ def has_permission(self, request, view):
                 or request.user.is_superuser
             )
             return can_get
-        elif request.method == "POST" or request.method == "PUT" or request.method == "DELETE":
+        elif (
+            request.method == "POST"
+            or request.method == "PUT"
+            or request.method == "PATCH"
+            or request.method == "DELETE"
```
beygorghor, line 2150: i'm wondering if we are using other methods

plugins/polio/js/src/pages/VaccineModule/Nopv2Authorisations/Table/Nopv2AuthorisationsTable.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { useGetLatestAuthorisations } from '../hooks/api';
+import { useNopv2AuthTableColumns } from './useNopv2AuthTableColumns';
+import { VaccineAuthParams } from '../types';
+
+type Props = { params: VaccineAuthParams };
+
+export const Nopv2AuthorisationsTable: FunctionComponent<Props> = ({
+    params,
+}) => {
+    // Waiting for backend API fix to uncomment
```
beygorghor: ?

plugins/polio/js/src/pages/VaccineModule/Nopv2Authorisations/hooks/api.ts
```diff
@@ -0,0 +1,108 @@
...
+            // TODO refactor when useSnackMtation refactor is merged
+            onSuccess: () => {
+                queryClient.invalidateQueries('nopv2-auth');
+                queryClient.invalidateQueries('latest-nopv2-auth');
+            },
+        },
+    });
+};
+
+const createEditNopv2Authorisation = body => {
```
beygorghor: no typings for this ?

plugins/polio/js/src/pages/VaccineModule/Nopv2Authorisations/hooks/statuses.ts
```diff
@@ -0,0 +1,24 @@
...
+    'expired',
+];
+
+export const useStatusOptions = (): DropdownOptions<string>[] => {
+    const { formatMessage } = useSafeIntl();
+    return useMemo(() => {
+        return statuses.map(status => {
+            return {
+                value: status,
+                label: formatMessage(MESSAGES[status]),
```
beygorghor, line 20: better return the key if the backend return unknown status no ?

plugins/polio/js/src/pages/VaccineModule/Nopv2Authorisations/hooks/validation.ts
```diff
@@ -0,0 +1,50 @@
...
+            const { path, createError, parent } = context;
+            const now = moment();
+            const newExpiryDate = moment(value);
+            const { status } = parent;
+
+            let errorMessage;
+
+            if (newExpiryDate?.isAfter(now) && status === 'expired') {
+                errorMessage = formatMessage(MESSAGES.dateForExpired);
+            }
```
beygorghor, line 21: for this, you are using local time with a date saved in DB, are you sure this is going to work while the DB date has been saved in an other time zone ?

beygorghor (changes requested):
Some minor remarks, but i could not test the front locally
I don't have vaccination module in my DB and the admin is not loading correctly to add it manualy. Am I missing something ?

plugins/polio/api.py
quang-le: ping @SimonDus

quang-le: But my stance would be: we'll add other methods when we need them

plugins/polio/js/src/pages/VaccineModule/Nopv2Authorisations/hooks/statuses.ts
quang-le: I don't know, we don't fetch the statuses from the backend so, it will not really be dynamic anyway

plugins/polio/js/src/pages/VaccineModule/Nopv2Authorisations/hooks/validation.ts
quang-le: We only compare dates from the browser so I think it's fine

PR #767, opened 2023-08-28
IA-1976: Enable personalized user guide per IASO account

hakifran:
- For each acccount, only IASO user manual was accessible even for POLIO account
- This ticket is to enable personalized user guide per account
Related JIRA tickets : [IA-1976](https://bluesquare.atlassian.net/browse/IA-1976)

- [ ] Are there enough tests

## Changes

- Create a migration to add user_manual_path to account
- Add user_manual_path field to account as_dict method 
- On the front check if the account user_manual_path is null, if yes display the default user manual if not display the account user manual path

## How to test
- Edit an account by adding a personalized user guide link
- Go on the bottom of the sidebar and check if the user manual is the one added on the account

[IA-1976]: https://bluesquare.atlassian.net/browse/IA-1976?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hakifran:
@cypress

beygorghor (approved):

hat/assets/js/apps/Iaso/domains/app/components/SidebarMenuComponent.js
```diff
@@ -79,10 +80,13 @@ const styles = theme => ({
 });
 
 const localizedManualUrl = (locale, account) => {
-    if (locale === 'fr' && account === 'RDC') {
+    if (locale === 'fr' && account.name === 'RDC') {
```
mathvdh, line 83: Why do we have this specific case ? I know you didn't write it in the first place but I'm just wondering. Shouldn't we have a way to have specific user manuals for specific locales. Like multiple user manuals per account (one for fr, one for en .. etc ?)

mathvdh (approved):
Seems fine to me, except for my question above, but I guess this should be handled separetely

mathvdh (approved):
Seems fine to me, except for my question above, but I guess this should be handled separetely

hat/assets/js/apps/Iaso/domains/app/components/SidebarMenuComponent.js
hakifran: I didn't change that code to avoid that persons who use IASO RDC don't miss the user guide. And for multiple languages for user manual , on my opinion it should be treated later.

PR #768, opened 2023-08-29
POLIO-1162: Fix LQAS Afro Map district view

quang-le:
When switching the map to "distrcit" view without zooming in, the API would return an empty list on big screens

Related JIRA tickets :POLIO-1162

- [ ] Are there enough tests

## Changes

- The culprit was `Polygon.from_bbox` which would change the bounds when longitude would cover more than 180°. 
- Add `make_safe_bounds` function and use it to generate a tuple that is used to create the `Polygon`

## How to test

Go to Afro map and chack the "District" radio button

## Print screen / video

Screenshot 2023-08-29 at 10 16 07

##Notes

Thanks @edarchis  who actually found the solution of this one

### DIFF
--- a/plugins/polio/api.py
+++ b/plugins/polio/api.py
@@ -77,6 +77,7 @@ from .helpers import (
     CustomFilterBackend,
     calculate_country_status,
     determine_status_for_district,
+    make_safe_bbox,
 )
 from .vaccines_email import send_vaccines_notification_email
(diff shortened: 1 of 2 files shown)


mathvdh (approved):
Seems fine, following the conversation with Eric.

kemar (approved):
LGTM

PR #777, opened 2023-09-01
IA-2379: Completeness stats: page crashes when trying to show parent of top org unit

beygorghor:
When “Digging down” the table then back up, the button to show parent is still shown when the top org unit is reached. This results in undefined being passed as org unit id to the API, which crashes  the page.

Probably related: on the map, there seems to be an edge case where the map wrongly remove the button to show parent

Related JIRA tickets : IA-2379

- [ ] Are there enough tests

## Changes

`undefined` was added as a string

## How to test

Open completeness stats, select a form. Drill down and up in the health pyramid

## Print screen / video

https://github.com/BLSQ/iaso/assets/12494624/fa10e817-c7d4-4d4c-804c-0f62da75bc41


quang-le (approved):
Code looks OK to me 😂

PR #779, opened 2023-09-01
Polio 1164 add endpoint for lqas round data

quang-le:
OpenHexa pipeline needed an endpoint to safely PATCH lqas districts passed and failed to round in DB.

Related JIRA tickets : POLIO-1164

- [ ] Are there enough tests

## Changes

- Add `RoundViewSet` and  `api/polio/rounds` endpoint. It's only used for the custom PATCH request for now.
- Add `updatelqasfields` method to `RoundViewset`. It takes the round number, campaign obr_name and the values of the lqas fields to update.
- Override `partial_update` method in `CampaignViewSet` to prevent unsafe updates.

## How to test

Go to polio> Campaigns
Choose and obr_name
Go to `/swagger` and find `/api/polio/rounds/updatelqasfields`
Click try it out
Fill in all fields and send the request

## Print screen / video

https://github.com/BLSQ/iaso/assets/38907762/ea75d61b-58e0-4927-b52d-3772d0f1fe44

## Notes

Requires https://github.com/BLSQ/iaso-pipelines/pull/10 to be merged and run as pipeline in OpenHexa

### DIFF
--- a/plugins/polio/api.py
+++ b/plugins/polio/api.py
@@ -27,6 +27,7 @@ from django.utils.timezone import now, make_aware
 from django.views.decorators.cache import cache_page
 from django_filters.rest_framework import DjangoFilterBackend  # type: ignore
 from gspread.utils import extract_id_from_url  # type: ignore
+from hat.menupermissions import models as permission
 from openpyxl.writer.excel import save_virtual_workbook  # type: ignore
 from requests import HTTPError
 from iaso.api.serializers import OrgUnitDropdownSerializer
@@ -39,6 +40,7 @@ from rest_framework.request import Request
 from rest_framework.response import Response
 from iaso.api.common import (
     CSVExportMixin,
+    HasPermission,
     ModelViewSet,
     DeletionFilterBackend,
     CONTENT_TYPE_XLSX,
@@ -50,8 +52,10 @@ from plugins.polio.serializers import (
     ConfigSerializer,
     CountryUsersGroupSerializer,
     ExportCampaignSerializer,
+    LqasDistrictsUpdateSerializer,
     RoundDateHistoryEntrySerializer,
     PowerBIRefreshSerializer,
+    RoundSerializer,
 )
 from plugins.polio.serializers import (
     OrgUnitSerializer,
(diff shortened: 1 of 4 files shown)


mathvdh (approved):
No comments, except maybe that the tests seems to fail in CI

plugins/polio/api.py
```diff
@@ -513,6 +517,15 @@ def send_notification_email(self, request, pk, **kwargs):
 
         return Response({"message": "email sent"})
 
+    # We need to authorize PATCH request to enable restore_deleted_campaign endpoint
+    # But Patching the campign directly is very much error prone, so we disable it indirectly
+    def partial_update(self):
```
madewulf, line 522: This works, and is ok for this, but I'm quite sure that there is a cleaner way to disallow a method on a viewset: https://stackoverflow.com/questions/9549867/django-rest-framework-limit-the-allowed-methods-to-get

plugins/polio/api.py
```diff
@@ -2082,6 +2095,40 @@ def get_queryset(self):
             )
 
 
+@swagger_auto_schema(tags=["rounds"], request_body=LqasDistrictsUpdateSerializer)
+class RoundViewset(ModelViewSet):
+    # Patch should be in the list to allow updatelqasfields to work
+    http_method_names = ["patch"]
```
madewulf, line 2101: You used it here, the cleaner method to limit authorized actions.

plugins/polio/api.py
```diff
@@ -2082,6 +2095,40 @@ def get_queryset(self):
...
+    @action(detail=False, methods=["patch"], serializer_class=LqasDistrictsUpdateSerializer)
+    def updatelqasfields(self, request):
+        round_number = request.data.get("number", None)
+        obr_name = request.data.get("obr_name", None)
+        if obr_name is None:
+            raise serializers.ValidationError({"obr_name": "This field is required"})
+        if round_number is None:
+            raise serializers.ValidationError({"round_number": "This field is required"})
+        try:
+            round_instance = Round.objects.get(campaign__obr_name=obr_name, number=round_number)
```
madewulf, line 2122: Looks to me that this might allow to bypass limitations by country (say, you are allowed to modify only a country, you could still modify the rounds of another country with this method, I think

PR #785, opened 2023-09-04
Polio 1151 lqas afro map round data

quang-le:
LQAS Afro Map was not filtering future rounds properly

Related JIRA tickets : POLIO-1151

- [ ] Are there enough tests

## Changes

- Add method get_latest_active_campaign_and_rounds` to `LqasAfroViewset` to filter out:
    - future rounds
    - rounds with `lqas_ended_at` in the future (i.e: LQAs not over). If `lqas_ended_at` is null, we use `ended_at` + 10 days (same buffer as in the openhexa pipeline)

## How to test

- run iaso locally and use ngrok to allow openhexa to access your localhost
- clone https://github.com/BLSQ/iaso-pipelines/. checkout to version 63 (via git tags)
- Run the pipeline with your localhost as target
- In iaso, go to campaigns
- In an existing campaign, add a round with an end date in the future
- Go to afro map and look for the country of the campaign you just updated: the future round should not be displayed
- repeat with: 
    -  an lqas end date in the future (should not appear on the map)
    - no lqas end date and an end date more than 10 days in the past (should appear)
    - no lqas end date and and end date less than 10 days in the past (should not appear)
    - an lqas end date and a round end date less than 10 days in the past (should appear) 

## Print screen / video

https://github.com/BLSQ/iaso/assets/38907762/b322d9c2-52da-4e95-8ede-a1985e41eb9c


quang-le:
The constraints on dates are more strict, so the tests fail. I'll look into it

plugins/polio/helpers.py
```diff
@@ -201,3 +202,45 @@ def get_datastores(self):
...
+            .filter(campaign__deleted_at__isnull=True)
+            .exclude(campaign__is_test=True)
+            .order_by("-started_at")[:1]
+        )
+        latest_active_campaign = (
+            Campaign.objects.filter(id__in=Subquery(latest_active_round_qs.values("campaign")))
+            .filter(deleted_at=None)
+            .exclude(is_test=True)
+            .prefetch_related("rounds")
+            .first()
```
mathvdh, line 242: @quang-le I dont really get why you filter with a `id_in= Subquery` and then do `first()` above if anyway the subquery itself is supposed to return only one object (cf `.order_by("-started_at")[:1]`)

I'm not saying this is wrong I'm just wondering ... Or I'm not getting a subtility

plugins/polio/api.py
```diff
@@ -1930,36 +1918,34 @@ def list(self, request):
...
                 continue
-            sorted_rounds = sorted(latest_campaign.rounds.all(), key=lambda round: round.number, reverse=True)
             if requested_round == "latest":
-                round_number = sorted_rounds[0].number if len(sorted_rounds) > 0 else None
-            elif requested_round == "penultimate" and len(sorted_rounds) > 1:
-                round_number = sorted_rounds[1].number if len(sorted_rounds) > 1 else None
+                round_number = (
+                    latest_active_campaign_rounds[0].number if latest_active_campaign_rounds.count() > 0 else None
+                )
+            elif requested_round == "penultimate":
```
mathvdh: I know it's all over the place in this file (or in general in the project), but in general let's avoid using string directly in the code like "latest" or "penultimate"

You could use a Enum or models.TextChoices or just constants, for example with String Enums it could look like this

```python
from enum import Enum

class RoundNumber(str, Enum):
    Latest = "latest"
    Penultimate = "penultimate"
```

Then use `RoundNumber.Penultimate` instead of "penultimate"
or something similar with models.TextChoices  I guess (depending if it used with django models also).

Main reason are that if you make a spelling mistake it might create a difficult to see bug and also IDE auto completion.

plugins/polio/api.py
```diff
@@ -2037,14 +2023,17 @@ def get_queryset(self):
             )
         )
         # TODO see if we need to filter per user as with Campaign
-        return (
+        qs = (
             OrgUnit.objects.filter(org_unit_type__category="COUNTRY")
             .exclude(simplified_geom__isnull=True)
             .filter(simplified_geom__intersects=bounds_as_polygon)
         )
+        print("Query", qs.query)
```
mathvdh, line 2046: Do we have to keep this print on ?

mathvdh:
Also it seems there are some conflicts ...

plugins/polio/api.py
mathvdh: This can also be shared by multiple files if used in multiple places ...

plugins/polio/helpers.py
quang-le: my understanding is that `.order_by("-started_at")[:1])` still returns a queryset, whereas `first` does not and I wanted to access the instance

mathvdh (approved):
👍🏼

PR #786, opened 2023-09-04
WC2-252 enable export of workflows to another environment

mathvdh:
Enable the export of a (entity) workflow to a JSON
And then import of this JSON into (a possibly different instance of iaso).
(Example from staging to prod)

Related JIRA tickets : WC2-252

- [x] Are there enough tests

## Changes

Add endpoints for import export.
Add UUIDs on workflows, versions, changes, follow_ups in order to handle the case of "re-import" ..
Cleanup repeated code in tests
Add some code in test_utils.py

## How to test

After creating a workflow, got to `/api/workflows/export/{workflow_id}/` and get the JSON.
Then try importing it on `/api/workflows/import/` (no need for is as there are UUIDs in the JSON).
(You can try to modify some parts of the JSON and see if they are applied)
(See the tests for some example)

## Notes

This DOES not import and export the entities, entity types or forms used in the workflows, these have to be migrated independently.


iaso/api/workflows/import_export.py
```diff
@@ -0,0 +1,152 @@
...
+from iaso.models import Account, Workflow
+from iaso.models.entity import EntityType
+from iaso.models.forms import Form
+from iaso.models.workflow import Workflow, WorkflowChange, WorkflowFollowup, WorkflowVersion
+
+
+def export_workflow_real(workflow: Workflow) -> typing.Dict:
+    """Exports the given workflow as a dict
+    This version implies that an entity type with the same name and the same reference form exists.
+    It also implies that the forms in the follow ups and changes exist.
```
bmonjoie:
```suggestion
    This version assumes that an entity type with the same name and the same reference form exists.
    It also assumes that the forms in the follow-ups and changes exist.
```

iaso/api/workflows/import_export.py
```diff
@@ -0,0 +1,152 @@
...
+
+        workflow_data["versions"].append(version_data)  # type: ignore
+
+    return workflow_data
+
+
+def import_workflow_real(workflow_data: typing.Dict, account: Account) -> Workflow:
+    """Imports a workflow from a dict and returns a Workflow object
+    This version implies that an entity type with the same name and the same reference form exists.
+    It also implies that the forms in the follow ups and changes exist.
```
bmonjoie:
```suggestion
    This version assumes that an entity type with the same name and the same reference form exists.
    It also assumes that the forms in the follow-ups and changes exist.
```

bmonjoie (approved):

PR #787, opened 2023-09-05
IA-2342: Filter OrgUnits by Project

bmonjoie:
Allow to filter OrgUnitTypes by project
Allow to filter OrgUnits by project
Add "Project" select dropdown in OrgUnits' filters component

Related JIRA tickets : IA-2342

- [X] Are there enough tests

## Changes

When listing the OrgUnits, there could be multiple OrgUnitTypes in the dropdown with the same name (because they belong to different projects) and we don't know which one to select. This change adds a "Project" dropdown which filters the OrgUnitTypes dropdown list but also applies to the OrgUnit list search results.

## How to test

1. Ensure you have more than one project. 
1. Ensure you have OrgUnit types linked to your projects.
1. Go to the OrgUnits list screen
1. Open the OrgUnitTypes dropdown field, see that they are all listed
1. Select a project in the Project dropdown field
1. Open the OrgUnitTypes dropdown field, see that only the OrgUnitTypes of the selected project are all listed
1. Play with the search to only display the OrgUnits you want


hat/assets/js/apps/Iaso/domains/orgUnits/components/OrgUnitsFilters.tsx
```diff
@@ -274,14 +292,14 @@ export const OrgUnitFilters: FunctionComponent<Props> = ({
...
                             type="select"
-                            disabled={isFetchingOrgunitTypes}
+                            disabled={isFetchingOrgUnitTypes}
                             keyValue="version"
                             onChange={handleChange}
                             value={sourceVersionId}
                             label={MESSAGES.sourceVersion}
                             options={versionsDropDown}
                             clearable={false}
-                            loading={isFetchingOrgunitTypes}
```
hakifran, line 284: @bmonjoie is this means that the displaying of sources is related to the loading of the orgUnitTypes?

beygorghor (approved):
Working for me, it makes me think we don't have enough cypress test on this search...

hakifran (approved):
Seems good to me apart from the question I added to know if the display of the source dropdown is related to the loading of the orgUnitTypes.

hat/assets/js/apps/Iaso/domains/orgUnits/components/OrgUnitsFilters.tsx
bmonjoie: The only thing I did here is move the logic of fetching OrgUnitTypes within the component instead of outside it like it used to be. You see a line change here because I have fixed the case: `Orgunit` became `OrgUnit`.
I have no idea why the two were linked and I haven't changed that logic.

iaso/api/org_unit_types/viewsets.py
```diff
@@ -32,14 +35,14 @@ def destroy(self, request, pk):
 
     def get_queryset(self):
         queryset = OrgUnitType.objects.filter_for_user_and_app_id(
-            self.request.user, self.request.query_params.get("app_id")
+            self.request.user, self.request.query_params.get(APP_ID)
```
SimonDus, line 38: Is it normal that the params are in uppercase ? We do it lowercase everywhere. Same for every params in the PR

mathvdh (approved):
All good on the backend side for me.
Just a suggestion for the future (that we should all start follow).
Either implement a custom backend filter and filter there or implement the `filter_queryset` method instead of filtering in the `get_queryset` method

iaso/api/org_unit_types/viewsets.py
SimonDus: My bad it's friday

PR #793, opened 2023-09-08
POLIO-1190 link to dashboard were missing from mail alert in nOPV2 vaccine authorization

SimonDus:
The dashboard link to the platform was missing.

Related JIRA tickets : POLIO-1190

- [ ] Are there enough tests

## How to test

Expired : 

Create a vacc auth expired since 1 day exactly. As it's a cron job the easiest way to try the function is to run a python shell in iaso and run expired_vaccine_authorizations_email_alert(). 

You should see in the console that the mail have been sent and the content
you need a team of users named nOPV2 vaccine authorization alerts in order to make it work.

To Expire : 

Create a vacc auth with an expiration day equal to today's date + 60 days.  As it's a cron job the easiest way to try the function is to run a python shell in iaso and run vaccine_authorizations_60_days_expiration_email_alert(). 

You should see in the console that the mail have been sent and the content
you need a team of users named nOPV2 vaccine authorization alerts in order to make it work.

### DIFF
--- a/plugins/polio/tasks/vaccine_authorizations_mail_alerts.py
+++ b/plugins/polio/tasks/vaccine_authorizations_mail_alerts.py
@@ -3,12 +3,15 @@ import datetime as dt
 from datetime import timedelta
 
 from django.contrib.auth.models import User
+from django.contrib.sites.models import Site
 from django.core.mail import send_mail
 from django.shortcuts import get_object_or_404
 
 from hat import settings
 from iaso.models import Team
 from plugins.polio.models import VaccineAuthorization
+
+vaccine_dashboard_link = f"{Site.objects.get_current()}/dashboard/polio/vaccinemodule/nopv2authorisation/accountId/1/order/-current_expiration_date/pageSize/20/page/1"
 
 
 def vaccine_authorizations_60_days_expiration_email_alert():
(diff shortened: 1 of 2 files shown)


plugins/polio/tasks/vaccine_authorizations_mail_alerts.py
```diff
@@ -3,13 +3,16 @@
...
 from django.contrib.auth.models import User
+from django.contrib.sites.models import Site
 from django.core.mail import send_mail
 from django.shortcuts import get_object_or_404
 
 from hat import settings
 from iaso.models import Team
 from plugins.polio.models import VaccineAuthorization
 
+vaccine_dashboard_link = f"{Site.objects.get_current()}/dashboard/polio/vaccinemodule/nopv2authorisation/accountId/1/order/-current_expiration_date/pageSize/20/page/1"
```
quang-le: Do we hard code the account id everywhere in polio? if not maybe get it from the user?

It will work in prod, but if we get multiple testing environments, this may not work for future testers.

SimonDus: I changed it

quang-le (approved):

PR #795, opened 2023-09-11
New APIs proposition

bmonjoie:
Proposition to implement a new feature related to OrgUnit registry.

Related JIRA tickets : IA-2229


docs/pages/dev/reference/API/org_unit_registry.md
```diff
@@ -0,0 +1,436 @@
...
+  "page": "Long",
+  "pages": "Long",
+  "limit": "Long",
+  "results": [
+    {
+      "id": "Int - id in the database, I know Martin wants it",
+      "uuid": "UUID - provided by the client when request was created",
+      "status": "Enum<Status> - one of `new`, `validated`, `rejected`",
+      "org_unit_name": "String - name of the OrgUnit",
+      "org_unit_type_name": "String - name of the current OrgUnitType",
```
beygorghor, line 118: i would add org unit type id too here, or a small object sith id and name

docs/pages/dev/reference/API/org_unit_registry.md
```diff
@@ -0,0 +1,436 @@
...
+  "update_by": "String? - username of the user who accepted/rejected that request",
+  "updated_at": "Timestamp?",
+  "requested_fields": "Array<String> - name of the properties that were requested to change", 
+  "approved_fields": "Array<String>? - name of the properties that were approved to change",
+  "rejection_comment": "String? - Comment about why the changes were rejected",
+  "org_unit": {
+    "parent": "String - Name of the parent OrgUnit.",
+    "name": "String - Name of the OrgUnit.",
+    "status": "Enum<Status>? - Status of the OrgUnit, One of `open` or `close`, may be null or omitted.",
+    "org_unit_type": "String - Name of the OrgUnitType",
```
beygorghor: same here

docs/pages/dev/reference/API/org_unit_registry.md
```diff
@@ -0,0 +1,436 @@
...
+          }
+        ]
+      }
+    ]
+  },
+  "changes": {
+    "parent": "String? - Name of the new parent OrgUnit, may be null or omitted.",
+    "name": "String? - New name of the OrgUnit, may be null or omitted.",
+    "status": "Enum<Status>? - New status of the OrgUnit, One of `open` or `close`, may be null or omitted.",
+    "org_unit_type": "String? - Name of the new OrgUnitType, may be null or omitted",
```
beygorghor: ...

beygorghor (commented):
What about other fields like: `has_geo_json`, `validation_status`, `groups`, `creator`, `source`. Also if you want to use it for the searches, don't forget that we have multiple searches at the same time.

docs/pages/dev/reference/API/org_unit_registry.md
```diff
@@ -0,0 +1,436 @@
...
+some changes still. Therefore, the IDs can be either the DB id or the UUID generated by the client.
+Changes that are not provided are considered not being part of the change request.
+
+## Permissions
+
+User must be authenticated.
+
+## Query parameters
+
+- app_id: String - Must be provided, project for which this is created.
```
madewulf: I'm always wondering if it would not be cleaner to have the app_id in the POST for post requests

docs/pages/dev/reference/API/org_unit_registry.md
```diff
@@ -0,0 +1,436 @@
...
+```json
+{
+  "uuid": "UUID - Client generated UUID",
+  "org_unit_id": "String - id or UUID of the OrgUnit to change, if UUID doesn't exist in DB, it's a new OrgUnit.",
+  "created_at": "Timestamp - format same as /api/instances",
+  "updated_at": "Timestamp - format same as /api/instances",
+  "changes": { "": "Changes is not null and must contain at least one field",
+    "parent_id": "String? - id or UUID of the parent OrgUnit, may be null or omitted.",
+    "name": "String? - Name of the OrgUnit, may be null or omitted.",
+    "status": "Enum<Status>? - One of `open` or `close`",
```
madewulf: we already have validation_status, and consequently, to avoid confusion, I would call this one opening_status

docs/pages/dev/reference/API/org_unit_registry.md
```diff
@@ -0,0 +1,436 @@
...
+    "geo_point": { "": "New geopoint for the OrgUnit, may be null or omitted",
+      "latitude": "Double - New latitude of the OrgUnit",
+      "longitude": "Double - New longitude of the OrgUnit",
+      "altitude": "Double - New altitude of the OrgUnit",
+      "accuracy": "Double - New accuracy of the OrgUnit"
+    },
+    "geo-shape": "String? - GeoJSON representation, may be null or omitted",
+    "forms": [ "Array of form objects? - may be null or omitted, cannot be empty",
+      {
+        "form_id": "id of the form",
```
madewulf: Why send the form id? This can be inferred from the instance id. My intuition is that this should be a "instances" array, not a "forms" array

docs/pages/dev/reference/API/org_unit_registry.md
```diff
@@ -0,0 +1,436 @@
...
+- `app_id` was not provided
+- A not nullable field was null or omitted
+- `org_unit_id` is a new OrgUnit and the following fields were null or omitted: `parent_id`, `name`, `geo_point` and `org_unit_type_id`
+- A `String` field has an empty value
+- `parent_id` is not a valid OrgUnit
+- `status` was not one of the enum values.
+- `org_unit_type_id` is not a valid OrgUnitType
+- One of the `groups` id is not a valid Group
+- `forms` is empty
+- `form_id` is not a reference form for the OrgUnitType
```
madewulf: I'm wondering if we really need to force this. In the backend, if the instance is for one of the reference forms, we can just ask if the validator wants to replace the existing one, don't you think?

madewulf: Otherwise, it's just a submission, like any other.

docs/pages/dev/reference/API/org_unit_registry.md
```diff
@@ -0,0 +1,436 @@
...
+- No authentication token or an invalid one was provided
+
+### 403 - Forbidden
+
+- `org_unit_type_id` is not a valid OrgUnitType to be created for the given `parent_id`
+- `org_unit_type_id` is not a valid OrgUnitType to be moved to the given `parent_id`
+
+### 404 - Not found
+
+- `form_id` is not found
```
madewulf: Ideally, the API should indicate which field is creating the 404 (this is something that we're lacking in other endpoints)

docs/pages/dev/reference/API/org_unit_registry.md
```diff
@@ -0,0 +1,436 @@
...
+- `status` was `rejected` but `rejection_comment` was null, omitted or empty
+- `status` of the change to be patched is not `new`
+
+
+
+
+
+# GET /api/mobile/orgunits/changes/?app_id=..
+
+API to list all the change requests for the authorized user. Used by admin in the web UI.
```
madewulf: /mobile used by admin? I'm not sure this phrase makes sense

docs/pages/dev/reference/API/org_unit_registry.md
```diff
@@ -0,0 +1,436 @@
...
+
+- No authentication token or an invalid one was provided
+
+
+
+
+
+# GET /api/mobile/orgunits/{id or UUID}/reference_instances?app_id=...
+
+Returns the reference forms for an OrgUnit from newest to oldest.
```
madewulf: reference submissions or instances (not reference forms, which is confusing)

docs/pages/dev/reference/API/org_unit_registry.md
```diff
@@ -0,0 +1,436 @@
...
+
+## Permission
+
+Same as downloading instances
```
madewulf: This should be limited by the app_id too. Forms are associated with projects. If a given instance is not from a form that is in the project of the current app, it should not be downloadable

madewulf (dismissed):

docs/pages/dev/reference/API/org_unit_registry.md
madewulf:
```suggestion
    "opening_status": "Enum<Status>? - One of `open` or `close`",
```

docs/pages/dev/reference/API/org_unit_registry.md
bmonjoie: I think that for consistency, it's great it's in the query parameter. Also, I have been thinking that I should add an OkHttp interceptor to add it automatically so I don't have to think about it anymore.

docs/pages/dev/reference/API/org_unit_registry.md
kemar: > we already have validation_status

What about this existing status? Shouldn't the new workflow replace it?

docs/pages/dev/reference/API/org_unit_registry.md
bmonjoie: I think we should, if it's not, it should be part of a Change Request.

docs/pages/dev/reference/API/org_unit_registry.md
bmonjoie: Copy-pasta gone wrong

docs/pages/dev/reference/API/org_unit_registry.md
bmonjoie: It already is, `app_id` is required.

docs/pages/dev/reference/API/org_unit_registry.md
madewulf: Good point, since it's possible to have multiple org unit types with the same name

docs/pages/dev/reference/API/org_unit_registry.md
madewulf: No, these are two separate things: I would say that "validation_status" is what's happening BEFORE the org unit opens, and opening status, what is happening once it has been opened.

docs/pages/dev/reference/API/org_unit_registry.md
```diff
@@ -0,0 +1,427 @@
...
+- `instances` is empty
+- `org_unit_type_id` is not a valid sub OrgUnitType for the given parent --> TO BE CONFIRMED
+
+### 401 - Unauthorized
+
+- No authentication token or an invalid one was provided
+
+### 403 - Forbidden
+
+- `org_unit_type_id` is not a valid OrgUnitType to be created for the given `parent_id`
```
mestachs: It's a bit strange to send a 403 - Forbidden for
- org_unit_type_id is not a valid OrgUnitType to be created for the given parent_id
- org_unit_type_id is not a valid OrgUnitType to be moved to the given parent_id
I would have considered it as a bad request, same for the "Not found"

IMO for me I would keep
- forbidden : you are not allowed (you don't have the permission or you are referencing a object out of your permission (something belonging to another account or project))
- bad request : you ask something against the "integrity of the data"
 - not found : 
     - either it doesn't exist
     - either we "hide a forbidden" behind a "not found" (a bit like github not wanting to show you that a private repo actually exist)
         - for the orgunit type id I would expect a  "bad request" ' or  forbidden

docs/pages/dev/reference/API/org_unit_registry.md
```diff
@@ -0,0 +1,423 @@
...
+    "name": "String? - Name of the OrgUnit, may be null or omitted.",
+    "opening_status": "Enum<Status>? - One of `open` or `close`",
+    "org_unit_type_id": "Int? - id of the OrgUnitType, may be null or omitted",
+    "groups": "Array of Group ids? - can be empty, null or omitted. Empty means we want to remove all values",
+    "geo_point": { "": "New geopoint for the OrgUnit, may be null or omitted",
+      "latitude": "Double - New latitude of the OrgUnit",
+      "longitude": "Double - New longitude of the OrgUnit",
+      "altitude": "Double - New altitude of the OrgUnit",
+      "accuracy": "Double - New accuracy of the OrgUnit"
+    },
```
kemar:
```suggestion
    "location": { "": "New geopoint for the OrgUnit, may be null or omitted",
      "latitude": "Double - New latitude of the OrgUnit",
      "longitude": "Double - New longitude of the OrgUnit",
      "altitude": "Double - New altitude of the OrgUnit"
    },
```

This field is actually called [`OrgUnit.location`](https://github.com/BLSQ/iaso/blob/f9568a8fa40a25f7544bf320f09ff3b246a08098/iaso/models/org_unit.py#L285) and implemented as a `PointField` which does not include `accuracy`.

I suggest renaming the field and getting rid of `accuracy`?

docs/pages/dev/reference/API/org_unit_registry.md
```diff
@@ -0,0 +1,423 @@
...
+```json
+{
+  "uuid": "UUID - Client generated UUID",
+  "org_unit_id": "String - id or UUID of the OrgUnit to change, if UUID doesn't exist in DB, it's a new OrgUnit.",
+  "created_at": "Timestamp - format same as /api/instances",
+  "updated_at": "Timestamp - format same as /api/instances",
+  "changes": { "": "Changes is not null and must contain at least one field",
+    "parent_id": "String? - id or UUID of the parent OrgUnit, may be null or omitted.",
+    "name": "String? - Name of the OrgUnit, may be null or omitted.",
+    "opening_status": "Enum<Status>? - One of `open` or `close`",
```
kemar: ~~There is no `opening_status` field on `OrgUnit.location`. Should we remove this field?~~

docs/pages/dev/reference/API/org_unit_registry.md
```diff
@@ -0,0 +1,423 @@
...
+- `parent_id`
+- `name`
+- `opening_status`
+- `org_unit_type_id`
+- `groups`
+- `geo_point`
+- `shape`
+- `instances`
+
+# POST /api/orgunits/changes/?app_id=..
```
kemar: I'm thinking of naming the model `OrgUnitUpsertRequest`. Sounds better suited for both INSERT and UPDATE operations.

@bmonjoie WDYT about `POST /api/orgunits/upsert/?app_id=..`?

docs/pages/dev/reference/API/org_unit_registry.md
bmonjoie: Why are you looking for an `opening_status` on the `location` property? I'm confused

kemar: There is no `opening_status` field on `OrgUnit`. Should we remove this field?

docs/pages/dev/reference/API/org_unit_registry.md
bmonjoie: I don't have any opinion on this. To my understanding, `upsert` is a made-up word by PostgreSQL, but it's clear. @madewulf, what is your take on this?

docs/pages/dev/reference/API/org_unit_registry.md
bmonjoie: I thought it was already there. There is a will to import the status from DHIS2, if I understand correctly, and this would allow us to bring changes to that property. @mestachs might have more information as to where this property currently is.

docs/pages/dev/reference/API/org_unit_registry.md
bmonjoie: `accuracy` is sent when creating instances. Therefore, I think we should leave it to stay consistent even if it is not used.

kemar: Yes, `accuracy` is sent when creating instances, but not when creating org units (unless you prove me wrong). And here we're talking about org units.

bmonjoie: That's why I'm talking about consistency. I meant consistency between APIs

kemar: I can add a "fake" `accuracy` field, but it feels like just covering up weaknesses and it may be misleading…

docs/pages/dev/reference/API/org_unit_registry.md
```diff
@@ -0,0 +1,423 @@
...
+    "opening_status": "Enum<Status>? - One of `open` or `close`",
+    "org_unit_type_id": "Int? - id of the OrgUnitType, may be null or omitted",
+    "groups": "Array of Group ids? - can be empty, null or omitted. Empty means we want to remove all values",
+    "geo_point": { "": "New geopoint for the OrgUnit, may be null or omitted",
+      "latitude": "Double - New latitude of the OrgUnit",
+      "longitude": "Double - New longitude of the OrgUnit",
+      "altitude": "Double - New altitude of the OrgUnit",
+      "accuracy": "Double - New accuracy of the OrgUnit"
+    },
+    "geo-shape": "String? - GeoJSON representation, may be null or omitted",
```
kemar: After another look at whimsical, it looks like the `geo-shape` cannot be updated:

map

The user can signal an outdated map:

> You cannot edit a shape. Instead, you can raise the fact that this shape is outdated.
> Do you confirm the shape is outdated and needs an update?

Therefore we may need a boolean here instead of a GeoJSON representation.

bmonjoie: My understanding is that we cannot edit a shape because we don't have a proper tool to do it correctly. Still, it should be possible, I guess.
But we can add a boolean on top of it.

docs/pages/dev/reference/API/org_unit_registry.md
```diff
@@ -0,0 +1,423 @@
...
+## Query parameters
+
+- app_id: String - Must be provided, project for which this is created.
+
+## Body
+
+```json
+{
+  "uuid": "UUID - Client generated UUID",
+  "org_unit_id": "String - id or UUID of the OrgUnit to change, if UUID doesn't exist in DB, it's a new OrgUnit.",
```
bmonjoie:
```suggestion
  "org_unit_id": "String - id or UUID of the OrgUnit to change",
```

docs/pages/dev/reference/API/org_unit_registry.md
kemar: Clarified in meeting. `ChangeRequest` is good enough.

docs/pages/dev/reference/API/org_unit_registry.md
kemar:
```suggestion
```

We won't implement `opening_status` in V1.

docs/pages/dev/reference/API/org_unit_registry.md
kemar: `accuracy` will stay in `ChangeRequest` as an "helper for validation". A change request could be refused if `accuracy` is too low.

docs/pages/dev/reference/API/org_unit_registry.md
kemar:
```suggestion
```

We won't implement `geo-shape`.

docs/pages/dev/reference/API/org_unit_registry.md
```diff
@@ -0,0 +1,423 @@
...
+    "org_unit_type_id": "Int? - id of the OrgUnitType, may be null or omitted",
+    "groups": "Array of Group ids? - can be empty, null or omitted. Empty means we want to remove all values",
+    "geo_point": { "": "New geopoint for the OrgUnit, may be null or omitted",
+      "latitude": "Double - New latitude of the OrgUnit",
+      "longitude": "Double - New longitude of the OrgUnit",
+      "altitude": "Double - New altitude of the OrgUnit",
+      "accuracy": "Double - New accuracy of the OrgUnit"
+    },
+    "geo-shape": "String? - GeoJSON representation, may be null or omitted",
+    "instances":  "Array of instance ids? - may be null or omitted, cannot be empty"
```
kemar:
```suggestion
  "uuid": "UUID - Client generated UUID",
  "org_unit_id": "String - id or UUID of the OrgUnit to change",
  "created_at": "Timestamp - format same as /api/instances",
  "updated_at": "Timestamp - format same as /api/instances",
  "new_parent_id": "String? - id or UUID of the parent OrgUnit, may be null or omitted.",
  "new_name": "String? - Name of the OrgUnit, may be null or omitted.",
  "new_org_unit_type_id": "Int? - id of the OrgUnitType, may be null or omitted",
  "new_groups": "Array of Group ids? - can be empty, null or omitted. Empty means we want to remove all values",
  "new_geo_point": { "": "New geopoint for the OrgUnit, may be null or omitted",
    "latitude": "Double - New latitude of the OrgUnit",
    "longitude": "Double - New longitude of the OrgUnit",
    "altitude": "Double - New altitude of the OrgUnit",
    "accuracy": "Double - New accuracy of the OrgUnit",
  },
  "new_reference_instances":  "Array of instance ids? - may be null or omitted, cannot be empty"
```

As discussed this morning, I suggest using a flat structure in the returned payload.

It makes the code clearer and more succinct on the back-end side (more meaningful variable names and no need of nested Serializers).

> [Flat is better than nested.](https://peps.python.org/pep-0020/) 🤣

docs/pages/dev/reference/API/org_unit_registry.md
```diff
@@ -0,0 +1,423 @@
...
+}
+```
+
+## Possible responses
+
+### 201 - Created
+
+### 400 - Bad request
+
+- `app_id` was not provided
```
bmonjoie: Since this endpoint is also used by the backend, this parameter is optional
```suggestion
```

docs/pages/dev/reference/API/org_unit_registry.md
```diff
@@ -0,0 +1,423 @@
...
+      "org_unit_type_name": "String - name of the current OrgUnitType",
+      "groups": "Array<String> - names of the current Groups",
+      "created_by": "String - username of the user who created that request",
+      "created_at": "Timestamp",
+      "update_by": "String? - username of the user who accepted/rejected that request",
+      "updated_at": "Timestamp?",
+      "requested_fields": "Array<String> - name of the properties that were requested to change",
+      "approved_fields": "Array<String>? - name of the properties that were approved to change",
+      "rejection_comment": "String? - Comment about why the changes were rejected",
+      "instances": "Array of instance ids? - may be null or omitted, cannot be empty"
```
kemar:
```suggestion
```

As discussed, this may be a copy and paste error.

docs/pages/dev/reference/API/org_unit_registry.md
```diff
@@ -0,0 +1,423 @@
...
+        "values": [
+          {
+            "key": "String",
+            "label": "String",
+            "value": "String"
+          }
+        ]
+      }
+    ]
+  }
```
kemar:
```suggestion
  "new_parent": "String? - Name of the new parent OrgUnit, may be null or omitted.",
  "new_name": "String? - New name of the OrgUnit, may be null or omitted.",
  "new_org_unit_type_id": "Int? - id of the new OrgUnitType, may be null or omitted",
  "new_org_unit_type_name": "String? - Name of the new OrgUnitType, may be null or omitted",
  "new_groups": "Array of String? - can be empty, null or omitted. Names of the new groups",
  "new_geo_point": { "": "New GeoPoint? for the OrgUnit, may be null or omitted",
    "latitude": "Double - New latitude of the OrgUnit",
    "longitude": "Double - New longitude of the OrgUnit",
    "altitude": "Double - New altitude of the OrgUnit",
    "accuracy": "Double - New accuracy of the OrgUnit"
  },
  "new_reference_instances": [ "Array of form objects? - may be null or omitted, cannot be empty",
    {
      "form_id": "id of the form",
      "form_name": "Name of the form",
      "instance_id": "id of the instance",
      "values": [
        {
          "key": "String",
          "label": "String",
          "value": "String"
        }
      ]
    }
  ]
```

`changes` has been flattened, plus:

- Naming:
    - `instances` => `new_reference_instances`.
- Deletion:
    - `"opening_status"`
    - `"geo-shape"`

docs/pages/dev/reference/API/org_unit_registry.md
```diff
@@ -0,0 +1,423 @@
...
+some changes still. Therefore, the IDs can be either the DB id or the UUID generated by the client.
+Changes that are not provided are considered not being part of the change request.
+
+## Permissions
+
+User must be authenticated.
+
+## Query parameters
+
+- app_id: String - Must be provided, project for which this is created.
```
bmonjoie:
```suggestion
- app_id: String - Optional, project for which this is created.
```

docs/pages/dev/reference/API/org_unit_registry.md
```diff
@@ -0,0 +1,423 @@
...
+- A not nullable field was null or omitted
+- `org_unit_id` is a new OrgUnit and the following fields were null or omitted: `parent_id`, `name`, `geo_point` and `org_unit_type_id`
+- A `String` field has an empty value
+- `parent_id` is not a valid OrgUnit
+- `opening_status` was not one of the enum values.
+- `org_unit_type_id` is not a valid OrgUnitType
+- One of the `groups` id is not a valid Group
+- `instances` is empty
+- `org_unit_type_id` is not a valid sub OrgUnitType for the given `parent_id` --> TO BE CONFIRMED
+- `org_unit_type_id` is not a valid sub OrgUnitType to be moved to the given `parent_id` --> TO BE CONFIRMED
```
bmonjoie:
```suggestion
- `new_parent_id` is not a valid OrgUnit
- `new_org_unit_type_id` is not a valid OrgUnitType
- One of the `new_groups` id is not a valid Group
- `new_reference_instances` is empty
- `new_org_unit_type_id` is not a valid sub OrgUnitType for the given `parent_id` --> TO BE CONFIRMED
- `new_org_unit_type_id` is not a valid sub OrgUnitType to be moved to the given `parent_id` --> TO BE CONFIRMED
```

docs/pages/dev/reference/API/org_unit_registry.md
```diff
@@ -0,0 +1,423 @@
...
+
+### 401 - Unauthorized
+
+- No authentication token or an invalid one was provided
+
+### 404 - Not found
+
+- one or more of `instances` ids is not found
+- `parent_id` is not found
+- `org_unit_type_id` is not found
```
bmonjoie:
```suggestion
- one or more of `new_reference_instances` ids is not found
- `new_parent_id` is not found
- `new_org_unit_type_id` is not found
```

docs/pages/dev/reference/API/org_unit_registry.md
```diff
@@ -0,0 +1,423 @@
...
+    "org_unit_type_name": "String - Name of the OrgUnitType",
+    "groups": "Array of String - can be empty. Names of the new groups",
+    "geo_point": { "": "Geopoint for the OrgUnit",
+      "latitude": "Double - New latitude of the OrgUnit",
+      "longitude": "Double - New longitude of the OrgUnit",
+      "altitude": "Double - New altitude of the OrgUnit",
+      "accuracy": "Double - New accuracy of the OrgUnit"
+    },
+    "geo-shape": "String? - GeoJSON representation, may be null or omitted",
+    "instances": [ "Array of form objects - can be empty",
```
bmonjoie:
```suggestion
    "reference_instances": [ "Array of form objects - can be empty",
```

docs/pages/dev/reference/API/org_unit_registry.md
```diff
@@ -0,0 +1,423 @@
...
+  "created_at": "Timestamp",
+  "update_by": "String? - username of the user who accepted/rejected that request",
+  "updated_at": "Timestamp?",
+  "requested_fields": "Array<String> - name of the properties that were requested to change", 
+  "approved_fields": "Array<String>? - name of the properties that were approved to change",
+  "rejection_comment": "String? - Comment about why the changes were rejected",
+  "org_unit": {
+    "parent": "String - Name of the parent OrgUnit.",
+    "name": "String - Name of the OrgUnit.",
+    "opening_status": "Enum<Status>? - Status of the OrgUnit, One of `open` or `close`, may be null or omitted.",
```
bmonjoie:
```suggestion
```

docs/pages/dev/reference/API/org_unit_registry.md
```diff
@@ -0,0 +1,423 @@
...
+    "org_unit_type_id": "Int - id of the OrgUnitType",
+    "org_unit_type_name": "String - Name of the OrgUnitType",
+    "groups": "Array of String - can be empty. Names of the new groups",
+    "geo_point": { "": "Geopoint for the OrgUnit",
+      "latitude": "Double - New latitude of the OrgUnit",
+      "longitude": "Double - New longitude of the OrgUnit",
+      "altitude": "Double - New altitude of the OrgUnit",
+      "accuracy": "Double - New accuracy of the OrgUnit"
+    },
+    "geo-shape": "String? - GeoJSON representation, may be null or omitted",
```
bmonjoie:
```suggestion
```

docs/pages/dev/reference/API/org_unit_registry.md
```diff
@@ -0,0 +1,423 @@
...
+    "name": "String - Name of the OrgUnit.",
+    "opening_status": "Enum<Status>? - Status of the OrgUnit, One of `open` or `close`, may be null or omitted.",
+    "org_unit_type_id": "Int - id of the OrgUnitType",
+    "org_unit_type_name": "String - Name of the OrgUnitType",
+    "groups": "Array of String - can be empty. Names of the new groups",
+    "geo_point": { "": "Geopoint for the OrgUnit",
+      "latitude": "Double - New latitude of the OrgUnit",
+      "longitude": "Double - New longitude of the OrgUnit",
+      "altitude": "Double - New altitude of the OrgUnit",
+      "accuracy": "Double - New accuracy of the OrgUnit"
```
bmonjoie:
```suggestion
```

docs/pages/dev/reference/API/org_unit_registry.md
```diff
@@ -0,0 +1,423 @@
...
+        "groups": "Array of Group ids? - can be empty, null or omitted. Empty means we want to remove all values",
+        "geo_point": { "": "New geopoint for the OrgUnit, may be null or omitted",
+          "latitude": "Double - New latitude of the OrgUnit",
+          "longitude": "Double - New longitude of the OrgUnit",
+          "altitude": "Double - New altitude of the OrgUnit",
+          "accuracy": "Double - New accuracy of the OrgUnit"
+        },
+        "geo-shape": "String? - GeoJSON representation, may be null or omitted",
+        "instances": "Array of instance ids? - may be null or omitted, cannot be empty"
+      }
```
bmonjoie:
```suggestion
      "new_parent_id": "String? - id or UUID of the parent OrgUnit, may be null or omitted.",
      "new_name": "String? - Name of the OrgUnit, may be null or omitted.",
      "new_org_unit_type_id": "Int? - id of the OrgUnitType, may be null or omitted",
      "new_groups": "Array of Group ids? - can be empty, null or omitted. Empty means we want to remove all values",
      "new_geo_point": { "": "New geopoint for the OrgUnit, may be null or omitted",
        "latitude": "Double - New latitude of the OrgUnit",
        "longitude": "Double - New longitude of the OrgUnit",
        "altitude": "Double - New altitude of the OrgUnit",
        "accuracy": "Double - New accuracy of the OrgUnit"
      },
      "new_reference_instances": "Array of instance ids? - may be null or omitted, cannot be empty"
```

docs/pages/dev/reference/API/org_unit_registry.md
```diff
@@ -0,0 +1,406 @@
...
+                "uuid": "UUID - provided by the client",
+                "form_id": "Int",
+                "form_version_id": "Int",
+                "created_at": "Timestamp in double",
+                "updated_at": "Timestamp in double",
+                "json":  "JSONObject - contains the key/value of the instance"
+            }
+        ]
+    }
+  ]
```
bmonjoie:
```suggestion
  "results": [
    {
        "id": "Int",
        "uuid": "UUID - provided by the client",
        "form_id": "Int",
        "form_version_id": "Int",
        "created_at": "Timestamp in double",
        "updated_at": "Timestamp in double",
        "json":  "JSONObject - contains the key/value of the instance"
    }
  ]
```

kemar (approved):

docs/pages/dev/reference/API/org_unit_registry.md
```diff
@@ -0,0 +1,437 @@
...
+      "new_groups": "Array of Group ids? - can be empty, null or omitted. Empty means we want to remove all values",
+      "new_location": { "": "New geopoint for the OrgUnit, may be null or omitted",
+        "latitude": "Double - New latitude of the OrgUnit",
+        "longitude": "Double - New longitude of the OrgUnit",
+        "altitude": "Double - New altitude of the OrgUnit"
+      },
+      "new_location_accuracy": "Double - New accuracy of the OrgUnit",
+      "new_opening_date": "Timestamp in double",
+      "new_closed_date": "Timestamp in double",
+      "new_reference_instances": "Array of instance ids? - may be null or omitted, cannot be empty"
```
bmonjoie, line 210: @kemar As discussed, this should return the UUIDs and not the ID Ints.

PR #797, opened 2023-09-13
POLIO-1197 make expiration_date required

SimonDus:
As expiration_date is required for calculation purposes, we should not be allowed to create a vacc_auth without it. 

Related JIRA tickets : POLIO-1197

- [ ] Are there enough tests

## Changes

Front end and backend changes to make expiration_date required

## How to test

Go to vaccine authorization tab in polio and create one. Expiration Date must be required.

## Print screen / video

https://github.com/BLSQ/iaso/assets/45785235/fa28660a-0165-4083-858f-d2f4d682d0dc


kemar (approved):
LGTM

plugins/polio/migrations/0145_alter_vaccineauthorization_expiration_date.py
```diff
@@ -0,0 +1,26 @@
...
+
+from django.db import migrations, models
+import datetime
+
+
+# handle old records with a None date.
+def populate_expiration_date(apps, schema_editor):
+    VaccineAuthorization = apps.get_model("polio", "VaccineAuthorization")
+
+    VaccineAuthorization.objects.filter(expiration_date=None).update(expiration_date=datetime.datetime.now())
```
mathvdh, line 11: Est-ce qu'on veut vraiment mettre "now" comme date d'expiration à tout ce qui existe déjà ? Je sais pas si c'est des vrais datas dedans ou quoi mais je me demande ?

plugins/polio/api.py
```diff
@@ -2175,6 +2175,10 @@ def create(self, validated_data):
         user = self.context["request"].user
         validated_data["account"] = user.iaso_profile.account
 
+        expiration_date = validated_data.get("expiration_date")
+        if expiration_date is None:
```
mathvdh: On checkerait pas aussi que c'est pas une string vide ou qqch ? Je sais pas trop ce que balance le frontend .. ?

plugins/polio/migrations/0145_alter_vaccineauthorization_expiration_date.py
quang-le, line 11: I'm not sure about this. it will lead to incorrect data. Doesn't really matter if there's no data in prod though

plugins/polio/tests/test_vaccine_authorizations.py
```diff
@@ -113,6 +113,22 @@ def test_user_can_post(self):
...
+            data={
+                "country": self.org_unit_DRC.pk,
+                "quantity": 12346,
+                "status": "ONGOING",
+                "comment": "waiting for approval.",
+            },
+        )
+
+        self.assertEqual(response.status_code, 400)
+
```
mathvdh, line 132: Ce serait pas mal de rajouter un check un peu plus complet genre :

```python
self.assertEqual(response.data["detail"], "Expiration date is required.")
```

mathvdh (commented):
De manière générale il ne serait pas possible de partir de la branche `POLIO-1188-remove-the-vaccine-management-tab-in-campaigns` pour faire ça ?
Si ce n'est pas urgent, cela pourrait être mergé après que POLIO-1188 soit mergée
(C'est pas obligatoire mais juste une demande ..)
Sinon je ferai le merge manuel dans POLIO-1188 après que t'aies mergé ceci

plugins/polio/api.py
```diff
@@ -2175,6 +2175,10 @@ def create(self, validated_data):
         user = self.context["request"].user
         validated_data["account"] = user.iaso_profile.account
 
+        expiration_date = validated_data.get("expiration_date")
```
madewulf: This will work, but it seems to me that we should use the serializers to do this kind of validation, like here: https://stackoverflow.com/questions/30953481/django-rest-framework-how-to-make-model-serializer-fields-required 

(not a blocking remark)

mathvdh:
Egalement je comprends pas trop que le `expiration_date = TimestampField(read_only=True)` soit `read_only` sur le serializer et qu'en même temps il soit required et qu'on l'update ... ?

plugins/polio/api.py
madewulf: Actually, I'm wondering if changing at the database level does not already imply that the ModelSerializer will send back that the field is required if provided empty. That's what would make sense to me

mathvdh: @madewulf I was looking into it and it seems something like this on the serializer would work

```pythin
extra_kwargs = {"expiration_date": {"required": True, "allow_null": False, "allow_blank": False}}
```

Not sure because it seems to depend on the version of DRF

mathvdh: @madewulf yes I was thinking that too, I guess it needs to be tested ? Not sure how it goes while defining the field as a `read_only` TimestampField on the serializer as it is at the moment ....

plugins/polio/migrations/0145_alter_vaccineauthorization_expiration_date.py
SimonDus: Il n'y a pas d'entrée sur la prod c'est uniquement pour les environnement de tests.  Ca peut être enlevé plus tard tbh.

SimonDus: It will only run at migration if there is no date which is impossible now as we dont allow it. Without this, I don't see how to rung the migrations on tests environments. It can be removed later tho

plugins/polio/tests/test_vaccine_authorizations.py
SimonDus: Done

plugins/polio/api.py
SimonDus: Indeed, the check from DB is enough. I removed it

mathvdh (approved):
👍🏼

SimonDus:
@mathvdh Je voudrais merger ça directement pour corriger ce bug avant que ça parte en prod.

mathvdh:
Go ahead I will merge it manually it’s not much

plugins/polio/api.py
SimonDus: le front envoit null si y'a rien

PR #798, opened 2023-09-13
IA-2296 completeness v5

beygorghor:
- when there is no submission expected, the shape should appear grey, not red
- filter planning with selected forms
- have an error message when there is no shape to explain why nothing shows

Related JIRA tickets : IA-2388, IA-2354, IA-2389

- [x] Are there enough tests

## Changes

- filter plannings with planning using selected forms
- compare map an list results to display a warning that one or more org unit doesn't have a geolocation
- get assignments for selected planning and check that the org units are used in the planning to change the color 

## How to test

Make sure you have a planning with org unit assigned to a user or a team
Go to completeness stats page, select the form used in the planning.
You should have only planning using this form in plannings dropdown .
Select the initial planning and drill down to the assignment level.
Org unit without assignment should be grey.
Unselect the planning, all org units should be red, green or orange.

Add a children org unit without geolocation to an existing org unit with a shape or a marker.
Visit completeness on the parent org unit. 
You should see a snackbar saying that an org unit is missing geolocation.

## Print screen / video

https://github.com/BLSQ/iaso/assets/12494624/a97755c9-9fb9-4044-a88e-50c23627d1e5
https://github.com/BLSQ/iaso/assets/12494624/e9e87cac-7e02-40c9-b4c3-eea5e97f82ce

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/app/translations/fr.json
+++ b/hat/assets/js/apps/Iaso/domains/app/translations/fr.json
@@ -24,6 +24,7 @@
     "iaso.assignment.baseOrgUnitsType": "Types d'unités d'org. de base",
     "iaso.assignment.clickRowToUnAssign": "Sélectionnez la ligne pour désassigner l'unité d'org.",
     "iaso.assignment.count": "Nombre d'assignations",
+    "iaso.completenessStats.completenessMapWarning": "Unité d'org. trouvée(s) sans donnée géographique",
     "iaso.assignment.inAnotherTeam": "dans une autre équipe",
     "iaso.assignment.label": "Affectation",
     "iaso.assignment.map.disabled": "Déjà assigné",
--- a/hat/assets/js/apps/Iaso/domains/app/translations/fr.json
+++ b/hat/assets/js/apps/Iaso/domains/app/translations/fr.json
@@ -24,7 +24,7 @@
     "iaso.assignment.baseOrgUnitsType": "Types d'unités d'org. de base",
     "iaso.assignment.clickRowToUnAssign": "Sélectionnez la ligne pour désassigner l'unité d'org.",
     "iaso.assignment.count": "Nombre d'assignations",
-    "iaso.completenessStats.completenessMapWarning": "Unité d'org. trouvée(s) sans donnée géographique",
+    "iaso.completenessStats.completenessMapWarning": "Unité(s) d'org. trouvée(s) sans données géographiques",
     "iaso.assignment.inAnotherTeam": "dans une autre équipe",
     "iaso.assignment.label": "Affectation",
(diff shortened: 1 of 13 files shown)


kemar:
Almost no tests, is it on purpose?

hat/assets/js/apps/Iaso/domains/app/translations/fr.json
```diff
@@ -24,6 +24,7 @@
     "iaso.assignment.baseOrgUnitsType": "Types d'unités d'org. de base",
     "iaso.assignment.clickRowToUnAssign": "Sélectionnez la ligne pour désassigner l'unité d'org.",
     "iaso.assignment.count": "Nombre d'assignations",
+    "iaso.completenessStats.completenessMapWarning": "Unité d'org. trouvée(s) sans donnée géographique",
```
quang-le: Unité(s) d'org. trouvée(s) sans données géographiques

hat/assets/js/apps/Iaso/domains/completenessStats/CompletenessStatsFilters.tsx
```diff
@@ -135,19 +135,28 @@ export const CompletenessStatsFilters: FunctionComponent<Props> = ({
...
+    const handleChangeForm = useCallback(
+        (keyValue, value) => {
+            if (keyValue === 'formId') {
+                handleChange('planningId', undefined);
+            }
+            handleChange(keyValue, value);
+        },
+        [handleChange],
+    );
+
```
quang-le: Since you only use it with the `formId` component, you can probabmy do without the if statement. Just nitpicking

quang-le (changes requested):
I'll just block on the french "Orthographe" in the translations 😉

Other wise works fine for me.

Code looks good to me, but a pythonista's opinion on the backend would be good

quang-le (approved):

iaso/api/completeness_stats.py
```diff
@@ -226,16 +227,21 @@ def list(self, request: Request, *args, **kwargs) -> Response:
...
-        planning = params.get("planning", None)
+        planning_id = params.get("planning_id", None)
         org_unit_validation_status = params["org_unit_validation_status"]
+        as_location = params.get("as_location", None)
 
         instance_qs = Instance.objects.all()
+
+        planning = None
+        if planning_id:
+            planning = get_object_or_404(Planning, id=planning_id)
```
SimonDus, line 238: As I don't have all the context I'm not sure if we really want to return a 404 at this line. Maybe a queryset would be better ? At line 243 and 302 there is a check for planning so I guess it can still run without a planning. An empty queryset is evaluated as None so it would not break anything I guess

PR #801, opened 2023-09-13
Polio-1191 add endpoint for openhexa task

quang-le:
To enable users to refresh lqas data in the UI, we need an API endpoint that can launch the OpenHexa pipeline for a given country

Related JIRA tickets :  POLIO-1191

- [] Are there enough tests - safeguards to prevent non external tasks to be patched are not tested
- [ ] Documentation has been included (for new feature) - Coming soon (probably in another PR)

## Changes

- Add `external` boolean field to `Task`model. When `True` the `Task` can be PATCHed through the new endpoint.
- Add `/api/polio/tasks/refreshlqas/` endpoint:
    - POST: will create an `external` `Task` and launch the OpenHexa lqas pipeline
    - PATCH: meant to be used by the OpenHexa pipeline to update the `Task` (status, message, and progress related fields) 
 - Add endpoint `/api/polio/tasks/refreshlqas/last_run_for_country/` to get last task run for a given country. Not part of regular `task` API so we don't have to give polio users the `iaso_tasks` permission for the page to work.
 - Since the external `Task` does not run in the worker, changing its `should_be_killed` field to `True` will change  the status to `KILLED`immediately.
 - Added environment variables to store API token and to enable local testing
 - Added dependency to `gql[requests]` and fixed dependencies conflicts

## How to test

- go to Django api and try POST and PATCH
- Also try GET on the `last_run_for_country` action

## Print screen / video

https://github.com/BLSQ/iaso/assets/38907762/c1dfe512-0166-4981-b1cc-724f0afc4296

## Notes

To test the pipeline integration, the corresponding PR in iaso-pipelines is required:https://github.com/BLSQ/iaso-pipelines/pull/14


plugins/polio/tasks/api/refresh_lqas_data.py
```diff
@@ -0,0 +1,218 @@
...
+    ordering_fields = ["created_at", "ended_at", "name", "started_at", "status"]
+    ordering = ["-started_at"]
+
+    def get_serializer_class(self):
+        if self.request.method == "POST":
+            return RefreshLQASDataSerializer
+        return ExternalTaskSerializer
+
+    def get_queryset(self):
+        return Task.objects.all()
```
mathvdh: Shouldn't we be filtering by `account` ? and maybe `launcher` ?

plugins/polio/tasks/api/refresh_lqas_data.py
```diff
@@ -0,0 +1,218 @@
...
+        transport = RequestsHTTPTransport(
+            url=settings.OPENHEXA_URL,
+            verify=True,
+            headers={"Authorization": f"Bearer {settings.OPENHEXA_TOKEN}"},
+        )
+        client = Client(transport=transport, fetch_schema_from_transport=True)
+        get_runs = gql(
+            """
+        query pipeline {
+            pipeline(id: "e5b9a3bb-a22c-461f-b000-4dfa8d1cc445"){
```
mathvdh: Does the pipeline `uuid` needs to be hardcoded there ?

mathvdh (changes requested):
Except for my 2 comments, not much to say ! Good job !!
I don't know much about the details of how openhexa works so I can't really comment on that part ...

plugins/polio/tasks/api/refresh_lqas_data.py
mathvdh: At least put it in a constant somewhere or better in a settings var (or the env)

quang-le: Good catch, it's already in `settings.LQAS_PIPELINE`, I missed this occurence it when I refactored

plugins/polio/tasks/api/refresh_lqas_data.py
```diff
@@ -0,0 +1,227 @@
...
+        if country_id is not None:
+            user = request.user
+            country_id = int(country_id)
+            user_has_access = OrgUnit.objects.filter_for_user(user).filter(id=country_id).count() > 0
+            if not user_has_access:
+                raise serializers.ValidationError(NO_AUTHORIZED_COUNTRY_ERROR)
+        return validated_data
+
+
+class ExternalTaskSerializer(TaskSerializer):
```
madewulf, line 42: This should be in the Iaso codebase, not in the polio plugin

hat/settings.py
```diff
@@ -519,6 +519,16 @@ def sentry_tracer_sampler(sampling_context):
 THEME_SECONDARY_COLOR = os.environ.get("THEME_SECONDARY_COLOR", "#0066CC")
 THEME_PRIMARY_BACKGROUND_COLOR = os.environ.get("THEME_PRIMARY_BACKGROUND_COLOR", "#F5F5F5")
 SHOW_NAME_WITH_LOGO = os.environ.get("SHOW_NAME_WITH_LOGO", "yes")
+# OpenHexa API url
```
madewulf, line 522: I think that this should live in the Config model, not in the server settings. Changing the OpenHexa token should not require to redeploy the server, I think.

madewulf (approved):

PR #803, opened 2023-09-14
IA-2345: Permissions grouping for IASO admin

hakifran:
- Group permissions into modules
- Limit the access of permissions bye  modules

Related JIRA tickets : [IA-2345](Permissions grouping for IASO admin)

- [x] Are there enough tests

## Changes
- Added models (Module and ModulePermission(extends Auth/Permission model))
- Link Modules to ModulePermission throught ModulePermission
- Create Modules https://docs.google.com/spreadsheets/d/1zzCD19HJLm3xkDfS3HrZonMvbBPsY9YEXhywDCpoFzU/edit#gid=7786233 throught migrations
- Link Account with modules
- Add modules JsonField on setup_account
- Edit filter_permissions to take in account modules
- Display on the front end modules(those activated for the account and the other not activated

## How to test
- Run migrations(to create the modules and their related permissions)
- Go in admin/iaso/account/
- Link an account to modules
- Go to fronte and try to edit permissions linked to a user or userrole
- Go again the admin menu
- Go to modules menu 
- Check which modules are active and not for the account

## Print screen / video

Screenshot from 2023-09-14 15-41-39

Screenshot from 2023-09-14 15-43-01

## Notes

[IA-2345]: https://bluesquare.atlassian.net/browse/IA-2345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hat/assets/js/apps/Iaso/domains/modules/hooks/requests/useGetModules.ts
```diff
@@ -0,0 +1,40 @@
...
+};
+
+export const useGetModules = (
+    options: ModuleParams | ModulesFilterParams,
+): UseQueryResult<ModulesList, Error> => {
+    const { select } = options as Record<string, any>;
+    return useSnackQuery({
+        queryKey: ['modulesList', options],
+        queryFn: () => getModules(options),
+        snackErrorMsg: undefined,
```
beygorghor: no need to specify this

hat/menupermissions/apps.py
```diff
@@ -1,5 +1,58 @@
...
+                else:
+                    filtered_module = list(
+                        filter(lambda module: module["codename"] == module_codename, MODULES)
+                    )  # get the module to be created
+                    return iaso_module.objects.create(
+                        name=filtered_module[0]["name"], codename=filtered_module[0]["codename"]
+                    )
+
+
+def my_callback(sender, **kwargs):
```
quang-le: maybe change that name 😉

beygorghor (changes requested):
Nice job, very clean !!
I have two remarks:
- can we provide some Cypress tests for this page ?
- also we should add this module dropdown on the react setup account side.
- To see it you should use a user (staff or super user) without iaso profile and try to connect to Iaso with this user.
Screenshot 2023-09-19 at 09 59 36

quang-le (commented):
Didn't review in details, but I want to congratulate you on the quality of your front-end code. It's very clean and legible, and it totally follows the guidelines!

Congrats!

hakifran:
@beygorghor j'ai un soucis dans 
`testTablerender({
            baseUrl,
            rows: listFixture.results.length,
            columns: 4,
            apiKey: 'modules',
        });`

iaso/api/modules.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def get_queryset(self) -> QuerySet[Module]:
+        queryset = Module.objects.all()  # type: ignore
+        search = self.request.GET.get("search", None)
+        orders = self.request.GET.get("order", "name").split(",")
+
+        if search:
+            queryset = queryset.filter(Q(name__icontains=search)).distinct()
+        if orders:
+            queryset = queryset.order_by(*orders)
```
madewulf: This works, but I'm under the impression that there are ways to specify sorting and searching through the DRF (this is a general comment, this does not mean that this needs to change)

madewulf:
I'm a bit confused with this quick test: 

https://github.com/BLSQ/iaso/assets/185797/f556dd13-8ac8-4a0e-9814-b269dd7036f6

With this scenario, I would expect to not have polio or storage in the menu. What did I miss? 

Notice that the video shows to other small problems: changing language after submitting the form reinitialize the form. I think we can forget that, but it surprised me. 

Also, there is a small typo in the json for the modules (visible at 1:30).

The problem is this: permissions_to_add = CustomPermissionSupport.get_full_permission_list() should be filtered depending on the chosen modules, I would say

iaso/api/setup_account.py
```diff
@@ -36,6 +37,11 @@ def validate_user_username(self, value):
             raise serializers.ValidationError("user_name_already_exist")
```
madewulf, line 37: Just spotting that it should be "account_name_already_exists"

madewulf:
One important security point: permissions set to a user should be limited to the ones in the module. This probably happens in the profile API and needs to be fixed. Warning that permissions from the batch excel import should be limited too.

madewulf:
One important security point: permissions set to a user should be limited to the ones in the module. This probably happens in the profile API and needs to be fixed. Warning that permissions from the batch excel import should be limited too.

beygorghor (approved):

iaso/api/profiles.py
```diff
@@ -407,13 +407,23 @@ def update_user_profile(self, request, pk):
...
-            user.user_permissions.add(get_object_or_404(Permission, codename=permission_codename))
+            if permission_codename in self.module_permissions(current_account):
+                CustomPermissionSupport.assert_right_to_assign(request.user, permission_codename)
+                user.user_permissions.add(get_object_or_404(Permission, codename=permission_codename))
         user.save()
 
+    @staticmethod
+    def module_permissions(current_account):
+        # Get all modules linked to the current account
+        modules = list(Module.objects.filter(account_modules=current_account).values_list("id", flat=True))
```
madewulf: Wondering why you needed a list() here, this will most probably add one sql request while using directly the queryset in the following line would have probably given a nested select (so, just  one request). 

These are still probably very small performance impacts, so this does not really need to change.

iaso/api/profiles.py
```diff
@@ -407,13 +407,23 @@ def update_user_profile(self, request, pk):
...
     @staticmethod
-    def update_permissions(user, request):
+    def update_permissions(self, user, request):
         user.user_permissions.clear()
+        current_account = user.iaso_profile.account
+
         for permission_codename in request.data.get("user_permissions", []):
-            CustomPermissionSupport.assert_right_to_assign(request.user, permission_codename)
-            user.user_permissions.add(get_object_or_404(Permission, codename=permission_codename))
+            if permission_codename in self.module_permissions(current_account):
```
madewulf: Oh, this must have an impact on performance, you are calling module_permissions in a loop, and I fear that it means that you are going to do the same requests for eache persmission in the request data.

hat/menupermissions/apps.py
```diff
@@ -1,5 +1,58 @@
...
+def create_iaso_permissions_callback(sender, **kwargs):
+    from iaso.models import Permission as IasoPermission, Module
+    from django.contrib.auth.models import Permission as AuthPermission
+
+    if "plan" in kwargs:
+        plan = kwargs["plan"]
+        if plan:  # Check if there is a new migration
+            (plan_migration, plan_bool) = plan[0]
+            if not plan_bool:  # Check if it's not a rollback migration
+                operation = plan_migration.operations[0]
```
madewulf: I had this fail on a migration that was just a merge, which means, I think that there are no operations in it. It made me realize that this is probably called on ALL migrations. Is that what was intended?

hat/menupermissions/apps.py
```diff
@@ -1,5 +1,58 @@
...
+                                    )  # create the iaso permission
+                            module = get_module(permission[0], IasoPermission, Module)  # get the module
+                            if module:  # link the module to iaso permission if the module already exist
+                                iaso_permission.module = module
+                                iaso_permission.save()
 
 
 class MenupermissionsConfig(AppConfig):
     name = "hat.menupermissions"
+
```
madewulf: I'm under the impression that this will be run after ALL migrations, which is a lot

madewulf:
I think this is missing one last migration that would give all modules to all accounts as a start, so that nobody gets in the situation of having an empty permission list after upgrading their version of Iaso to include this permission module. 

And outside of my few comments, this is great work!

hat/menupermissions/apps.py
hakifran: Yes my intention was to run this code every time a new hat.menupermissions is added.

hakifran:
> I think this is missing one last migration that would give all modules to all accounts as a start, so that nobody gets in the situation of having an empty permission list after upgrading their version of Iaso to include this permission module.
> 
> And outside of my few comments, this is great work!

Already added

PR #807, opened 2023-09-15
IA-2411 Allow multiple reference `Form` and reference `Instance`

kemar:
Enable the possibility to have:

- multiple reference `Form`s per `OrgUnitType`
- multiple reference `Instance`s per `OrgUnit`

Related JIRA tickets :

- [IA-2411](https://bluesquare.atlassian.net/browse/IA-2411)
- [IA-2229](https://bluesquare.atlassian.net/browse/IA-2229) (Epic)

- [x] Are there enough tests

## Changes

- back-end
    - models
        - `OrgUnitType`
            - replace `reference_form` (FK) by `reference_forms` (M2M)
            - migrate DB data from `reference_form` to `reference_forms`
        - `OrgUnit`
            - replace `reference_instance` (FK) by `reference_instances` (M2M)
            - migrate DB data from `reference_form` to `reference_forms`
         - `OrgUnitReferenceInstance`
             - used as through table
         - `Instance`
             - add the logic to flag and unflag an instance as reference
    - APIs
        - mobile 
            - adapt `FormSerializer` ([per request](https://bluesquare.atlassian.net/browse/IA-2411?focusedCommentId=59205))
                - new `reference_form_of_org_unit_types` field
            - adapt `OrgUnitTypeSerializerV1` (nothing exposed as requested by the mobile app development team)
               - remove all `reference*` fields (unused by the mobile)
            - adapt `MobileOrgUnitSerializer`
                - new `reference_instances` field
        - web
            - adapt `InstancesViewSet`
                - `list()`and `retrieve()`
            - adapt `OrgUnitTypeSerializerV2`
                - add `reference_forms` and `reference_forms_ids` fields
                - new `validate_reference_forms()` method
            - adapt `OrgUnitViewSet`
                - this is what the front-end uses to flag an instance as reference
                - new logic in `partial_update()` based on an explicit code to trigger a `flag` or `unflag` action
    - adapt tests
    - adapt fixtures
- front-end
    - allow multiple reference forms per `OrgUnitType`
    - allow multiple reference instances per `OrgUnit`
    - new logic to flag or unflag a `reference instance` (`instance` is called `submission` in the front-end)
    - new component to display and switch between multiple `reference instances`s
    - deleted [`OrgUnitActions.tsx`](https://github.com/BLSQ/iaso/blob/b66a3f0e0ffdc0f9bcada7e0b50f9c74f81c4fa4/hat/assets/js/apps/Iaso/domains/orgUnits/components/OrgUnitActions.tsx) and the `SpeedDialsAction`from the `Org unit` details page
        - replaced by a link to the `Instance` details page
    - adapt fixtures
    - adapt tests

## How to test

Front-end (watch the video below for help):

1. Go to `Org Units > Organisation unit type`
    - edit a `Org unit type`
    - set some `Reference form`s
1. Go to `Forms > Submissions`
    - filter by `Forms` (choose those flagged above)
    - `link` some `Forms` to one `Org unit` to create `Reference submission`s
-  Go to `Org unit` details
    - you can see one or more `Reference submission`s and switch between them
    - each `Reference submission` has a link to its details

## Video

https://github.com/BLSQ/iaso/assets/281139/a2ab7dad-6c53-4b46-a324-ca33d55ed204

## Notes

[IA-2411]: https://bluesquare.atlassian.net/browse/IA-2411?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[IA-2229]: https://bluesquare.atlassian.net/browse/IA-2229?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kemar:
Almost no test, is it on purpose?

kemar:
@cypress

kemar:
@cypress

hat/assets/js/apps/Iaso/domains/instances/components/ActionTableColumnComponent.js
```diff
@@ -121,8 +129,14 @@ const ActionTableColumnComponent = ({ settings }) => {
 
     const confirmLinkOrUnlink = isItLinked => {
         return !isItLinked
-            ? linkOrgUnitToReferenceSubmission(null)
-            : linkOrgUnitToReferenceSubmission(settings.row.original.id);
+            ? linkOrgUnitToReferenceSubmission(
+                  settings.row.original.id,
+                  'unflag',
```
bmonjoie: https://www.typescriptlang.org/docs/handbook/enums.html

hat/assets/js/apps/Iaso/domains/orgUnits/components/OrgUnitMultiReferenceInstances.tsx
```diff
@@ -0,0 +1,92 @@
...
+        label: instance.form_name,
+    }));
+    const url = `${baseUrls.instanceDetail}/instanceId/${active.id}/referenceFormId/${active.form_id}`;
+    const title = `${formatMessage(MESSAGES.detailTitle)} - ${
+        active.form_name
+    }`;
+
+    const handleChange = value => {
+        setActive(referenceInstances.filter(i => i.id === value)[0]);
+    };
```
bmonjoie, line 53: I think this should be wrapped into a `useCallback` not to redeclare the function on every render.

hat/assets/js/apps/Iaso/domains/orgUnits/components/OrgUnitMultiReferenceInstances.tsx
```diff
@@ -0,0 +1,92 @@
...
+    const [active, setActive] = useState<Instance>(referenceInstances[0]);
+
+    const options = referenceInstances.map(instance => ({
+        value: instance.id,
+        label: instance.form_name,
+    }));
+    const url = `${baseUrls.instanceDetail}/instanceId/${active.id}/referenceFormId/${active.form_id}`;
+    const title = `${formatMessage(MESSAGES.detailTitle)} - ${
+        active.form_name
+    }`;
```
bmonjoie, line 49: Those should probably be wrapped into `useMemo` not to compute them on every render.

iaso/api/org_units.py
```diff
@@ -428,25 +429,20 @@ def partial_update(self, request, pk=None):
...
+                            "errorMessage": e.messages[0],
                         }
                     )
-                else:
-                    org_unit.reference_instance = instance
-            else:
-                instance = None
-                org_unit.reference_instance = instance
+            elif request.data["reference_instance_action"] == Instance.REFERENCE_UNFLAG_CODE:
+                instance.unflag_reference_instance(org_unit)
```
bmonjoie, line 445: There should probably be a `else` to add an error to the `errors` dict because a wrong action was provided

iaso/api/org_units.py
```diff
@@ -615,17 +612,13 @@ def create_org_unit(self, request):
...
-            # Check if the instance has as form the reference_form for the orgUnittype
-            # if the reference_form is the same as the form related to the instance one,
-            # assign the instance to the orgUnit as a reference instance
-            if org_unit_type.reference_form == instance.form:
-                org_unit.reference_instance = instance
-
         org_unit.save()
         org_unit.groups.set(new_groups)
 
+        if reference_instance_id and org_unit_type:
```
bmonjoie, line 618: When creating an Org Unit, shouldn't we be allowed to pass a table of `reference_instance_ids`?

iaso/models/base.py
```diff
@@ -885,6 +888,37 @@ class Instance(models.Model):
...
+        if not self.org_unit:
+            return False
+        return self.org_unit.reference_instances.filter(orgunitreferenceinstance__instance=self).exists()
+
+    def flag_reference_instance(self, org_unit: "OrgUnit") -> "OrgUnitReferenceInstance":
+        if not self.form:
+            raise ValidationError(_("The Instance must be linked to a Form."))
+        if not org_unit.org_unit_type:
+            raise ValidationError(_("The OrgUnit must be linked to a OrgUnitType."))
+        if not org_unit.org_unit_type.reference_forms.filter(id=self.form_id).exists():
```
bmonjoie, line 911: Could be replaced with the property:
```suggestion
        if not self.is_instance_of_reference_form():
```

bmonjoie:
As discussed with @kemar, I believe the goal was to be able to see the history of the reference instances. Therefore, the unicity on `org_unit_id` and `form_id` in the join table would be problematic.
@madewulf Should the code allow to see the history of the reference instances, or only seeing the current one is enough?

iaso/models/base.py
kemar: In the current state of affairs, it cannot because `org_unit` is an argument.

The reason why it's an argument are:

- the `Instance.org_unit` field is nullable (I don't konw if it's always set AND set to the right one?)
- the previous version of the code was acting like this

iaso/api/forms.py
```diff
@@ -139,6 +143,10 @@ def get_latest_form_version(obj: Form):
     def get_org_unit_types(obj: Form):
         return [t.as_dict() for t in obj.org_unit_types.all()]
 
+    @staticmethod
+    def get_reference_form_of_org_unit_types(obj: Form):
+        return [org_unit.as_dict() for org_unit in obj.reference_of_org_unit_types.all()]
+
```
madewulf, line 149: Looks to me that these are not org_unit but reference forms (the name org_unit is wrong?)

iaso/api/instances.py
```diff
@@ -314,9 +314,9 @@ def list(self, request):
         # 2. Prepare queryset (common part between searches and exports)
         queryset = self.get_queryset()
         queryset = queryset.exclude(file="").exclude(device__test_device=True)
-        queryset = queryset.prefetch_related("org_unit")
-        queryset = queryset.prefetch_related("org_unit__version", "org_unit__version__data_source")
-        queryset = queryset.prefetch_related("org_unit__org_unit_type")
+        queryset = queryset.prefetch_related("org_unit__reference_instances")
```
madewulf: I'm a tad worried about performance issues linked to this for large exports to csv (which is something that happens a lot, since this is the mechanism used to produce dashboards out of the data)

madewulf: Looks like if you have 200 000 instances that are reference instances, we are going to fetch them all at once for this?

madewulf: For csv exports, we're streaming the content to avoid loading too much data at once in memory and this might defeat the purpose. On the other hand the django ORM is maybe smart enough to prefetch multiple times the data  needed when streaming, but I'm not sure.

iaso/models/base.py
```diff
@@ -885,6 +888,37 @@ class Instance(models.Model):
...
+    def is_instance_of_reference_form(self) -> bool:
+        if not self.org_unit or not self.org_unit.org_unit_type:
+            return False
+        return self.org_unit.org_unit_type.reference_forms.filter(id=self.form_id).exists()
+
+    @property
+    def is_reference_instance(self) -> bool:
+        if not self.org_unit:
+            return False
+        return self.org_unit.reference_instances.filter(orgunitreferenceinstance__instance=self).exists()
```
madewulf, line 904: I'm a bit worried about making  n sql requests if serializing n instances when using this. Has this been checked?

madewulf:
Functionally, this looks exactly like what I expected. I'm just worried about some possibles performance impacts. 
If these are unfounded, you can merge the PR for me

hat/assets/js/apps/Iaso/domains/orgUnits/components/OrgUnitMultiReferenceInstances.tsx
kemar: Re-renders aren't a big deal in most cases. I don't think that a simple CPU bound operation could lead to performance problems here.

iaso/api/forms.py
kemar: It looks a bit strange when seen out of context, but it's right.

I'm adding this method to `FormSerializer` to populate `reference_form_of_org_unit_types`:

```python
serialized_form = FormSerializer(form)
serialized_form.reference_form_of_org_unit_types
```

So it returns `org_unit`s, [as asked by @bmonjoie ](https://bluesquare.atlassian.net/browse/IA-2411?focusedCommentId=59205).

Can't think of a better name…

iaso/api/instances.py
kemar: > if you have 200 000 instances that are reference instances

Is this a real life use-case?

iaso/models/base.py
kemar: Yup, this is why I'm using `prefetch_related` in various Serializers.

iaso/api/org_units.py
kemar: I asked myself this question, but it would have required too many changes on the front-end.

And since the front-end is only flagging references one by one, I sticked with one `reference_instance_ids`…

iaso/api/instances.py
kemar: Anyway I moved this part explicitly out https://github.com/BLSQ/iaso/pull/807/commits/5f87799890735652ac1fa916384799ca6d62c33b

hat/assets/js/apps/Iaso/domains/instances/components/ActionTableColumnComponent.js
kemar: Fixed https://github.com/BLSQ/iaso/pull/807/commits/69804c211a3c4b1170e4ad3f8530c99cd18a5d7f (but not in TS because it's still a JS file).

kemar:
@cypress

hat/assets/js/apps/Iaso/domains/instances/hooks/speeddials.ts
```diff
@@ -23,20 +24,24 @@ export const useLinkOrgUnitToReferenceSubmission = ({
...
     return useCallback(
-        (currentInstance: Instance, isOrgUnitAlreadyLinked: boolean) => {
-            const { org_unit: orgUnit, id: instanceId } = currentInstance ?? {};
-            const id = isOrgUnitAlreadyLinked ? null : instanceId;
+        (currentInstance: Instance) => {
+            const {
+                org_unit: orgUnit,
+                id: instanceId,
+                is_reference_instance: isReferenceInstance,
+            } = currentInstance ?? {};
```
beygorghor, line 38: Not sure this will work if `currentInstance` is empty has you are trying to get org_unit, id, ... of an empty object

hat/assets/js/apps/Iaso/domains/orgUnits/components/OrgUnitMultiReferenceInstances.tsx
beygorghor: useMemo is not required while getting a string or a boolean FYI

hat/assets/js/apps/Iaso/domains/orgUnits/components/OrgUnitMultiReferenceInstances.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+    const handleChange = value => {
+        setActive(referenceInstances.filter(i => i.id === value)[0]);
+    };
+
+    return (
+        <>
```
beygorghor: useless fragment

hat/assets/js/apps/Iaso/domains/orgUnits/types/orgunitTypes.ts
```diff
@@ -30,7 +31,7 @@ export type OrgunitType = {
     units_count: number;
     sub_unit_types: OrgunitType[];
     allow_creating_sub_unit_types: OrgunitType[];
-    reference_form: any;
+    reference_forms: Form[];
```
beygorghor, line 34: ❤️

beygorghor (approved):
I have a warning while editing reference form on an org unit type just:
Screenshot 2023-10-04 at 13 46 38

hat/assets/js/apps/Iaso/domains/orgUnits/components/OrgUnitMultiReferenceInstances.tsx
bmonjoie: what about the `options` variable?
Also, could you explain why it's not needed for strings and booleans?
It seems to me that `url` and `title` are both computing strings, which can be costly.

kemar: > In my personal opinion, it would be a waste of time to wrap every single object/array/function in these hooks. The benefits are negligible in most cases; React is highly optimized, and re-renders often aren't as slow or expensive as we often think they are!

https://www.joshwcomeau.com/react/usememo-and-usecallback/#when-to-use-these-hooks-6

kemar:
> I have a warning while editing reference form on an org unit type just: Screenshot 2023-10-04 at 13 46 38

@beygorghor thanx for the review! But I can't reproduce this. Can you give me the exact URL please?

PR #808, opened 2023-09-15
TRIA-17 use trypelim reducers in app

beygorghor:
Trypelim is using a bunch of reducers in order to work, we should be able to add manually reducers to the existing one's so we d'ont have to rewrite all the logic of trypelim pages

Related JIRA tickets :[TRIA-17](https://bluesquare.atlassian.net/browse/TRIA-17)

- [ ] Are there enough tests

## Changes

- extend store to allow plugin reducers to be added
- `useInjectedStore` hook to inject a reducer

## How to test

Wrap a component qith the code in useInjectedStore and try to inject a existing reducer with a different key, it should work

[TRIA-17]: https://bluesquare.atlassian.net/browse/TRIA-17?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/hat/assets/js/apps/Iaso/redux/createStore.js
+++ b/hat/assets/js/apps/Iaso/redux/createStore.js
@@ -6,12 +6,17 @@ import {
 } from 'redux';
 import { routerReducer } from 'react-router-redux';
 
-export default (initialState = {}, reducers = {}, middleWare = []) =>
-    _createStore(
-        combineReducers({
-            routing: routerReducer,
-            ...reducers,
-        }),
+const createReducer = (appReducers, pluginReducers) => {
+    return combineReducers({
+        routing: routerReducer,
+        ...appReducers,
+        ...pluginReducers,
+    });
+};
+
+export default (initialState = {}, reducers = {}, middleWare = []) => {
+    const store = _createStore(
+        createReducer(reducers),
         initialState,
         compose(
             applyMiddleware(...middleWare),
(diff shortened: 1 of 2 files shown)


hat/assets/js/apps/Iaso/redux/createStore.js
quang-le, line 1: Wouldn't we need an `initialState` from the plugins as well?

hat/assets/js/apps/Iaso/redux/useInjectedStore.ts
```diff
@@ -0,0 +1,63 @@
...
+        },
+    ]);
+    if (!store) return null;
+    return (
+        <Provider store={store}>
+            <CUSTOM_COMPONENT />
+        </Provider>
+    );
+ *
+ */
```
quang-le, line 24: This makes me think we could go further than the hook and provide a custom `Provider` where the reducers are passed as props. Not blocking though, especially given the intended use:
```typescript
export const CombinedStoreProvider = ({reducers: Reducer[], children:ReactNode}) => {
    const store = useInjectedStore(reducers);
    if (!store) return null;
    return (
        <Provider store={store}>
           {children}
        </Provider>
    );
}

```

quang-le (commented):
Haven't live tested, but here are my comments on the code. 
Neat trick!

hat/assets/js/apps/Iaso/redux/createStore.js
beygorghor: In fact the initialState is already set by default in the reducer...

bramj:
I'm not yet well-versed enough in React to give meaningful comments on the code structure, but it sure is working great on Trypeliaso, so 👍 for me

bramj (approved):
LGTM 👍

PR #810, opened 2023-09-18
Polio-1124 add lqas refresh button

quang-le:
LQAs country map needed a button to refresh data from country. This implies launching an OpenHexa pipeline

Related JIRA tickets :  POLIO-1124

- [ ] Are there enough tests

## Changes

- Add `RefreshLqas` component
    -  The component checks in the tasks lists the latest task for the country and displays its datetime info+status. It keeps polling the API every second to update the status and inform the user if a manual update is ongoing
    - If a manual update is ongoing, the user will not be able to launch a new task until the manual update is over
    - Else it will create a new refresh task and disable the update button until the task is done
- Add `taskMonitor.ts` with hooks to create tasks and poll `tasks` api for task update
- Add `useGetLatestLQASIMUpdate` to show latest manual update and its status
- Rearranged the LQAs Afro map filters to have a better layout on small screens. No visible difference otherwise

### Pre-requisites
- Admin access to the [OpenHexa workspace](https://app.openhexa.org/workspaces/polio-3688d8)
- An `lqas-config` json config file to get lqas data 
- ngrok

### Setup
- Run `docker-compose up` or `docker-compose up iaso`
- Start an ngrok tunnel on localhost:8081
- In OpenHexa: 
    - Change the `iaso-tester` connection with the ngrok url and your local Db credentials
    - Go to "Pipelines", click "Create" and copy your API token
- In your `.env` : 
    - Set `OPENHEXA_TOKEN` to the value just copied from OpenHexa 
    - Set `OH_PIPELINE_TARGET` to `"custom"`
    - Set `OPENHEXA_URL` to  `"https://api.openhexa.org/graphql/"`
    - Set `LQAS_PIPELINE_VERSION` to 74
    - Set `LQAS_PIPELINE`: see setup for https://github.com/BLSQ/iaso-pipelines/pull/15
- If you already have lqas data in your DB:
    - Go to django admin > Jsondatastore and pick an lqas datastore (naming convention `lqas_<country_id >`)
    - In the json, delete entries at the `data` level of the nicely nested object. This will allow you to see the changes when launching the update.

### Test
- Press the button, see what happens.

## Print screen / video

https://github.com/BLSQ/iaso/assets/38907762/91aaeb7d-1b6b-4120-a8f3-8cdee174f71f

## Notes

Merges in https://github.com/BLSQ/iaso/tree/POLIO-1191_add_endpoint_for_openhexa_task
Requires https://github.com/BLSQ/iaso-pipelines/pull/15

VERY IMPORTANT: environment variables need to be added to polio and polio staging. The `OPENHEXA_TOKEN`should be taken from the `OpenHexa API` connection in OpenHexa's polio workspace


hat/assets/js/apps/Iaso/hooks/taskMonitor.ts
```diff
@@ -0,0 +1,83 @@
...
+            onSuccess: data => {
+                if (!data) {
+                    setEnabled(false);
+                    if (invalidateQueries.length > 0) {
+                        invalidateQueries.forEach(queryKey =>
+                            queryClient.invalidateQueries(queryKey),
+                        );
+                    }
+                }
+            },
```
beygorghor, line 55: ❤️

plugins/polio/js/src/components/LQAS-IM/RefreshLqasData.tsx
```diff
@@ -0,0 +1,149 @@
...
+    }
+    if (lastUpdate?.status === 'SUCCESS' && Boolean(lastUpdate?.ended_at)) {
+        result.message = `${moment
+            .unix(lastUpdate.ended_at as number)
+            .format('LTS')}`;
+    }
+    if (lastUpdate?.status) {
+        const updateStatus = formatMessage(MESSAGES[lastUpdate.status])
+            ? formatMessage(MESSAGES[lastUpdate.status])
+            : '';
```
beygorghor: this will raise a warning anyway, you should check the presence of the entry in the object MESSAGE
```suggestion
        const updateStatus = MESSAGES[lastUpdate.status]
            ? formatMessage(MESSAGES[lastUpdate.status])
            : '';
```

beygorghor (commented):
Just reviewed the code, need to configure the pipeline with Open Hexa...

iaso/models/base.py
```diff
@@ -294,6 +300,10 @@ def report_success(self, message=None):
         self.result = {"result": SUCCESS, "message": message}
         self.save()
 
+    def kill_if_external(self):
+        if self.external:
+            self.stop_if_killed()
```
bramj: This method seems not used

iaso/api/tasks/__init__.py
```diff
@@ -35,6 +36,8 @@ class Meta:
     def update(self, task, validated_data):
         if validated_data["should_be_killed"] is not None:
             task.should_be_killed = validated_data["should_be_killed"]
+            if task.external:
+                task.status = KILLED
```
bramj: Should this be `task.stop_if_killed()`?

plugins/polio/js/src/domains/LQAS-IM/shared/Filters.tsx
```diff
@@ -125,22 +118,15 @@ const Filters: FunctionComponent<Props> = ({
...
-                            onClick={handleRefresh}
-                        >
-                            <Box mr={1} pt={1}>
-                                <RefreshIcon fontSize="small" />
-                            </Box>
-                            {formatMessage(MESSAGES.refreshPage)}
-                        </Button>
-                    </Box>
-                </Grid> */}
+                {/* remove condition when IM pipeline is ready */}
```
bramj, line 121: What is the context on this? Will there be other categories?

plugins/polio/js/src/domains/LQAS-IM/shared/RefreshLqasData.tsx
```diff
@@ -0,0 +1,151 @@
...
+        result.updateStatus = updateStatus;
+    }
+    return result;
+};
+
+export const RefreshLqasData: FunctionComponent<Props> = ({
+    countryId,
+    category,
+}) => {
+    const taskUrl = category === 'lqas' ? LQAS_TASK_ENDPOINT : undefined;
```
bramj, line 58: This feels a bit weird?

hat/assets/js/apps/Iaso/hooks/taskMonitor.ts
```diff
@@ -0,0 +1,83 @@
+import { useEffect, useState } from 'react';
+import { UseMutationResult, UseQueryResult, useQueryClient } from 'react-query';
+import { getRequest, postRequest } from '../libs/Api';
+import { useSnackMutation, useSnackQuery } from '../libs/apiHooks';
+
+const TASK_ENDPOINT = '/api/tasks/';
+const TASK_CREATE_ENDPOINT = '/api/tasks/create/';
+
+const getTasks = (id?: number, endpoint = TASK_ENDPOINT) => {
+    const url = id ? `${endpoint}${id}` : endpoint;
```
bramj: Missing `/` after the id

plugins/polio/js/src/domains/LQAS-IM/shared/RefreshLqasData.tsx
```diff
@@ -0,0 +1,151 @@
...
+    const { formatMessage } = useSafeIntl();
+    const [taskId, setTaskId] = useState<number>();
+    const [lastTaskStatus, setlastTaskStatus] = useState<string | undefined>();
+    const queryClient = useQueryClient();
+    const { data: isDataUpdating, isFetching: isFetchingTaskStatus } =
+        useTaskMonitor({
+            taskId,
+            endpoint: taskUrl,
+            invalidateQueries: [category, 'get-latest-task-run'],
+        });
```
bramj, line 68: Is it possible that we don't need the `useTaskMonitor`? We already have the `useGetLatestLQASIMUpdate` polling continuously on the page. Since we can only run 1 refresh task for a given country, I think the `useTaskMonitor` and the latest update will always be the same. In that case might the latest update polling be sufficient?

plugins/polio/tasks/api/refresh_lqas_data.py
```diff
@@ -0,0 +1,227 @@
...
+    country_id = serializers.IntegerField(required=False)
+
+    def validate(self, attrs):
+        validated_data = super().validate(attrs)
+        request = self.context["request"]
+        country_id = request.data.get("country_id", None)
+        if request.method == "GET":
+            country_id = request.query_params.get("country_id", None)
+        # It seems a bit stange to limit the access on country id
+        # but to launch the refresh for all countries if no id is passed
```
bramj: Where do we call the refresh without country_id?

plugins/polio/tasks/api/refresh_lqas_data.py
```diff
@@ -0,0 +1,227 @@
...
+                        run_id
+                        status
+                        config
+                    }
+                }
+            }
+        }
+        """
+            % (settings.LQAS_PIPELINE)
+        )
```
bramj: Next time we add a connection to OpenHexa we should extract these calls into an OpenHexa API client "wrapper"

plugins/polio/tasks/api/refresh_lqas_data.py
```diff
@@ -0,0 +1,227 @@
...
+        try:
+            latest_runs = client.execute(get_runs)
+            # Warning the query will only return the last 10 runs
+            active_runs = [
+                run
+                for run in latest_runs["pipeline"]["runs"]["items"]
+                if (run["status"] != "queued" and run["status"] != "success" and run["status"] != "failed")
```
bramj:
```suggestion
                if (run["status"] not in ["queued", "success", "failed"])
```

plugins/polio/tasks/api/refresh_lqas_data.py
```diff
@@ -0,0 +1,227 @@
...
+            task.progress_message = validated_data["progress_message"]
+        if has_progress_value:
+            task.progress_value = validated_data["progress_value"]
+        if has_end_value:
+            task.end_value = validated_data["end_value"]
+        task.save()
+        return task
+
+
+class CustomTaskSearchFilterBackend(filters.BaseFilterBackend):
```
bramj: Just a question out of curiosity: why do we need this?

bramj (changes requested):
Nice work! Definitely not the easiest feature, I tried to take my time to test it well locally and it's working well! 👍 

Some comments, mostly small and one thing where I think we can simplify quite a bit (unless I'm missing something, which is very possible 😅 )

A more general comment: I suppose it's already been discussed, but for this sort of refresh button we could use Websockets to avoid potentially DOSing our own server. Has there already been talk of adding it to Iaso in general?

quang-le:
> Nice work! Definitely not the easiest feature, I tried to take my time to test it well locally and it's working well! 👍
> 
> Some comments, mostly small and one thing where I think we can simplify quite a bit (unless I'm missing something, which is very possible 😅 )
> 
> A more general comment: I suppose it's already been discussed, but for this sort of refresh button we could use Websockets to avoid potentially DOSing our own server. Has there already been talk of adding it to Iaso in general?

Websockets is a route we don't want to travel 😁

plugins/polio/js/src/domains/LQAS-IM/shared/Filters.tsx
quang-le: Yes, there will be at least one other pipeline for IM (Independent Monitoring) which is a similar process, but gets the data from other forms

iaso/api/tasks/__init__.py
quang-le: No, external tasks are handled differently

plugins/polio/tasks/api/refresh_lqas_data.py
quang-le: We may do it through a cron.
We don't do it yet here but the pipeline can run without it

plugins/polio/js/src/domains/LQAS-IM/shared/RefreshLqasData.tsx
quang-le: That's because the filter is shared with IM, which has no pipeline associated to it yet

plugins/polio/tasks/api/refresh_lqas_data.py
quang-le: Agreed

plugins/polio/tasks/api/refresh_lqas_data.py
quang-le: We don't really. At some stage I toyed with the idea of using such filter, so the front-end would send search parameters to filter tasks by status. Then I just left it in

iaso/models/base.py
quang-le: True, leftover from a road not traveled

plugins/polio/js/src/domains/LQAS-IM/shared/RefreshLqasData.tsx
quang-le: It needs to be tested, but it sounds about right. Also, since the loading state would be inferred from the latest running task, if we land on the page and another task in running, the button would be disabled, so that would be an, improvement of the feature.
Nice catch!

bramj: Nice! :D I noticed because that's actually already sort of the situation when you click the button and then refresh the page. You still get the correct state because of the continuous fetching of the last task.

bramj:
Wrt Websockets: I agree it adds a complexity that we might want to avoid. But we should reconsider the question the next time we add this sort of functionality. Short polling like this can be pretty heavy on the DB since we're not just doing one query but also checking the user permissions, potentially loading org units and whatnot.

plugins/polio/tasks/api/refresh_lqas_data.py
bramj: Ah yes that makes sense 👍

plugins/polio/js/src/domains/LQAS-IM/shared/RefreshLqasData.tsx
quang-le: If it's OK for you, I'll spin it in another ticket, so I can change it without rush

quang-le:
> Wrt Websockets: I agree it adds a complexity that we might want to avoid. But we should reconsider the question the next time we add this sort of functionality. Short polling like this can be pretty heavy on the DB since we're not just doing one query but also checking the user permissions, potentially loading org units and whatnot.

ping @madewulf . In the short term, we can probably increase the interval of the polling

PR #812, opened 2023-09-19
IA-2437: wrap comment in tooltip (budget tool)

quang-le:
Full comments are not visible in budget tool table

Related JIRA tickets : IA-2437

- [ ] Are there enough tests

## Changes

 -Add a tooltip around the comment

## How to test

- Go to polio budget, create a step with a long comment, save, then hover over it on the table

## Print screen / video

https://github.com/BLSQ/iaso/assets/38907762/8ca0e1ed-1ab7-4436-9cf1-f3ea230b5e3e


beygorghor (approved):
Not blocking but a separate component is better, also we put always the tooltip at the bottom...

quang-le:
> Not blocking but a separate component is better, also we put always the tooltip at the bottom...

Yeah, but since it can be shown for every row of the table, I put it on the side in order not to block the row below

hakifran (approved):

PR #820, opened 2023-09-20
POLIO-1205- Correct filters on vaccine authorisations

SimonDus:
It was not only EXPIRED but also Signature filters that were broken.
As the status are required to compute the correct response, filtering in the get_queryset method was not possible for this action. So I moved the filtering after calculation of the response. 

Related JIRA tickets : POLIO-1205

- [x] Are there enough tests

## How to test

Go to vaccine authorisation tab in Polio and user the status filter. The result should be correct.  A team named " nOPV2 vaccine authorization alerts " is required with your user in it to see the results. 

## Print screen / video

https://github.com/BLSQ/iaso/assets/45785235/3ca43d6a-ae11-4d7f-b736-ae956711b28d

### DIFF
--- a/plugins/polio/api/vaccines/vaccine_authorization.py
+++ b/plugins/polio/api/vaccines/vaccine_authorization.py
@@ -143,7 +143,9 @@ class VaccineAuthorizationViewSet(ModelViewSet):
             queryset = queryset.filter(country__pk__in=ou_pk_list)
         if search:
             queryset = queryset.filter(country__name__icontains=search)
-        if auth_status:
+
+        # Expired status need to be treated differently as others status are mandatory to calculate the get_most_recent
+        if auth_status and auth_status != "EXPIRED":
             auth_status = auth_status.split(",")
             queryset = queryset.filter(status__in=auth_status)
 
@@ -159,6 +161,20 @@ class VaccineAuthorizationViewSet(ModelViewSet):
 
         return super().create(request)
 
+    def list(self, request: Request, *args, **kwargs):
+
+        auth_status = self.request.query_params.get("auth_status", None)
+
+        if auth_status == "EXPIRED":
+
+            queryset = self.get_queryset().filter(status="EXPIRED")
+
+            serializer = VaccineAuthorizationSerializer(queryset, many=True)
+
+            return Response(serializer.data)
+
+        return super().list(request, *args, **kwargs)
+
     @action(detail=False, methods=["GET"])
     def get_most_recent_authorizations(self, request):
         """
@@ -169,6 +185,7 @@ class VaccineAuthorizationViewSet(ModelViewSet):
         response = []
 
         ordering = request.query_params.get("order", None)
+        auth_status = request.query_params.get("auth_status", None)
 
         for auth in queryset:
             if auth.country not in country_list:
@@ -192,6 +209,9 @@ class VaccineAuthorizationViewSet(ModelViewSet):
             last_entry_date_check = VaccineAuthorization.objects.filter(
                 country=country, account=self.request.user.iaso_profile.account, deleted_at__isnull=True
             ).last()
+
+            if auth_status == "EXPIRED":
+                response = [entry for entry in response if entry["status"] == "EXPIRED"]
 
             if last_entry_date_check and last_entry:
                 if last_entry_date_check.expiration_date > last_entry.expiration_date:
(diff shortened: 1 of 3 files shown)


quang-le (dismissed):
It seems a side effect of the fix was to break the filter on country groups
https://github.com/BLSQ/iaso/assets/38907762/64930046-e096-468a-9589-34f6fbfb29c3

SimonDus:
@quang-le Indeed, it's fixed

plugins/polio/api/vaccines/vaccine_authorization.py
```diff
@@ -247,6 +253,24 @@ def get_most_recent_authorizations(self, request):
 
                 response.append(vacc_auth)
 
+        if auth_status:
+            response = [entry for entry in response if entry["status"] in auth_status.split(",")]
+
+        if block_country:
+            block_country = block_country.split(",")
+            block_country = Group.objects.filter(pk__in=block_country)
+            org_units = [
```
hakifran: There are three imbricated for loops on one line. could it be possible to make separated for loops and each in its variable to make the code more readable

plugins/polio/api/vaccines/vaccine_authorization.py
```diff
@@ -247,6 +253,24 @@ def get_most_recent_authorizations(self, request):
...
+                ou_queryset
+                for ou_queryset in [
+                    ou_group_queryset for ou_group_queryset in [country.org_units.all() for country in block_country]
+                ]
+            ]
+            ou_pk_list = []
+            for ou_q in org_units:
```
hakifran: I maybe wrong, but this can be done on one line of code. thank you

hakifran (changes requested):

plugins/polio/api/vaccines/vaccine_authorization.py
SimonDus: To me it's one of the best advantage of the list comprehension in Python is that you can shrink simple for loop.

plugins/polio/api/vaccines/vaccine_authorization.py
SimonDus: indeed it can be done with a list comprehension

plugins/polio/api/vaccines/vaccine_authorization.py
SimonDus: Indeed after reading it I kinda wrote it too quickly and can simplify it

hakifran (approved):
Seems good for me

plugins/polio/api/vaccines/vaccine_authorization.py
```diff
@@ -162,9 +162,15 @@ def create(self, request: Request, *args: Any, **kwargs: Any) -> Response:
...
     def get_most_recent_authorizations(self, request):
         """
-        Returns the most recent validated or expired authorization or the most recent ongoing or signature if the first case does not exists.
+        Compute the most recent vaccine authorization per country.
         """
-        queryset = self.get_queryset()
+        # Filters are done after calculation as all the status are required in order to compute the correct response
+        user = self.request.user
+        user_access_ou = OrgUnit.objects.filter_for_user_and_app_id(user, None)
+        user_access_ou = user_access_ou.filter(org_unit_type__name="COUNTRY")
```
mathvdh: Should "COUNTRY" be hardcoded, is it always the correct name for it ? At least put it in a constant or a setting

plugins/polio/api/vaccines/vaccine_authorization.py
```diff
@@ -247,6 +253,18 @@ def get_most_recent_authorizations(self, request):
...
+        if auth_status:
+            response = [entry for entry in response if entry["status"] in auth_status.split(",")]
+
+        if block_country:
+            block_country = block_country.split(",")
+            block_country = Group.objects.filter(pk__in=block_country)
+            org_units = [country.org_units.all() for country in block_country]
+            ou_pk_list = []
+            for ou_q in org_units:
+                ou_pk_list = [ou.pk for ou in ou_q]
```
mathvdh: Something like this would be more efficient and more readable (I think) and will avoid duplicated pks
```python
import itertools
org_units_ids = [ country.org_units.all().values_list('pk', flat=True) for coutry in block_country ]
ou_pk_list = set(itertools.chain.from_iterable(org_units_ids))
```
If you are sure there will be no duplicates you can use `list` instead of `set`

```

mathvdh (changes requested):
Except for my little comment I think it's good

plugins/polio/api/vaccines/vaccine_authorization.py
SimonDus: I already simplified it, no ?

mathvdh (approved):

beygorghor (approved):

PR #826, opened 2023-09-22
Removing mypy checks from CI

madewulf:
They have been an annoyance rather than helpful globally


mathvdh:
Agreed. The only result is that I often add `# type: ignore` just to have it go through

mathvdh (approved):

PR #828, opened 2023-09-25
IA-2392:Org unit logs add link to parentorg unit and translate status

beygorghor:
In the history of an organisation unit, the information is not always clear and sometimes not translated.

For example :

the modification of a parent is noted according to the id of the parent and not its name. The name of the parent should appear clearly.

ditto for the path

the modification of the status of an OU is noted in English and should be translated into French when the language of the IASO instance is French (NEW → Nouveau, VALID → Validé)

Related JIRA tickets : IA-2392

- [ ] Are there enough tests

## Changes

- use LinkToOrgUnit to display parent log (+ fetch parent org unit detail)
- translate status

## How to test

- Create a new org unit, without parent and save
- add the parent and save
- change status and save
- go to to history tab and check that parent org unit log and status are correctly displayed

## Print screen / video

Screenshot 2023-09-25 at 12 29 41
Screenshot 2023-09-25 at 12 29 28


SimonDus (approved):
Looks good to me

PR #829, opened 2023-09-25
IA-2432 add an unique constraint on form version

mathvdh:
Due to some concurrency issues, some clients managed to create Form Versions which have the same id even thou there is already a check about that when parsing the XLS file after upload via API.

Related JIRA tickets : IA-2432

- [x] Are there enough tests

## Changes

Adding of a unique_together DB contraint on FormVersion <form_id, version_id>

Adding a manual migration to check if some of those cases are there before adding the constraint and handling them by renaming them (increment the previous name).

## How to test

Not easy to test as it's protected at 2 levels (API and DB)
But see attached tests


mathvdh:
@cypress

mathvdh:
@cypress

mathvdh:
@cypress

mathvdh:
@cypress

mathvdh:
@cypress

mathvdh:
@cypress

iaso/migrations/0232_check_no_duplicated_version_id.py
```diff
@@ -0,0 +1,64 @@
...
+
+            print("Updating other version to use version_id", vv.version_id, "\n")
+
+            # update corresponding instances
+
+            instances = Instance.objects.filter(form_version=vv)
+            for instance in instances:
+                print("Updating instance", instance.id, "for version", vv.version_id)
+
+                instance.json["_version"] = vv.version_id
```
madewulf, line 55: This won't fix the xml of the form, and I'm guessing that editing an instance with enketo would rewrite that field back to the version still in the xml

PR #830, opened 2023-09-25
Polio-1125 backend multi round for budget

SimonDus:
Added a Model named Process to allows budget to be multiround.
A process is linked to round(s) that are linked to a campaign.

Related JIRA tickets : [POLIO-1125](https://bluesquare.atlassian.net/browse/POLIO-1125)

- [x] Are there enough tests

## Changes

There is no interface. you must go to /api/polio/budgetprocesses/ or django admin and create a process after creating a budget
You can also check the response from the budget and there should be the proccess info in the json.
/api/polio/budget/
The setup is quite cumbersome.

## Print screen / video
image

[POLIO-1125]: https://bluesquare.atlassian.net/browse/POLIO-1125?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


plugins/polio/api/urls.py
```diff
@@ -21,6 +20,7 @@
 from plugins.polio.api.vaccines.vaccine_authorization import VaccineAuthorizationViewSet
 from plugins.polio.api.dashboards.vaccine_stocks import VaccineStocksViewSet
 from plugins.polio.budget.api import BudgetCampaignViewSet, BudgetStepViewSet, WorkflowViewSet
```
hakifran: I think this line should be removed as it is repeated at line 23 with BudgetProcessViewset

plugins/polio/budget/api.py
```diff
@@ -12,6 +12,9 @@
 
 from iaso.api.common import CSVExportMixin, ModelViewSet, DeletionFilterBackend, HasPermission
```
hakifran: Same here, this line is duplicated at 15, it should be removed

plugins/polio/budget/api.py
```diff
@@ -12,6 +12,9 @@
...
 from plugins.polio.budget.models import BudgetStep, MailTemplate, get_workflow, BudgetStepFile
```
hakifran: same here, this line is duplicated at line 17

plugins/polio/budget/api.py
```diff
@@ -197,18 +193,38 @@ def mail_template(self, request, pk):
...
+
+    permission_classes = [HasPermission(permission.POLIO_BUDGET)]  # type: ignore
+    serializer_class = BudgetProcessSerializer
+    http_method_names = ["get", "head", "delete", "patch", "post"]
+    ordering_fields = ["created_at", "updated_at", "rounds", "teams"]
+    filter_backends = [filters.OrderingFilter, DjangoFilterBackend, DeletionFilterBackend]
+    results_key = "results"
+    remove_results_key_if_paginated = True
+    pagination_class = Paginator
+    user_team = Team.objects.filter()
```
hakifran: Why is the filter added on the teams objects but there is no a defined criteria on which the filter will take in account?

plugins/polio/budget/models.py
hakifran, line 1: There is an import **from plugins.polio.models import Campaign** which is defined inside **def filter_for_user(self, user: Union[User, AnonymousUser]):** method. is it possible to move at the line 20 with the import of **Round**

plugins/polio/budget/test.py
```diff
@@ -475,3 +476,72 @@ def test_csv_export_date(self):
...
+        data = {
+            "created_by": self.test_user.pk,
+            "created_by_team": self.team1.pk,
+            "rounds": [self.rounds_0.pk, self.rounds_1.pk, self.rounds_2.pk],
+            "teams": [self.team1.pk],
+        }
+
+        response = self.client.post("/api/polio/budgetprocesses/", data=data)
+
+        print(response.data)
```
hakifran: I don't know if this is necessary here

hakifran (changes requested):

plugins/polio/api/urls.py
SimonDus: Yea there was some duplicated/unused import in this file. Forgot to remove it.

plugins/polio/budget/api.py
SimonDus: Same as previous comment

plugins/polio/budget/api.py
SimonDus: Still

plugins/polio/budget/api.py
SimonDus: I think it's something that I wrote in case I'll need it but there was no use of it at the end, I'll remove it

plugins/polio/budget/models.py
SimonDus: done

plugins/polio/migrations/0147_update_budget_processes.py
```diff
@@ -0,0 +1,26 @@
...
+    BudgetStep = apps.get_model("polio", "BudgetStep")
+
+    budget_steps = BudgetStep.objects.all()
+
+    for step in budget_steps:
+        process = BudgetProcess.objects.create(
+            created_by_team=step.created_by_team,
+        )
+
+        process.rounds.set([step.campaign.rounds])
```
mathvdh: Did you actually try this with a campaign that already has multiple rounds ?

plugins/polio/budget/serializers.py
```diff
@@ -71,6 +78,26 @@ class TimelineSerializer(serializers.Serializer):
...
+class ProcessesForCampaignBudgetSerializer(serializers.ModelSerializer):
+    rounds = serializers.SerializerMethodField()
+
+    class Meta:
+        model = BudgetProcess
+        fields = ["id", "rounds", "teams"]
+
+    @staticmethod
+    def get_rounds(obj):
+        round_pks = obj.rounds.all().values_list("pk", flat=True)
```
mathvdh: I don't understand why we have to do this ?
Like obj.rounds.all().value_list("pk") then load Round object using the pk then get the "number" of each list .. ?
Can't we just :
```python
return [round.number for round in obj.rounds.all()]
```
Or something ..

plugins/polio/budget/test.py
```diff
@@ -475,3 +476,70 @@ def test_csv_export_date(self):
...
+        cls.rounds_1 = Round.objects.create(campaign=cls.campaign_1)
+        cls.rounds_2 = Round.objects.create(campaign=cls.campaign_1)
+
+        cls.campaign_2 = Campaign.objects.create(obr_name="test campaign_2", account=cls.test_user.iaso_profile.account)
+
+        cls.project1 = cls.account.project_set.create(name="project1")
+        cls.team1 = Team.objects.create(project=cls.project1, name="budget_test_team", manager=cls.test_user)
+        cls.team1.users.set([cls.test_user])
+
+    def test_post_process(self):
```
mathvdh, line 503: You are only testing the "happy" path, what about malformed requests ... Like posting a process without rounds or without teams, or trying to "create_by_team" with a team which I'm not part of, or from another account ...

plugins/polio/budget/models.py
```diff
@@ -96,6 +96,23 @@ def __repr__(self):
...
+class BudgetProcess(SoftDeletableModel):
+    class Meta:
+        ordering = ["-updated_at"]
+        verbose_name_plural = "Budget Processes"
+
+    created_at = models.DateTimeField(auto_now_add=True)
+    created_by = models.ForeignKey("auth.User", on_delete=models.PROTECT)
+    updated_at = models.DateTimeField(auto_now=True)
+    # Keep trace of the Team the user was acting on behalf of in case they get remove from it.
+    created_by_team = models.ForeignKey("iaso.Team", on_delete=models.PROTECT, blank=True, null=True)
```
mathvdh, line 108: What is the difference between `"iaso.Team"` (line 108)
And just `Team` (line 110)
?

plugins/polio/budget/serializers.py
```diff
@@ -94,22 +122,28 @@ class Meta:
...
-
     next_transitions = serializers.SerializerMethodField()
     # will need to use country__name for sorting
     country_name: serializers.SlugRelatedField = serializers.SlugRelatedField(
         source="country", slug_field="name", read_only=True
     )
 
+    @staticmethod
+    def get_processes(campaign: Campaign):
+        processes = BudgetProcess.objects.filter(rounds__in=Round.objects.filter(campaign=campaign))
```
mathvdh, line 142: Did you try
```python
BudgetProcess.objects.filter(rounds__campaign=campaign))
```
Honest question : Does it give the same results ?

mathvdh (changes requested):
A few questions and comments ... Please add some more tests

plugins/polio/budget/serializers.py
SimonDus: indeed

plugins/polio/budget/models.py
SimonDus: It's a mistake here. Using it as a string when the model is in a different app is the right way to import a model in Django... It can also avoid circular import. But from django doc, importing as a string when the model is in a different app is the right way to do it.

SimonDus:
I close this PR due to failed merge  with main and everything is broken. I'll make a new one.

PR #832, opened 2023-09-26
IA-1787 Form list: Form in multiple projects are counted multiple times in the submissions count

SimonDus:
Instances count were multiplied by the number of projects if a same form is shared between projects.  It results in 
instance_count was instance X projects which was wrong. Simply adding distinct to the querysets solve this bug. 

Related JIRA tickets : IA-1787

- [x] Are there enough tests

## How to test

Have a form with multiple submissions from multiple projects.
For example if you have 2 instances, in different projects you should have 2 submissions for the form. But the list will show 4.

## Print screen / video

https://github.com/BLSQ/iaso/assets/45785235/5d25262b-dbaf-472b-8dc7-8dc47374ffdb

### DIFF
--- a/iaso/tests/api/test_forms.py
+++ b/iaso/tests/api/test_forms.py
@@ -1,10 +1,11 @@
 import typing
 
+from django.core.files import File
 from django.utils.timezone import now
 
 from iaso import models as m
 from iaso.api.common import CONTENT_TYPE_XLSX
-from iaso.models import Form
+from iaso.models import Form, OrgUnit, Instance
 from iaso.test import APITestCase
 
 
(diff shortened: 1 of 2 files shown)


iaso/tests/api/test_forms.py
```diff
@@ -33,6 +41,15 @@ def setUpTestData(cls):
 
         cls.form_1 = m.Form.objects.create(name="Hydroponics study", created_at=cls.now)
 
+        cls.project_3 = m.Project.objects.create(name="Kotor", app_id="knights.of.the.old.republic", account=star_wars)
+
+        cls.jedi_council_corruscant = m.OrgUnit.objects.create(
```
kemar, line 46:
```suggestion
        cls.org_unit = m.OrgUnit.objects.create(
```

It would be easier to understand with this kind of variable names :)

kemar (approved):
LGTM

PR #844, opened 2023-09-29
Adding a reason for delay for polio campaigns: VRF_NOT_SIGNED

madewulf:


plugins/polio/models.py
quang-le, line 1: I think you need a migration for this change

quang-le (changes requested):

quang-le (approved):

PR #851, opened 2023-10-04
IA-2473:Optimizing campaigns endpoint

madewulf:
Added a lot of prefetch_related
Ensured that we didn't make sql queries for data that we already had in memory
Prevented the loading of geometries from the database when they are not needed. 

Result: from 26 seconds to 0.9 seconds to get all the campaigns on my Mac M1 and from 3699 sql requests to 16 (or 22 when logged in)

IA-2473


plugins/polio/api/campaigns/campaigns.py
```diff
@@ -134,7 +133,9 @@ def get_top_level_org_unit_id(self, campaign):
 
     def get_general_status(self, campaign):
         now_utc = timezone.now().date()
-        for round in campaign.rounds.all().order_by("-number"):
+        ordered_rounds = list(campaign.rounds.all())
+        ordered_rounds.sort(key=lambda x: x.number, reverse=True)
```
quang-le, line 137: I would never have thought it was cheaper to do the ordering this way

plugins/polio/api/lqas_im/lqasim_zoom_in_map.py
```diff
@@ -124,12 +124,12 @@ def list(self, request):
                 continue
             if latest_active_campaign.separate_scopes_per_round:
                 scope = latest_active_campaign.get_districts_for_round_number(round_number)
-
             else:
                 scope = latest_active_campaign.get_all_districts()
             # Visible districts in scope
+            scope_qs = OrgUnit.objects.filter(id__in=[ou.id for ou in scope])
```
quang-le, line 130: I don't understand this one: both `get_districts_for_round_number` and `get_all_districts` return a queryset that already filtered the OrgUnits in the scope

plugins/polio/api/campaigns/campaigns.py
madewulf: It prevents additional requests to the database server while we already have the data. 

Going through the network is usally an order of magnitude slower than accessing memory. 

Here it works because I prefetched all the rounds in prefetch_related. If that was not there, we would do one request per campaign

hakifran (approved):
Good for me

PR #852, opened 2023-10-04
POLIO-1223 Expiry date is past but the status is APPROVED and not EXPIRED

SimonDus:
The vaccine Authorisation status job was looking exactly for a date expired since 1 day. Normally it’s impossible to create an authorisation with a past expiration_date so this case should not happen in production. However I did  change the script to look for all the past dates instead of strictly today - 1.   

Related JIRA tickets : POLIO-1223

- [x] Are there enough tests

## How to test

You can run the function from a python shell in iaso environment. As it'as a tasks triggered by EB I don't see how you can try it on a local env.


kemar (approved):
LGTM

PR #859, opened 2023-10-12
Polio 1217 reasons for delay interface

quang-le:

Related JIRA tickets :  POLIO-1217

- [ ] Are there enough tests

## Changes

- Add dependency to django-translated-fields
- Add `ReasonForDelay`model:
- Add `reason_for_delay` ForeignKeyField to `RoundDatesHistoryEntry`
- Deprecate `reason`field.
- Add API endpoint to GET, POST and PATCH
- Replace `reason` with `reason_for_delay` in existing front-end code (`ReasonForDelayModal` mostly)
- Split Polio > Config menu entry in 2 sub menus: country and reasons for delay
- Add Reasons for delay page:
   - Create button: create new reason. `
       - Key` field is mandatory and has regex validation and an info popper to clarify constraints to users
       - English translation is mandatory and will be used as fallback when no translation for French is provided
   - Table view: edit button available. Same as Create, without the info popover. Key field is not editable 

## How to test

- Go to Polio > Configurations > Reasons for delay. Try:
    - Adding
    - Editing
    - Saving without FR translation
- Go to Polio > Campaigns. try to edit a round with the reason just added
- Also check the round dates history table

## Print screen / video

https://github.com/BLSQ/iaso/assets/38907762/ee07c37b-cdd5-4a0b-b407-378daf2ac4f0

## Notes

**IMPORTANT** PowerBI dashboards will need to be updated to use `reason_for_delay` i.o. `reason`. Other than that nothing should change

Reviewers: Please have a careful look at the migration as it was a bit tricky and there's prod data to migrate


plugins/polio/models.py
```diff
@@ -174,12 +176,31 @@ class RoundDateHistoryEntry(models.Model):
...
+    reason_for_delay = models.ForeignKey(
+        "ReasonForDelay", on_delete=models.PROTECT, null=True, blank=True, related_name="round_history_entries"
+    )
     round = models.ForeignKey("Round", on_delete=models.CASCADE, related_name="datelogs", null=True, blank=True)
     modified_by = models.ForeignKey("auth.User", on_delete=models.PROTECT, null=True, blank=True)
     created_at = models.DateTimeField(auto_now_add=True)
 
 
+class ReasonForDelay(SoftDeletableModel):
+    name = TranslatedField(models.CharField(_("name"), max_length=200), {"fr": {"blank": True, "null": True}})
```
kemar:
```suggestion
    name = TranslatedField(models.CharField(_("name"), max_length=200), {"fr": {"blank": True}})
```

It's better to avoid using [`null` on string-based fields](https://docs.djangoproject.com/en/4.2/ref/models/fields/#django.db.models.Field.null).

plugins/polio/models.py
```diff
@@ -174,12 +176,31 @@ class RoundDateHistoryEntry(models.Model):
...
+    key_name = models.CharField(null=True, blank=True, max_length=200, validators=[RegexValidator(r"^[A-Z_]+$")])
```
kemar: Do we really need this `key_name` field?
Why not just a unique `name` by account?
Unless I missed something, I think the code could be simpler without it.

Otherwise:

```suggestion
    key_name = models.CharField(blank=True, max_length=200, validators=[RegexValidator(r"^[A-Z_]+$")])
```

plugins/polio/migrations/0149_create_reasons_for_delay.py
```diff
@@ -0,0 +1,69 @@
...
+        "fr": "Intervalle de 4 semaines entre les rounds à respecter",
+    },
+    "OTHER_VACCINATION_CAMPAIGNS": {"en": "Other vaccination campaigns", "fr": "Autres campagnes de vaccination"},
+}
+
+
+def create_reasons_for_delay(apps, schema_editor):
+    ReasonForDelay = apps.get_model("polio", "ReasonForDelay")
+    text_choice_keys = [choice.name for choice in DelayReasons]
+    for choice_key in text_choice_keys:
```
kemar:
```suggestion
    for choice_key in DelayReasons.values:
```

This would give you the same result.

plugins/polio/models.py
```diff
@@ -174,12 +176,31 @@ class RoundDateHistoryEntry(models.Model):
     previous_ended_at = models.DateField(null=True, blank=True)
     started_at = models.DateField(null=True, blank=True)
     ended_at = models.DateField(null=True, blank=True)
+    # Deprecated.
```
kemar: Couldn't we delete the field then?

plugins/polio/admin.py
```diff
@@ -130,11 +131,16 @@ class VaccineAuthorizationsAdmin(admin.ModelAdmin):
     raw_id_fields = ("country",)
 
 
+class ReasonForDelayAdmin(admin.ModelAdmin):
```
kemar, line 135:
```suggestion
@admin.register(ReasonForDelay)
class ReasonForDelayAdmin(admin.ModelAdmin):
```

[This decorator](https://docs.djangoproject.com/en/4.2/ref/contrib/admin/#the-register-decorator) allows to keep things closer instead of putting everything at the end of the file:

~`admin.site.register(ReasonForDelay, ReasonForDelayAdmin)`~

plugins/polio/api/campaigns/campaigns.py
```diff
@@ -258,10 +258,19 @@ def update(self, instance: Campaign, validated_data):
             # we pop the campaign since we use the set afterward which will also remove the deleted one
             round_data.pop("campaign", None)
             scopes = round_data.pop("scopes", [])
+
+            # # Replace ReasonForDelay instance with key_name to avoid type error when calling is_valid
```
kemar:
```suggestion
            # Replace ReasonForDelay instance with key_name to avoid type error when calling is_valid
```

BTW, not sure to understand this comment. Let's talk about it.

plugins/polio/api/rounds/reasons_for_delay.py
```diff
@@ -0,0 +1,95 @@
...
+    def get_times_selected(self, reason_for_delay):
+        return len(list(reason_for_delay.round_history_entries.all()))
+
+    def validate(self, data):
+        key_name = data.get("key_name", None)
+        name_en = data.get("name_en", None)
+        if key_name is None:
+            raise serializers.ValidationError("requiredField")
+        if name_en is None:
+            raise serializers.ValidationError("requiredField")
```
kemar, line 58: Checking `name_en` could be redundant, since this its mandatory at the model level and DRF should validate it "out of the box".

You can try in a Python shell like that:

```python
reason_serializer = ReasonForDelaySerializer(data={"foo":1})
reason_serializer.is_valid()
reason_serializer.errors
```

plugins/polio/api/rounds/reasons_for_delay.py
```diff
@@ -0,0 +1,95 @@
...
+    times_selected = serializers.SerializerMethodField()
+
+    def get_times_selected(self, reason_for_delay):
+        return len(list(reason_for_delay.round_history_entries.all()))
+
+    def validate(self, data):
+        key_name = data.get("key_name", None)
+        name_en = data.get("name_en", None)
+        if key_name is None:
+            raise serializers.ValidationError("requiredField")
```
kemar, line 56: If `key_name` is required, why not making it required at the model level?

plugins/polio/api/rounds/reasons_for_delay.py
```diff
@@ -0,0 +1,95 @@
...
+        if name_en is None:
+            raise serializers.ValidationError("requiredField")
+        # Checking that updates are made from user from the same accont as the Reason for delay
+        if self.instance is not None:
+            user_account = self.context["request"].user.iaso_profile.account
+            if user_account != self.instance.account:
+                raise serializers.ValidationError("wrongUserAccount")
```
kemar: Checking perms/authorizations shouldn't be the responsability of the Serializer. Its purpose is to convert Python into JSON and vice versa.

This is better done in the ViewSet.

kemar (commented):
First comments after a preliminary reading. Let's talk about it tomorrow :)

plugins/polio/models.py
quang-le: Yeah, for 3 reasons:
- The powerBi dashboards rely on them, at least for the moment
- I need to filter out `INITIAL_DATA` in the front-end, and it seems more reliable to do it this way than based on name (which is editable) or `id`
- If someone changes `INITIAL_DATA` to "Yabadabadoo", we kinda get a hint what it was rezally supposed to be. this argument is not so strong, given the presence of the logs. Still, it kinda locks the original intent.

plugins/polio/models.py
quang-le: only after we're sure the powerBi dashboards have been updated accordingly

plugins/polio/api/campaigns/campaigns.py
quang-le: Because the serializer is nested, the data is transformed several times, and the `key_name` is replaced with `name_en` I think. In an earlier version I had the same problem with a PrimaryKeyRelatedField (I think), where the pk was replaced by the instance at the validation step. When I passed the data to the serializer, I got type error because it expected a pk an received the instance instead. Same general idea here.

plugins/polio/api/rounds/reasons_for_delay.py
quang-le: True, but I want to error message to be "requiredField" specifically, because I can use it in the front-end and display a translated message

plugins/polio/api/rounds/reasons_for_delay.py
quang-le: fair enough

plugins/polio/api/rounds/reasons_for_delay.py
quang-le: I think it was causing me headaches in migrations at some point. i should try it again

quang-le: Also, I couldn't think of a suitable default value

plugins/polio/api/rounds/reasons_for_delay.py
quang-le: It's also useless. I noticed, while writing tests, that trying to PATCH from the wrong account will result in a 404 before even reaching that code

plugins/polio/js/src/domains/Config/ReasonsForDelay/CreateEdit/CreateEditReasonForDelay.tsx
```diff
@@ -0,0 +1,182 @@
...
+                onClose={() => resetForm()}
+                confirmMessage={MESSAGES.confirm}
+                cancelMessage={MESSAGES.cancel}
+                allowConfirm={allowConfirm}
+                closeOnConfirm={false}
+            >
+                <Divider />
+                <Box mt={2}>
+                    {!id && (
+                        <InputWithInfos
```
beygorghor, line 120: instead of repeating this code we can add a prop hide to this component ?

beygorghor (approved):
Great, job, very clean !

PR #861, opened 2023-10-13
POLIO-1220: Make IASO work on chrome 56

beygorghor:
IASO calendar page should work on Samsung TV's using Tyzenbrowser 4.2.11030
WhatsApp Image 2023-09-29 at 15 54 26 (1)
On this version the web engine used is Chrome 56, according to their [doc](https://developer.samsung.com/smarttv/develop/specifications/web-engine-specifications.html).

I used the first version of Chromium 56 available [here](https://vikyd.github.io/download-chromium-history-version/#/).

Version [56.0.2885.0](https://commondatastorage.googleapis.com/chromium-browser-snapshots/index.html?prefix=Mac/424072/)

Related JIRA tickets :POLIO-1220

- [ ] Are there enough tests

## Changes

- adapt **dev** and **prod** Webpack configs to generate compatible js ( i had to include two packages from the node modules too, has they use ES2020 to compile their code)
- installed multiple polyfills (url-search-params-polyfill, abortcontroller-polyfill, polyfill-object.fromentries, array-flat-polyfill) to provide modern javascript to an old browser
- Rewrite useCurrentUser, useless in this case but much better
- added the `header` `credentials=same-origin` to all our requests, in order to use auth cookies on the requests we are doing on old browsers
- review a bit CSS on calendar page 

## How to test

- install chromium 56
- visit ` /dashboard/polio/embeddedCalendar/`

## Print screen / video
Capture d’écran 2023-10-13 à 09 56 02

Capture d’écran 2023-10-13 à 09 56 31


hat/assets/js/apps/Iaso/domains/app/index.js
```diff
@@ -19,6 +17,7 @@ import {
 } from '../../constants/routes';
 
 import ProtectedRoute from '../users/components/ProtectedRoute';
+import { useGetCurrentUser } from '../users/hooks/useGetCurrentUser.ts';
```
quang-le: I think it should be renamed. 2 hooks with the same name, that do pretty much the same thing will be confusing. I'm mostly thinking stupid wrong auto-imports.

hat/assets/js/apps/Iaso/domains/users/hooks/useGetCurrentUser.ts
```diff
@@ -0,0 +1,36 @@
...
+import { useDispatch } from 'react-redux';
+import { useSnackQuery } from '../../../libs/apiHooks';
+import { getRequest } from '../../../libs/Api';
+import { Profile } from '../../teams/types/profile';
+import { setCurrentUser } from '../actions';
+
+const getCurrentUser = (): Promise<Profile> => {
+    return getRequest('/api/profiles/me/');
+};
+export const useGetCurrentUser = (
```
quang-le: Re name change: it doesn't only get the user, it sets it in redux, so it's a bit misleading

hat/assets/js/apps/Iaso/libs/polyfills.ts
quang-le, line 1: so.many.polyfills

hat/assets/js/apps/Iaso/domains/users/hooks/useGetCurrentUser.ts
```diff
@@ -0,0 +1,36 @@
...
+    enabled: boolean,
+): UseQueryResult<Profile, Error> => {
+    const queryKey: any[] = ['currentUser'];
+    const dispatch = useDispatch();
+    // @ts-ignore
+    return useSnackQuery({
+        queryKey,
+        queryFn: () => getCurrentUser(),
+        dispatchOnError: false,
```
quang-le: Why not dispatch on Error for this one? I don't think we have any front-end that manages this error

hat/assets/js/apps/Iaso/libs/Api.js
```diff
@@ -37,7 +37,10 @@ export const iasoFetch = async (resource, init = {}) => {
     const url = resource.url ?? resource;
     const method = init?.method ?? 'GET';
     try {
-        response = await fetch(resource, init);
+        response = await fetch(resource, {
+            ...init,
+            credentials: 'same-origin',
```
quang-le, line 42: I remember that one

quang-le (approved):
Well done! What an ungrateful task that was.

A few non-blocking comments about the code.

The feature works well on my machine.

I think at minimum we should deploy on staging early and go through all of iaso, not just the one page we need to display on the TV.

I'm not comfortable with the amount of polyfills and babel plugins this introduces for such a edge case. Maybe we can consider making a separate build for iaso-TV and dedicate a server (if it makes sense cost wise). The impact on the iaso build and setup seems disproportionate considering the intended use in my opinion.
But that's a wider discussion that shouldn't block the PR going forward at this stage

PR #868, opened 2023-10-16
IA-2425 Add a `ThreeDimPointField`

kemar:
Proposition for a new `ThreeDimPointField` serializer field.

It should be reviewed after #864.

## Screenshot

point


kemar:
We decided to leave everything flat.

PR #887, opened 2023-10-20
PoC to parse accuracy in the same object as location

bmonjoie:
This PR offers a simple and generic way to parse `new_accuracy` within the `new_location` object while still store them in different attributes of the model class.

Related JIRA tickets : IA-2425

- [X] Are there enough tests

### DIFF
--- a/iaso/api/org_unit_change_requests/serializers.py
+++ b/iaso/api/org_unit_change_requests/serializers.py
@@ -1,11 +1,66 @@
 from rest_framework import serializers
 
 from iaso.models import OrgUnitChangeRequest
+from collections import OrderedDict
+
+from django.contrib.gis.geos import GEOSGeometry
+from django.contrib.gis.geos.error import GEOSException
+from django.utils.translation import gettext_lazy as _
+
+
+class PointFieldWithAccuracy(serializers.Serializer):
+    """
+    Expected input:
+    {
+        "latitude": 49.8782482189424,
+        "longitude": 24.452545489,
+        "altitude": 5.1,
+        "accuracy": 1.2
+    }
+    """
+
+    default_error_messages = {
+        "invalid": _("Enter a valid location."),
+    }
+
+    point_field: str = "new_location"
+    accuracy_field: str = "new_accuracy"
+
+    def __init(self, point_field="new_location", accuracy_field="new_accuracy"):
+        self.point_field = point_field
+        self.accuracy_field = accuracy_field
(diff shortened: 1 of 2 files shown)


iaso/api/org_unit_change_requests/serializers.py
```diff
@@ -1,11 +1,68 @@
...
+            }
+        return None
 
 
 class MobileOrgUnitChangeRequestListSerializer(serializers.ModelSerializer):
-    org_unit_id = serializers.IntegerField(source="org_unit.id")
-    org_unit_uuid = serializers.UUIDField(source="org_unit.uuid")
+    org_unit_id = serializers.IntegerField(source="org_unit.id", allow_null=True)
+    org_unit_uuid = serializers.UUIDField(source="org_unit.uuid", allow_null=True)
+    new_location = PointFieldWithAccuracy(source="*")
```
kemar, line 65: [This works yes](https://www.django-rest-framework.org/api-guide/fields/#using-source), but it makes it **implicit** that `new_accuracy ` will be populated too.

Someone reading the code should be capable of understanding what it does as fast as possible.

bmonjoie: I can turn it around, and say that someone reading the API must be able to understand as fast as possible which fields are which. Also, the current implementation makes it very easy not to provide `new_accuracy` when `new_location` is provided. Yet, they should all be provided or none. The solution I'm offering allows us, very easily, to force `new_accuracy` when `new_location` is provided.

Finally, if you are really scared someone will miss that `new_accuracy` is updated alongside the `new_location` property, we can add a comment above :wink:

kemar: Well, I guess I'm out of arguments… It wouldn't even be worth it 😄

iaso/api/org_unit_change_requests/serializers.py
```diff
@@ -1,11 +1,68 @@
...
+    """
+
+    default_error_messages = {
+        "invalid": _("Enter a valid location."),
+    }
+
+    point_field: str = "new_location"
+    accuracy_field: str = "new_accuracy"
+
+    def __init(self, point_field="new_location", accuracy_field="new_accuracy"):
```
kemar, line 29: Making `point_field` and `accuracy_field` mandatory would probably make it more explicit.

kemar (approved):

PR #889, opened 2023-10-20
IA-2487: Provide groups in `/api/mobile/orgUnits`

bmonjoie:
The mobile application needs to know to which groups the OrgUnits are linked.
This adds the information to the `/api/mobile/orgUnits` endpoint.

Related JIRA tickets: IA-2487

- [X] Are there enough tests

## How to test

- Go to `/api/mobile/orgUnits/?app_id=...` with a proper app_id and sufficient permissions. *Advice: use `&page=1&limit=10` to load the page faster*
- See that groups are empty for all the OrgUnit
- Add a group and attach it to some OrgUnit
- Go back to the endpoint and see that the array is populated with the right id

## Notes

At first, I had reused the query from [groups.py#L65](/BLSQ/iaso/tree/main/iaso/api/mobile/groups.py#L65) but I noticed it wasn't necessary as there is a constraint forcing the OrgUnit to be attached to a group with the same `source_version`.

### DIFF
--- a/iaso/api/mobile/groups.py
+++ b/iaso/api/mobile/groups.py
@@ -55,11 +55,7 @@ class MobileGroupsViewSet(ListModelMixin, GenericViewSet):
         return super().list(request, *args, **kwargs)
 
     def get_queryset(self) -> QuerySet:
-        app_id_serializer = AppIdSerializer(data=self.request.query_params)
-        app_id_serializer.is_valid(raise_exception=True)
-        self.app_id_param = app_id_serializer.data[APP_ID]
-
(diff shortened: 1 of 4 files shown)


kemar:
> but I noticed it wasn't necessary as there is a constraint forcing the OrgUnit to be attached to a group with the same source_version

Sounds about right:

- https://github.com/BLSQ/iaso/blob/9cc04848ddd1bf48de121cba33a56d450d125bb5/iaso/models/base.py#L402-L404
- https://github.com/BLSQ/iaso/blob/main/iaso/migrations/0086_add_version_constraints.py

kemar (approved):
LGTM

PR #891, opened 2023-10-23
POLIO-1236 add automatic check on the quantity entered in campaigns per country vs quantity authorized

SimonDus:
Check if the quantity of requested doses in Risk Assessment campaign is inferior or equal to the authorized doses in vaccine authorization for the country. When creating a campaign, if there is an authorization the check operate. If not, a mail is sent to warn the NOPV2 Team that a campaign without authorization has been created. 

When updating a campaign the check also operate. 

There was also a bug if no NOPV2 team was created, when sending the mail it was making the campaign creation crash. ( Wont happen in prod but we never know... )

Related JIRA tickets : POLIO-1236

- [x] Are there enough tests

## How to test

You need to have a team named " nOPV2 vaccine authorization alerts " in order to carry the test.
First scenario : 
Create a campaign without vaccine authorization approved, add a round and in risk assessement add a quantity for this round. You should be able to create the campaign. 
A mail should be sent to warn that a campaign with vaccine haut has been created.
Then create  a vaccine authorization, add X quantity. Go back to the campaign, add a round and test the quantity. If the quantity is superior to the vacc auth, an error should be triggered. 

I also suggest you to test all possibilities of the campaign as it touch directly the campaign API code. 

Second Scenario: 

Create an approved vaccine auth first with X quantity. Create a campaign, add a round and add requested doses superior to the vacc auth qty. An error should be triggered. 

https://github.com/BLSQ/iaso/assets/45785235/9057f9c0-337b-4750-814b-d4d61112f822


plugins/polio/api/campaigns/campaigns.py
```diff
@@ -153,6 +175,29 @@ def get_has_data_in_budget_tool(self, campaign):
...
+        if initial_org_unit:
+            for d in rounds:
+                round_count = d["number"]
+                for element in d["vaccines"]:
+                    if element.get("name") == "nOPV2":
+                        nOPV2_rounds.append(round_count)
+                        nopv2_vacc_found = True
+
+            if nopv2_vacc_found:
+                initial_org_unit = OrgUnit.objects.get(pk=initial_org_unit.pk)
```
quang-le: shouldn't we handle the case where `get` doesn't find the org unit?

plugins/polio/api/campaigns/campaigns.py
```diff
@@ -153,6 +175,29 @@ def get_has_data_in_budget_tool(self, campaign):
...
+                vaccine_authorization = VaccineAuthorization.objects.filter(
+                    country=initial_org_unit, status="VALIDATED"
+                )
+                if vaccine_authorization:
+                    check_total_doses_requested(vaccine_authorization[0], rounds, nOPV2_rounds)
```
quang-le: Are we sure they come ordered by descending expiration date? It seems to me there's a risk we'll get the last created or last updated, which could be historical data and not the real latest authorization

plugins/polio/api/campaigns/campaigns.py
```diff
@@ -218,8 +263,29 @@ def create(self, validated_data):
     @atomic
     def update(self, instance: Campaign, validated_data):
         old_campaign_dump = serialize_campaign(instance)
-        rounds = validated_data.pop("rounds", [])
         campaign_scopes = validated_data.pop("scopes", [])
+        rounds = validated_data.pop("rounds", [])
+        initial_org_unit = validated_data.get("initial_org_unit")
```
quang-le, line 289: Shouldn't we check that it matches `self.initial_org_unit`?

plugins/polio/api/campaigns/campaigns.py
```diff
@@ -218,8 +263,29 @@ def create(self, validated_data):
...
+        if initial_org_unit:
+            for d in rounds:
+                round_count = d["number"]
+                for element in d["vaccines"]:
+                    if element.get("name") == "nOPV2":
+                        nOPV2_rounds.append(round_count)
+                        nopv2_vacc_found = True
+
+            if nopv2_vacc_found:
+                initial_org_unit = OrgUnit.objects.get(pk=initial_org_unit.pk)
```
quang-le: same comment as for `create`

plugins/polio/api/campaigns/campaigns.py
```diff
@@ -218,8 +263,29 @@ def create(self, validated_data):
...
+                vaccine_authorization = VaccineAuthorization.objects.filter(
+                    country=initial_org_unit, status="VALIDATED"
+                )
+                if vaccine_authorization:
+                    check_total_doses_requested(vaccine_authorization[0], rounds, nOPV2_rounds)
```
quang-le: same as above

plugins/polio/api/campaigns/vaccine_authorization_missing_email.py
```diff
@@ -0,0 +1,29 @@
...
+        message = f"""
+        Dear team,
+
+        Please note that the campaign {obr_name} for {org_unit} has been created into the poliooutbreaks platform. 
+
+        Be aware that {org_unit} does not have any nOPV2 authorization recorded into the platform yet.
+
+        To add one, please follow this link: https://www.poliooutbreaks.com/dashboard/polio/vaccinemodule/nopv2authorisation/accountId/1/order/-current_expiration_date/pageSize/20/page/1
+
+        This is an automated message from the poliooutbreaks platform.
```
quang-le: I would change the wording a bit:

Dear team,

The campaign {obr_name} for {org_unit} has been created in the poliooutbreaks platform.

Please note that no nOPV2 authorization has been recorded in the platform for  {org_unit} yet.

etc

plugins/polio/api/campaigns/vaccine_authorization_missing_email.py
```diff
@@ -0,0 +1,29 @@
...
+        subject = f"Vaccine Authorization missing for campaign OBR Name {obr_name}"
+        from_email = (settings.DEFAULT_FROM_EMAIL,)
+        message = f"""
+        Dear team,
+
+        Please note that the campaign {obr_name} for {org_unit} has been created into the poliooutbreaks platform. 
+
+        Be aware that {org_unit} does not have any nOPV2 authorization recorded into the platform yet.
+
+        To add one, please follow this link: https://www.poliooutbreaks.com/dashboard/polio/vaccinemodule/nopv2authorisation/accountId/1/order/-current_expiration_date/pageSize/20/page/1
```
quang-le: You can just use : https://www.poliooutbreaks.com/dashboard/polio/vaccinemodule/nopv2authorisation/accountId/1/

Also: I wouldn't hard code the account. I would get it from the serializer and pass it as arg

plugins/polio/api/campaigns/vaccine_authorization_missing_email.py
```diff
@@ -0,0 +1,29 @@
...
+
+        This is an automated message from the poliooutbreaks platform.
+            """
+
+        send_mail(subject, message, from_email, recipient_list)
+    except ObjectDoesNotExist:
```
quang-le, line 30: Shouldn't we log that somewhere or have someone notified internally? 
If I understand correctly, if there's an errror with the sending of the email, it fails silently.

plugins/polio/api/campaigns/vaccine_authorization_missing_email.py
```diff
@@ -0,0 +1,29 @@
...
+from django.core.mail import send_mail
+
+from hat import settings
+from iaso.models import Team
+from plugins.polio.settings import NOPV2_VACCINE_TEAM_NAME
+
+
+def missing_vaccine_authorization_for_campaign_email_alert(obr_name, org_unit):
+    try:
+        users = [user for user in Team.objects.get(name=NOPV2_VACCINE_TEAM_NAME).users.all()]
```
quang-le: This will work because of the polio setup, but normally we should  to filter by account

plugins/polio/api/vaccines/vaccine_authorization.py
```diff
@@ -17,6 +17,15 @@
 from plugins.polio.settings import COUNTRY
 
 
+def check_for_already_validated_authorization(status, country):
+    if status == "VALIDATED":
+        validated_vaccine_auth = VaccineAuthorization.objects.filter(status="VALIDATED", country=country)
```
quang-le: filter by account

plugins/polio/api/campaigns/campaigns.py
quang-le: Also, shouldn't we match the round dates against the authorization dates to make sure we get the right one?

plugins/polio/api/campaigns/campaigns.py
```diff
@@ -153,6 +175,29 @@ def get_has_data_in_budget_tool(self, campaign):
...
+                round_count = d["number"]
+                for element in d["vaccines"]:
+                    if element.get("name") == "nOPV2":
+                        nOPV2_rounds.append(round_count)
+                        nopv2_vacc_found = True
+
+            if nopv2_vacc_found:
+                initial_org_unit = OrgUnit.objects.get(pk=initial_org_unit.pk)
+                vaccine_authorization = VaccineAuthorization.objects.filter(
+                    country=initial_org_unit, status="VALIDATED"
```
quang-le: I would filter that by account as well

quang-le (changes requested):
If i start encoding requested doses before I encode the vaccine authorization, the already encoded doses are not taken into account:

https://github.com/BLSQ/iaso/assets/38907762/c684cc6e-e22d-4bd0-aa7c-947c82b9c967


The normal case also doesn't seem to work:

Uploading Screen Recording 2023-11-03 at 13.03.40.mov…

plugins/polio/api/campaigns/vaccine_authorization_missing_email.py
quang-le:
```suggestion
        users = [user for user in Team.objects.get(name=NOPV2_VACCINE_TEAM_NAME, account=account).users.all()]
```

plugins/polio/api/campaigns/campaigns.py
```diff
@@ -263,6 +237,29 @@ def create(self, validated_data):
         campaign.update_geojson_field()
         campaign.save()
         log_campaign_modification(campaign, None, self.context["request"].user)
+
+        # check if the quantity of the vaccines requested is not superior to the authorized vaccine quantity
+        c_rounds = [r for r in campaign.rounds.all()]
+        nOPV2_rounds = []
+        for r in c_rounds:
+            if r.vaccine_names() == "nOPV2":
```
quang-le: So we don't check if there's multiple vaccines for the round? (Seems ok to me, just checking to confirm)

plugins/polio/api/campaigns/campaigns.py
```diff
@@ -263,6 +237,29 @@ def create(self, validated_data):
...
+                nOPV2_rounds.append(r)
+
+        if initial_org_unit and len(nOPV2_rounds) > 0:
+            try:
+                initial_org_unit = OrgUnit.objects.get(pk=initial_org_unit.pk)
+                vaccine_authorization = VaccineAuthorization.objects.filter(
+                    country=initial_org_unit, status="VALIDATED", account=account
+                )
+                if vaccine_authorization:
```
quang-le, line 273: Please add comment explaining why it's safe to use `vaccine_authorization[0]`

quang-le (approved):

PR #892, opened 2023-10-24
IA-2492:  display default permission error or back-end error on 403

beygorghor:
While having a permission error on save, we should display better error 

Related JIRA tickets : IA-2492

- [ ] Are there enough tests

## Changes

Adapt `useSnackMutation` to display custom error message while having a 403

## How to test

Try to save something on iaso while you don't the permission to do it.

I the backend provide a messge you should see it in the snackbar, otherwise you should see a default message

## Print screen / video

https://github.com/BLSQ/iaso/assets/12494624/4a8624a8-98d3-4a0c-8659-2d21e91b768c


hat/assets/js/apps/Iaso/libs/apiHooks.ts
```diff
@@ -14,12 +14,17 @@ import {
     QueryFunction,
 } from 'react-query';
 import { defineMessages } from 'react-intl';
-import { IntlMessage } from 'bluesquare-components';
+import { IntlMessage, useSafeIntl } from 'bluesquare-components';
 import { isArray } from 'lodash';
 import { enqueueSnackbar } from '../redux/snackBarsReducer';
 import { errorSnackBar, succesfullSnackBar } from '../constants/snackBars';
 
 const MESSAGES = defineMessages({
```
hakifran, line 22: It's not blocking but if we could manage MESSAGES in separate file. But this can be for the future refactoring

hakifran (approved):

PR #906, opened 2023-10-31
POLIO-1289 Edit content of automatic emails on Fridays

SimonDus:
Change of the footer of the weekly mail in polio. Nothing complicated and nothing to test. Just check that the translation are corrects. 

Related JIRA tickets : POLIO-1289

- [x] Are there enough tests


quang-le (commented):
I would put Carolyne's family name as well

quang-le (changes requested):
I would put Carolyne's surname as well

quang-le (approved):

PR #908, opened 2023-10-31
POLIO-1277 Add GPEI Coordinator to mailing list of vacc auth emails.

SimonDus:
**CAN BE HOTFIXED**
Add the GPEI coordinator per country to the vaccine authorization mails. Before it was only sent to the NOPV2 team. 

Related JIRA tickets : POLIO-1277

- [x] Are there enough tests

## Changes

The GPEI coordinator is added to the mailing list for the vaccine mail alert. We retrieve the country of the vaccine_auth then search the coordinator ( username starts with GPEI ) then check if the org_unit of the profile match with the country. 

## How to test

Kinda tricky to test as the mailing list is build when the task is called. So I copied the code that build the mailing list and test it in unit test. I don't really see another way to test it.

### DIFF
--- a/plugins/polio/tasks/vaccine_authorizations_mail_alerts.py
+++ b/plugins/polio/tasks/vaccine_authorizations_mail_alerts.py
@@ -4,12 +4,13 @@ from datetime import timedelta
 
 from django.contrib.auth.models import User
 from django.contrib.sites.models import Site
+from django.core.exceptions import ObjectDoesNotExist
 from django.core.mail import send_mail
 from django.shortcuts import get_object_or_404
 
 from beanstalk_worker import task_decorator
 from hat import settings
-from iaso.models import Team
+from iaso.models import Team, Profile
 from plugins.polio.models import VaccineAuthorization
 from logging import getLogger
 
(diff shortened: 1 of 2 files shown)


plugins/polio/tasks/vaccine_authorizations_mail_alerts.py
```diff
@@ -83,6 +86,11 @@ def send_email_vaccine_authorizations_60_days_expiration_alert(task=None):
     email_sent = 0
 
     for i, vacc_auth in enumerate(vaccine_auths):
+        try:
+            gpei_coordinator = Profile.objects.get(user__username__istartswith="gpei", org_units=vacc_auth.country)
+            mailing_list.append(gpei_coordinator.user.email)
```
quang-le, line 91: There is a GPEI coordinator field in the campaign object. It seems more reliable to me that filtering by user name. Or we could add a field to set the GPEI coordinator in the polio country config

quang-le (approved):
Looks OK for now, but we shouldn't rely on user name in the long term

Related ticket https://bluesquare.atlassian.net/browse/POLIO-1295

quang-le (approved):
OK for now, but we shouldn't rely on the name in the long run.; Related ticket https://bluesquare.atlassian.net/browse/POLIO-1295

PR #910, opened 2023-11-02
Add vaccine field in the scope export

madewulf:
Optimizing requests a little, too.

We were missing a vaccine field in the export of the scope

- [x ] Did I use eslint and black formatters
- [ x] Are there enough tests

## Changes

## How to test
access this url with the id of a round 
http://localhost:8081/api/polio/campaigns/csv_campaign_scopes_export/?&round=48

### DIFF
--- a/plugins/polio/api/campaigns/campaigns.py
+++ b/plugins/polio/api/campaigns/campaigns.py
@@ -960,17 +960,26 @@ class CampaignViewSet(ModelViewSet, CSVExportMixin):
         round = Round.objects.get(pk=request.GET.get("round"))
         campaign = round.campaign
         org_units_list = []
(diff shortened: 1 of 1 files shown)


plugins/polio/api/campaigns/campaigns.py
```diff
@@ -960,17 +960,26 @@ def csv_campaign_scopes_export(self, request, **kwargs):
         round = Round.objects.get(pk=request.GET.get("round"))
         campaign = round.campaign
         org_units_list = []
-        org_units = campaign.get_districts_for_round(round)
```
quang-le, line 963: I wonder if it's worth keeping these methods in the model since they assume a lot of stuff has already been prefetched

quang-le (approved):

PR #928, opened 2023-11-21
POLIO-1309 [Line-list] Polio notifications API

kemar:
API for Polio notifications.

Related JIRA tickets : [POLIO-1309](https://bluesquare.atlassian.net/browse/POLIO-1309)

## Self proofreading checklist

- [x] Did I use eslint and black formatters
- [x] Is my code clear enough and well documented
- [x] My migrations file are included
- [x] Are there enough tests
- [x] Documentation has been included (for new feature)

## Changes

- API for CRUD on `Notification`s
- additional route to download a sample of the expected XSLX format
- additional route to upload and import an XSLX file:
    - a `NotificationImport` instance is created
    - then an async task is triggered

## How to test

Building on #904, you can test the API via the DRF UI:

- `http://localhost:8081/api/polio/notifications/`

You can filter results, e.g.:

- `http://localhost:8081/api/polio/notifications/?vdpv_category=cvdpv2&source=accute_flaccid_paralysis&country=1245455&date_of_onset_after=2019-04-05&date_of_onset_before=2019-04-06`

## Print screen

api

---

api_filters

---

api_actions

[POLIO-1309]: https://bluesquare.atlassian.net/browse/POLIO-1309?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #941, opened 2023-11-29
POLIO-1276 [Line-list] Polio notifications front-end

kemar:
Front-end for Polio notifications.

Related JIRA tickets : [POLIO-1309](https://bluesquare.atlassian.net/browse/POLIO-1276)

## Changes

Front-end for [Polio notifications API](https://github.com/BLSQ/iaso/pull/928).

New UIs for notifications:

- CRUD
- bulk update

## Related PR (`bluesquare-components`)

https://github.com/BLSQ/bluesquare-components/pull/135/files

## How to test

- ensure to be admin or to have the right perms (`POLIO_NOTIFICATIONS`)
- create notifications:
    1. either manually
    1. or by importing an .xlsx file ([sample available here](https://bluesquare.atlassian.net/browse/POLIO-1275?focusedCommentId=59457))
        - the backend's import is async, so a worker must be running
        - `docker-compose run iaso manage tasks_worker`
- edit notifications
- delete notifications

## Print screen / video

https://github.com/BLSQ/iaso/assets/281139/91b6b929-dd58-4d8e-af3e-604dfc71e5bc

[POLIO-1309]: https://bluesquare.atlassian.net/browse/POLIO-1309?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


plugins/polio/js/src/domains/Notifications/Modals/NotificationsBulkImportModal.tsx
```diff
@@ -0,0 +1,151 @@
...
+        const formikError = errors.file;
+        const apiError = apiErrors.error;
+        if (formikError) {
+            result.push(formikError);
+        }
+        if (apiError) {
+            result.push(apiError);
+        }
+        return result;
+    }, [apiErrors.error, errors.file]);
```
quang-le, line 84: It's nice but we won't have the errors translated, if I'm not mistaken

plugins/polio/js/src/domains/Notifications/Filters/NotificationsFilters.tsx
```diff
@@ -0,0 +1,96 @@
...
+    params,
+    dropdownContent,
+}) => {
+    const { filters, handleSearch, handleChange, filtersUpdated } =
+        useFilterState({ baseUrl, params });
+    const { formatMessage } = useSafeIntl();
+
+    return (
+        <Grid container spacing={2}>
+            <Grid item xs={12} lg={3}>
```
quang-le: Please put a value for `md`as well: until we deploy MUI5, the breakpoint for `lg` is 1260, so full width would be too big for screens with 1200 width

plugins/polio/js/src/domains/Notifications/hooks/api.ts
```diff
@@ -0,0 +1,119 @@
...
+        options: {
+            staleTime: 1000 * 60 * 15, // in ms
+            cacheTime: 1000 * 60 * 5,
+            select: data => {
+                const metadata = data.actions.POST;
+                const mapChoices = choices =>
+                    choices.map(choice => ({
+                        label: choice.display_name,
+                        value: choice.value,
+                    }));
```
quang-le, line 77: If we're going to start using OPTION requests regularly, this function would fit well in a `utils` file somewhere in (vanilla) Iaso, like `/iaso/hat/assets/js/apps/Iaso/libs/utils.tsx`

plugins/polio/js/src/domains/Notifications/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+            {!isFetchingDropdownContent && (
+                <>
+                    <TopBar
+                        title={formatMessage(MESSAGES.notificationsTitle)}
+                        displayBackButton={false}
+                    />
+                    <Box className={classes.containerFullHeightNoTabPadded}>
+                        <NotificationsFilters
+                            params={paramsNew}
+                            dropdownContent={dropdownContent}
```
quang-le: You've got a TS error here

plugins/polio/js/src/domains/Notifications/Table/NotificationsTable.tsx
```diff
@@ -0,0 +1,137 @@
...
+import { EditNotificationModal } from '../Modals/NotificationsCreateEditModal';
+import { NOTIFICATIONS_BASE_URL } from '../../../constants/routes';
+import { useGetNotifications } from '../hooks/api';
+import {
+    ApiNotificationsParams,
+    DropdownsContent,
+    NotificationsParams,
+} from '../types';
+
+type Props = { params: NotificationsParams; dropdownContent: DropdownsContent };
```
quang-le: I think you need to either make `dropDownContent`optional: 
```suggestion
type Props = { params: NotificationsParams; dropdownContent?: DropdownsContent };
```

Or cast the type when you pass the props (see suggestion in index.tsx)

plugins/polio/js/src/domains/Notifications/index.tsx
quang-le:
```suggestion
                            dropdownContent={dropdownContent as DropdownsContent}
```

Not blocking but the TS compiler is not smart enough to realize you won't pass the prop if `dropdownContent` is `undefined`

plugins/polio/js/src/domains/Notifications/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+                        />
+                        <NotificationsTable
+                            params={paramsNew}
+                            dropdownContent={dropdownContent}
```
quang-le: same stuff about the TS compiler

plugins/polio/js/src/domains/Notifications/Table/NotificationsTable.tsx
```diff
@@ -0,0 +1,137 @@
...
+}) => {
+    const apiParams: ApiNotificationsParams = {
+        ...params,
+        limit: params.pageSize || '20',
+        order: params.order || 'id',
+        page: params.page || '1',
+    };
+    const { data, isFetching } = useGetNotifications(apiParams);
+    const { formatMessage } = useSafeIntl();
+    const columns: Array<Column> = [
```
quang-le: Not blocking but we usually extract that in a hook to void cluttering the table component

plugins/polio/js/src/domains/Notifications/hooks/validation.ts
```diff
@@ -0,0 +1,57 @@
...
+import { useSafeIntl } from 'bluesquare-components';
+
+import { ValidationError } from '../../../../../../../hat/assets/js/apps/Iaso/types/utils';
+import { useAPIErrorValidator } from '../../../../../../../hat/assets/js/apps/Iaso/libs/validation';
+
+import MESSAGES from '../messages';
+
+export const useNotificationSchema = () => {
+    const { formatMessage } = useSafeIntl();
+    return yup.object().shape({
```
quang-le, line 15: It seems the check on the org unit is missing here

plugins/polio/js/src/domains/Notifications/Modals/NotificationsBulkImportModal.tsx
```diff
@@ -0,0 +1,151 @@
...
+        onSuccess: () => closeDialog(),
+    });
+
+    const validationSchema = useBulkImportNotificationSchema(
+        apiErrors,
+        payload,
+    );
+
+    const formik = useFormik<XlsxFile>({
+        initialValues: { file: undefined },
```
quang-le, line 52: TS error here

plugins/polio/js/src/domains/Notifications/Modals/NotificationsBulkImportModal.tsx
```diff
@@ -0,0 +1,151 @@
...
+    };
+
+    const accept: Accept = {
+        'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet': [
+            '.xslx',
+        ],
+    };
+
+    return (
+        <>
```
quang-le: The fragment is not needed here since `SimpleModal` is the only "top" component

plugins/polio/js/src/domains/Notifications/Modals/NotificationsBulkImportModal.tsx
```diff
@@ -0,0 +1,151 @@
...
+            <SimpleModal
```
quang-le: `dataTestId` is missing

quang-le (changes requested):
Very nice job. A few remarks, but it's all details really. 
The only blocking changes are:
- Layout changes
    - Action buttons should be inline in the table
    - Column selector should be "button" i.o " icon (but only the action buttons are blocking)
linelist_review

- Filters should be split in 2 lines and `DateRange` size adapted to match other inpout components':

linelist_review2

- Also on the pic above: I've this weird bug when creating an entry manually, where the country list stays empty

Not blocking, but should be put in separate tickets if not dealt with in this PR:

- Validation: Confirm button should be disabled if user didn't pick an org unit
    
line list review 3

- There's no column sort on country:
- trying to drop several files will enable the confirm button:

Uploading Screen Recording 2023-12-13 at 17.31.08.mov…



- File restriction:
   - restricting to xlsx files doesn't seem to work front-end side (I did use your branch of bluesquare-components)

plugins/polio/js/src/domains/Notifications/Modals/NotificationsBulkImportModal.tsx
kemar: What is the purpose of `dataTestId`?

plugins/polio/js/src/domains/Notifications/Modals/NotificationsBulkImportModal.tsx
kemar: TBH, that's a copy and paste from another piece of code. What are the consequences and what should I do here?

kemar:
> Action buttons should be inline in the table

What do you mean exactly?

> Column selector should be "button" i.o " icon

Do you mean: use `columnSelectorButtonType="button"`?

> Also on the pic above: I've this weird bug when creating an entry manually, where the country list stays empty

That's correct in fact :) Here's how it works:

- I didn't want to display all available countries in the menu in the filter
- because it wouldn't make sense to filter on "unused" countries
- so I'm only fetching countries used in notifications
- since you only have "one" notification without any country, the filter is empty

This shouldn't be the case for our customer because a lot of data will be preloaded.

But I can change the behaviour if it leads to thinking that the UI is buggy.

> There's no column sort on country

- nope because I'm annotating the queryset on the basis of the org unit chain (e.g. `org_unit__parent__parent__name`)
- so filtering on country, province or district would be a bit more work
- I suggest handling it in another PR if necessary
- meanwhile, the "country" filter can do the job

> Trying to drop several files will enable the confirm button

How can I fix that?

> restricting to xlsx files doesn't seem to work front-end side (I did use your branch of bluesquare-components)

- Strange, because it works for me…
- there is a backend check on the file format anyway

plugins/polio/js/src/domains/Notifications/Modals/NotificationsBulkImportModal.tsx
quang-le: It's to make the component easy to target in cypress testing

beygorghor (approved):

PR #942, opened 2023-11-30
WC2-305: beneficiaries fixes

beygorghor:
- location of a beneficiary is not present on the dashboard, we have the org unit but no geo location of the instance
- multi duplicates links in beneficiaries list is not working
- launch a search in beneficiaries list remove the selected beneficiary type
- beneficiaries default order should be recent first

Related JIRA tickets :WC2-306

- [ ] Are there enough tests

## Changes

- display submissions locations not only org units
- allow to switch between org unit locations or submissions locations
- fix routing issues
- fetch all the entities locations to display it on the map
- improve a bit entities api

## How to test

- Make sure you activate SHOW_BENEFICIARY_TYPES_IN_LIST_MENU account feature flag
- Make also sure you have entities with locations and or org unit with locations
- open entities for a specific beneficiary type
- try the map tab, you should have all the possible locations and switch between submissions or org unit locations
- switch from one type to another and make sure that search is relevant

## Print screen / video

https://github.com/BLSQ/iaso/assets/12494624/e0ac1acf-66e3-423c-8550-4344d074f172

## Notes

Call to entities API to fetch locations is very slow, i made a dedicated ticket for that [WC2-307](https://bluesquare.atlassian.net/browse/WC2-307)

[WC2-307]: https://bluesquare.atlassian.net/browse/WC2-307?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hat/assets/js/apps/Iaso/domains/entities/components/Filters.tsx
```diff
@@ -67,6 +68,18 @@ const Filters: FunctionComponent<Props> = ({ params, isFetching }) => {
...
+        setFilters({
+            search: params.search,
+            location: params.location,
+            dateFrom: params.dateFrom,
+            dateTo: params.dateTo,
+            submitterId: params.submitterId,
+            submitterTeamId: params.submitterTeamId,
+            entityTypeIds: params.entityTypeIds,
+        });
+    }, [params]);
```
quang-le, line 82: We have `useFilterState` for this

hat/assets/js/apps/Iaso/domains/entities/components/LocationSwitch.tsx
```diff
@@ -0,0 +1,62 @@
...
+export type Legend = {
+    value: string;
+    label: string;
+    color: string; // has to be an hexa color
+};
+
+type Props = {
+    displayedLocation: DisplayedLocation;
+    setDisplayedLocation: Dispatch<SetStateAction<DisplayedLocation>>;
+};
```
quang-le: I think if we pass it an array of `possibleValues` and map it, it would be reusable. Then we could keep it in /components/maps/

iaso/api/entity.py
```diff
@@ -347,91 +347,54 @@ def list(self, request: Request, *args, **kwargs):
             total_count = entities.count()
             num_pages = math.ceil(total_count / limit_int)
             entities = entities[start_int:end_int]
-            results_count = entities.count()
-
-        if entity_type_ids is None or (entity_type_ids is not None and len(entity_type_ids.split(",")) > 1):
```
quang-le, line 352: Why is it safe to remove this check?

quang-le (commented):
code looks OK, couldn't test all the features because I have a lousy DB. 

Just a question about a backend change

iaso/api/entity.py
beygorghor: this check is still done at line 384

hat/assets/js/apps/Iaso/domains/entities/components/Filters.tsx
beygorghor: yes, but has this is working i will not rewrite the it i think

quang-le (approved):

PR #953, opened 2023-12-04
TRIA-37: Use Iaso JsonDataStore to add "extra fields" to OrgUnit

bramj:
This PR is needed for the Trypelim plugin. 
Context: Geo structures in Trypelim are very similar to Iaso org units (with a name and location as a point or geometric shape), but they also have extra data such as `contact_emails`, `population`, etc.

Related JIRA tickets : https://bluesquare.atlassian.net/browse/TRIA-37

- [x] Are there enough tests

## Changes

To avoid adding new fields on org unit, I store this extra data in the `JsonDataStore`. For this to work I've added:

- A foreign key to link `JsonDataStore` to `OrgUnit`
- Some helper methods on the `OrgUnit` model to facilitate getting and setting these extra fields


iaso/models/org_unit.py
```diff
@@ -548,6 +549,61 @@ def source_path(self):
...
+        return self.get_extra_fields().get("population_year", None)
+
+    @property
+    def population_source(self):
+        return self.get_extra_fields().get("population_source", None)
+
+    @property
+    def merged_to(self):
+        return self.get_extra_fields().get("merged_to", None)
+
```
bramj: I'm still a bit thorn on doing this. It's really nice for me, but it does pollute org_unit with Trypelim-specific data.

iaso/models/org_unit.py
```diff
@@ -548,6 +549,37 @@ def source_path(self):
...
+    def set_extra_fields(self, content):
+        from iaso.models import Account
+        from iaso.models.data_store import JsonDataStore
+
+        try:
+            datastore = self.jsondatastore_set.get(
+                account=Account.objects.filter(default_version=self.version).first(),
+                slug="extra_fields",
+            )
+            datastore.content = {**datastore.content, **content}
```
kemar, line 574: TIL: the second dictionary's values overwriting those from the first!
It might be worth a comment and small test?

kemar (approved):
LGTM

iaso/models/org_unit.py
bramj: Yep I love that splat thingy. And you're right! I'll add a test :)

kemar (approved):
LGTM

PR #964, opened 2023-12-07
POLIO-1332 Models and migration file

mathvdh:
Vaccine stock management page.
Models for the stock management and related migration file.

Related JIRA tickets : POLIO-1332, POLIO-1148

## Changes

A few new models, namely `VaccineStock` which is a pair Country, Vaccine (unique together).
`OutgoingStockMovement` which represents a "Form A" and is link to a `VaccineStock` and a campaign.
Then `DestructionReport` and `IncidentReport` which are only linked to a `VaccineStock`.

Also add admin  classes for those models

## How to test

You can try to create some models manually but there is not much to test at the moment.


plugins/polio/models.py
```diff
@@ -1054,6 +1054,65 @@ class VaccineArrivalReport(SoftDeletableModel):
...
+    form_a_reception_date = models.DateField()
+    usable_vials_used = models.PositiveIntegerField()
+    unusable_vials = models.PositiveIntegerField()
+    lot_numbers = ArrayField(models.CharField(max_length=200, blank=True), default=list)
+    missing_vials = models.PositiveIntegerField()
+
+
+class DestructionReport(models.Model):
+    vaccine_stock = models.ForeignKey(VaccineStock, on_delete=models.CASCADE)
+    action = models.TextField()
```
quang-le, line 1080: I put a comment in the mock up but I think we need to clarify this one. If we leave it as a text field we won't be able to translate it, and from the mockup it's only "Destruction + district"

plugins/polio/models.py
```diff
@@ -1054,6 +1054,65 @@ class VaccineArrivalReport(SoftDeletableModel):
...
+    rrt_destruction_report_reception_date = models.DateField()
+    destruction_report_date = models.DateField()
+    unusable_vials_destroyed = models.PositiveIntegerField()
+    lot_number = models.CharField(max_length=200)
```
quang-le, line 1084: We should confirm that there cannot be multiple lot numbers like formA

plugins/polio/models.py
```diff
@@ -1054,6 +1054,65 @@ class VaccineArrivalReport(SoftDeletableModel):
...
+
+
+class IncidentReport(models.Model):
+    class StockCorrectionChoices(models.TextChoices):
+        VVM_REACHED_DISCARD_POINT = "vvm_reached_discard_point", _("VVM reached the discard point")
+        VACCINE_EXPIRED = "vaccine_expired", _("Vaccine expired")
+        LOSSES = "losses", _("Losses")
+        RETURN = "return", _("Return")
+        STEALING = "stealing", _("Stealing")
```
quang-le, line 1093: "THEFT" sounds like better English to me, but maybe the client wants it that way

plugins/polio/models.py
```diff
@@ -1054,6 +1054,65 @@ class VaccineArrivalReport(SoftDeletableModel):
...
+    account = models.ForeignKey("iaso.account", on_delete=models.CASCADE, related_name="vaccine_stocks")
+    country = models.ForeignKey(
+        "iaso.orgunit",
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name="vaccine_stocks",
+        help_text="Unique (Country, Vaccine) pair",
+    )
+    vaccine = models.CharField(max_length=5, choices=VACCINES)
```
quang-le, line 1058: Shouldn't there also be a link with the VAR from "SupplyChain", or is it enough with the country to retrieve them?

I also wonder about the dashboards and the long term: we have a chance to have the same sort of problems as preparedness, i.e: the data will just grow and grow for each country+vaccine and will take longer an longer to compute in the API. We should probably have some way of archiving results after a year or something and use the last year's balance as a starting point for the computation (like any accounting software I guess). This is probably another ticket to put in specs, but I think it's good to already have that in mind

plugins/polio/admin.py
```diff
@@ -170,6 +171,35 @@ class VaccineRequestFormAdmin(admin.ModelAdmin):
     readonly_fields = ["created_at", "updated_at"]
     list_display = ["campaign", "get_country", "count_pre_alerts", "count_arrival_reports", "created_at"]
 
+    def save_related(self, request, form, formsets, change):
+        for formset in formsets:
+            if not formset.is_valid():
+                print(f"Formset errors: {formset.errors}")
```
kemar, line 177: Shouldn't we do something more useful than a `print()` here?(e.g. showing an error to the use or logging the error)

plugins/polio/api/vaccines/supply_chain.py
```diff
@@ -40,8 +40,11 @@ def validate_rounds_and_campaign(data, current_user=None):
         raise forms.ValidationError("A campaign must be attached.")
 
     try:
-        new_campaign = Campaign.objects.get(obr_name=campaign_obr_name)
-        data["campaign"] = new_campaign
+        if isinstance(campaign_obr_name, Campaign):
+            new_campaign = campaign_obr_name
```
kemar, line 44: Not blocking but reading this is a bit misleading.
It looks just like `campaign_obr_name` is a string while in fact it's a `Campaign` instance.
I had to go back the chain up to the model to understand.

plugins/polio/models.py
```diff
@@ -1054,6 +1054,65 @@ class VaccineArrivalReport(SoftDeletableModel):
...
+
+    class Meta:
+        unique_together = ("country", "vaccine")
+
+
+# Form A
```
kemar, line 1064: What does "Form A" mean? Is it a business-related term?
It's a bit meaningless out of context.
I suggest removing it or explaining with a bit more details in the docstring?

kemar (approved):
LGTM

A few comments but nothing blocking.

You should treat some of the comments if you care about "quality" though ;)

PR #970, opened 2023-12-08
IA-2552 Allow to tune sentry error sample rate

mestachs:
Sentry quota is regularly reached  and this PR try to help reducing sentry usage (temporarily or more definitively)

Related JIRA tickets : IA-2552

- [ ] Are there enough tests

## Changes

Sentry quota is regularly reached to help reducing sentry usage : 

- don't send `django.security.DisallowedHost` (happening often during security scanning, this has been partially fixed by setting a Web Application Firewall, but not all environments have that)
- adds sample rate limit to not send every errors to sentry (by default send everything)
- since we rely on some external api not under our control make a special case of HTTPError with a lower rate (by default send 80% of them)
- but still allow to override that via env variables
  - SENTRY_TRACES_SAMPLE_RATE (was already existing )
  - SENTRY_ERRORS_SAMPLE_RATE (new, equivalent but for the errors see the doc, currently send everything)
  - SENTRY_ERRORS_HTTPERROR_SAMPLE_RATE (new equivalent for the http error, allow to overide the default without code modification in case we decide to send every error again or even lower the sampling rate)

I've upgraded the sentry sdk and companion extra django integration, the error_sampler is "new" wasn't supported by our current version.

## How to test

this is a bit complicated but doable
 - modify your docker compose to add a valid  SENTRY_URL` (you can the same url as your staging/production env)
 -  in my case I've seeded an account via
```
 docker-compose run --rm iaso manage seed_test_data --mode=seed --dhis2version=2.37.10
```
 - then modified on the django api I know (here iaso/api/instances.py#retrieve function) and put a snippet like this
```
    import requests
    resp = requests.get("https://httpbin.org/status/401")
    resp.raise_for_status()
```
or 
```
   raise Exception("Oups local error triggering a 500")
```
- finally access the feature through the ui to trigger the HTTPError
- check the results in sentry UI

## Notes

we will work on the root causes in separate PR.

### DIFF
--- a/docs/pages/dev/reference/env_variables/env_variables.md
+++ b/docs/pages/dev/reference/env_variables/env_variables.md
@@ -1,0 +1,58 @@
+# Environnement variables
+
+## DB connection related
+
+the url is build based on the following env variables
+```
+RDS_USERNAME
+RDS_PASSWORD
+RDS_HOSTNAME
+RDS_DB_NAME
+RDS_PORT
+```
+
+the SQL dashboard use a dedicated user/password with readonly access to the data
+
+```
+DB_READONLY_USERNAME 
+DB_READONLY_PASSWORD
+```
+
+## AWS related
+
+Storing the various files like
+ - js/css/... static assets
+ - raw forms (xlsform), submissions (xml and media),... is done in s3 (or an s3 compatible api like minio)
+
+```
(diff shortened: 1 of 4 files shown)


kemar (approved):
LGTM

madewulf (approved):

PR #980, opened 2023-12-12
IA-2489: Selector of org unit should display type

beygorghor:
Here we are in district of Dedougou, Commune of Dedougou and I’m picking the centre de distribution of Dedougou. 

Not having the type displayed is confusing

Related JIRA tickets : IA-2489

- [ ] Are there enough tests

## Changes

- adapt api to return org unit type short name
- switch to show or hide org unit type
- tweek label to display or not type

## How to test

Open  org unit tree view on subissions page and play with the switch

## Print screen / video

https://github.com/BLSQ/iaso/assets/12494624/9903fdda-c51a-42af-be88-25e489c3f2a5


hat/assets/js/apps/Iaso/domains/orgUnits/components/TreeView/utils.js
```diff
@@ -95,20 +95,25 @@ const makeOrgUnistStatusIcon = (classes, orgUnit, formatMessage) => {
     return null;
 };
 
-export const makeTreeviewLabel = (classes, withStatusIcon) => orgUnit => {
-    const { formatMessage } = useIntl();
+export const makeTreeviewLabel =
+    (classes, withStatusIcon, withType = false) =>
+    orgUnit => {
+        const { formatMessage } = useIntl();
```
quang-le: useSafeIntl?

hat/assets/js/apps/Iaso/domains/orgUnits/components/TreeView/OrgUnitTreeviewModal.js
```diff
@@ -45,7 +46,10 @@ const OrgUnitTreeviewModal = ({
     allowedTypes,
     errors,
 }) => {
+    const theme = useTheme();
+    const { formatMessage } = useSafeIntl();
     const classes = useStyles();
+    const [displayTypes, setDisplayTypes] = useState(true);
```
quang-le, line 52: I'm not sure about having this feature enabled by default. I think in polio it would needlessly clutter the interface. Or maybe we could pass a prop to set the initial value of the switch?

quang-le (changes requested):
I think having a prop to set the switch initial value would really be a plus

hat/assets/js/apps/Iaso/domains/orgUnits/components/TreeView/OrgUnitTreeviewModal.js
beygorghor: the initial request was to display it without the switch option... let's try it like that

quang-le (approved):
ok nevermind, we can add the props later

PR #982, opened 2023-12-12
POLIO-1339: use supply chain permissions

quang-le:
- replace iaso-polio with supply chain read and write

Related JIRA tickets : POLIO-1339

- [ ] Are there enough tests

## Changes
- block edition if no write access
- block access if no read access (treating write as write only i. o read/write)
- Add tooltip and translation
- Update polio module

## How to test

- Create regular user without polio supply chain read access: should not see the feature in the menu
- Create regular user with read access only: should see list but no action buttons in table and no create button
- Create regular user with read+write: should have access to all supplychain features

## Print screen / video


plugins/polio/js/config.js
```diff
@@ -588,7 +588,9 @@ const menu = [
                         label: MESSAGES.vaccineSupplyChain,
                         key: 'supplychain',
                         dev: true,
```
hakifran: Is this menu will stay in dev mode?

hakifran (approved):
Everything is correct apart from the dev mode for the vaccin supply chain menu which I'm not sure if it should stay like that.

plugins/polio/js/config.js
quang-le: No it shouldn't, thanks fro pointing it out

PR #983, opened 2023-12-12
IA-2536 Modify permissions for Org Unit Change Request API

kemar:
Modify permissions of Org Unit Change Request API.

Related JIRA tickets : [IA-2536](https://bluesquare.atlassian.net/browse/IA-2536)

## Changes

The main goal is to use only one permission:

- most of the CRUD API must be available for authenticated users
- only `ORG_UNITS_CHANGE_REQUEST_REVIEW` is required to approve or reject a change request
- changes are reflected in the documentation

Other enhancements for ease of use by the front-end:

- use a proper pagination
- use `filter_backends`

## How to test

Use the Rest framework UI to test pagination and filters:

- http://localhost:8081/api/orgunits/changes/?order=id
- http://localhost:8081/api/orgunits/changes/?order=-id
- etc.

[IA-2536]: https://bluesquare.atlassian.net/browse/IA-2536?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iaso/api/org_unit_change_requests/permissions.py
```diff
@@ -5,4 +5,13 @@
 
 class HasOrgUnitsChangeRequestPermission(permissions.BasePermission):
     def has_permission(self, request, view):
-        return request.user.is_authenticated and request.user.has_perm(iaso_permission.ORG_UNITS_CHANGE_REQUEST)
+        if request.user.is_superuser:
+            return True
+        return request.user.is_authenticated
```
bmonjoie: Nitpicking here but it can be simplified to:
```suggestion
        return request.user.is_authenticated
```

iaso/api/org_unit_change_requests/views.py
```diff
@@ -20,8 +27,16 @@
...
+    filter_backends = [filters.OrderingFilter, django_filters.rest_framework.DjangoFilterBackend]
     filterset_class = OrgUnitChangeRequestListFilter
+    pagination_class = OrgUnitChangeRequestPagination
+
+    def get_permissions(self):
+        if self.action == "partial_update":
+            permission_classes = [HasOrgUnitsChangeRequestReviewPermission]
+        else:
+            permission_classes = [HasOrgUnitsChangeRequestPermission]
+        return [permission() for permission in permission_classes]
```
bmonjoie, line 39: I may have misunderstood here but I believe this listing is only relevant for reviewer so I would change it to:
```suggestion
    permission_classes = [HasOrgUnitsChangeRequestReviewPermission]
```

iaso/api/org_unit_change_requests/permissions.py
kemar: Are you sure of your comment? Because `request.user.is_superuser` != `request.user.is_authenticated`

iaso/api/org_unit_change_requests/views.py
kemar: No we can't do that because we are in the context a viewset and the other views (`list`, `update`, etc.) only require `HasOrgUnitsChangeRequestPermission`.

iaso/api/org_unit_change_requests/permissions.py
bmonjoie: If you are a superuser, you are authenticated, aren't you?

iaso/api/org_unit_change_requests/views.py
bmonjoie: You are right, I forgot that the `POST` and `PUT` were used by regular users. I would change it like so:
```suggestion
    def get_permissions(self):
        if self.action == "insert":
            permission_classes = [HasOrgUnitsChangeRequestPermission]
        else:
            permission_classes = [HasOrgUnitsChangeRequestReviewPermission]
        return [permission() for permission in permission_classes]
```
There is no need for "regular" users to list all the change requests. They need only to create new ones.

iaso/api/org_unit_change_requests/permissions.py
kemar: Yes, but the inverse is not true.
Being authenticated doesn't imply having the [`is_superuser` flag](https://docs.djangoproject.com/en/3.2/ref/contrib/auth/#django.contrib.auth.models.User.is_superuser).

iaso/api/org_unit_change_requests/views.py
kemar: This wouldn't work because `insert` is not a [valid action on a viewset](https://www.django-rest-framework.org/api-guide/viewsets/#viewset-actions).

My code is actually correct and conform to [what we said in IA-2536](https://bluesquare.atlassian.net/browse/IA-2536).

Trust me on this one ;)

iaso/api/org_unit_change_requests/permissions.py
bmonjoie: I know, but if they are not a super user, it will return `is_authenticated`; it doesn't matter if the user has the `is_superuser` flag.

You can see this as :
- `is_superuser = a && b`
- `is_authenticated = a`

Your code does:
```python
if a && b:
    return True
return a
```
It doesn't matter what `b` (`is_superuser`) is since you'll always return the value of `a` (`is_authenticated`)

iaso/api/org_unit_change_requests/views.py
bmonjoie: I wrote pseudocode, in the sense that I didn't check the possible actions, I thought the intention was clear as is.
If you want, you can replace `insert` by `create`.

Anyway, I'm not arguing whether your code is correct or wrong. I'm just asking if it makes sense to expose `list`, `retrieve`,  `destroy`, and `update` to regular users since we only need them to create new change requests and not act upon them.

iaso/api/org_unit_change_requests/permissions.py
kemar: I guess your suggestion is wrong then, what you mean is:

```python
return request.user.is_superuser or request.user.is_authenticated
```

iaso/api/org_unit_change_requests/views.py
kemar: > I'm just asking if it makes sense to expose list, retrieve,  destroy, and update to regular users since we only need them to create new change requests and not act upon them.

That's what we agreed on in [IA-2536](https://bluesquare.atlassian.net/browse/IA-2536). Shall we change that?

iaso/api/org_unit_change_requests/permissions.py
bmonjoie: Ok, let's deconstruct this for the two cases:
A)
```python
if request.user.is_superuser:
     return True
return request.user.is_authenticated
```
B)
  ```python
return request.user.is_authenticated
```
```
|| Type of User          ||  A  ||   B  ||
| super user             | True  | True  |
| is authenticated       | True  | True  |
| is *not* authenticated | False | False |
```

As you can see, the condition about `is_superuser` is unnecessary.

iaso/api/org_unit_change_requests/views.py
bmonjoie: That's what I'm asking you :slightly_smiling_face:

iaso/api/org_unit_change_requests/permissions.py
kemar: Ok got it, you're right.

iaso/api/org_unit_change_requests/views.py
kemar: Let's stick to what we agreed on.

bmonjoie: As discussed, I believe it's "dangerous" to let `update` and `delete` open. I also don't think it makes much sense to leave `list` and `retrieve` accessible, but they are not a threat.
Anyway, I have raised the concern, it's not my decision, the code is fine by me.

bmonjoie (approved):

PR #984, opened 2023-12-13
POLIO-1327 key error preparedness

mestachs:
When the template isn't respected, unexpected things happen.
Try to handle them and return warnings

Related JIRA tickets : POLIO-1327

- [x] Are there enough tests

## Changes

I've captured a googlesheet as json with a bad shaped sheet and reproduced the problem in the test. Now I'm just handling the fact that we might not have found the data, and putting a warning when the data isn't found.

I'm now display that some warnings at the end of the caption.
Since the warnings are not translated, I've put them as a tool tip.

## How to test

Since it's quite specific to that sheet and campaign it's hard to test except messing up the google sheet your self. In this case the "region name" in C4 and the first column with the regional info and district have been modified by hand/copy pasting. Initially it's a formula referring to C4 but here it was a simple text (not even matching the name)

## Print screen / video

image

## Notes

Since the preparedness data can be "cached" some of them might not have the warnings field.
That's why I've made the client a bit more cautious.


plugins/polio/tests/test_preparedness_parser.py
```diff
@@ -0,0 +1,28 @@
...
+        with open(filename, "rb") as text:
+            return json.load(text)
+
+    def test_reproduce_bad_district_naming(self):
+        self.maxDiff = None
+        PREPAREDNESS_SHEET = "plugins/polio/tests/fixtures/preparedness_sheet.json"
+        PREPAREDNESS_EXPECTED_STATS = "plugins/polio/tests/fixtures/preparedness_expected_stats.json"
+
+        stats = get_preparedness(CachedSpread(self.load_fixture(PREPAREDNESS_SHEET)))
+        print(json.dumps(stats, indent=4))
```
kemar, line 27: This `print` could be removed?

kemar (approved):

PR #993, opened 2023-12-19
POLIO-1340 Display details of notification import tasks

kemar:
Display details of notification import tasks.

Related JIRA tickets : [POLIO-1340](https://bluesquare.atlassian.net/browse/POLIO-1340)

## Changes

- add Polio Import Notification details in the Tasks UI
- include a `polio_notification_id` in the output of `TaskSerializer`
- include new fields (like `errors`) in the output of `NotificationImportSerializer`
- add Cypress tests

## How to test

1. ensure to:
    - be admin
    - **or** to have the perm `menupermissions | custom permission support | Polio notifications`
1. navigate to `Polio > Virus notification`
2. click on `CREATE FROM FILE`
1. import an .xlsx file ([sample available here](https://bluesquare.atlassian.net/browse/POLIO-1275?focusedCommentId=59457))
        - the backend's import is async, so a worker must be running
        - `docker-compose run iaso manage tasks_worker`
1. navigate to `Admin > Tasks`
1. you must have 2 perms:
        1. `menupermissions | custom permission support | Tâches`
        1. `menupermissions | custom permission support | Polio notifications`
1. you should be able to view the details of the import

## Print screen / video

https://github.com/BLSQ/iaso/assets/281139/9aa7e374-c66e-405a-89d9-9c5b4eea97b3

[POLIO-1340]: https://bluesquare.atlassian.net/browse/POLIO-1340?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/app/translations/en.json
+++ b/hat/assets/js/apps/Iaso/domains/app/translations/en.json
@@ -1098,6 +1098,8 @@
     "iaso.tasks.launcher": "User",
     "iaso.tasks.message": "Message",
     "iaso.tasks.name": "Name",
+    "iaso.tasks.polioNotificationImport.details": "Polio Notifications Import Details",
+    "iaso.tasks.polioNotificationImport.errors": "{count} error(s). The following data couldn't be imported.",
     "iaso.tasks.progress": "Progress",
     "iaso.tasks.queued": "Queued",
     "iaso.tasks.running": "Running",
--- a/hat/assets/js/apps/Iaso/domains/tasks/components/NotificationImportDetailModal.tsx
+++ b/hat/assets/js/apps/Iaso/domains/tasks/components/NotificationImportDetailModal.tsx
@@ -1,0 +1,78 @@
+import React, { FunctionComponent } from 'react';
+
+import {
+    LoadingSpinner,
+    makeFullModal,
+    useSafeIntl,
+} from 'bluesquare-components';
+import { AlertModal } from '../../../components/AlertModal/AlertModal';
+
+import MESSAGES from '../messages';
+import { NotificationImportDetailButton } from './NotificationImportDetailButton';
+import { Task } from '../types';
+import { useGetPolioNotificationImport } from '../hooks/api';
+
+type Props = {
+    task: Task<any>;
+    isOpen: boolean;
+    closeDialog: () => void;
+};
+
+const NotificationImportDetailModal: FunctionComponent<Props> = ({
+    task,
+    isOpen,
+    closeDialog,
+}) => {
+    const { formatMessage } = useSafeIntl();
+
+    const {
+        data: notificationImport,
+        isFetching: isFetchingNotificationImport,
+    } = useGetPolioNotificationImport(task.polio_notification_id);
+
+    return (
+        <AlertModal
+            isOpen={isOpen}
+            closeDialog={closeDialog}
+            titleMessage={formatMessage(
+                MESSAGES.polioNotificationImportDetails,
+            )}
+            maxWidth="md"
+        >
+            {isFetchingNotificationImport && <LoadingSpinner />}
+            {!isFetchingNotificationImport && (
+                <>
+                    <p>{task.result.message}</p>
+                    {notificationImport &&
+                        notificationImport.errors.length > 0 && (
+                            <>
+                                <p>
+                                    {formatMessage(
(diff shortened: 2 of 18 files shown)


hat/assets/js/apps/Iaso/domains/app/translations/en.json
```diff
@@ -1098,6 +1098,8 @@
     "iaso.tasks.launcher": "User",
     "iaso.tasks.message": "Message",
     "iaso.tasks.name": "Name",
+    "iaso.tasks.polioNotificationImport.details": "Polio Notifications Import Details",
```
hakifran, line 1101: Just a question, why in this translation each word begin with a capital letter?

hat/assets/js/apps/Iaso/domains/tasks/components/NotificationImportDetailModal.tsx
```diff
@@ -0,0 +1,79 @@
...
+                                <p>
+                                    {formatMessage(
+                                        MESSAGES.polioNotificationImportErrors,
+                                        {
+                                            count: notificationImport.errors
+                                                .length,
+                                        },
+                                    )}
+                                </p>
+                                <pre style={{ maxHeight: '550px' }}>
```
hakifran, line 59: I don't know if it should be like that, but I checked the [mockup](https://whimsical.com/polio-line-list-import-error-check-DDJfGsn9n4yMpsJvo19KCY) and find that the error data or updated ones should not appear like json objects but instead like a list.

hat/assets/js/apps/Iaso/domains/app/translations/en.json
kemar: Because [title case](https://en.wikipedia.org/wiki/Title_case) is a common rule of most English style guides for rendering the titles.

Should I switch to normal case?

hat/assets/js/apps/Iaso/domains/tasks/components/NotificationImportDetailModal.tsx
kemar: Yes, because the solution in the mockup would have required a too complex implementation.

So I suggested using a JSON field to store the errors.

iaso/tests/api/test_tasks.py
```diff
@@ -1,26 +1,127 @@
...
+                "started_at": None,
+                "ended_at": None,
+                "progress_value": 1,
+                "end_value": 1,
+                "launcher": {
+                    "first_name": "",
+                    "last_name": "",
+                    "username": "user",
+                },
+                "result": "324.49 sec, processed 5262 org units",
```
hakifran, line 72: I may be wrong, but I think that this value should gotten from self.task like self.task.pk

iaso/tests/api/test_tasks.py
```diff
@@ -1,26 +1,127 @@
...
+                "started_at": None,
+                "ended_at": None,
+                "progress_value": 1,
+                "end_value": 1,
+                "launcher": {
+                    "first_name": "",
+                    "last_name": "",
+                    "username": "user",
+                },
+                "result": "3768 polio notifications created.",
```
hakifran, line 97: Same here

hat/assets/js/apps/Iaso/domains/app/translations/en.json
hakifran: I don't know, but I'm wondering if the rule is applied on all title in english in IASO. @quang-le  can may be can help to clarify it?

hakifran (approved):
For me everything is fine apart from some comment for clarification(not blocking)

iaso/tests/api/test_tasks.py
kemar: I don't understand your comment.
`result` is stored as text in `Task.result` and defined just above in `setUpTestData()`.

hakifran: I was just wondering if instead of writing "324.49 sec, processed 5262 org units" we can write self.task.result so that when its value change it doesn't break the test. but it's not blocking

kemar: Ah ok, understood. Nope, in fact I want the test to break explicitly if the value changes!

PR #999, opened 2023-12-20
POLIO-1365: Track dashboard use: Move the analytics script field from page to account

hakifran:
- The analytics script should be one for each account
- For that, the field related to it should be moved from page to account
Related JIRA tickets : [POLIO-1365](https://bluesquare.atlassian.net/browse/POLIO-1365)

- [ ] Are there enough tests

## Changes
- Remove the analytics_script field from page model
- Add it to account model
- Add the field to api/setaccount

## How to test
- Setaccount or edit on from the admin
- Add an analytics script sample
- Go in the embedded links
- Check in inspect element if the script is in the head tag

## Print screen / video
Screenshot from 2023-12-20 17-17-15
Screenshot from 2023-12-20 17-16-47

[POLIO-1365]: https://bluesquare.atlassian.net/browse/POLIO-1365?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hakifran:
@madewulf , should I add the analytics script field in /dashboard/setupAccount too?

iaso/migrations/0237_add_modules_to_all_existing_accounts.py
```diff
@@ -7,7 +7,8 @@
 
 def add_modules_to_accounts(apps, schema_editor):
     modules = [module["codename"] for module in MODULES]
-    all_accounts = Account.objects.all()
+    account = apps.get_model("iaso", "Account")
+    all_accounts = account.objects.all()
```
madewulf, line 11: Why is this in this PR?

hakifran: It's because, the migration was generating an error that the field analytics_script didn't exist. So I request the model Account before the field to be added to fix the error.

madewulf: I think this was only necessary because you were in a strange migration state (having the field analytics_script in the db already due to a previous migration then running that migration). Do you think that's possible?

hakifran: First I made a migration to remove the analytics_script from the Page model **iaso/migrations/0252_remove_analytics_script_field_from_page.py**, after I added the analytics_script to the Account model **iaso/migrations/0253_add_analytics_script_field_into_account.py** and in local everything worked fine. It's when I pushed on github that the migration precisely on this migration and on the line **all_accounts = Account.objects.all()** saying that **analytics_script does not exist yet**.

madewulf (approved):

PR #1015, opened 2024-01-05
Polio 1396 add lqas country endpoint

quang-le:
lqas front-end was using datastore endpoint directly, but by design we don't give that permission to users, so the page was basically broken

Related JIRA tickets : POLIO-1396

- [ ] Are there enough tests

## Changes

- Delete old `/lqasstats` endpoint as it's not used anymore
- Add `LQASCountryViewset` which extends `JsonDataStoreViewset` but changes the permissions, has narrower queryset (only lqas data) and is GET only
- Replace url in front-end
- Add `blank=true` to `org_unit` field in `JsonDataStore` model to enable creating new datastores without org unit in the django admin 

Basically, it's just the same endpoint with only a GET method, other permissions and a restricted queryset

## How to test

- Setup json datastores in your local polio DB:
    - get at least 1 lqas datastore and 1 im datastore (can be found on polio staging)
    - copy them in your DB via django admin
    - double check that th org unit ids match

- Create a non-super user account with polio permission, and another one with polio+polio_config permission
- as regular user, log in, go to lqas page, select country with the datastore: you should see the results on the map (+graphs)
- repeat with polio admin user. Result should be the same and the refresh button should also be visible
- In Django API, try to access
    - `/api/polio/lqasmap/country/ ` : you should only see the lqas data store
    - `/api/polio/lqasmap/country/ <im-datastore-slug>`: you should get a 404

## Print screen / video

https://github.com/BLSQ/iaso/assets/38907762/2d7eaf1b-2f69-4241-b0a1-18c4d3d46366

## Notes


plugins/polio/api/lqas_im/lqas_country.py
```diff
@@ -0,0 +1,30 @@
+from iaso.api.data_store import DataStoreViewSet
+from hat.menupermissions import models as permission
```
kemar:
```suggestion
from hat.menupermissions import models as menu_permission
```

Just to avoid an easy confusion between `permission` and `permissions` just below.

kemar (approved):
LGTM, clear and well tested 👏

PR #1021, opened 2024-01-08
Polio 1358 vaccine module format thousand

quang-le:
We needed number formatting for vaccine supplychain

Related JIRA tickets : POLIO-1358

- [ ] Are there enough tests

## Changes

- Import `NumberInput` and `FormattedNumber` from `bluesquare-components`
- Refactor `InputComponent` to use `NumberInput`. Add props and get thousand and decimal separators from the locale by default
- Add `NumberCell` which uses `FormattedNumber` with the same defaults as `InputComponent` but can be used in tables and text.
- Update all components in polio that used `TextInput` to show numbers so they use `NumberInput` instead
- Update Vaccine puppy chain and vaccine stock management
- Harmonize layout in polio campaigns modal: empty fields don't shrink the label anymore

## How to test
- run `npm ci`
- replace dep to bluesquare-components in package.json :
`"bluesquare-components": "github:BLSQ/bluesquare-components#POLIO-1358_thousand_separator",`
- run `npm update bluesquare-components && npm run dev` (you will run webpack outside of docker)
- Go through iaso and check that numbers are formatted where needed. 

What you should see: when setting language to english, thousands should be separated by commas, and decimals by a dot. When setting language to French it should be the opposite.

So when checking high integers, check in english, when checking floats (like latitude/longitude), check in french.

## Print screen / video

https://github.com/BLSQ/iaso/assets/38907762/561121e6-9b7b-47de-912f-d75ed5c983c7

## Notes
⚠️Depends on https://github.com/BLSQ/bluesquare-components/pull/137

TESTS WILL FAIL UNTIL https://github.com/BLSQ/bluesquare-components/pull/137 is merged and the dep updated


hat/assets/js/apps/Iaso/types/general.ts
```diff
@@ -7,6 +7,11 @@ export type PostArg = {
     signal?: AbortSignal | null;
 };
 
+export type Locale = {
+    code: string;
+    label: string;
+};
+
```
beygorghor, line 14: ❤️

hat/assets/js/apps/Iaso/hooks/useNumberSeparatorsFromLocale.tsx
```diff
@@ -0,0 +1,15 @@
...
+export const useNumberSeparatorsFromLocale = (): {
+    thousand: '.' | ',';
+    decimal: '.' | ',';
+} => {
+    // @ts-ignore
+    const activeLocale = useSelector(state => state.app.locale);
+    return useMemo(
+        () => determineSeparatorsFromLocale(activeLocale),
+        [activeLocale],
+    );
```
beygorghor: As we discussed, we can use intl.formatNumber to get separators.

```
  const decimalSeparator = intl.formatNumber(1.1).charAt(1);
  const thousandSeparator = intl.formatNumber(1000).charAt(1);
```
For the group styles we have to hardcode it anyway:

```
const thousandsGroupStyles = {
  // Grouping by thousands (most Western and many other countries)
  'US': 'thousand', // United States
  'GB': 'thousand', // United Kingdom
  'CA': 'thousand', // Canada
  'FR': 'thousand', // France
  'DE': 'thousand', // Germany
  'AU': 'thousand', // Australia
  'BR': 'thousand', // Brazil
  'RU': 'thousand', // Russia
  'JP': 'thousand', // Japan
  // ...

  // Grouping by lakhs (South Asian countries)
  'IN': 'lakh', // India
  'PK': 'lakh', // Pakistan
  'BD': 'lakh', // Bangladesh
  'LK': 'lakh', // Sri Lanka
  'NP': 'lakh', // Nepal
  // ...

  // Grouping by wan (China)
  'CN': 'wan', // China
  // Note: 'wan' is not a standard option in react-number-format, this is for illustrative purposes
};

// Usage example
const country = 'IN'; // Assume this is dynamically determined by the user's locale
const groupStyle = thousandsGroupStyles[country] || 'thousand'; // Default to 'thousand'
```

hat/assets/js/apps/Iaso/utils/dataManipulation.ts
```diff
@@ -9,3 +11,17 @@ export const stringToBoolean = (str: string): boolean | undefined => {
...
+    // using a switch to add more locales easily
+    switch (activeLocale.code) {
+        case 'fr':
+            return { thousand: '.', decimal: ',' };
+        case 'en':
+            return { thousand: ',', decimal: '.' };
+        default:
+            return { thousand: ',', decimal: '.' };
+    }
+};
```
beygorghor: This can be removed if we use react-intl

plugins/polio/js/src/domains/VaccineModule/StockManagement/Details/Table/useVaccineStockManagementDetailsColumns.tsx
```diff
@@ -22,18 +23,30 @@ export const useVaccineStockManagementDetailsColumns = (tab): Column[] => {
             {
                 Header: formatMessage(MESSAGES.vials_in),
                 accessor: 'vials_in',
+                Cell: settings => (
+                    <NumberCell value={settings.row.original.vials_in} />
+                ),
```
beygorghor: check usage of UserCell, you can avoid a function here

plugins/polio/js/src/domains/Campaigns/Rounds/RoundsForm.js
```diff
@@ -203,7 +203,7 @@ export const RoundsForm = () => {
                 )}
             </Box>
             {currentRoundNumber !== undefined && (
-                <Box mt={2}>
+                <Box mt={2} style={{ width: '100%' }}>
```
beygorghor: you can use width props directly 
```suggestion
                <Box mt={2} width="100%" >
```

beygorghor (changes requested):
This is looking very good, even while typing the number is auto formatted.
Just use react-intl to get separators

plugins/polio/js/src/domains/VaccineModule/StockManagement/Details/Table/useVaccineStockManagementDetailsColumns.tsx
quang-le: It's to avoid a ts error. we should probably update the `Column` type as well. Also, I wanted to be able to pass additional props. There are cases where you need to pass the `key` prop to avoid nasty bugs when the table updates

beygorghor (approved):

PR #1027, opened 2024-01-10
IA-2602: Fix GET /api/mobile/orgunits/changes/

bmonjoie:
The endpoint ‘GET /api/mobile/orgunits/changes/?app_id=…’ is used by the mobile application to keep track of the changes made to the user’s change requests. It therefore doesn’t have to return the whole list of change request but only the ones of the authorized user.

Also, the endpoint should return the reference instances' data to recreate them when they are not present on the mobile device.

Related JIRA tickets : IA-2602

- [X] Are there enough tests

## Changes

- Return only the changes of the authenticated user
- Return instances as objects and not as UUIDs

## How to test

Send a change request linked to an instance, then request your change request list.


kemar (approved):
LGTM

PR #1032, opened 2024-01-10
WC2-301 initial proposal to fix the two bugs in WC2-301

mathvdh:
This fixes 2 bugs.
1) There was a problem when in some occasions after exporting a workflow from an instance to another one, there were multiple versions of the workflow_version with status "PUBLISHED" which shouldn't happen.
2) Other bug is more subtile and basically if you do an export of a workflow from instance I1 to instance I2 and somebody deleted a workflow_version on I2 but not on I1 (and it had been already exported in the pas from I1 to I2) the import on I2 would bug because it would search for a workflow_version by UUID but it was (soft) deleted in the meanwhile, then it wouldn't find it and would try to recreate it with the same UUID but as it was only soft deleted a workflow version with this UUID would actually already exist in DB and the DB constraint on unicity of the UUIDs for the workflows version would make it crash.

Related JIRA tickets : WC2-301

- [ ] Are there enough tests

## Changes

We verify that the imported payload doesn't include multiple workflow_versions with status = "PUBLISHED" and return an error if so. We checked before overriding the actual version to put older versions that would have status UNPUBLISHED to PUBLISHED.

## How to test

See the attached tests for an example. But basically for 1) you can just modify an exported JSON payload and have 2 workflows versions with status PUBLISHED. You could also change a status randomly to PUBLISHED on the destination before importing from source with another version with status PUBLISHED.
for 2) soft delete a workflow version on destination only and reimport from source where this workflow version is not deleted.


iaso/api/workflows/import_export.py
```diff
@@ -76,9 +76,15 @@ def import_workflow_real(workflow_data: typing.Dict, account: Account) -> Workfl
     except Workflow.DoesNotExist:
         wf = Workflow.objects.create(uuid=workflow_data["uuid"], entity_type=entity_type)
 
+    # Set all related WorkflowVersion objects' status to UNPUBLISHED if they are currently PUBLISHED
+    published_versions = wf.workflow_versions.filter(status=WorkflowVersionsStatus.PUBLISHED.value)
+    for version in published_versions:
+        version.status = WorkflowVersionsStatus.UNPUBLISHED.value
```
madewulf: This assumes that the import always include a published workflow that will replace the previously published ones, I think. Is it always true?

kemar (approved):

PR #1033, opened 2024-01-10
IA-2604 Replace `reviewed_*` fields by `updated_*` fields.

kemar:
Fix bug for `created_at > updated_at`.

Related JIRA tickets : [IA-2604](https://bluesquare.atlassian.net/browse/IA-2604)

## Changes

I thought it would be a smart idea to have `reviewed_at` and `reviewed_by` fields. It turns out it wasn't such a smart idea:

- these fields were not exposed by the API (nor documented)
- they were easily confused with `updated_at` and `updated_by`
- they were not used by the `last_sync` filter

Changes:

- `created_at` and `created_by` are now read only in admin
- `reviewed_at` and `reviewed_by` are deleted
- `updated_at` and `updated_by` are now set at review time (approval or rejection)

## Notes

[`Field.default`](https://docs.djangoproject.com/en/4.2/ref/models/fields/#django.db.models.Field.default):

> If callable it will be called every time a new object is created.

[`DateField.auto_now`](https://docs.djangoproject.com/en/4.2/ref/models/fields/#django.db.models.DateField.auto_now):

> The field is only automatically updated when calling `Model.save()`.

`updated_at = models.DateTimeField(auto_now=True)` would populate `reviewed_at` **even** at creation time and I'd rather let it empty (see unit tests).

[IA-2604]: https://bluesquare.atlassian.net/browse/IA-2604?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bmonjoie (approved):
LGTM

PR #1047, opened 2024-01-15
IA-2611 Keep old values for change requests

kemar:
Keep old values for change requests.

Related JIRA tickets : [IA-2611](https://bluesquare.atlassian.net/browse/IA-2611).

## Changes

When a Change Request is approved:

- we keep the old values as well as the new ones
- in order to make it easier to display a diff in the UI

## How to test

1. create a Change Request
2. approve it
3. you can see old values:
    1. in admin `http://localhost:8081/admin/iaso/orgunitchangerequest/` + **ID** + `/change/`
    2. in the DRF UI `/api/orgunits/changes/` + **ID** + `/`

## Print screen / video

admin

api

[IA-2611]: https://bluesquare.atlassian.net/browse/IA-2611?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mathvdh (commented):
Comments :

- Opening date field
When I create a change request via Django ADMIN, it doesn't want to save it, if I don't set a `Opening Date` value (is that correct ?) Maybe I have old data in local DB but I didn't have any value for this. What's weird is that in initial Change Request I did setup this date and when I go and see the Org Unit after applying the CR it's there with the new value.
But When I create a new CR after that, it's not there and asking me AGAIN about this date
(not sure this is related to this PR exactly but with change requests in general). 
After checking the code a bit more it might be related to the fact that I didn't have this field setup initially. But what bothers me is that it still re-did the same after I applied it and got a value in DB for this field ..

- Codewise
I'm not a big fan of the old_this and old_that (no particular reason except adding more fields to the model and feeling "dirty") BUT it **does** work.

- 👍🏼 for updating the documentation.

Otherwise it's quite straightforward.

kemar:
@mathvdh 

> * Codewise
>   I'm not a big fan of the old_this and old_that (no particular reason except adding more fields to the model and feeling "dirty") BUT it **does** work.

I didn't come up with anything better. Do you have a better suggestion?

mathvdh:
@kemar Not really, I have a tendency to over-engineer so I would maybe have another table and copy the whole "change request" there and point it to this one. Or keep it there (add a field with a date or like is_old and point it to the new one ..)
But honestly it doesn't matter and I do think that your solution is more "efficient" and fast.

kemar:
@mathvdh 

> * Opening date field
>   When I create a change request via Django ADMIN, it doesn't want to save it, if I don't set a `Opening Date` value (is that correct ?)

Nope it was a bug fixed in https://github.com/BLSQ/iaso/pull/1047/commits/2ce5b31fc260509e730faad50050ffc35e8cb12b

mathvdh:
Good for me ! Bug with opening date is fixed

mathvdh (approved):

PR #1050, opened 2024-01-16
IA-2565: Org units search change source should reset parent org unit

beygorghor:
If source field isn't actively cleared before switching source, then parent selection doesn't take into account the source switch
In Org unit search, if source field isn't actively cleared before switching source, then parent selection doesn't take into account the source switch (see attached video).

https://github.com/BLSQ/iaso/assets/12494624/e04a1149-0cf2-4cbe-aec2-a55d455b4ad4

Related JIRA tickets : IA-2565

- [ ] Are there enough tests

## Changes

- fix typings issues
- delete levels in search if source is changed

## How to test

Make a string search leading to few results on a specific source. (take a print screen to compare results)
Change the source and choose a  parent org unit. and keep the same string search and search.
List should be empty or different.
Switch back to first source, keep search string and search again, results should be the same as the print screen.

## Print screen / video

https://github.com/BLSQ/iaso/assets/12494624/228a1d4b-4d64-44c3-ad40-3a592c5488b6

## Notes

Cypress test are not implemented here, we should keep in mind those race conditions and test it !


hat/assets/js/apps/Iaso/components/filters/DatesRange.d.ts
quang-le, line 1: How was this generated?

hat/assets/js/apps/Iaso/domains/orgUnits/components/OrgUnitsFilters.tsx
```diff
@@ -117,20 +117,22 @@ export const OrgUnitFilters: FunctionComponent<Props> = ({
...
     } = useGetValidationStatus(true);
     const handleChange = (key, value) => {
+        const newFilters: Record<string, unknown> = {
+            ...filters,
+        };
         if (key === 'version') {
             setSourceVersionId(parseInt(value, 10));
         }
         if (key === 'source') {
+            console.log('Reset');
```
quang-le: 🚮

hat/assets/js/apps/Iaso/domains/orgUnits/components/OrgUnitsFilters.tsx
```diff
@@ -258,7 +260,7 @@ export const OrgUnitFilters: FunctionComponent<Props> = ({
                     label={MESSAGES.search}
                     blockForbiddenChars
                     onEnterPressed={onSearch}
-                    onErrorChange={setTextSearchError}
+                    onErrorChange={(hasError) => setTextSearchError(hasError)}
```
quang-le, line 263: This is just to make the TS compiler happy isn't it?😁

quang-le (changes requested):
I noticed that in the url, when we change source, the `orgUnitParentId` is not cleared
https://github.com/BLSQ/iaso/assets/38907762/30b7fd61-f9a7-449f-9607-ff379b537ccb
It doesn't break anything, but since we're cleaning up...

hat/assets/js/apps/Iaso/components/filters/DatesRange.d.ts
beygorghor: by hand

hat/assets/js/apps/Iaso/domains/orgUnits/components/OrgUnitsFilters.tsx
beygorghor: yep

quang-le (approved):

PR #1062, opened 2024-01-18
Polio-1320 move pipeline version to config

quang-le:
storing this data in an env variable was not practical as it needs frequent updates

Related JIRA tickets : POLIO-1320

- [ ] Are there enough tests

## Changes

- Get all Openhexa settings from a `Config` object i.o from `settings.py` (env var)
- Remove obsolete variables from settings.py

## How to test

- Go to django admin, and add a `Config` called "lqas-pipeline-config"
- Fill it with the OpenHexa settings and save
- Log in to OpenHexa and open lqas pipeline
- In the interface, go to polio>lqas>country view
- Choose a campaign and click Refresh
- Go to OpenHexa: the pipeline should have been launched with the proper params

## Print screen / video

No visible changes

### DIFF
--- a/plugins/polio/tasks/api/refresh_lqas_data.py
+++ b/plugins/polio/tasks/api/refresh_lqas_data.py
@@ -1,4 +1,5 @@
 import logging
+from django.shortcuts import get_object_or_404
 from gql import Client, gql
 from gql.transport.requests import RequestsHTTPTransport
 from datetime import datetime
@@ -13,6 +14,8 @@ from rest_framework.response import Response
 from django_filters.rest_framework import DjangoFilterBackend  # type:ignore
 from django.db.models import Q
 from rest_framework.decorators import action
+
+from plugins.polio.models import Config
 
 logger = logging.getLogger(__name__)
 TASK_NAME = "Refresh LQAS data"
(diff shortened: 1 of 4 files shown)


plugins/polio/tasks/api/refresh_lqas_data.py
```diff
@@ -136,10 +139,21 @@ def create(self, request):
...
+            pipeline_config = get_object_or_404(Config, slug="lqas-pipeline-config")
+        except:
+            logger.exception("Could not fetch openhexa config")
+            return ERRORED
+        lqas_pipeline_version = pipeline_config.content["lqas_pipeline_version"]
+        oh_pipeline_target = pipeline_config.content["oh_pipeline_target"]
+        lqas_pipeline = pipeline_config.content["lqas_pipeline"]
+        openhexa_url = pipeline_config.content["openhexa_url"]
+        openhexa_token = pipeline_config.content["openhexa_token"]
+
```
mathvdh: I think this could crash if one of these keys is not present in the config.content JSON thing.
Maybe put a big try catch around it or something as I don't think there is anywhere where the structure of this config is verified. 

Alternatively and also for documentation purpose, you could use the jsonschema lib to verify this structure (See `iaso/tests/api/workflows/test_workflows.py` for an full example)

mathvdh (changes requested):
Please see my comment. Otherwise it's good !

mathvdh (approved):
Great ! jsonschema is cool for these usecases

PR #1069, opened 2024-01-22
Update the setuper doc

kemar:
Update the setuper doc


quang-le (changes requested):
Could you also put that in `/docs/pages/dev/how-to`
in the medium term, I would like to trim down the README a bit and make the doc the source of truth for such things

quang-le (approved):

PR #1071, opened 2024-01-23
IA-2468: make pipeline code generic

quang-le:
Make code used for updating lqas via openhexa pipeline reusable

Related JIRA tickets : IA-2468, POLIO-1401

- [X] Are there enough tests

## Doc

https://wiki.bluesquare.org/en/Dev-team/product-management/iaso/Polio/refresh_lqas_reference

## Changes

- Add `ExternalTaskModelViewSet`, a subclass of iaso's `ModelViewSet`with all the pipeline related code from `RefreshLQASData` made more generic
- Add `ExternalTaskPostSerializer` to handle external task creation
- Move `ExternalTaskSerializer` iaso.api.tasks
- Refactor `RefreshLQASData` to subclass `ExternalTaslModelViewSet`, with some overrides in validation to take into account the specificities of the LQAS use case.
- Update front-end to use the modified API
- Refactor `test_refresh_lqas`

**Note**: `test_refresh_lqas` actually tests the external task's flow. It should also be moved and renamed, but first we need to move the `Config` object out of the polio plugin, which is another ticket: [POLIO_1439](https://bluesquare.atlassian.net/browse/POLIO-1439). Moving the tests will be done as part of that ticket

## How to test

I have tested the feature on staging (see screenshots)

To test locally:
- Set up openhexa local testing:https://github.com/BLSQ/iaso-pipelines/wiki/Setup-Iaso-to-test-OpenHexa-Polio-integration
- Go to django Admin>JsonDatastores
- Chose a an lqas  datastore
- Chose a campaign. Delete some district data
- Go to polio>lqas>country
- Select a campaign and click refresh:
    - The button should enter loading state and the last run status should be "Running" 
    - Go to Tasks: you should see the task running with the name: "lqas-pipeline-config-<countryID>"
    - Go to OpenHexa: you should see a pipeline running. Open the pipeline, it should have the same taskId as the task in iaso/polio
- Once the pipeline stops running (can take several miniutes, it's a separate issue):
    - Check in Openhexa that there were no errors, in particular no errors coming from iaso when:
        - Trying to write in the datastore
        - Trying to update the task status
        - Trying to write in the campaign 
    - Check in iaso:
        - Task status should be updated
        - The map should refresh without needing to reload 

## Print screen / video

The (renamed) task ran successfully (staging)
Screenshot 2024-01-23 at 15 00 21

It launched the pipeline successfully (statging)
Screenshot 2024-01-23 at 15 00 36

The data was refreshed (see status upper right)
Screenshot 2024-01-23 at 15 00 51

## Note 

Merges in https://github.com/BLSQ/iaso/pull/1062

The doc is a bit vague on some implementation details because I wanted it to fit the current implementation as well

Better not to merge before [POLIO_1439](https://bluesquare.atlassian.net/browse/POLIO-1439) is done, since the feature won't work outside of polio


iaso/api/tasks/__init__.py
```diff
@@ -70,3 +79,211 @@ def get_queryset(self):
...
+        if id_field_count > 1:
+            error["non_field_errors"] = f"id_field should only have one property, got {id_field_count}"
+
+        if error:
+            raise serializers.ValidationError(error)
+
+        res = {**validated_data}
+        res["slug"] = slugify(slug)
+        res["config"] = config
+        res["id_field"] = id_field
```
mathvdh, line 140: I think this is not required as this data is in `validated_data` and has the same name

iaso/api/tasks/__init__.py
```diff
@@ -70,3 +79,211 @@ def get_queryset(self):
...
+        """
+            % (pipeline)
+        )
+        try:
+            latest_runs = client.execute(get_runs)
+            # Warning the query will only return the last 10 runs
+            active_runs = [
+                run
+                for run in latest_runs["pipeline"]["runs"]["items"]
+                if (run["status"] not in ["queued", "success", "failed"])
```
mathvdh: What about using constants instead of string for "queued" et al.

iaso/api/tasks/__init__.py
```diff
@@ -70,3 +79,211 @@ def get_queryset(self):
...
+    slug = serializers.CharField()
+    config = data = serializers.JSONField()
+    id_field = data = serializers.JSONField()
+
+    def validate(self, attrs):
+        validated_data = super().validate(attrs)
+        request = self.context["request"]
+        slug = request.data.get("slug", None)
+        config = request.data.get("config", None)
+        id_field = request.data.get("id_field", None)
```
mathvdh: Are these fields optional ?

iaso/api/tasks/__init__.py
```diff
@@ -70,3 +79,211 @@ def get_queryset(self):
...
+        error = {}
+
+        if slug is None:
+            error["slug"] = "This field is mandatory"
+
+        id_field_count = len(id_field.keys())
```
mathvdh, line 130: This will crash if id_field is None.

iaso/api/tasks/__init__.py
mathvdh: In this case you don't need to unpack them at the beginning of this function ...

plugins/polio/tasks/api/refresh_lqas_data.py
```diff
@@ -42,34 +33,40 @@ def validate(self, attrs):
...
+        slug = validated_data.get("slug", None)
+        config = validated_data.get("config", None)
+        id_field = validated_data.get("id_field", None)
+        error = {}
+        if slug is None or slug != LQAS_CONFIG_SLUG:
+            error["slug"] = f"Wrong config slug. Expected {LQAS_CONFIG_SLUG}, got {slug}"
+        if config is None:
+            error["config"] = "This field is mandatory"
+        if id_field is None:
+            error["id_field"] = "This field is mandatory"
```
mathvdh, line 62: Aren't these checks already done by `ExternalTaskPostSerializer` ?

mathvdh (changes requested):
A few minor issues otherwise it's good

iaso/api/tasks/__init__.py
quang-le: Well spotted, thanks!

plugins/polio/tasks/api/refresh_lqas_data.py
quang-le: Only for `slug` really. `config` and `id_field` are not mandatory, as their value depend on the OpenHexa pipeline. 
I added stricter checks here, which are specific to the lqas pipeline

iaso/api/tasks/__init__.py
quang-le: only `slug` is always mandatory.
`id_field` is used when using OpenHexa API to check for active runs for a specific country, in this case `id_field` is `{country_id:12646}`, but we may not always have to perform such a specific check

`config` is a dict regrouping the arguments of that need to be passed to the pipeline. It's possible to have pipeline that don't take arguments

iaso/api/tasks/__init__.py
quang-le: It comes from OpenHexa and is only used there (Iaso's are uppercase), so I don't know if there's much to gain

iaso/api/tasks/__init__.py
quang-le: It breaks the test if I remove them. I'm getting the values from the request data, not from validated_data. I think it's because the fields don't exist on the model or something

mathvdh (approved):
Ok .. Good for me

PR #1094, opened 2024-01-31
POLIO-1461: The analytics script is not executed on the embedded calendar

hakifran:
- The analytics script is not executed on the embedded calendar
- This PR is about to fix it
Related JIRA tickets : [POLIO-1461](https://bluesquare.atlassian.net/browse/POLIO-1461)

- [ ] Are there enough tests

## Changes
- Get all analytics from all accounts
- Add the analytics script to django template
## How to test
-Go on the embedded calendar and check if the analytics script is running
## Print screen / video
[Screencast from 2024-01-31 14-59-50.webm](https://github.com/BLSQ/iaso/assets/19631540/f1752d8b-963d-45ec-a282-786040fe2f6b)

[POLIO-1461]: https://bluesquare.atlassian.net/browse/POLIO-1461?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/hat/dashboard/views.py
+++ b/hat/dashboard/views.py
@@ -7,22 +7,28 @@ from django.shortcuts import render
 from django.views.decorators.http import require_http_methods
 
 from hat.__version__ import VERSION
+from iaso.models.base import Account
 
 
-def _base_iaso(request: HttpRequest) -> HttpResponse:
+def _base_iaso(request: HttpRequest, analytics_scripts: list[str] = []) -> HttpResponse:
     try:
         USER_HOME_PAGE = request.user.iaso_profile.home_page if request.user.is_authenticated else ""
     except ObjectDoesNotExist:
         USER_HOME_PAGE = ""
+    variables_to_render = {
+        "PLUGINS_ENABLED": settings.PLUGINS,
+        "STATIC_URL": settings.STATIC_URL,
+        "USER_HOME_PAGE": USER_HOME_PAGE,
+        "VERSION": VERSION,
+    }
+
+    if analytics_scripts:
+        variables_to_render["ANALYTICS_SCRIPTS"] = list(analytics_scripts)
+
     return render(
         request,
         "iaso/index.html",
-        {
-            "PLUGINS_ENABLED": settings.PLUGINS,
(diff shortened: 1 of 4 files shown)


hat/templates/iaso/base.html
```diff
@@ -26,6 +26,11 @@
   <meta property="og:image" content="" />
   <meta property="og:url" content="{{ request.build_absolute_uri }} " />
   {% endif %}
+  {% if ANALYTICS_SCRIPTS %}
+    {% for analytic_script in ANALYTICS_SCRIPTS %}
```
madewulf, line 30: You can't have all the analytics script of all accounts displayed here, that defeats the point of having a different script per account

hat/dashboard/views.py
```diff
@@ -35,7 +41,10 @@ def iaso(request: HttpRequest) -> HttpResponse:
 @require_http_methods(["GET"])
 def embeddable_iaso(request: HttpRequest) -> HttpResponse:
     """Embeddable iaso page without login requirement and with correct header"""
-    response = _base_iaso(request)
+    all_analytics_scripts = set(
+        [account.analytics_script for account in Account.objects.all() if account.analytics_script]
```
mathvdh, line 45: It's weird to have ALL the analytics_script from all the Accounts loaded ? Or I don't understand somelthing

mathvdh (changes requested):
See my comment, otherwise I think it's good.

hat/templates/iaso/base.html
mathvdh: Ah this seems to be about the same thing I'm commenting below

hakifran: As you can see the analytics script will be added only for the embedded calendar, which is accessible by not authenticated user. And on the embedded calendar there can be displayed campaigns from different accounts, that's why I tried to get all the analytics script from all accounts. I don't know if it's clear or if you want we can have a 5min call. thanks

madewulf (approved):

PR #1096, opened 2024-02-01
IA-2621:  Org unit change request parent shape

beygorghor:
Add alert on whether the change request location falls within the parent's shape or not

Related JIRA tickets : IA-2621

- [ ] Are there enough tests

## Changes

- fixed typescript issues for react-leaflet and removed ts-ignore
- added geo_json on parent org unit in api
- hook to see if a location is within a shape
- customise markermap component to get a parent org unit and display it on map
- add popup if marker is not inside his parent shape
- covering edge case when you change parent too

## How to test

- make sure you have an org unit with a parent
- create a change request with this org unit and change the location  with something like this SRID=4326;POINT Z (23.0449 -2.9820 0)
- open change request dialog and check that everything is fine
- if you location is within the parent shape, edit the change request new location value and change lat or long to 0

## Print screen / video

Screenshot 2024-02-01 at 12 51 05

### DIFF
--- a/hat/assets/js/apps/Iaso/components/maps/hooks/useMarkerWithinBounds.ts
+++ b/hat/assets/js/apps/Iaso/components/maps/hooks/useMarkerWithinBounds.ts
@@ -1,0 +1,20 @@
+import { useState, useEffect } from 'react';
+import L from 'leaflet';
+import { ShortOrgUnit } from '../../../domains/orgUnits/types/orgUnit';
+
+export const useMarkerWithinBounds = (
+    latitude: number | undefined,
+    longitude: number | undefined,
+    geoJson: ShortOrgUnit['geo_json'],
+): boolean => {
+    const [isMarkerInside, setIsMarkerInside] = useState(true);
+
+    useEffect(() => {
+        if (geoJson && latitude !== undefined && longitude !== undefined) {
+            const bounds = L.geoJSON(geoJson).getBounds();
+            setIsMarkerInside(bounds.contains(L.latLng(latitude, longitude)));
+        }
+    }, [latitude, longitude, geoJson]);
+
+    return isMarkerInside;
+};
--- a/hat/assets/js/apps/Iaso/components/maps/messages.js
+++ b/hat/assets/js/apps/Iaso/components/maps/messages.js
@@ -39,7 +39,8 @@ const MESSAGES = defineMessages({
     },
     locationNotInShape: {
         id: 'iaso.map.locationNotInShape',
-        defaultMessage: 'The location is not within the bounds of the shape.',
+        defaultMessage:
+            "The location is not within the bounds of the parent's shape.",
     },
 });
 
--- a/hat/assets/js/apps/Iaso/components/maps/messages.js
+++ b/hat/assets/js/apps/Iaso/components/maps/messages.js
@@ -33,6 +33,14 @@ const MESSAGES = defineMessages({
         id: 'iaso.map.title.markerClustering',
         defaultMessage: 'Clustering',
     },
+    parent: {
+        id: 'iaso.label.parent',
+        defaultMessage: 'Parent',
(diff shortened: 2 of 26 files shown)


iaso/api/org_unit_change_requests/serializers.py
quang-le, line 1: Isn't there a python test to update with this change?

hat/assets/js/apps/Iaso/domains/app/translations/en.json
```diff
@@ -717,6 +717,7 @@
     "iaso.map.location": "Location",
     "iaso.map.locationLimit": "Map results limit",
     "iaso.map.locationLimit.error": "Please enter a correct value",
+    "iaso.map.locationNotInShape": "The location is not within the bounds of the shape.",
```
quang-le: " the parent's shape"?

hat/assets/js/apps/Iaso/components/maps/messages.js
```diff
@@ -33,6 +33,14 @@ const MESSAGES = defineMessages({
         id: 'iaso.map.title.markerClustering',
         defaultMessage: 'Clustering',
     },
+    parent: {
+        id: 'iaso.label.parent',
+        defaultMessage: 'Parent',
+    },
+    locationNotInShape: {
+        id: 'iaso.map.locationNotInShape',
+        defaultMessage: 'The location is not within the bounds of the shape.',
```
quang-le: idem: parent shape

hat/assets/js/apps/Iaso/components/maps/hooks/useMarkerWithinBounds.ts
quang-le, line 1: ❤️

hat/assets/js/apps/Iaso/domains/orgUnits/reviewChanges/hooks/useNewFields.tsx
```diff
@@ -86,8 +108,8 @@ export const useNewFields = (
             setFields([]);
             return;
         }
-        const orgUnit = changeRequest.org_unit;
-        const fieldDefinitions = {
+        type FieldDefinitions = Record<string, FieldDefinition>;
```
quang-le: This type definition should be moved

hat/assets/js/apps/Iaso/domains/orgUnits/reviewChanges/hooks/useNewFields.tsx
```diff
@@ -139,32 +166,56 @@ export const useNewFields = (
...
+                    false,
+                );
+            }
+
+            return {
+                oldValue,
+                newValue,
+                isChanged,
+            };
+        };
```
quang-le: This function definition should be moved out of the `useMemo` hook

hat/assets/js/apps/Iaso/domains/orgUnits/reviewChanges/hooks/useNewFields.tsx
quang-le, line 1: The `useMemo`in this hook is used like a `useEffect`. 
`useMemo` is supposed to return a value, so it's confusing that it's not. 

From what I gather, it's used to calculate the value of `fields`, so it should return the value, and `setFields` should be called in a `useEffect` that depends on the return value of `useMemo`. 

That would make the code easier to understand I think

quang-le (changes requested):
Feature works like a charm!
I'm just blocking on the python api test and the useMemo thing

iaso/api/org_unit_change_requests/serializers.py
beygorghor: you are right, the tests are not using / testing nested org unit type. I'll see if i have time to do it

quang-le (approved):

PR #1117, opened 2024-02-07
IA-2679 PLANNING refactor

beygorghor:
Initial demand was to review the process of  parent picking mode for the list view.

Finally i reviewed the whole page to make it more easy to understand

Related JIRA tickets : IA-2679

- [ ] Are there enough tests

## Doc

## Changes

- pass org unit type id and name while getting an org unit with parents
- use org unit type id and name in list columns
- remove parent picking option
- split way to select parent in map view and list view (dropdown on map and check box on list)
- highlight column when came org unit type is selected
- move tabs at a lower level

## How to test

Open a planning, drill down to a lower level org unit (village, hopital)
Open list view and check one of the column, click on a row, you should be able to assign all the org unit in the org unit type selected to the current team / user.
Check the same behaviour on the map

## Print screen / video
[
](https://github.com/BLSQ/iaso/assets/12494624/8f170805-9bf1-4742-ab9e-f1f7cc3c642b)
## Notes

- right click should be removed and replaced by a sticky tooltip => https://bluesquare.atlassian.net/browse/IA-2685
- on a sortable column, when you leave a your mouse on a head cell, a title appears saying you can sort this column, even if you are hovering a button or a checkbox in this case


iaso/api/serializers.py
```diff
@@ -159,7 +159,13 @@ class Meta:
 class OrgUnitSearchParentSerializer(OrgUnitSerializer):
     class Meta:
         model = OrgUnit
-        fields = ["id", "name", "parent"]
+        fields = [
+            "id",
+            "name",
+            "parent",
+            "org_unit_type_id",
+            "org_unit_type_name",
```
madewulf, line 167: We need a prefetch_related on tha field

hat/assets/js/apps/Iaso/components/maps/tools/CustomTileLayer.tsx
```diff
@@ -1,16 +1,19 @@
 import React, { FunctionComponent, useRef, useEffect, RefObject } from 'react';
 import { TileLayer, useMap } from 'react-leaflet';
+import { SxProps, Theme } from '@mui/material';
 import { Tile, TilesSwitchControl } from './TilesSwitchControl';
 
 type Props = {
     currentTile: Tile;
     // eslint-disable-next-line no-unused-vars
     setCurrentTile: (newTile: Tile) => void;
+    styles?: SxProps<Theme>;
```
quang-le, line 10: nitpick: we could do like we used to with `className` and name this prop `sx`

hat/assets/js/apps/Iaso/components/nav/TopBarComponent.js
```diff
@@ -65,7 +65,12 @@ function TopBar(props) {
 
     return (
         <>
-            <AppBar position="relative" color="primary" id="top-bar">
+            <AppBar
+                position="relative"
+                color="primary"
+                id="top-bar"
+                sx={{ zIndex: 10 }}
```
quang-le, line 72: I wonder what kind of bug you had that made this necessary 🤔

hat/assets/js/apps/Iaso/domains/assignments/components/AssignmentsListTab.tsx
```diff
@@ -69,52 +83,96 @@ export const AssignmentsListTab: FunctionComponent<Props> = ({
     setParentSelected,
 }: Props) => {
     const columns = useColumns({
-        orgUnits,
+        orgUnits: orgUnits || [],
```
quang-le, line 86: type name should be upper case
```suggestion
        orgUnits: OrgUnits || [],
```

quang-le (approved):
Some minor bugs that I think were already there. I'll put them in new tickets.
The UI is much more legible now, nice job!

iaso/api/serializers.py
kemar: @beygorghor I think what we really need is a `select_related()`, not a `prefetch_related()`, because [`"org_unit_type_name"`](https://github.com/BLSQ/iaso/blob/38ea50276cd283c99623d9ceeca530e4a75c3dd3/iaso/api/serializers.py#L167) is calling [`def get_org_unit_type_name()`](https://github.com/BLSQ/iaso/blob/38ea50276cd283c99623d9ceeca530e4a75c3dd3/iaso/api/serializers.py#L84-L85) and it's a foreign key.

We need to optimize the view where [`OrgUnitSearchSerializer` is called](https://github.com/BLSQ/iaso/blob/c9f6f24001d7d2648f284d197bca4c63024ac391/iaso/api/org_units.py#L156).

The only problem is that this view is a bit of a mess in terms of where the query is performed…

So we could [replace this line](https://github.com/BLSQ/iaso/blob/c9f6f24001d7d2648f284d197bca4c63024ac391/iaso/api/org_units.py#L141):

```python
queryset = queryset.order_by(*order)
```

by:

```python
queryset = queryset.select_related("parent__org_unit_type").order_by(*order)
```

kemar: @beygorghor this wasn't the right place, my bad. I fixed it here https://github.com/BLSQ/iaso/pull/1117/commits/0464a4867ad98ffeaf957f727ec86d7996ac7b7e

Don't forget to pull.

PR #1120, opened 2024-02-07
POLIO-1474 For Vaccine Supply Chain a VRF shouldnt exist without rounds

mathvdh:
For Vaccine Supply Chain, a Vaccine Request Form without rounds shouldn't be able to exist.
Apparently on staging and in some local install it does and that crashes the Supply chain list page.
This is a fix to make it more robust ..

Related JIRA tickets : POLIO-1474

## Changes

Handle case when no rounds with `if` and `Coalesce`

## How to test

If polio staging is working after that, it's working.
Otherwise locally it's difficult to test as now even django admin doesn't allow you to create a VRF without rounds.


quang-le (approved):
It's a bit weird, but since neither the front nor the API will allow the case to happen...

PR #1122, opened 2024-02-07
POLIO-1458 add target population field to Vaccine Request Form

mathvdh:
The target population of the VRF is not the same as for the Risk assessment (which is stored in “Campaigns”) - it can vary

We should therefore distinguish it from the target pop of the RA is the field name and label and create a new field 

Related JIRA tickets : POLIO-1458

- [x] Are there enough tests

## Changes

- Add the field on the VRF model
- Related migrations
- Change APIs in accordance
- Found problem with PATCH api that forces you to change stuff (round/campaigns) that you might not want to change.
- Add tests

## How to test

See tests.


plugins/polio/api/vaccines/supply_chain.py
```diff
@@ -22,14 +22,15 @@
 AR_SET = "vaccinearrivalreport_set"
 
 
-def validate_rounds_and_campaign(data, current_user=None):
+def validate_rounds_and_campaign(data, current_user=None, force_rounds=True, force_campaign=True):
```
kemar, line 25: If I'm not mistaken, this is what you are doing:

```python
def validate_rounds_and_campaign(data, current_user=None, create=True):

    if create and not rounds_data:
        raise

    if create and not campaign_obr_name:
        raise
```

But I don't understand why rounds and campaing are required at creation time, but optional at update time?

kemar: > But I don't understand why rounds and campaing are required at creation time, but optional at update time?

So based on our discussion, I understand that `data.get("rounds")` and `data.get("campaign")` can be None in case of a `PATCH` request.

Suggestions to make it easier to understand:

- add a comment that explain this
- or replace `force_rounds=True` and `force_campaign=True` by a single argument `create=True`

kemar (approved):
LGTM

I have a comment but nothing blocking.

PR #1134, opened 2024-02-13
TRIA-117 Allow superuser to export mobile app setup for a user

bramj:
This PR adds a feature that allows a superuser to download a .zip file for a given user and project.
This zip file contains all the files needed to do the complete setup of the mobile app. It contains the user's profile, an access token and .json files for all resources such as the org unit pyramid and the entities, based on the enabled feature flags.

- ~My migrations file are included~
- [x] Are there enough tests

## Doc

## Changes

- Added an action (icon button) on the users list if the authorized user is a super user
- Generating the export will:
  1. Launch a Task
  2. Poll the task for it's status and progress
  3. When the task has completed successfully, request a signed url (S3) for the generated file
  4. Display the download button
- The task will call all API endpoints and write the responses to files in a temporary directory. When done it adds all files to an archive and uploads to S3.

## How to test

1. Set `DNS_DOMAIN` Env variable to your local machine. Note: I had to use `ngrok` to get it to work because it didn't find my localhost from inside the docker container. Using `ngrok` was the faster solution.
2. Launch the tasks_worker:
  ```
  docker-compose run --rm iaso bash
  python manage.py tasks_worker
  ```
3. Make sure you're a super user and use the dashboard to generate an export

## Print screen / video

https://github.com/BLSQ/iaso/assets/2719739/4774f879-a986-4714-9eaf-584e173e3ec3


bramj:
General thoughts:
- Personally I find the default MUI button less than great, maybe we can use the "contained" variant everywhere? https://mui.com/material-ui/react-button/#basic-button
- I think the title of our dialog is missing some style (e.g. the bold part): https://mui.com/material-ui/react-dialog/#alerts It's a detail, but it does look better on the MUI docs. But I'm not sure how to fix it

hat/assets/js/apps/Iaso/domains/tasks/components/TaskDetails.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+    const [presignedUrl, setPresignedUrl] = useState(null);
+
+    const taskHasDownloadableFile =
+        task.result && task.result.data && task.result.data.startsWith('file:');
+
+    useEffect(() => {
+        if (taskHasDownloadableFile) {
+            getRequest(`/api/tasks/${task.id}/presigned-url/`).then(resp =>
+                setPresignedUrl(resp.presigned_url),
```
beygorghor: Why not react-query hook enabled if task.id and watching it ?

hat/assets/js/apps/Iaso/domains/tasks/index.js
```diff
@@ -108,6 +109,10 @@ const Tasks = ({ params }) => {
                     baseUrl={baseUrl}
                     extraProps={{
                         loading: isLoading,
+                        SubComponent: task => {
+                            console.log('task', task);
```
beygorghor: 🗑️

hat/assets/js/apps/Iaso/domains/users/components/ExportMobileAppSetupDialog.tsx
```diff
@@ -0,0 +1,265 @@
...
+    onCreateExport: MutateFunction<any>;
+    isOpen: boolean;
+    closeDialog: () => void;
+};
+
+function LinearProgressWithLabel(
+    props: LinearProgressProps & { value: number },
+) {
+    return (
+        <Box sx={{ display: 'flex', alignItems: 'center' }}>
```
beygorghor: this is a FunctionComponent

beygorghor: also you can use proper sxStyles for it

hat/assets/js/apps/Iaso/domains/users/components/ExportMobileAppSetupDialog.tsx
```diff
@@ -0,0 +1,265 @@
...
+                                setSelectedProject({
+                                    id: value,
+                                })
+                            }
+                            value={selectedProject?.id}
+                            type="select"
+                            multi={false}
+                            label={MESSAGES.project}
+                            options={selectedUser.projects.map(project => {
+                                return {
```
beygorghor: maybe a useMemo for this

beygorghor (changes requested):
Some small changes

hat/assets/js/apps/Iaso/domains/users/components/ExportMobileAppSetupDialog.tsx
bramj: > this is a FunctionComponent

Sorry, but what do you mean by this comment? Not sure what I should change

iaso/tasks/export_mobile_app_setup_for_user.py
```diff
@@ -0,0 +1,282 @@
...
+        iaso_client.authenticate_with_token(str(refresh.access_token))
+
+    the_task.report_progress_and_stop_if_killed(progress_value=2)
+
+    for call in API_CALLS:
+        _get_resource(iaso_client, call, tmp_dir, project.app_id, feature_flags)
+        the_task.report_progress_and_stop_if_killed(progress_value=the_task.progress_value + 1)
+
+    s3_object_name = _compress_and_upload_to_s3(tmp_dir, export_name, password)
+    print("s3_object_name ")
```
mathvdh: I know we are in a task but do we want to keep "print" ? I mean don't we have to use the logging system instead ? 🤔

iaso/tasks/export_mobile_app_setup_for_user.py
```diff
@@ -0,0 +1,282 @@
...
+        response = None
+        # S3 urls contain a signature and don't work with additional auth headers
+        if "s3.amazonaws" in url:
+            response = requests.get(url)
+        else:
+            response = requests.get(url, headers=iaso_client.headers)
+
+        with open(os.path.join(tmp_dir, "forms", filename), mode="wb") as f:
+            f.write(response.content)
+
```
mathvdh: This could be eventually put in a function and reused on lines 246 -> 255

mathvdh (changes requested):
I think this is really great ! 👍🏼 
I have small comments and suggestions below, but otherwise it's really good.
Some questions I have : 
I like the API_CALLS file ... it's really SMART.
I'm not sure how it works on import in mobile app (like I can' t really test it).
BUT would it be possible in the future to import such a zip file locally ? Like from server A of iaso to server B (if we would add like a screen where you can select with check boxes ;; export this and that ;; then import them on another server it would be a killer feature

iaso/tasks/export_mobile_app_setup_for_user.py
bramj: You're right, this is actually a debug leftover, good catch

iaso/tasks/export_mobile_app_setup_for_user.py
bramj: 👍 done, I think I was a bit lazy because of the small difference with the url/path in case of a non S3 url

bramj:
> I think this is really great ! 👍🏼 I have small comments and suggestions below, but otherwise it's really good. Some questions I have : I like the API_CALLS file ... it's really SMART. I'm not sure how it works on import in mobile app (like I can' t really test it). BUT would it be possible in the future to import such a zip file locally ? Like from server A of iaso to server B (if we would add like a screen where you can select with check boxes ;; export this and that ;; then import them on another server it would be a killer feature

Thanks for the kind words and the review :) You mean if we could import such a zip on the backend of another server? I suppose there's not really a reason why not 🤔 Though if we would add such a feature maybe it would make more sense to enter an API endpoint (pointing to server A) on your server B and have it pull in the data with a sort of setuper.

But with this zip file the data is specific to the mobile app, so I suppose you'd also miss a lot of data like fields on the OrgUnits that are simply not in the mobile payload.

Do you think there's a demand for a server A to server B import?

bramj:
@mathvdh Okay for merge

mathvdh (approved):

beygorghor (approved):

PR #1141, opened 2024-02-15
WC2-331 optional name on Entity. Ameliorate Admin

mathvdh:
Impossibility to "undelete" a `Entity` via django admin.
When soft deleting via admin, force to fill in the "name" field of the `Entity` which is not used anyway (name is taken from attributes form instance).

Related JIRA tickets : WC2-331

## Changes

- Make the field "name" optional on `Entity` model (it's not used anyway).
- As a results, when soft deleting an object via django admin, you are not forced to enter a fake name to save it.
- Add possibility to list soft deleted objects in django admin and to undelete them.
- Add possibility to filter django admin list by this status.
- Add possibility to search in the entities and via their attributes values (json content). So you can now search by an entity's actual name.

## How to test

Explain how to test your PR.
Visit django admin, go to entities, soft delete an entity then un-soft delete it.
Search by name, filter, etc.

## Print screen / video

Screenshot 2024-02-15 at 16 30 30


iaso/models/entity.py
```diff
@@ -149,7 +150,7 @@ class Entity(SoftDeletableModel):
     submission
     """
 
-    name = models.CharField(max_length=255)
+    name = models.CharField(max_length=255, blank=True, null=True)
```
kemar: Not blocking but just to let you know: the Django convention is to avoid using `null=True` with string-based fields (https://docs.djangoproject.com/en/4.2/ref/models/fields/#null).

kemar (approved):

iaso/admin.py
```diff
@@ -414,6 +414,8 @@ class SourceVersionAdmin(admin.ModelAdmin):
 @admin.register(Entity)
 @admin_attr_decorator
 class EntityAdmin(admin.ModelAdmin):
+    search_fields = ["account__name", "entity_type__name", "attributes__json"]
```
bmonjoie, line 417: It could be really useful to search on `attributes__id` and `attributes__uuid` as well.

If I'm not mistaken, `attributes__uuid` will allow us to find entities by their UUIDs

iaso/migrations/0262_alter_entity_name.py
```diff
@@ -0,0 +1,17 @@
...
+    dependencies = [
+        ("iaso", "0261_config"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="entity",
+            name="name",
+            field=models.CharField(blank=True, max_length=255),
+        ),
```
bmonjoie, line 16: Shouldn't we remove it completely?

PR #1143, opened 2024-02-16
POLIO-1479 Correct default values for doses per vials

mathvdh:
Wrong value in default values for "doses per vials" for Polio.

But it should be 
nOPV2 : 50 doses
bOPV : 20 doses
mOPV2 : 20 doses

Related JIRA tickets : POLIO-1479

## Changes

Little correction in backend
Add a mechanism to get default value based on current vaccine in VRF for frontend.


hakifran (approved):
Seems good to me

PR #1152, opened 2024-02-21
POLIO-1488 hide lot numbers info in vaccine stock management

mathvdh:
In the current status it has been asked to hide all references to lot numbers from Vaccine Stock Management.

Related JIRA tickets : POLIO-1488

## Changes

Commented out the occurences in the frontend as it might be asked to put those back soon.
Didn't change the backend as this fields were already optional.

## How to test

Go to to Frontend > Vaccine Stock Management, Choose a Country/Vaccine line and click on the eye. Then on the next window, check the list, it shouldn't mention any lot number anywhere. Then click "Stock Variation" and again check everywhere and there should not be any mentions of lot numbers.

### DIFF
--- a/plugins/polio/js/src/domains/VaccineModule/StockManagement/StockVariation/Modals/validation.ts
+++ b/plugins/polio/js/src/domains/VaccineModule/StockManagement/StockVariation/Modals/validation.ts
@@ -15,8 +15,8 @@ yup.addMethod(
                 const valuesArray = Array.isArray(value)
                     ? value
                     : value
-                          .split(',')
-                          .map((v: string | number) => `${v}`.trim());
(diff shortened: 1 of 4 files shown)


plugins/polio/js/src/domains/VaccineModule/StockManagement/StockVariation/Modals/validation.ts
```diff
@@ -44,7 +44,7 @@ export const useFormAValidation = () => {
         lot_numbers: yup
```
hakifran, line 44: Why not comment the whole lot_numbers validation as it's not used for now? But not blocking

hakifran (approved):

PR #1154, opened 2024-02-22
POLIO-1443: Remove test campaigns from the Campaigns CSV extract 

hakifran:
- When there is no filter, it generates csv with test campaigns
- This PR is about to fix that

Related JIRA tickets : [POLIO-1443](https://bluesquare.atlassian.net/browse/POLIO-1443)

- [ ] Are there enough tests

## Changes
- Rename the campaign type 'Test campaigns' into 'Test campaign/on hold'
- Add filterLaunched param to confirm when the button filter has been pressed
- Make showTest to be false or true depending on the value of filterLaunched

## How to test
- Go on campaigns page 
- Download csv without any filter(there should not be test campaigns)
- Download csv with filter campaign_type at all(there should be test campaigns)

## Print screen / video
Screenshot from 2024-02-22 16-54-45

[POLIO-1443]: https://bluesquare.atlassian.net/browse/POLIO-1443?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


plugins/polio/js/src/domains/Calendar/campaignCalendar/Filters.js
```diff
@@ -113,9 +116,34 @@ const Filters = ({
...
+        search,
+        showOnlyDeleted,
+    ]);
     useEffect(() => {
         setFiltersUpdated(true);
+        if (filtersFilled()) {
+            setFilterLaunched(true);
+        } else {
+            setFilterLaunched(false);
+        }
```
beygorghor: This looks a bit overkill to use a local state and change it with a useeffect. You can set the url parma to true while you execute the search instead no ?

beygorghor (changes requested):

plugins/polio/js/src/domains/Calendar/campaignCalendar/Filters.js
hakifran: Thank you. Good observation

plugins/polio/js/src/domains/Calendar/campaignCalendar/Filters.js
```diff
@@ -63,6 +63,28 @@ const Filters = ({
         dateApiToDateRangePicker(params.roundStartTo),
     );
 
+    const filtersFilled = useCallback(() => {
```
beygorghor: cool, no i have the sentiment that you can remove totally the usecallback, the value returned is a boolean and don't even need to be memofied, just make it a bool

beygorghor (approved):

PR #1156, opened 2024-02-23
IA-2708 - Allow to retrieve OrgUnits via their IDs

bmonjoie:
In the application, it is possible to retrieve a Change Request that is linked to an OrgUnit not returned by the regular call to `/api/mobile/orgunits/?app_id=...`. This endpoint allows someone to retrieve OrgUnits by their ID.

Related JIRA tickets : IA-2708

- [X] Are there enough tests

## Doc
- Docstring

## Changes

Add a check for the query parameter `ids` to return only those listed in the parameter.

## How to test

Call `/api/mobile/orgunits/?app_id=...&ids=...` to retrieve only the list of OrgUnit specified.

## Notes

I haven't considered all the potential security risks of this endpoint. 
Someone with a better knowledge of that matter should probably have a look at this endpoint.


mathvdh (approved):
Ok

PR #1162, opened 2024-02-27
Cypress webpack preprocessor

beygorghor:
Cypress test are crashing [here](https://github.com/BLSQ/iaso/actions/runs/8064975732/job/22030059169) because new PhoneInput is using custom css and not specific loader is set to process it.

## Changes

- added a preprocessor to resolve css with style-loader and css-loader

## How to test

https://github.com/BLSQ/iaso/actions/runs/8067667341


mathvdh (approved):
Ok got it

PR #1166, opened 2024-03-04
POLIO-1126 Multi Round Budget (Frontend)

kemar:
Polio Multi Round Budget front-end (with a lot of back-end code though).

Related JIRA tickets : [POLIO-1126](https://bluesquare.atlassian.net/browse/POLIO-1126)

## Doc

https://whimsical.com/multi-round-budget-FdiBzSUsuozi8aAnuvJWAP

## Scope of this PR

As discussed, this PR **does NOT** include the `Campaign -> Edit Campaign -> Budget` tab.

It will be done in another PR.

done

## Changes

- `polio/campaigns` route
    - fix create/update for the new `Round.budget_process` field
    - prevent deletion of `Round` objects linked to budget processes

- Rename `BudgetCampaign` to `BudgetProcess`

- `BudgetProcess` payload
    - include nested `Round` objects VS just their numbers

- UI for `BudgetProcess`
    - *list*
        - fix filters
        - fix ordering
    - *read*
        - fix go to next step
        - fix override step
    - *create*
        - the UI includes a list of dependent dropdowns
    - *update*
        - the UI includes a list of dependent dropdowns
    - *soft* delete

## Print screen / video

https://github.com/BLSQ/iaso/assets/281139/4a277f8d-0e5e-471c-9a59-2f10c420bf28

[POLIO-1126]: https://bluesquare.atlassian.net/browse/POLIO-1126?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


plugins/polio/budget/api.py
```diff
@@ -41,43 +41,56 @@ class BudgetCampaignViewSet(ModelViewSet, CSVExportMixin):
...
+        """
+        # Soft delete `BudgetProcess`.
+        budget_process = self.get_object()
+        self.perform_destroy(budget_process)
+        # Soft delete `BudgetStep`s.
+        budget_process.budget_steps.all().delete()
+        # Reset `Rounds`s FKs so that they can be linked to a new `BudgetProcess`.
+        budget_process.rounds.update(budget_process=None)
+        return Response(status=status.HTTP_204_NO_CONTENT)
+
```
quang-le, line 78: Maybe we could have some safeguards to avoid deleting e.g. completed processes.
It's tricky because the exact state of the process is determined by the json file, and we want to keep that configurable.
Do you think we could add a `block_deletion` field in the json config (in the `nodes`) and have use that to check if we should allow to delete the process?

plugins/polio/js/src/domains/Budget/BudgetDetails/BudgetDetails.tsx
```diff
@@ -32,10 +32,10 @@ type Props = {
     router: any;
 };
 
-export const BudgetDetails: FunctionComponent<Props> = ({ router }) => {
+export const BudgetProcessDetails: FunctionComponent<Props> = ({ router }) => {
     const { params } = router;
+    const { campaignName, budgetProcessId, transition_key, ...rest } = params;
     const classes = useStyles();
-    const { campaignName, campaignId, transition_key, ...rest } = router.params;
     const [showHidden, setShowHidden] = useState<boolean>(
```
quang-le, line 39: I should have put a redirection here at the time. The hiding feature works, but it's not deep-linked. 
Not blocking. 
I mention it in the review so we can regroup these legacy bugs in a separate ticket

plugins/polio/js/src/domains/Budget/BudgetProcess/CreateBudgetProcessModal.tsx
```diff
@@ -0,0 +1,124 @@
...
+        [],
+    );
+    useEffect(() => {
+        const campaigns = dropdownsData?.campaigns || [];
+        const filtered = campaigns.filter(
+            i => String(i.country_id) === String(formik.values.country),
+        );
+        formik.setFieldValue('campaign', '');
+        setCampaignOptions(filtered);
+    }, [formik.values.country]);
```
quang-le: unless it breaks the interface, good practice would require to include `dropdownsData?.campaigns` in the dependencies array of `useEffect`
```suggestion
    }, [dropdownsData?.campaigns, formik.values.country]);
```

plugins/polio/js/src/domains/Budget/BudgetProcess/CreateBudgetProcessModal.tsx
```diff
@@ -0,0 +1,124 @@
...
+    // Filter "Rounds" values on "Campaign" change.
+    const [currentRoundsOptions, setRoundsOptions] = useState<Options[]>([]);
+    useEffect(() => {
+        const rounds = dropdownsData?.rounds || [];
+        const filtered = rounds.filter(
+            i => String(i.campaign_id) === String(formik.values.campaign),
+        );
+        formik.setFieldValue('round', '');
+        setRoundsOptions(filtered);
+    }, [formik.values.campaign]);
```
quang-le: Same

```suggestion
    }, [dropdownsData?.rounds, formik.values.campaign]);
```

plugins/polio/js/src/domains/Budget/BudgetProcess/CreateBudgetProcessModal.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+    const titleMessage = formatMessage(MESSAGES.createBudgetProcessTitle);
+
+    return (
+        <>
+            {isFetchingDropdownData && <LoadingSpinner />}
+            {!isFetchingDropdownData && (
+                <FormikProvider value={formik}>
```
quang-le: Nitpick, but I would replace the Fragment `<>` with the `FormikProvider`. That way the formik `Context` can already initialize while the dropdown data is fetched.
Not blocking, really splitting hairs here

plugins/polio/js/src/domains/Budget/BudgetProcess/EditBudgetProcessModal.tsx
```diff
@@ -0,0 +1,97 @@
...
+        !formik.isSubmitting &&
+        formik.isValid &&
+        isFormChanged &&
+        Boolean(formik.values.rounds);
+
+    return (
+        <>
+            {isFetchingAvailableRounds && <LoadingSpinner />}
+            {!isFetchingAvailableRounds && (
+                <FormikProvider value={formik}>
```
quang-le: same as the other modal

plugins/polio/js/src/domains/Budget/hooks/api/useDeleteBudgetProcess.ts
```diff
@@ -0,0 +1,13 @@
...
+
+const deleteBudgetProcess = (id: number) => {
+    return deleteRequest(`/api/polio/budget/${id}`);
+};
+
+export const useDeleteBudgetProcess = (): UseMutationResult =>
+    useSnackMutation({
+        mutationFn: deleteBudgetProcess,
+        invalidateQueryKey: 'budget',
+    });
```
quang-le: A `snackSuccessMessage` would be nice here.

plugins/polio/js/src/domains/Budget/hooks/api/useGetBudgetProcessAvailableRounds.ts
```diff
@@ -0,0 +1,75 @@
...
+                    label: formatRoundNumber(round.label),
+                    campaign_id: round.campaign_id,
+                };
+            });
+            result.countries = data.countries;
+            result.campaigns = data.campaigns;
+            result.rounds = formattedRounds;
+        }
+        return result;
+    }, []);
```
quang-le: You don't really need `useCallback` if the dependencies array is empty. You can simply hoist the function outside of the hook

plugins/polio/js/src/domains/Budget/hooks/api/useGetBudgetProcessAvailableRounds.ts
```diff
@@ -0,0 +1,75 @@
...
+            result = data.map((round: OptionsRounds) => {
+                return {
+                    value: round.value,
+                    label: formatRoundNumber(round.label),
+                    campaign_id: round.campaign_id,
+                };
+            });
+        }
+        return result;
+    }, []);
```
quang-le: same

quang-le (changes requested):
On the front-end side, the only blocking comment is the one about the dependencies array of `useEffect`.
About blocking deletion of finished/advanced processes, I think we can make another ticket for that.

Nice work!

plugins/polio/js/src/domains/Budget/BudgetProcess/CreateBudgetProcessModal.tsx
kemar: @quang-le I can see the ESLint warning but I don't understand this one.

Why is it a good practice to include `dropdownsData?.campaigns` in the dependencies array of `useEffect`?

What is the benefit of adding it?

quang-le (approved):
Awesome!

PR #1169, opened 2024-03-07
POLIO-1500 create notion of campaign type(s) backend

mathvdh:
Add the notion of "campaign type" to a Campaign (previously the only type was Polio).
It should be as transparent as possible and all the existing campaigns should have one type "Polio".
It should have the possibility to have multiple types.

Related JIRA tickets : POLIO-1500, POLIO-1483

- [x] Are there enough tests


quang-le (commented):
Maybe we should find another name or rename the current campaign type (preventive, test etc) to avoid confusion

mathvdh:
@quang-le `campaign_type` has been renamed to `campaign_category` for the filters. But on the model it was never a real "thing" just a bunch of booleans .. (like `is_preventive` ..)

beygorghor (approved):

PR #1173, opened 2024-03-08
IA-2772 Add a `unique=True` constraint on `OrgUnitChangeRequest.uuid`

kemar:
Add a `unique=True` constraint on `OrgUnitChangeRequest.uuid`

Related JIRA tickets : [IA-2772](https://bluesquare.atlassian.net/browse/IA-2772)

[IA-2772]: https://bluesquare.atlassian.net/browse/IA-2772?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iaso/migrations/0264_data_delete_change_requests_duplicates_uuid.py
```diff
@@ -0,0 +1,33 @@
...
+        .annotate(Count("uuid"))
+        .order_by()
+        .filter(uuid__count__gt=1)
+    )
+
+    for uuid in duplicated_uuids:
+        # Retrieve change requests sharing the same UUID except for the oldest one.
+        change_requests_to_delete = OrgUnitChangeRequest.objects.filter(uuid=uuid).order_by("created_at")[1:]
+        for change_request in change_requests_to_delete:
+            change_request.delete()
```
quang-le: Should we not check that they have the same status? E.g: if one of the dupes has been approved we should probably delete the other

quang-le: But I don't know if there's a big risk/probability to have that issue in prod

kemar: Good point. I investigated the data and I need to think a little bit more about how to handle duplicates.

kemar: Fixed in https://github.com/BLSQ/iaso/pull/1173/commits/2ccdcbca626e49ab56f8ac78c322fee46d6052f4

quang-le (approved):
👍

iaso/models/org_unit.py
```diff
@@ -625,7 +625,7 @@ class Statuses(models.TextChoices):
         REJECTED = "rejected", _("Rejected")
         APPROVED = "approved", _("Approved")
 
-    uuid = models.UUIDField(default=uuid.uuid4, editable=False)
+    uuid = models.UUIDField(default=uuid.uuid4, editable=False, unique=True)
```
bmonjoie, line 628: Does this mean that sending a new Change Request with the same UUID will produce a 400 or that it will update it?

kemar: In fact, sending a new Change Request with an existing UUID will trigger an `IntegrityError` in the serializer, resulting in a 500 HTTP response.

This will help us to detect the duplicate error source.

If you need something else, just tell me and I'll handle it in the code.

bmonjoie: To be confirmed with @madewulf, but we usually don't want synchronization from device to backend to be throwing errors (not to block people on the field). So I'm wondering if that could be an issue

PR #1174, opened 2024-03-08
IA-2777 Force latest version of Django 4 in requirements

kemar:
Force latest version of Django 4 in requirements

Related Ticket : IA-2777


mathvdh (approved):
👍🏼

PR #1178, opened 2024-03-12
IA-2699 Allow to limit authentication to third party authentication

madewulf:
Adding an environment variable DISABLE_PASSWORD_LOGINS that allows to disable logins using password (to only allow SSO )

Related JIRA tickets : IA-2699 (and sub stories)

I also removed a LOT of print() in the codebase so that the test outputs are clean. 

- [x] Are there enough tests

## Doc
- under /docs

## How to test
Activate the environment variable. 

Go to /login  or /admin and check that no password field is available to fill
 /api/token/ should give you a 404 too

## Print screen / video

CleanShot 2024-03-15 at 18 11 52@2x

### DIFF
--- a/hat/urls.py
+++ b/hat/urls.py
@@ -74,7 +74,6 @@ for plugin_name in settings.PLUGINS:
     urls_module_name = "plugins." + plugin_name + ".urls"
     urls_module = importlib.util.find_spec(urls_module_name)  # checking if the urls module exists for this plugin
     if urls_module:
-        print("importing urls for plugin", plugin_name)
         urlpatterns = urlpatterns + [
             path(plugin_name + "/", include(urls_module_name)),
         ]
--- a/hat/urls.py
+++ b/hat/urls.py
@@ -10,24 +10,37 @@ from drf_yasg.views import get_schema_view
 from rest_framework import permissions
 import importlib
 from iaso.views import health, page
+from django.views.generic import TemplateView
 
 
 admin.site.site_header = "Administration de Iaso"
 admin.site.site_title = "Iaso"
 admin.site.index_title = "Administration de Iaso"
 
-urlpatterns = [
+if settings.DISABLE_PASSWORD_LOGINS:
+    urlpatterns = [
+        path("login/", TemplateView.as_view(template_name="iaso/disabled_password_login.html"), name="login"),
+        path("admin/login/", TemplateView.as_view(template_name="iaso/disabled_password_login.html"), name="login"),
+    ]
+else:
+    urlpatterns = [
+        path("login/", auth.views.LoginView.as_view(template_name="iaso/login.html"), name="login"),
+        path("admin/login/", auth.views.LoginView.as_view(template_name="iaso/login.html"), name="login"),
+    ]
+
+urlpatterns = urlpatterns + [
     path("", RedirectView.as_view(pattern_name="dashboard:home_iaso", permanent=False), name="index"),
     path("_health/", health),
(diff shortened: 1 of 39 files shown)


hat/urls.py
```diff
@@ -10,24 +10,37 @@
...
+
+urlpatterns = urlpatterns + [
     path("", RedirectView.as_view(pattern_name="dashboard:home_iaso", permanent=False), name="index"),
     path("_health/", health),
     path("_health", health),  # same without slash otherwise AWS complain about redirect
     path("health/", health),  # alias since current apache config hide _health/
     path("accounts/", include("django.contrib.auth.urls")),
     path("accounts/", include("allauth.urls")),
+    path("login/", auth.views.LoginView.as_view(template_name="iaso/login.html"), name="login"),
+    path("admin/login/", auth.views.LoginView.as_view(template_name="iaso/login.html"), name="login"),
```
kemar:
```suggestion
```

It seems this should be removed as it's already defined in the above condition.

iaso/tests/api/test_token.py
```diff
@@ -6,6 +6,38 @@
 from iaso.test import APITestCase
 
 
+class DisableLoginTokenAPITestCase(APITestCase):
+    @classmethod
+    def setUpTestData(cls):
+        data_source = m.DataSource.objects.create(name="counsil")
+        version = m.SourceVersion.objects.create(data_source=data_source, number=1)
+        star_wars = m.Account.objects.create(name="Star Wars", default_version=version)
+        cls.yoda = cls.create_user_with_profile(username="yoda", account=star_wars)
```
kemar: You told me we should stop using names such as `star_wars` etc. ;)

iaso/tests/api/test_token.py
```diff
@@ -6,6 +6,38 @@
...
+
+    def test_acquire_token_and_authenticate_when_passwords_disabled(self):
+        """Test that token authentication is not possible when passwords are disabled"""
+
+        def reload_urlconf():
+            import importlib
+            from django.urls import clear_url_caches
+
+            clear_url_caches()
+            importlib.reload(importlib.import_module("hat.urls"))
```
kemar: I thought reloading `hat.urls` would be enough, but it seems not.

iaso/tests/api/test_token.py
```diff
@@ -6,6 +6,38 @@
...
+            importlib.reload(importlib.import_module("iaso.urls"))
+
+        with self.settings(DISABLE_PASSWORD_LOGINS=True):
+            reload_urlconf()
+            response = self.client.post(
+                f"/api/token/", data={"username": "yoda", "password": "IMomLove"}, format="json"
+            )
+            self.assertEqual(response.status_code, 404)
+
+        reload_urlconf()
```
kemar: This is useful and could be more generic.

I tried but It's harder than it looks.

I suggest this https://github.com/BLSQ/iaso/pull/1194

kemar (approved):
LGTM
I suggest one little improvement in https://github.com/BLSQ/iaso/pull/1194

iaso/tests/api/test_token.py
madewulf: Correct! I copy pasted these, I shouldn't have. Will correct

iaso/tests/api/test_token.py
madewulf: No, that was the problem we resolved with @mathvdh on Friday. Having the url cache reset after the test is also a challenge.

PR #1185, opened 2024-03-13
POLIO-1501,  POLIO-1502,  POLIO-1503 campaign type filters

beygorghor:
Campaign type has been renamed to campaign categories, types is used for multi diseases campaigns now

Related JIRA tickets : POLIO-1501,  POLIO-1502,  POLIO-1503

## Self proofreading checklist

- [x] Did I use eslint and black formatters
- [x] Is my code clear enough and well documented
- [x] Are my typescript files well typed
- [x] New translations have been added or updated if new strings have been introduced in the frontend
- [ ] My migrations file are included
- [ ] Are there enough tests
- [ ] Documentation has been included (for new feature)

## Changes

- renamed types filter to category filter
- hook to fetch types
- add new types filter

## How to test

Open campaigns or calendar page, and ply with types or categories filters

## Print screen / video

Screenshot 2024-03-13 at 15 54 31
Screenshot 2024-03-13 at 15 54 26
Screenshot 2024-03-13 at 15 54 18
Screenshot 2024-03-13 at 15 54 11

PR #1187, opened 2024-03-14
POLIO-1519: Adding campaign types columns on calendar and map

beygorghor:
Adding campaign types columns on calendar and map pages

Related JIRA tickets : POLIO-1519

- [ ] Are there enough tests

## Changes

Added columns to static columns on calendar page and campaign list page

## How to test

Open campaigns page and check that type columns is displaying correct infos, play with the sort.
Make the same thing on calendar page

## Print screen / video

Screenshot 2024-03-14 at 10 19 25
Screenshot 2024-03-14 at 10 19 20


plugins/polio/api/campaigns/campaigns.py
```diff
@@ -720,6 +734,7 @@ class CampaignViewSet(ModelViewSet):
         "first_round_started_at",
         "last_round_started_at",
         "country__name",
+        "campaign_types__name",
```
mathvdh:
```suggestion
```

Please let's remove this line as it doesn't make sense for the moment

mathvdh (approved):
Please apply the suggest change and we can merge it !
it's ok on the backend side for me

mathvdh:
I applied the suggestion myself. I think we can merge this

plugins/polio/js/src/domains/Campaigns/CampaignsList/useCampaignsTableColumns.tsx
```diff
@@ -69,8 +70,19 @@ export const useCampaignsTableColumns = ({
                 },
             },
             {
-                Header: formatMessage(MESSAGES.status),
+                Header: formatMessage(MESSAGES.campaign_types),
```
quang-le, line 73: This will lead to confusion with the existing  "campaign type"  filter, which filters on "regular", "test" and "preventive" campaign types. We need different words for the 2 types of types

quang-le (changes requested):

plugins/polio/js/src/domains/Campaigns/CampaignsList/useCampaignsTableColumns.tsx
beygorghor: This filter has been changed on this [PR](https://github.com/BLSQ/iaso/pull/1185)

quang-le (approved):

PR #1197, opened 2024-03-19
Updating feature flag migration to make a get_or_create and not fail …

madewulf:
…if we created the FF already

- [x ] My migrations file are included


bmonjoie (approved):
LGTM but I'm no authority in Django :sweat_smile:

madewulf:
> LGTM but I'm no authority in Django 😅

The migrations are passing during the test, I'm not worried.

PR #1199, opened 2024-03-20
IA-1301: 401 unauthorize

beygorghor:
When a project requires authentication (feature flag) or an endpoint does, if no authentication token is provided (or the authentication token is expired), the backend should return a 401 - Unauthorized to indicate the endpoint requires a valid authentication token to be provided.

Related JIRA tickets :IA-1301, WC2-353

- [ ] Are there enough tests

## Doc
- https://www.django-rest-framework.org/api-guide/authentication/#unauthorized-and-forbidden-responses
- https://auth0.com/blog/forbidden-unauthorized-http-status-codes/

## Changes

- just reorder AUTH_CLASSES with `CsrfExemptSessionAuthentication` at the end to return 401
- Tweak `iasoFetch` to redirect to login if any call receive a 401
- fix python tests 
- fix logout redirect url 

## How to test

Use PostMan to GET a secure API endpoint like `/api/payments/lots/` without auth token
You should receive a 401 not a 403

Open the dashboard, login on one tab and open a second one on the dashboard too.
Logout on one of those tab and try to make something on the second, you should be redirect to login page

## Print screen / video

Screenshot 2024-03-20 at 11 27 53


hat/assets/js/apps/Iaso/libs/Api.ts
```diff
@@ -68,6 +68,9 @@ export const iasoFetch = async (
         return emptyRes;
     }
     if (!response.ok) {
+        if (response.status === 401) {
+            window.location.href = '/login';
+        }
```
bmonjoie: You should probably add a `return` in the `if` since the case has been handled.

```suggestion
        if (response.status === 401) {
            window.location.href = '/login';
            return;
        }
```

bmonjoie: Also, do you think it could be possible to pass the current route as a query parameter?
Like, if you are currently on `"/form"` the window location would be:
`/login?redirect_to=/form`

hat/settings.py
```diff
@@ -355,8 +355,8 @@ def is_superuser(u):
 LOGIN_REDIRECT_URL = "/"
 
 AUTH_CLASSES = [
-    "iaso.api.auth.authentication.CsrfExemptSessionAuthentication",
     "rest_framework_simplejwt.authentication.JWTAuthentication",
+    "iaso.api.auth.authentication.CsrfExemptSessionAuthentication",
```
bmonjoie, line 359: CSRF usually is immensely important for security. Let us consider removing this entirely.
Also, I'm not sure, but I believe that putting CsrfExemptSessionAuthentication in second place breaks it. Do we know why we added that in the first place?
@madewulf

iaso/tests/api/test_storage.py
```diff
@@ -458,7 +458,7 @@ def test_post_existing_logs(self):
     def test_list_only_authenticated(self):
         """GET /api/storages/ is rejected if user is not authenticated."""
         response = self.client.get("/api/storages/")
-        self.assertEqual(response.status_code, 403)
+        self.assertEqual(response.status_code, 401)
         # TODO: according to the specs, it should be 401.
```
bmonjoie: Comment can be removed

mathvdh (approved):
Seems good to me. But are we still waiting for an answer from @madewulf about the concerns of @bmonjoie ??

madewulf (approved):

PR #1206, opened 2024-03-22
IA-2785 prevent user to pay themselves

quang-le:

Related JIRA tickets : IA-2785

- [X] Are there enough tests

## Doc
In the code

## Changes

- filter out change requests from the request user when generating `PotentialPayments`
- raise Validation error when a user updates their own `Payment`
- exclude `PotentialPayments `by task launcher (user) when running the payment lot creation task
- exclude `Payments` by the task launcher (user) when launching the payment bulk update task
- idem with the`mark_payments_as_read` task

## How to test

- Create change requests with 2 different users, and have a 3rd user with the payment permission
- Go to potential payments page with each user: only potential payments for the other user should be created when using the change request creators, but the 3rd user should see all
- Create payment lot for both users with change requests
- Go to Payment Lot page as one of the users with change requests:
    - Try updating  a payment for the same user you're logged as: you should get a 400
    - Try bulk updating payments for both users: only payments for the other user should be updated
    - Try using the "mark as read" button: only payments for the other user should be updated 

## Print screen / video
Sorry, maybe later

## Notes

Forked from IA-2779

### DIFF
--- a/iaso/api/payments/serializers.py
+++ b/iaso/api/payments/serializers.py
@@ -142,21 +142,18 @@ class PaymentSerializer(serializers.ModelSerializer):
             raise serializers.ValidationError("Invalid status")
         return status
 
-    def validate_user(self, user):
-        request = self.context["request"]
-        request_user = request.user
-        if user == request_user:
-            raise serializers.ValidationError("Users cannot modify their own payments")
-
     def get_change_requests(self, obj):
         change_requests = OrgUnitChangeRequest.objects.filter(payment=obj)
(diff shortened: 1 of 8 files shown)


beygorghor:
Martin : "I don’t think we should enforce this. Basically, if you have the control of the payment, that means you have some trust. You are able anyway to decide to pay a friend more and to ask him to give you a part if you want to abuse that trust. This feature does not seem to add a lot of security but does add complexity to the code"

PR #1211, opened 2024-03-25
IA-2811 Instance API should give a 404 if form_id is unknown, not a 500

madewulf:
Instance API should give a 404 if parameter form_id is unknown, not a 500
Related JIRA tickets : IA-2811

- [x] Are there enough tests


kemar (approved):
LGTM

PR #1212, opened 2024-03-25
POLIO-1524 calendar polio by default

beygorghor:
Campaing type Polio should be selected by default on calendar page

Related JIRA tickets : POLIO-1524

- [ ] Are there enough tests

## Changes

Use the slug as identifier
Set `campaingType` by default to polio in the url

## How to test

Open calendar page, polio should be selected by default

## Print screen / video

https://github.com/BLSQ/iaso/assets/12494624/af038206-46b0-415a-a142-3f31c63e1f4e

### DIFF
--- a/plugins/polio/js/src/domains/Campaigns/hooks/api/useGetCampaignTypes.ts
+++ b/plugins/polio/js/src/domains/Campaigns/hooks/api/useGetCampaignTypes.ts
@@ -20,7 +20,7 @@ export const useGetCampaignTypes = (): UseQueryResult<
                 return (
                     data?.map((campaignType: CampaignType) => {
                         return {
-                            value: campaignType.id,
+                            value: campaignType.slug,
                             label: campaignType.name,
                             original: campaignType,
                         };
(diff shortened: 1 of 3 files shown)


mathvdh (changes requested):
Looks good except needs to merge main first. All the python changes are already on main so thats weird

plugins/polio/js/src/domains/Campaigns/hooks/api/useGetCampaignTypes.ts
```diff
@@ -20,7 +20,7 @@ export const useGetCampaignTypes = (): UseQueryResult<
                 return (
                     data?.map((campaignType: CampaignType) => {
                         return {
-                            value: campaignType.id,
+                            value: campaignType.slug,
```
hakifran, line 23: useGetCampaignTypes is also used in BaseInfoForm, so by removing the compaignType.id it crush the OBR Name renaming when the campaignType isPolio
Screenshot from 2024-03-26 14-45-47

hakifran (changes requested):

PR #1213, opened 2024-03-26
TRIA-144 Add extra_fields JSON column on OrgUnits

bramj:
This was previously implemented using the JsonDataStore (TRIA-37).

While this works, it poses a serious performance challenge since it results in queries with JOINs on the JsonDataStore table + in a JSON blob. The resulting queries are really complicated and when needing to filter on these fields in the JSON payload, the resolving of N+1 queries is very complex.

Adding a simple JSON column directly on the OrgUnit seriously lowers complexity and improves performance.

Related JIRA tickets : https://bluesquare.atlassian.net/browse/TRIA-144

- ~~[ ] Did I use eslint and black formatters~~
- ~~[ ] Are my typescript files well typed~~
- ~~[ ] New translations have been added or updated if new strings have been introduced in the frontend~~
- [x] Are there enough tests
- ~~[ ] Documentation has been included (for new feature)~~

## Changes

A new column `extra_fields` with a GIN index (see https://www.postgresql.org/docs/current/datatype-json.html#JSON-INDEXING) was added to the `iaso_orgunit` table.


iaso/models/org_unit.py
```diff
@@ -548,36 +552,9 @@ def source_path(self):
...
-            datastore.content = {**datastore.content, **content}
-            datastore.save()
-        except JsonDataStore.DoesNotExist:
-            self.jsondatastore_set.create(
-                account=Account.objects.filter(default_version=self.version).first(),
-                slug="extra_fields",
-                content=content,
-            )
+    def set_extra_fields(self, fields):
+        self.extra_fields = {**self.extra_fields, **fields}
```
kemar:
```suggestion
        self.extra_fields = self.extra_fields | fields
```

Purely stylistic suggestion: this should work in Python 3.9 ([https://stackoverflow.com/a/26853961](https://stackoverflow.com/a/26853961)).

kemar (approved):
LGTM

iaso/models/org_unit.py
bramj: Ow nice I didn't know that! I like that syntax :+1:

PR #1216, opened 2024-03-26
POLIO-1498 Rename outgoing movement into form a unusable vials

mathvdh:
Rename "Outgoing movement" into "Form A - unusable vials" in Vaccine Stock Management > Unusable Vials.

Related JIRA tickets : POLIO-1498

## Changes

Rename one label

## How to test

Go to Vaccine Stock Management and find if there is no more "Outgoing Movement" in the list of unusable vials.

## Print screen / video

Screenshot 2024-03-26 at 16 17 53

### DIFF
--- a/plugins/polio/api/vaccines/stock_management.py
+++ b/plugins/polio/api/vaccines/stock_management.py
@@ -537,7 +537,7 @@ class VaccineStockManagementViewSet(ModelViewSet):
                 results.append(
                     {
                         "date": movement.report_date,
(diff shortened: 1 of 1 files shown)


plugins/polio/api/vaccines/stock_management.py
quang-le, line 1: It looks fine, but I would double check the front-end to make sure the `action` field is always used "as is"

kemar (approved):

PR #1225, opened 2024-03-28
IA-2710 Allow to erase a field in a Change Request

kemar:
Allow to erase a field in a Change Request.

Related JIRA tickets : [IA-2710](https://bluesquare.atlassian.net/browse/IA-2710)

- [x] Are there enough tests

## Problem

`requested_fields` is used by API clients to find out which fields need to be modified.

Until now:

- we couldn't know when a field needed to be set to `None` (e.g. no way to reset a `closed_date`)
- this is because `requested_fields` was dynamically computed based on the presence or absence of a value in `new_*` fields

## Changes

- the `@requested_fields` property is changed to a model field
- a data migration populate `requested_fields` for existing DB rows
- `requested_fields` is populated at creation time for new DB rows
- `requested_fields` is displayed in the admin

## Note for API clients

This means that at Change Request's creation time:

- fields we don't want to change must be excluded from the payload
- all `new_*` fields passed in the payload are those for which the user requests a change or a deletion

## How to test

See unit tests.

This should be transparent to API clients.

[IA-2710]: https://bluesquare.atlassian.net/browse/IA-2710?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


quang-le (approved):
works like a charm

PR #1226, opened 2024-03-29
POLIO-1508: campaign types 

beygorghor:
Campaign type edition should be possible on campaign dialog.

We have to keep in mind that campaign of polio types have more fields and tabs than a regular one

Related JIRA tickets : POLIO-1508

- [ ] Are there enough tests

## Changes

- passing components to TSX
- typing campaign form fields to make it easier
- dynamic validation based on campaign type
- relayout of fields to fit polio or not polio fields 
- keep dialog open after save

## How to test

Go to campaign page and edit a campaign.Add/remove polio form types, dialog should dynamically change.
Virus field is the only field required if the campaign is of polio type, it should not be possible to save if no virus is specified.

## Print screen / video

https://github.com/BLSQ/iaso/assets/12494624/2dbc8efa-3e2b-4460-ad71-d5d1d6c4d4dc

## Notes

Those fields will still be created with campaign even if it is not of polio type:        
```
        has_data_in_budget_tool: false,
        budget_current_state_key: '-',
        detection_status: 'PENDING',
        risk_assessment_status: 'TO_SUBMIT',

```


plugins/polio/js/src/constants/messages.js
```diff
@@ -350,9 +346,13 @@ const MESSAGES = defineMessages({
         defaultMessage: 'DG Authorization',
     },
     target_population: {
-        id: 'iaso.polio.form.label.targetpopulationRound',
+        id: 'iaso.polio.form.label.targetpopulation',
         defaultMessage: 'Target population Round',
```
kemar: `targetPopulationRound` sounded like a good name to avoid confusion with `targetPopulation` just below.

plugins/polio/js/src/domains/Calendar/campaignCalendar/popper/RoundPopper.tsx
```diff
@@ -25,6 +25,21 @@ const groupsForCampaignRound = (campaign, round) => {
...
+    popper: {
+        zIndex: 500,
+        width: 400,
+        backgroundColor: 'white',
+    },
+    popperClose: {
+        position: 'absolute',
+        top: theme => theme.spacing(1),
+        right: theme => theme.spacing(1),
+    },
```
kemar: `popper` and `popperClose` may be factorized between this and `plugins/polio/js/src/domains/Calendar/campaignCalendar/Styles.ts`?

plugins/polio/js/src/domains/Campaigns/hooks/useIsPolioCampaignCheck.ts
```diff
@@ -0,0 +1,25 @@
...
+            return (
+                campaignTypes?.some(
+                    type =>
+                        type.value === 'polio' &&
+                        campaign?.campaign_types?.includes(type.original?.id),
+                ) ?? false
+            );
+        };
+    }, [campaignTypes]);
+};
```
kemar, line 25: It's a bit strange to fetch campaign types each time we need to know if a campaign is of Polio type.

Maybe this should be computed by the campaign detail API to avoid a `useMemo` + an async call + a membership test.

kemar (approved):
LGTM AFAICT

plugins/polio/js/src/domains/Calendar/campaignCalendar/popper/RoundPopper.tsx
beygorghor: this is outdated after rebase 😉

plugins/polio/js/src/domains/Campaigns/hooks/useIsPolioCampaignCheck.ts
beygorghor: In fact react-query is keeping it in cache until cache time or invalidation, so this is the way to access data in the cache...

plugins/polio/js/src/constants/messages.js
beygorghor: you are right, but this is a roll back to how it was before, i broke something else i think, so i prefer to add another translation

PR #1228, opened 2024-04-01
IA-2830: Not allow the user with "Submission read only" permission to do batch delete of submisisons

hakifran:
- A user with "Submission read only" permission can do a batch delete and export requests
- This PR is about to avoid that
Related JIRA tickets : [IA-2830](https://bluesquare.atlassian.net/browse/IA-2830)

- [ ] Are there enough tests

## Changes

- Check if a user has the "Permission read and write" permission before adding the delete and export requests actions

## How to test
- Connect as a user with "Permission read only" permission
- Check if the user has access to batch delete and export requests

## Print screen / video

[Screencast from 2024-04-01 15-08-54.webm](https://github.com/BLSQ/iaso/assets/19631540/fa3d466e-ff22-4655-b5c9-8198757080eb)

[IA-2830]: https://bluesquare.atlassian.net/browse/IA-2830?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hat/assets/js/apps/Iaso/domains/instances/utils/index.tsx
```diff
@@ -410,11 +410,49 @@ export const getSelectionActions = (
...
+                setForceRefresh={setForceRefresh}
+                resetSelection={resetSelection}
+                isUnDeleteAction={isUnDeleteAction}
+            />
+        ),
+        label,
+        disabled: false,
+    };
+
+    const actions: any = [
```
beygorghor: we can already type the action on our side i think

beygorghor (approved):

PR #1235, opened 2024-04-04
POLIO-1531 scopes && campaign types

beygorghor:
Scope selection should come back on his own tab.

We need also to remove all vaccine notion if the campaign is not of Polio type.

Related JIRA tickets : POLIO-1531

- [ ] Are there enough tests

## Changes

- lots of typings fixes
- put back scopes in his own tab
- rewrite isPolio hooks
- migration to remove required from vaccine field
- update api and models to take into account that vaccine is not required anymore
- finalise front-end to be able to save

## How to test

Edit a campaign scope for polio and non polio campaigns

## Print screen / video

https://github.com/BLSQ/iaso/assets/12494624/22bb2880-b5ca-4225-96b3-a91c34decfea


hat/assets/js/apps/Iaso/types/utils.ts
```diff
@@ -4,10 +4,11 @@ export type DropdownOptions<T> = {
     original?: Record<any, any>;
     color?: string;
 };
-export type DropdownOptionsWithOriginal<T> = {
+
+export type DropdownOptionsWithOriginal<T, O = Record<any, any>> = {
     label: string;
     value: T;
-    original: Record<any, any>;
+    original: O;
```
quang-le, line 11: ???

plugins/polio/api/campaigns/campaigns.py
```diff
@@ -293,11 +295,11 @@ def update(self, instance: Campaign, validated_data):
         account = self.context["request"].user.iaso_profile.account
 
         for scope in campaign_scopes:
-            vaccine = scope.get("vaccine")
+            vaccine = scope.get("vaccine", None)
```
quang-le: It seems this loop is exactly the same as  the one line 351 `for scope in scopes`, apart from the way the name is formatted

plugins/polio/js/src/domains/Campaigns/Scope/ScopeForm.tsx
```diff
@@ -18,23 +18,15 @@ import { findRegion, findScopeWithOrgUnit } from './Scopes/utils';
...
 export const scopeFormFields = ['separate_scopes_per_round', 'scopes'];
 
-type Props = {
-    currentTab: string;
-};
-
-export const ScopeForm: FunctionComponent<Props> = ({ currentTab }) => {
-    const { values } = useFormikContext<Values>();
+export const ScopeForm: FunctionComponent = () => {
+    const { values } = useFormikContext<CampaignFormValues>();
```
quang-le, line 28: I really like this use of the `Context` API. We could use it more in our own components to limit props drilling

plugins/polio/js/src/domains/Campaigns/Scope/ScopeInput.tsx
```diff
@@ -70,17 +73,22 @@ export const ScopeInput: FunctionComponent<Props> = ({
...
+            } else {
+                // Find scope for vaccine
+                scope = newScopes.find(s => s.vaccine === selectedVaccine);
+            }
             if (!scope) {
                 scope = {
-                    vaccine: selectedVaccine,
                     group: {
                         org_units: [],
                     },
```
quang-le, line 87:
```typescript
const scope :Scope = (!isPolio 
    ? newScopes[0] 
    : newScopes.find(s => s.vaccine === selectedVaccine)) 
    ?? {
         vaccine: selectedVaccine,
         group: {
            org_units: [],
            }
        }
```


Not sure it's much more legible 😅

plugins/polio/js/src/domains/Campaigns/Scope/ScopeInput.tsx
```diff
@@ -117,24 +125,28 @@ export const ScopeInput: FunctionComponent<Props> = ({
...
             // check if a scope exists for currently selected vaccine
-            let scope: Scope | undefined = newScopes.find(
-                s => s.vaccine === selectedVaccine,
-            );
+            let scope: Scope | undefined;
+            if (!isPolio) {
+                [scope] = newScopes;
+            } else {
+                scope = newScopes.find(s => s.vaccine === selectedVaccine);
+            }
```
quang-le, line 139: same suggestion as above. 
Not blocking. I just don't like using `let` when I can avoid it

plugins/polio/js/src/domains/Campaigns/Scope/ScopeInput.tsx
```diff
@@ -205,6 +217,7 @@ export const ScopeInput: FunctionComponent<Props> = ({
                         isFetching={isFetching}
                         districtShapes={districtShapes || []}
                         selectedVaccine={selectedVaccine}
+                        isPolio={isPolio}
```
quang-le, line 220: If we put that in a Context like `formik` does, we wouldn't have to pass it

plugins/polio/js/src/domains/Campaigns/Scope/Scopes/DistrictScopeTable.tsx
```diff
@@ -21,25 +10,37 @@ import {
     TablePagination,
     TableRow,
     Typography,
-    Box,
 } from '@mui/material';
-import sortBy from 'lodash/sortBy';
+import {
+    // @ts-ignore
```
quang-le: These can be removed now, no?

plugins/polio/js/src/domains/Campaigns/Scope/Scopes/MapScope.tsx
```diff
@@ -18,20 +19,26 @@ import {
...
+    Scope,
+    Vaccine,
+} from '../../../../constants/types';
+import { Shape } from './types';
+import { findScopeWithOrgUnit } from './utils';
 
 type Props = {
     field: FieldInputProps<Scope[]>;
-    values: Values;
+    values: CampaignFormValues;
```
quang-le, line 33: ❤️

hat/assets/js/apps/Iaso/domains/orgUnits/components/OrgUnitLocationIcon.tsx
quang-le, line 1: Is that part of the scope PR?

plugins/polio/models.py
```diff
@@ -126,7 +126,7 @@ class RoundScope(models.Model):
     )
     round = models.ForeignKey("Round", on_delete=models.CASCADE, related_name="scopes")
 
-    vaccine = models.CharField(max_length=5, choices=VACCINES)
+    vaccine = models.CharField(max_length=5, choices=VACCINES, blank=True, null=True)
```
quang-le: I think we're trying to follow good practice and avoid using `null=True` for `CharField`: [django doc](https://docs.djangoproject.com/en/5.0/ref/models/fields/#null)

On the other hand we would have to check what the front-end sends

plugins/polio/models.py
```diff
@@ -144,7 +144,7 @@ class CampaignScope(models.Model):
         Group, on_delete=models.CASCADE, related_name="campaignScope", default=make_group_campaign_scope
     )
     campaign = models.ForeignKey("Campaign", on_delete=models.CASCADE, related_name="scopes")
-    vaccine = models.CharField(max_length=5, choices=VACCINES)
+    vaccine = models.CharField(max_length=5, choices=VACCINES, blank=True, null=True)
```
quang-le: idem

plugins/polio/models.py
```diff
@@ -262,10 +262,14 @@ def vaccine_names(self):
...
         if campaign.separate_scopes_per_round:
-            scopes_with_orgunits = filter(lambda s: len(s.group.org_units.all()) > 0, self.scopes.all())
+            scopes_with_orgunits = filter(
+                lambda s: len(s.group.org_units.all()) > 0 and s.vaccine is not None, self.scopes.all()
+            )
             return ", ".join(scope.vaccine for scope in scopes_with_orgunits)
         else:
-            scopes_with_orgunits = filter(lambda s: len(s.group.org_units.all()) > 0, self.campaign.scopes.all())
+            scopes_with_orgunits = filter(
+                lambda s: len(s.group.org_units.all()) > 0 and s.vaccine is not None, self.campaign.scopes.all()
```
quang-le, line 271:
```python
self.campaign.scopes
    .annotate(org_unit_count=Count("group__org_units"))
    .filter(Q(org_unit_count__gt=0) &Q( vaccine__isnull=False))
```

I don't know if it will be much more performant, but if I remember correctly, using list methods on the whole queryset is not optimal

quang-le: maybe a backend guru can weight in

plugins/polio/models.py
```diff
@@ -761,14 +766,14 @@ def update_geojson_field(self):
...
-                            "vaccine": scope.vaccine,
+                            "vaccine": scope.vaccine if scope.vaccine else None,
                             "scope_key": f"campaignScope-{scope.id}",
                             "top_level_org_unit_name": scope.campaign.country.name,
                         },
                     }
                     features.append(feature)
         else:
-            round_scopes = RoundScope.objects.filter(round__campaign=campaign)
+            round_scopes = RoundScope.objects.filter(round__campaign=campaign).all()
```
quang-le, line 778: is that necessary? `filter` should already return a queryset

quang-le (changes requested):
Creating a non-polio campaign without scope, saving, and adding the scope later generates a 500 on save and on all subsequent GET calls:

https://github.com/BLSQ/iaso/assets/38907762/4f4bdf07-de8c-4351-8f2f-9a025408d673

hat/assets/js/apps/Iaso/types/utils.ts
beygorghor: now we can type the original item of a dropdown list

plugins/polio/js/src/domains/Campaigns/Scope/ScopeInput.tsx
beygorghor: no

hat/assets/js/apps/Iaso/domains/orgUnits/components/OrgUnitLocationIcon.tsx
beygorghor: yes this is the icon use in scope; it was not center

plugins/polio/js/src/domains/Campaigns/MainDialog/CreateEditDialog.tsx
```diff
@@ -68,8 +71,11 @@ const CreateEditDialog: FunctionComponent<Props> = ({
     const handleSubmit = useCallback(
         (values, helpers) => {
             saveCampaign(convertEmptyStringToNull(values), {
-                onSuccess: () => {
+                onSuccess: result => {
                     setIsUpdated(true);
+                    if (!selectedCampaignId) {
+                        setSelectedCampaignId(result.id);
+                    }
```
quang-le, line 78: you could use react-query to save the result directly in the cache. It would save a `useState` and maybe an API call, but this works fine

plugins/polio/js/src/domains/Campaigns/Scope/Scopes/DistrictScopeTable.tsx
```diff
@@ -169,17 +166,15 @@ export const DistrictScopeTable: FunctionComponent<Props> = ({
...
-                                    variant="head"
-                                    style={{ cursor: 'pointer' }}
-                                >
-                                    <Typography>
-                                        {formatMessage(MESSAGES.vaccine)}
-                                    </Typography>
-                                </TableCell>
-                            )}
+                            <TableCell
+                                onClick={() => handleSort('VACCINE')}
```
quang-le: I would disable the sort if `!isPolio`

plugins/polio/js/src/domains/Campaigns/Scope/Scopes/DistrictScopeTable.tsx
```diff
@@ -271,7 +261,7 @@ export const DistrictScopeTable: FunctionComponent<Props> = ({
                                                 }
                                             />
                                         )}
-                                        {isShapeInScope && (
+                                        {shape.vaccineName && (
```
quang-le: shape.vaccineName is never true for non polio scopes, so the toggle works, but the UI always displays the "add" icon

plugins/polio/js/src/domains/Campaigns/hooks/useValidateCampaign.tsx
```diff
@@ -22,15 +22,12 @@ export const useValidateCampaign = (): ((
...
             return {};
         } catch (error) {
-            // Check if the error is a Yup ValidationError and has an inner array
             if (error instanceof ValidationError && error.inner) {
                 return error.inner.reduce((acc, err) => {
                     const path = err.path || 'unknownPath';
-                    if (!acc[path]) {
-                        acc[path] = err.message;
-                    }
+                    acc[path] = err.message;
```
quang-le, line 28: this will overwrite the existing message in case of multiple errors, no?

quang-le (approved):

plugins/polio/models.py
kemar: > I think we're trying to follow good practice and avoid using `null=True` for `CharField`: [django doc](https://docs.djangoproject.com/en/5.0/ref/models/fields/#null)

I approve this message 😄

```suggestion
    vaccine = models.CharField(max_length=5, choices=VACCINES, blank=True)
```

plugins/polio/models.py
kemar:
```suggestion
    vaccine = models.CharField(max_length=5, choices=VACCINES, blank=True)
```

plugins/polio/models.py
kemar: I know the code was already like that, but I think @quang-le is right and his suggestion more readable.

I (personally) don't like [the built-in Python `filter()`](https://docs.python.org/3/library/functions.html#filter) because:

- I have to think about what is happening every time I see it
- it's easily confused with Django `queryset.filter`

As far as I understand, inside this `filter()`:

- `s.group.org_units.all()` will be triggered
- for every item in  `self.campaign.scopes.all()`

A proper unit test of `vaccine_names()` would have allowed to:

-  confirm this with absolute certainty
- refactor with confidence (knowing the expected output)

but unfortunately there isn't any…

PR #1237, opened 2024-04-07
WC2-352 Adding endpoint to get an API token

madewulf:
Adds an endpoint to get an API token

Related JIRA tickets : WC2-352

- [x] Are there enough tests

## Doc
-  in Swagger

## Changes

## How to test

Login with a password, access  /api/apitoken/ and if you feel like it, test the token with the sample code found here: https://iaso.readthedocs.io/en/latest/pages/dev/how_to/use_iaso_apis/use_iaso_apis.html

### DIFF
--- a/iaso/api/api_tokens.py
+++ b/iaso/api/api_tokens.py
@@ -1,0 +1,21 @@
+from rest_framework import serializers, viewsets
+from rest_framework.response import Response
+from rest_framework_simplejwt.tokens import RefreshToken  # type: ignore
+
+
+class APITokenSerializer(serializers.Serializer):
+    token = serializers.CharField()
+
+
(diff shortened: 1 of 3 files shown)


iaso/api/api_tokens.py
```diff
@@ -0,0 +1,21 @@
...
+from rest_framework.response import Response
+from rest_framework_simplejwt.tokens import RefreshToken  # type: ignore
+
+
+class APITokenSerializer(serializers.Serializer):
+    token = serializers.CharField()
+
+
+class APITokenViewSet(viewsets.ViewSet):
+    f"""
```
kemar:
```suggestion
    """
```

kemar (approved):
LGTM

PR #1238, opened 2024-04-08
POLIO-1497: SIA Calendar: dotted lines for Test/on hold campaigns

beygorghor:
When the tick box “Test campaigns/on hold” is selected on the Campaigns edit view, the campaigns should show on the calendar as dotted lines (or other way if easier, but one should be able to differentiate)

Related JIRA tickets POLIO-1497

- [ ] Are there enough tests

## Changes

- force categories to all while loading campaigns or calendar
- return is_test on calendar viewset
- display a dashed red line on rounds if campaign is  test/onHold 
- add a small legend to explain the dashed red line
- review style a bit
- moved toggle period buttons to a specific component

## How to test

Open calendar, make sure you have a test campaign, navigate to the round dates and check if red dashed line is present.
We should also test pdf export (only when deployed, i have cross origin problem locally) and embedded calendar

## Print screen / video

Screenshot 2024-04-08 at 12 14 07


plugins/polio/js/src/domains/Calendar/campaignCalendar/cells/RoundCell.tsx
```diff
@@ -87,7 +87,10 @@ export const RoundCell: FunctionComponent<Props> = ({
         >
             <Box
                 className={classes.coloredBox}
-                style={{ backgroundColor: campaign.color }}
+                sx={{
+                    // backgroundColor: campaign.color,
```
quang-le: 🚮

quang-le (approved):
It works fine, but the dots don't stand out very much when there's a lot of campaigns.
I would say "let's play with opacity instead", but we have to make sure it works on all screens, as we already had issues with colors that were not sufficiently contrasted.

Screenshot 2024-04-09 at 12 24 11

PR #1242, opened 2024-04-10
IA-2710 (bis) Prevent Change Request's payload validation to fail for fields without value

kemar:
Prevent Change Request's payload validation to fail for fields with no value.

Fix for [IA-2710](https://bluesquare.atlassian.net/browse/IA-2710)

## Changes

This was triggering a validation error because `new_closed_date` was `null`:

```
{
  "uuid":"f02ad3ff-6bc3-43c2-aa44-ba258356505e",
  "org_unit_id":"1972317",
  "created_at":1.712741023886E9,
  "updated_at":1.712741023886E9,
  "new_closed_date":null
}
```

The validation logic is fixed to consider `null` as valid (as this can be used to erase a field).

An error is raised only when none of the new fields is present.

[IA-2710]: https://bluesquare.atlassian.net/browse/IA-2710?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bmonjoie (approved):
LGTM

PR #1248, opened 2024-04-12
Ia 2845 500 login

madewulf:
Ensuring there is a single url with name "login" preventing resolution errors. 
Ensuring also that the URLS of allauth are not loaded when we don't use single sign on, and further than that, ensuring that only the URLs used for SSO are made available. 

Adding also a variable in settings.py ACTIVATE_SOCIAL_ACCOUNT that is used to deactivate these allauth.urls

## How to test
Activate the various feature flags. Run manage show_urls to verify that nothing too much is here.


kemar (approved):
LGTM

quang-le (approved):

PR #1251, opened 2024-04-16
IA-2776 change request approved comment

beygorghor:
In the below GIF the first use case we accept partially the modifications, but it would be great to be able to send out a comment to warn the mobile app user that he needs to re-submit with the missing part? …
demopoliogeoregistry

Related JIRA tickets : IA-2776

- [ ] Are there enough tests

## Changes

- fix group field color while approved, it was always green even if change was not applied
- rename comment dialog
- allow to display comment dialog if partially approved
- adapt backend to save comment

## How to test

You need a new org unit change request with multiple changes.
Open the dialog to approve or reject, select only one field and click on approve selected changes, you should be able to enter a a comment.
Save and check that comment has been saved

## Print screen / video

https://github.com/BLSQ/iaso/assets/12494624/bd1a4a96-bfa0-4b51-bf56-cac4c75a202e

## Notes

We should check that mobile app is displaying also comment if change request is partially approved @bmonjoie


iaso/models/org_unit.py
```diff
@@ -712,10 +713,11 @@ def reject(self, user: User, rejection_comment: str) -> None:
         self.rejection_comment = rejection_comment
         self.save()
 
-    def approve(self, user: User, approved_fields: typing.List[str]) -> None:
+    def approve(self, user: User, approved_fields: typing.List[str], rejection_comment: str = "") -> None:
         self.__apply_changes(user, approved_fields)
         self.updated_by = user
         self.status = self.Statuses.APPROVED
+        self.rejection_comment = rejection_comment
```
kemar, line 720: This is the only thing that bugs me.

`rejection_comment` can now contain rejection or approval comments.

We should rename `rejection_comment` to `comment` because the name of the field is now misleading.

The renaming could be done in another PR (or in this one):

1. add a `comment` field
2. migrate data from `rejection_comment` to `comment`
3. delete `rejection_comment`

@bmonjoie is this easily doable in the mobile app or should we keep `rejection_comment`?

bmonjoie:
@Kemar, we should keep `rejection_comment`. It can be returned alongside `comment` with the same value if you want.

Though, I agree with @beygorghor that even if the status is `approved`, it's still a comment about what was rejected, so leaving it as `rejection_comment` is not shocking.

kemar:
> Though, I agree with @beygorghor that even if the status is `approved`, it's still a comment about what was rejected, so leaving it as `rejection_comment` is not shocking.

Ok, that makes sense. Thx.

kemar (approved):
LGTM

iaso/models/org_unit.py
kemar: Let's keep `rejection_comment` see https://github.com/BLSQ/iaso/pull/1251#issuecomment-2060696621

PR #1253, opened 2024-04-16
IA-2884 improve dhis2 error message parsing when conflicts don't have…

mestachs:
Sometime the dhis2 api doesn't return conflicts "with a description"
Ex in this jira the valueType of the data element don't match the value sent by iaso.

Related JIRA tickets : IA-2884 

- [x] Are there enough tests

## How to test

Don't think it's easy to test, but perhaps using the seed command and changing the valueType of the program used by the seed command


mathvdh (approved):
👍🏼

PR #1254, opened 2024-04-16
POLIO-1102: scope only valid org units

beygorghor:
In polio prod go to campaign CHD-2023-05-1_nOPV

In scope tab, the list will show both MICHEMIRE and MICHEMERE. The second one has been rejected. the name is actually an alias of the first one

Related JIRA tickets :POLIO-1102

- [ ] Are there enough tests

## Doc

## Changes

only fetch valid org units

## How to test

Select a campaign, edit the scope and get the name of an org unit.
Edit the org unit and set the status to rejected.
Refresh the campaign scope, org unit should be present anymore

## Print screen / video

https://github.com/BLSQ/iaso/assets/12494624/bbefcff1-dd2f-4480-901e-e3c8a903c380


quang-le (changes requested):
If an org unit is already in the scope but changes status, it will not be displayed on the map or in the list, though it's still part of the scope.
I think this problem is adjacent to "what do we do if an org unit is split in 2"

https://github.com/BLSQ/iaso/assets/38907762/38251caa-5da9-466a-8c96-82e5b3694264

beygorghor:
@quang-le   Now it displays REJECTED or NEW org unit if they are already part of the scope.

plugins/polio/js/src/domains/Campaigns/Scope/ScopeForm.tsx
```diff
@@ -77,44 +77,47 @@ export const ScopeForm: FunctionComponent = () => {
...
-                );
-            }
+        const orgUnitIdToVaccine = new Map();
+        if (isPolio && scopes) {
+            scopes.forEach(scope => {
+                scope.group.org_units.forEach(ouId => {
+                    orgUnitIdToVaccine.set(ouId, scope.vaccine);
+                });
+            });
+        }
```
quang-le, line 89: If we ever remake the API, it would be great to have that map coming straight from the backend

plugins/polio/js/src/domains/Campaigns/Scope/ScopeForm.tsx
```diff
@@ -77,44 +77,47 @@ export const ScopeForm: FunctionComponent = () => {
...
+                ...cloneDeep(district),
+                region: findRegion(district, regionShapes),
+                scope,
+                vaccineName,
+            });
+
+            return acc;
+        }, [] as FilteredDistricts[]);
+
+        return filtered;
```
quang-le: This would be much easier to read as a combination of `filter` and `map`

plugins/polio/js/src/domains/Campaigns/Scope/ScopeForm.tsx
```diff
@@ -77,44 +77,47 @@ export const ScopeForm: FunctionComponent = () => {
...
+                        .includes(debouncedSearch.toLowerCase()))
+            ) {
+                return acc;
             }
-            return filtered;
-        }
-        return undefined;
+
+            acc.push({
+                ...cloneDeep(district),
```
quang-le: There's now `structuredClone` that comes free with JS

https://developer.mozilla.org/en-US/docs/Web/API/structuredClone

plugins/polio/js/src/domains/Campaigns/Scope/ScopeForm.tsx
```diff
@@ -77,44 +77,47 @@ export const ScopeForm: FunctionComponent = () => {
...
-            if (debouncedSearch !== '') {
-                filtered = filtered.filter(d =>
-                    d.name
+        const filtered = districtShapes.reduce((acc, district) => {
+            const scope = findScopeWithOrgUnit(scopes, district.id);
+            const vaccineName =
+                orgUnitIdToVaccine.get(district.id) || undefined;
+            const isInScope = scopes.some(sc =>
+                sc.group.org_units.includes(district.id),
+            );
```
quang-le: Why is this check needed? We already look for a scope line 92

plugins/polio/js/src/domains/Campaigns/Scope/Scopes/MapScope.tsx
```diff
@@ -92,24 +92,26 @@ export const MapScope: FunctionComponent<Props> = ({
...
-    const districts = useMemo(
-        () =>
-            districtShapes.filter(
+    const filterOrgUnits = useCallback(
+        (orgUnits: OrgUnit[]) =>
+            orgUnits.filter(
                 ogrUnit =>
-                    ogrUnit.has_geo_json ||
-                    (ogrUnit.latitude && ogrUnit.longitude),
+                    (ogrUnit.has_geo_json ||
```
quang-le: `ogr_unit`  lol

quang-le (commented):
If the `reduce` can be refactored as `filter().map()` without breaking the feature, I think that would be great for future maintenance

plugins/polio/js/src/domains/Campaigns/Scope/ScopeForm.tsx
beygorghor: we need at least node 17 or to install a polyfill for that

quang-le (approved):

PR #1258, opened 2024-04-17
POLIO-1541 add stocks fields to api for dashboarding

mathvdh:
Add stocks fields to API for vaccine stock dashboarding

Fields added : 

- Stock in hand - computed field “Usable vials” for the country
- Form A reception (RRT) date
- Destruction Report Received by RRT date

Related JIRA tickets : POLIO-1541

- [x] Are there enough tests

## Changes

Add some fields on the serializer and cache some calculations

## How to test

Call the APIs and see if the fields are present.
Check the test code.

### DIFF
--- a/plugins/polio/api/dashboards/supply_chain.py
+++ b/plugins/polio/api/dashboards/supply_chain.py
@@ -1,12 +1,24 @@
 from rest_framework import serializers
+
 from iaso.api.common import ModelViewSet
+from plugins.polio.admin import DestructionReport
+from plugins.polio.api.vaccines.stock_management import VaccineStockCalculator
 from plugins.polio.api.vaccines.supply_chain import VaccineSupplyChainReadWritePerm
-from plugins.polio.models import VaccineArrivalReport, VaccinePreAlert, VaccineRequestForm
+from plugins.polio.models import (
+    OutgoingStockMovement,
+    VaccineArrivalReport,
+    VaccinePreAlert,
+    VaccineRequestForm,
+    VaccineStock,
+)
 
 
 class VaccineRequestFormDashboardSerializer(serializers.ModelSerializer):
     obr_name = serializers.CharField(source="campaign.obr_name")
     country = serializers.SerializerMethodField()
+    stock_in_hand = serializers.SerializerMethodField()
+    form_a_reception_date = serializers.SerializerMethodField()
+    destruction_report_reception_date = serializers.SerializerMethodField()
 
     class Meta:
         model = VaccineRequestForm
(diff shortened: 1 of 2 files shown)


plugins/polio/api/dashboards/supply_chain.py
```diff
@@ -1,12 +1,33 @@
...
+)
+
+
+def get_or_create_vaccine_stock(ser_obj, obj):
+    accnt = ser_obj.context["request"].user.iaso_profile.account
+    vaccine_stock, _ = VaccineStock.objects.get_or_create(
+        country=obj.campaign.country, vaccine=obj.vaccine_type, account=accnt
+    )
+
+    return vaccine_stock
```
kemar: This is called 3 times in the Serializer, generating 3 different SQL queries.

One way to deal with that is to [override serialization behavior](https://www.django-rest-framework.org/api-guide/serializers/#overriding-serialization-and-deserialization-behavior).

Broadly speaking, I think this could be implemented like this in order for `get_or_create` to be called only once:

```python
    def to_representation(self, instance):
        representation = super().to_representation(instance)

        accnt = self.context["request"].user.iaso_profile.account
        vaccine_stock, _ = VaccineStock.objects.get_or_create(
            country=instance.campaign.country, vaccine=instance.vaccine_type, account=accnt
        )

        representation["stock_in_hand"] = self.get_stock_in_hand(vaccine_stock)
        representation["form_a_reception_date"] = self.get_form_a_reception_date(vaccine_stock)
        representation["destruction_report_reception_date"] = self.get_destruction_report_reception_date(vaccine_stock)

        return representation
```

plugins/polio/tests/test_supply_chain_dashboards.py
```diff
@@ -100,6 +114,44 @@ def test_vrf_filters_by_account(self):
...
+
+        destruction_report = DestructionReport.objects.create(
+            vaccine_stock=self.vaccine_stock,
+            rrt_destruction_report_reception_date=date.today(),
+            destruction_report_date=date.today(),
+            unusable_vials_destroyed=10,
+            action="destroyed",
+        )
+
+        response = self.client.get(self.vrf_url)
```
kemar:
```suggestion
        with self.assertNumQueries(MAKE_ME_DREAM_NUMBER):
            response = self.client.get(self.vrf_url)
```

kemar (changes requested):
I think you should fix some SQL performance issue.

plugins/polio/api/dashboards/supply_chain.py
```diff
@@ -1,29 +1,85 @@
...
+)
 
 
 class VaccineRequestFormDashboardSerializer(serializers.ModelSerializer):
     obr_name = serializers.CharField(source="campaign.obr_name")
-    country = serializers.SerializerMethodField()
+    country = serializers.IntegerField(source="campaign.country.pk")
+    stock_in_hand = serializers.IntegerField(read_only=True)
+    form_a_reception_date = serializers.DateField(read_only=True)
+    destruction_report_reception_date = serializers.DateField(read_only=True)
```
kemar, line 21:
```suggestion
```

Those fields can be removed now that they are computed in `to_representation`!

kemar (approved):
LGTM

PR #1269, opened 2024-04-22
SLEEP-1337 Add support for mobile bulk uploads

bramj:
In Trypelim, we have performance issues on the form uploads.
To improve the performance of the uploads, we're grouping forms and org units into zip files (batches of 2000 forms) and upload these with only 1 API call per zip file.

This PR adds a job to process these bulk uploads:
1. A new API endpoint `/api/mobile/bulkupload/` that takes a `zip_file`. This endpoint simply uploads the file to S3, launches an async job, and returns HTTP 204.
2. The process job creates a record in the `APIImport` for future reference (and to be able to tag as `has_problem`) and processes all files in the zip in a db transaction.

Related JIRA tickets : https://bluesquare.atlassian.net/browse/SLEEP-1337

- ~~[ ] Did I use eslint and black formatters~~
- ~~[ ] Are my typescript files well typed~~
- [x] Are there enough tests

image


iaso/tasks/process_mobile_bulk_upload.py
```diff
@@ -0,0 +1,163 @@
...
+                fi.file = File(f)
+                fi.instance_id = instance.id
+                fi.name = instance_file.name
+                fi.save()
+                instance_files.append(fi)
+
+    return instance_files
+
+
+# Trypelim-specific
```
bramj, line 120: Main issue I see for merging this is how this code is Trypelim specific (although it won't affect any form that doesn't have a `serie_id` attribute)

bramj:
@mathvdh assigning you for review, I think Martin is too busy at the moment

iaso/api/mobile/bulk_uploads.py
```diff
@@ -0,0 +1,95 @@
...
+logger = logging.getLogger(__name__)
+
+
+class ZipFileSerializer(serializers.Serializer):
+    zip_file = serializers.FileField()
+
+    def validateZipFile(self):
+        if self.is_valid():
+            return self.validated_data["zip_file"]
+        raise ValueError("Zip file not valid")
```
madewulf, line 30: Does this do something for real, outside of checking that there is a file?

iaso/tasks/process_mobile_bulk_upload.py
```diff
@@ -0,0 +1,163 @@
...
+    the_task = task
+    the_task.report_progress_and_stop_if_killed(
+        progress_value=0,
+        progress_message=_("Starting"),
+        end_value=100,
+    )
+    user = User.objects.get(id=user_id)
+    project = Project.objects.get(id=project_id)
+
+    api_import = APIImport.objects.create(
```
madewulf: I would have said that this should be the very first thing happening with the zip file, before any check on the user or other treatments. The goal is to store the data before any other bug can happen, and there, it seems to happen relatively late (after even the pickup by the worker if I understand correctly, which means that if the worker is off, the payload can be lost?)

iaso/tasks/process_mobile_bulk_upload.py
```diff
@@ -0,0 +1,163 @@
...
+        api_import.exception = format_exc()
+        api_import.save()
+        raise e
+
+    the_task.report_success_with_result(
+        message=result_message(user, project, start_date, start_time, stats),
+    )
+
+
+def process_instance_xml(directory, zip_ref, user):
```
madewulf, line 99: Isn't this duplicated code from /sync/views.py 

In this case, we should make it a function as to ensure that when we update the code somewhere with new features, we don't forget to update it everywhere

madewulf (changes requested):

iaso/api/mobile/bulk_uploads.py
bramj: No, just a check of the file

iaso/tasks/process_mobile_bulk_upload.py
bramj: The file is uploaded to S3 before being passed to the worker, so we wouldn't really lose the data. That said, you're totally right we should create the import object first, which will make it a lot easier to track when something goes wrong. I'll fix it.

iaso/tasks/process_mobile_bulk_upload.py
bramj: :+1: agreed, I've extracted a function

madewulf (approved):

PR #1276, opened 2024-04-25
TRIA-7 Small fixes

bramj:
- Some typos in messages.js
- Small improvements on naming for org unit dicts (reason: I'm using these methods in Trypelim so I've changed the naming to be more generic)
- Small readme update
- ~~Sentry: Ignore django.security.DisallowedHost error: I believe these failed hacking attempts are not relevant.~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -843,29 +843,22 @@ Alternatively this can be done outside of docker by running:
 
 # Background tasks & worker
 
-Iaso queue certain functions (task) for later execution, so they can run
-outside an HTTP request. This is used for functions that take a long time to execute,
-so they don't canceled in the middle by a timeout of a connection closed.
-e.g: bulk import, modifications or export of OrgUnits.  Theses are the functions
-marked by the decorator @task_decorator, when called they get added to a Queue
-and get executed by a worker.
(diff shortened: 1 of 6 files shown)


hat/settings.py
```diff
@@ -529,6 +529,7 @@ def sentry_error_sampler(_, hint):
         integrations=[DjangoIntegration()],
         send_default_pii=True,
         release=VERSION,
+        ignore_errors=["django.security.DisallowedHost"],
```
kemar: I don't know if this will work?
Let's see.

Last time I did this, [I used `ignore_logger`](https://docs.sentry.io/platforms/python/integrations/logging/#ignoring-a-logger).

```python
from sentry_sdk.integrations.logging import ignore_logger

ignore_logger("django.security.DisallowedHost")
```

kemar (approved):
LGTM

hat/settings.py
bramj: Thanks! I think it worked for me on Trypelim, but let me double-check that

bramj: @kemar I went to dig a little deeper and it works on the version we're using (https://github.com/getsentry/sentry-python/blob/1.38.0/sentry_sdk/client.py#L429). However it appears undocumented and so it'll probably cause issues next library upgrade, so I'll change it to using an `ignore_logger`.

bramj: The `ignore_logger` was already added after a rebase, so I've just removed it.

PR #1297, opened 2024-05-07
WC2-367: Form detail label key name

beygorghor:
When using the “Default fields to display” parameter in the advanced settings of a form, the list offers all question names that ever existed but only displays their label.

@Michael Matiashe configured it to one, but the value didn’t appear in the application because it was wrong.

It would be better to display the question name along the label to remove the confusion.

Related JIRA tickets : WC2-367

- [ ] Are there enough tests

## Changes

just adding field name in label

## How to test

Open detail of a forms, click on advanced settings and open default field dropdown

## Print screen / video

Screenshot 2024-05-07 at 14 52 26


bmonjoie (approved):
LGTM

PR #1313, opened 2024-05-21
POLIO-1540 Create new tab sub activities for all campaigns polio and non polio

mathvdh:
Introduce SubActivity Model in Polio Plugin and the related APIs

Related JIRA tickets : POLIO-1540

- [x] Are there enough tests

## Doc

In the `docs/pages/dev/campaigns/subactivities.md` file included in this PR.

## Changes

- Added a new model `SubActivity` which represents a sub-activity within a round of a campaign. It has fields for name, age unit, minimum age, maximum age, start date, and end date.
- Added `SubActivityScope` model to represent the scope of a SubActivity with respect to vaccines.
- The `SubActivity` model is linked to the `Round` model via a foreign key, allowing each round to have multiple sub-activities.
- The `SubActivityScope` model is linked to the `SubActivity` model via a foreign key, allowing each sub-activity to have multiple scopes.

## How to test

As this only includes the APIs and no frontend code, you can use /swagger to test and/or use django admin.


plugins/polio/api/campaigns/subactivities.py
```diff
@@ -0,0 +1,121 @@
...
+        model = SubActivityScope
+        fields = ["group", "vaccine"]
+
+    group = GroupSerializer(required=False)
+
+
+class SubActivityCreateUpdateSerializer(serializers.ModelSerializer):
+    round_number = serializers.IntegerField(write_only=True, required=False)
+    campaign = serializers.CharField(write_only=True, required=False)
+    scopes = SubActivityScopeSerializer(many=True, required=False)
```
quang-le, line 22: shouldn't the scopes be mandatory?

quang-le (commented):
I tried the POST and PUT in the front-end, it works fine, but shouldn't we check the values against those of the round? eg:
- start and end date should be within the round dates
- scope should be a subset of round scopes
- vaccines should be a subset of round vaccines

plugins/polio/api/campaigns/subactivities.py
madewulf: I think not, you could have first just the encoding of dates and reason of activity and the scope coming later

madewulf (approved):

PR #1316, opened 2024-05-22
IA-2900: registry view permission

beygorghor:
With this permission, users would not be able to:

- see other Forms than the Reference forms
- edit form submissions via Enketo
- consult the Org unit details
- download data

Therefore the corresponding buttons should be hidden from the interface

Related JIRA tickets : IA-2900

- [ ] Are there enough tests

## Changes

- adding read and write permissions
- adapt backend to filter instances and get only reference instances
- display /hide buttons regarding permissions
- use `DisplayIfUserHasPerm`
- add strict mode to `DisplayIfUserHasPerm`and create a new `userHasAllPermissions`function

## How to test

You need a non superadmin user having only registry_read rights.
Connect to a registry page with this user and check that buttons are not present

## Print screen / video

Screenshot 2024-05-22 at 10 46 06


hat/assets/js/apps/Iaso/domains/registry/components/Instances.tsx
```diff
@@ -219,12 +228,9 @@ export const Instances: FunctionComponent<Props> = ({
                             columns={tableColumns}
                             count={data?.count ?? 0}
                             params={params}
-                            onTableParamsChange={p =>
-                                dispatch(
-                                    redirectToReplace(baseUrls.registry, p),
-                                )
-                            }
-                            extraProps={{ loading: isFetchingList }}
```
quang-le: ❤️

hat/assets/js/apps/Iaso/domains/registry/components/selectedOrgUnit/OrgUnitTitle.tsx
```diff
@@ -55,24 +57,30 @@ export const OrgUnitTitle: FunctionComponent<Props> = ({ orgUnit, params }) => {
...
-                    <IconButton
-                        url={`${baseUrls.orgUnitDetails}/orgUnitId/${orgUnit.id}`}
-                        color="secondary"
-                        icon="edit"
-                        tooltipMessage={MESSAGES.editOrgUnit}
-                        iconSize="small"
-                        size="small"
-                    />
+                    <DisplayIfUserHasPerm
+                        permissions={[Permissions.REGISTRY_WRITE]}
```
quang-le, line 61: Not blocking but maybe needs to be clarified: we use the registry permission to give access to org unit edition?

So if I have the registry read permission and org unit permission, I cannot edit the org unit via the registry page, but I can via the org unit page (and vice-versa)

hat/assets/js/apps/Iaso/domains/users/utils.js
quang-le, line 1: ❤️

quang-le (approved):
Just one general comment about permissions but nothing blocking.

I didn't review the backend though

hat/assets/js/apps/Iaso/domains/registry/components/selectedOrgUnit/OrgUnitTitle.tsx
beygorghor: Screenshot 2024-05-24 at 11 50 18

PR #1321, opened 2024-05-22
IA-2991 Fix for bug only when no limit parameter was given

mathvdh:
When calling the `/api/profiles` without any parameters (no `?limit=xxx` or else) it was creating a bug when serializing the phone numbers of users. 

Related JIRA tickets : IA-2991

## Changes

Very small changes copied from the code that was run with a `?limit=` parameter was added.

## How to test

Just call `/api/profiles` with at least one user having a phone number in the DB it should NOT return an error

### DIFF
--- a/iaso/models/base.py
+++ b/iaso/models/base.py
@@ -1350,8 +1350,8 @@ class Profile(models.Model):
             "email": self.user.email,
             "language": self.language,
             "user_id": self.user.id,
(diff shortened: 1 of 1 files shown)


mathvdh:
Already pushed as a hotfix on iaso prod

beygorghor (approved):

PR #1324, opened 2024-05-23
Performance improvements (mainly on entities dashboard)

bramj:
- [Improve performance of dashboard entities page](https://github.com/BLSQ/iaso/commit/603ed793ee8f37c2403b85f478071873cf27276c). There were a few issues with the /entities/list page:

  - Missing prefetch_relateds resulting in n+1 calls
  - No pagination on the api call for the map display
  - Unnecessary data (org unit parents) being fetched on the list and map page

- [Small improvement on /profiles API endpoint](https://github.com/BLSQ/iaso/commit/651cb9326a15d80240c127741075c03220147434)
  - Added missing prefetch for `user` model.
  - No need to load projects in the small dict (used to load profiles/users in select filters for example) **let me know if you disagree**
- Took out `simplified_geom` from the `defer` clause since it's used in the base OrgUnitSerializer for the field `has_geo_json`. Because it was in deferred, it would launch a new query for every record. **let me know if you disagree**

### DIFF
--- a/iaso/api/entity.py
+++ b/iaso/api/entity.py
@@ -234,14 +234,10 @@ class EntityViewSet(ModelViewSet):
         queryset = Entity.objects.filter(account=self.request.user.iaso_profile.account)
 
         queryset = queryset.prefetch_related(
-            "attributes",
-            "attributes__created_by",
             "attributes__created_by__teams",
             "attributes__form",
-            "attributes__org_unit",
             "attributes__org_unit__org_unit_type",
             "attributes__org_unit__parent",
-            "attributes__org_unit__version",
             "attributes__org_unit__version__data_source",
             "entity_type",
         )
--- a/iaso/api/entity.py
+++ b/iaso/api/entity.py
@@ -87,7 +87,7 @@ class EntitySerializer(serializers.ModelSerializer):
 
     def get_org_unit(self, entity: Entity):
         if entity.attributes and entity.attributes.org_unit:
-            return entity.attributes.org_unit.as_location(with_parents=True)
+            return entity.attributes.org_unit.as_location(with_parents=False)
         return None
 
     def get_submitter(self, entity: Entity):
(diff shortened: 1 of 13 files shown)


iaso/models/org_unit.py
```diff
@@ -206,7 +206,7 @@ def filter_for_user_and_app_id(
         if user and user.is_anonymous and app_id is None:
             return self.none()
 
-        queryset: OrgUnitQuerySet = self.defer("geom", "simplified_geom")
+        queryset: OrgUnitQuerySet = self.defer("geom")
```
bramj, line 209: Do you think this is an okay change? See description of the PR.

kemar: I fully agree with this change.

A `has_geo_json` annotation might be even more performant, freeing ourselves for fetching `simplified_geom`. Let's try this in a future version!

iaso/models/base.py
```diff
@@ -1364,7 +1364,6 @@ def as_short_dict(self):
             "user_id": self.user.id,
             "phone_number": self.phone_number.as_e164 if self.phone_number else None,
             "country_code": region_code_for_number(self.phone_number).lower() if self.phone_number else None,
-            "projects": [p.as_dict() for p in self.projects.all().order_by("name")],
```
kemar, line 1367: Since this isn't breaking any unit test, I assume it has no impact :)

iaso/api/entity.py
```diff
@@ -233,6 +233,19 @@ def get_queryset(self):
 
         queryset = Entity.objects.filter(account=self.request.user.iaso_profile.account)
 
+        queryset = queryset.prefetch_related(
+            "attributes",
+            "attributes__created_by",
+            "attributes__created_by__teams",
```
kemar:
```suggestion
            "attributes__created_by__teams",
```

`attributes__created_by__teams` already contains all the needed information, so the previous arguments `attributes` and `attributes__created_by` are actually redundant.

kemar (approved):
LGTM Good work!

PR #1328, opened 2024-05-24
IA-3024: Not displaying permissions related to modules which are not linked to the account

hakifran:
- Not displaying permissions related to modules which are not linked to the account
Related JIRA tickets : [IA-3024](https://bluesquare.atlassian.net/browse/IA-3024)

- [x] Are there enough tests

## Changes
- Check if grouped permissions is empty and skip if it is

## How to test
- Remove some modules from the current account
- Check if some permissions groups still appearing

## Print screen / video
[Screencast from 2024-05-24 16-02-47.webm](https://github.com/BLSQ/iaso/assets/19631540/d81f613e-be07-48cb-8587-d8154bf0f007)

[IA-3024]: https://bluesquare.atlassian.net/browse/IA-3024?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iaso/api/permissions.py
```diff
@@ -38,9 +38,13 @@ def grouped_permissions(self, request):
...
-            result[group] = []
-            for permission in perms.filter(codename__in=PERMISSIONS_PRESENTATION[group]):
-                result[group].append({"id": permission.id, "name": _(permission.name), "codename": permission.codename})
+            permissions = perms.filter(codename__in=PERMISSIONS_PRESENTATION[group])
+            if permissions:
+                result[group] = []
+                for permission in permissions:
+                    result[group].append(
+                        {"id": permission.id, "name": _(permission.name), "codename": permission.codename}
+                    )
```
kemar:
```suggestion
            for permission in perms.filter(codename__in=PERMISSIONS_PRESENTATION[group]):
                result[group] = []
                result[group].append({"id": permission.id, "name": _(permission.name), "codename": permission.codename})
```

I think you can keep the previous code and just move the `result[group] = []` line inside the `for` loop.

This could save a bunch of statements.

iaso/tests/api/test_permissions.py
```diff
@@ -0,0 +1,61 @@
...
+from iaso import models as m
+from django.contrib.auth.models import Permission
+
+from iaso.utils.module_permissions import account_module_permissions
+
+
+class OrgUnitAPITestCase(APITestCase):
+    @classmethod
+    def setUpTestData(cls):
+        cls.star_wars = star_wars = m.Account.objects.create(name="Star Wars")
```
kemar:
```suggestion
        cls.account = account = m.Account.objects.create(name="Account")
```

As a best practice, we try to use more meaningful names in unit tests to make them easier to understand.

iaso/tests/api/test_permissions.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def test_get_all_permissions_without_grouping(self):
+        """Get /permissions all permissions without grouping them"""
+
+        self.client.force_authenticate(self.yoda)
+
+        response = self.client.get("/api/permissions/")
+        self.assertJSONResponse(response, 200)
+        account = self.yoda.iaso_profile.account
+        account_modules = account.modules if account.modules else []
```
kemar:
```suggestion
        account_modules = account.modules
```

This should never be an empty `[]` because it's explicitely defined in `setUpTestData`.

You could use `account.modules` directly without an additional variable.

kemar (approved):
LGTM
I have a few comments on the form that you can fix if you agree with them.

iaso/api/permissions.py
hakifran: But at each loop it will remove the previous added permission and at the end result[group] will have the last permission

kemar: Yes you're right!
But that's strange because I ran the unit tests with this modified code and they were successful 🤔

hakifran: `self.assertTrue(
            set([permission["codename"] for permission in response.json()["permissions"]["org_units"]])
            <= set(PERMISSIONS_PRESENTATION["org_units"]),
        )`. Maybe I did it wrong here?

PR #1335, opened 2024-05-28
IA-3039 Entities dashboard list: filter results on user profile geo restrictions

bramj:
- Generalized the nice floating alert from the form submissions page
- Refactored so the dashboard api endpoint filters based on the user profile (just like the mobile entities endpoint)

Screenshot 2024-05-28 at 16 12 43


kemar (approved):
LGTM

PR #1339, opened 2024-05-30
IA-2985: org unit tree use new api

beygorghor:
- use the new backend APIs to display and search the tree by one or multiple status

- use the DataSource config (if it exists) to customize the display of the tree

Related JIRA tickets : IA-2985

- [ ] Are there enough tests

## Changes

- pass org unit tree modal and component to tsx
- use new api to get result
- link default visible validation status to the source config
- add infos over the source an the version used
- allow to hide valid org unit too

## How to test

Edit 3 sources and add different validation status, one should be the default one

Open submissions page and click on org unit select. Check that the correct source and version is displayed. try to find n org unit and select it.
Make the same thing while trying to export a data source (last icon on a source row). There you can select a version of a source.

## Print screen / video
Screenshot 2024-05-30 at 14 40 35

Screenshot 2024-05-30 at 14 40 10

⚠️ This is working with this [PR](https://github.com/BLSQ/bluesquare-components/pull/157) ⚠️


hat/assets/js/apps/Iaso/domains/orgUnits/components/TreeView/OrgUnitTreeviewModal.tsx
```diff
@@ -147,7 +175,7 @@ const OrgUnitTreeviewModal = ({
         if (multiselect) {
             onConfirm([]);
         } else {
-            onConfirm(null);
+            onConfirm(undefined);
```
kemar, line 181: I could learn something here, what is the rationale behind using `undefined` rather than `null` here?

hat/assets/js/apps/Iaso/domains/orgUnits/components/OrgUnitsSmallInfos.js
```diff
@@ -1,17 +1,14 @@
...
-    useGetStatusMessage,
-    getOrgUnitGroups,
-} from '../utils';
+import { getOrgUnitGroups, useGetStatusMessage } from '../utils';
 
 import OrgUnitsSmallInfosRow from './OrgUnitsSmallInfosRow';
+import { getOrgUnitParentsString } from './TreeView/utils';
 
-const OrgUnitsSmallInfos = ({ orgUnit }) => {
+function OrgUnitsSmallInfos({ orgUnit }) {
```
quang-le: Why the change from the arrow function? We use arrow functions for all our function components so I don't understand the reason behind the change

hat/assets/js/apps/Iaso/domains/orgUnits/components/TreeView/OrgUnitTreeviewModal.tsx
quang-le: Not sure it's the case here, but sometimes using `null` will prompt React to complain that we're switching between controlled and uncontrolled inputs, so most times using `undefined` will have the desired effect without popping errors in the console.

I used to use `null` when the value was set by my components, to differentiate from cases where no value was received but, in practice, we sometimes receive `null` from the backend, so 🤷‍♂️.

Also `typeof null` is `object`, so there's that as well... [insert JS rant here]

hat/assets/js/apps/Iaso/domains/orgUnits/components/TreeView/OrgUnitTreeviewModal.tsx
```diff
@@ -180,7 +208,20 @@ const OrgUnitTreeviewModal = ({
             resetSelection();
         }
     }, [resetTrigger, hardReset, resetSelection]);
+
+    useEffect(() => {
+        if (sourceSettings && !isFetchingSource) {
+            setSettings({
+                ...settings,
+                statusSettings: sourceSettings,
+            });
```
quang-le: Wouldn't using a function solve the exhaustive deps problem?

```suggestion
            setSettings(value => ({...value, statusSettings: sourceSettings}));
```

hat/assets/js/apps/Iaso/domains/orgUnits/components/TreeView/requests.ts
```diff
@@ -0,0 +1,199 @@
...
+            break;
+        case 'source':
+            typePrefix = 'data_source_id';
+            break;
+        default:
+            break;
+    }
+
+    const idParam = id && typePrefix ? `&${typePrefix}=${id}` : '';
+    return `${baseApiUrl}?ignoreEmptyNames=true${idParam}${validationStatusString}`;
```
quang-le: Unless the order of the params is important, I think we could use early returns in the switch statement. 
I prefer when we can avoid re-assigning value, but I can also live with it

```suggestion
    const defaultUrl = `${baseApiUrl}?ignoreEmptyNames=true${validationStatusString}`;
    if(!id) return defaultUrl;
    switch (type) {
        case 'version':
            return `${defaultUrl}&version=${id};
        case 'source':
            return `${defaultUrl}&data_source_id=${id};
        default:
            return  defaultUrl;
    }
```

hat/assets/js/apps/Iaso/domains/orgUnits/components/TreeView/requests.ts
```diff
@@ -0,0 +1,199 @@
...
+    const validationStatusString = getValidationStatus(statusSettings);
+    let typeParam = '';
+
+    if (type === 'version') {
+        typeParam = `&version=${input2}`;
+    } else if (type === 'source') {
+        typeParam = `&data_source_id=${input2}`;
+    }
+
+    return `search=${input1}${typeParam}${validationStatusString}`;
```
quang-le: Same as above

hat/assets/js/apps/Iaso/domains/orgUnits/components/TreeView/requests.ts
```diff
@@ -0,0 +1,199 @@
...
+            statusSettings,
+        });
+        const result = await getRequest(url);
+        return result.results;
+    } catch (error: unknown) {
+        dispatch(
+            enqueueSnackbar(errorSnackBar('searchOrgUnitsError', null, error)),
+        );
+        console.error('Error while searching org units:', error);
+        throw error;
```
quang-le: Should we still `throw`? We already handle the snack bar here and the internal hook of the treeview doesn't do anything `onError`

hat/assets/js/apps/Iaso/domains/orgUnits/components/TreeView/useSourceConfig.ts
```diff
@@ -0,0 +1,86 @@
...
+        } else if (versionId && version && !isFetchingVersion) {
+            if (version?.tree_config_status_fields?.length > 0) {
+                sourceSettings = version.tree_config_status_fields;
+            }
+            sourceInfos = {
+                sourceName: version.data_source_name,
+                sourceId: version.data_source,
+                versionNumber: version.number,
+                versionId: version.id,
+            };
```
quang-le: shouldn't the version have priority on the source?

hat/assets/js/apps/Iaso/domains/registry/hooks/useGetOrgUnit.ts
quang-le, line 1: I think `useFetchOrgUnits` should be moved out of registry

hat/assets/js/apps/Iaso/domains/orgUnits/components/TreeView/OrgUnitTreeviewModal.tsx
```diff
@@ -231,83 +275,59 @@ const OrgUnitTreeviewModal = ({
...
+                                withStatusIcon={showStatusIconInTree}
+                                withType={displayTypes}
+                            />
+                        )}
+                        toggleOnLabelClick={toggleOnLabelClick}
+                        onSelect={onOrgUnitSelect}
+                        request={searchOrgUnitsWithSource}
+                        makeDropDownText={orgUnit => (
+                            <OrgUnitLabel orgUnit={orgUnit} />
+                        )}
```
quang-le: This doesn't need to be a component. The `DynamicSelect` in bluesquare-components just puts its return value in a `Typography`.
 (es-lint complains that we pass a component that way)

hat/assets/js/apps/Iaso/domains/orgUnits/components/TreeView/OrgUnitTreeviewModal.tsx
```diff
@@ -231,83 +275,59 @@ const OrgUnitTreeviewModal = ({
...
+                            getChildrenData(id, statusSettings)
+                        }
+                        getRootData={getRootDataWithSource}
+                        label={(orgUnit: OrgUnit) => (
+                            <TreeViewLabel
+                                orgUnit={orgUnit}
+                                withStatusIcon={showStatusIconInTree}
+                                withType={displayTypes}
+                            />
+                        )}
```
quang-le, line 295: This one is a bit harder to get rid off 😅

quang-le (changes requested):
Great job, it was not the easiest code to modify. Thanks for converting everything to TS.

In the org unit search, if the source or version is changed, the treeview is empty, and the footer with the source+version only updates the source (not the version)

https://github.com/BLSQ/iaso/assets/38907762/93e79097-ca54-490d-ab7f-277b325f784a

The text search works though:

https://github.com/BLSQ/iaso/assets/38907762/eebd24c5-fbda-42b5-b1c4-5994ff282633

hat/assets/js/apps/Iaso/domains/orgUnits/components/OrgUnitsSmallInfos.js
beygorghor: auto complete ...

hat/assets/js/apps/Iaso/domains/orgUnits/components/TreeView/useSourceConfig.ts
beygorghor: it was implemented like that before...

beygorghor:
@kemar I have some strange behaviour in the result while changing default data source and filtering org unit on version id



https://github.com/BLSQ/iaso/assets/12494624/80f58c2b-d0da-4f8a-8d4e-d909a1e71705

quang-le (approved):

PR #1340, opened 2024-05-30
Polio-1556 sub activities front end

quang-le:
Add new tab to polio

Related JIRA tickets : POLIO-1556

- [ ] Are there enough tests

## Doc
comments in the code

## Changes

- Add useTableState hook
- fix query key in useSaveCampaigns
- Pass campaign as prop i.o getting it via formik context in ScopeInput
- Add useFilteredDistricts hook to allow filtering on different scopes
- Pass availableVaccines to scopes map to enable more flexible vaccine selection
- Add new components
- update polio dialog tabs
- Modal validation:
    - name, start and end date are mandatory, other fields are optional
    - start and end date are checked against each other and against round dates
- Scope: districts available for selection are computed from round scopes
- Vaccines: for ppolio campaigns, the vaccines available are computed from the round vaccines
- Deactivated snack bar on save because it would trigger a full re-render of the page which would reset the modal state and lose the active tab. Instead, the (sub)modal will stay open while saving (with a spinner), then close so the user returns to the tab with updated data

## How to test

With a polio then a non-polio campaign:
- go to sub-activities tab:
    - create, edit and delete sub-activities

## Print screen / video

https://github.com/BLSQ/iaso/assets/38907762/f80b9847-7b4a-4f81-a437-e3a008116ecd

## Notes

@mathvdh the sorting doens't seem to work. The params are sent to the API, but no sorting happens


plugins/polio/js/src/domains/Campaigns/MainDialog/usePolioDialogTabs.tsx
```diff
@@ -41,13 +42,22 @@ export const usePolioDialogTabs = (
                 title: formatMessage(MESSAGES.rounds),
                 form: RoundsForm,
                 key: 'rounds',
-                diabled: !formik.values.initial_org_unit,
```
beygorghor, line 44: 😉

plugins/polio/js/src/domains/Campaigns/Scope/ScopeForm.tsx
```diff
@@ -76,52 +75,15 @@ export const ScopeForm: FunctionComponent = () => {
...
-    ]);
+    const filteredDistricts: FilteredDistricts[] | undefined =
+        useFilteredDistricts({
+            isPolio,
+            scopes,
+            districtShapes,
+            regionShapes,
+            search: debouncedSearch,
+            searchScope,
+        });
```
beygorghor, line 86: ❤️

beygorghor (changes requested):
Nice ! Very clean  a few remarks:

- Can we add some min and max date using round date ?
Screenshot 2024-06-03 at 09 26 03
- This filed is on error on focus, it should go on error while focu
- s, edited and emptied
Screenshot 2024-06-03 at 09 27 25
- Space between inputs is too large:
Screenshot 2024-06-03 at 09 27 51
- I have some inconsistency with scope round s and sub activities scopes:
https://github.com/BLSQ/iaso/assets/12494624/ebde03de-5db4-4cde-aed2-0bf3ebdbedbf

beygorghor (approved):

PR #1349, opened 2024-06-03
IA-2967 Add calculated fields to submissions filter

bramj:
In the dashboard, on the form submissions page, when you filter results on a specific form, you can choose to filter on attributes in the forms (modal with title "Search in submitted fields").

Here, I've now made the calculated fields also available since it makes a lot of sense to search on these in the light of searching in reference forms for entities.

image

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/forms/fields/constants.ts
+++ b/hat/assets/js/apps/Iaso/domains/forms/fields/constants.ts
@@ -198,7 +198,18 @@ export const iasoFields: Field[] = [
     },
     {
         type: 'calculate',
-        disabled: true,
+        queryBuilder: {
+            type: 'text',
+            excludeOperators: [
+                'proximity',
+                'ends_with',
+                'starts_with',
+                'like',
+                'not_like',
+                'is_empty',
+                'is_not_empty',
+            ],
+        },
     },
     {
         type: 'acknowledge',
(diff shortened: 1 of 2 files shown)


beygorghor (changes requested):
This will work in most of the cases, anyway i think now we have the type of the result that is present in the suffix of the label. See related ticket in JIRA

bramj:
@beygorghor Is this supposed to be in the `possible_fields`? :thinking: 
I can't really see any type suffix on Trypelim, e.g.:
```json
[
    { "name": "trypelim_profile", "type": "calculate", "label": "Trypelim profile" },
    { "name": "serie_id", "type": "calculate", "label": "Serie ID" },
    { "name": "absent", "type": "select one", "label": "Absent" },
    { "name": "position", "type": "select one", "label": "position" },
    { "name": "result", "type": "select one", "label": "Résultat" },
    { "name": "is_suspect", "type": "calculate", "label": "Est suspect (Sérologie Positive)" },
    { "name": "is_not_done", "type": "calculate", "label": "Pas fait" },
    { "name": "first_test_done", "type": "calculate", "label": "Premier test" },
    { "name": "is_absent", "type": "calculate", "label": "Est absent" },
    { "name": "instanceID", "type": "calculate", "label": "" }
]
```

beygorghor:
So let's merge this ! We can always make it better after !

beygorghor (approved):

PR #1351, opened 2024-06-03
IA-2881: Completeness should be limited to the user assigned geography

hakifran:
- Completeness should be limited to the user assigned geography
- The PR is about to fix it
Related JIRA tickets : [IA-2881](https://bluesquare.atlassian.net/browse/IA-2881)

- [x] Are there enough tests

## Changes
- Added a filter on org_unit_path_descendants if the user has assigned org units

## How to test
- Create or edit a user and assign org units to him
- Check on completeness data if only submissions related to the org units(assigned to the user) and its descendants are the only being displayed

## Print screen / video
[Screencast from 2024-06-03 15-35-45.webm](https://github.com/BLSQ/iaso/assets/19631540/5b8283c1-1eb0-42fc-a8c1-c11a0908e438)
[Screencast from 2024-06-03 15-36-40.webm](https://github.com/BLSQ/iaso/assets/19631540/e037361e-e793-45b0-abc8-f73374a8ce94)

[IA-2881]: https://bluesquare.atlassian.net/browse/IA-2881?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iaso/api/completeness.py
```diff
@@ -19,12 +21,18 @@ class CompletenessViewSet(viewsets.ViewSet):
 
     def list(self, request):
         profile = request.user.iaso_profile
+        profile_org_units = OrgUnit.objects.filter(id__in=profile.org_units.all())
```
kemar: Is there a reason for not using `OrgUnit.objects.filter_for_user(request.user)` instead?

iaso/api/completeness.py
```diff
@@ -19,12 +21,18 @@ class CompletenessViewSet(viewsets.ViewSet):
...
+
         queryset = (
             Instance.objects.filter(project__account=profile.account)
             .exclude(deleted=True)
             .exclude(file="")
             .with_status()
         )
+
+        if profile_org_units:
+            queryset = queryset.filter(org_unit__path__descendants=ArraySubquery(profile_org_units.values("path")))
```
kemar: You can have something simpler if you use `filter_for_user()`:

```suggestion
            queryset = queryset.filter(org_unit__in=profile_org_units))
```

kemar (changes requested):
One question and a possible way to have a simpler query.

iaso/api/completeness.py
hakifran: This proposition **OrgUnit.objects.filter_for_user(request.user)** will return all the org units in the pyramid(I mean from the root orgunit related to the user) and if I apply the filter on all the org units like this **queryset = queryset.filter(org_unit__in=profile_org_units))**, from my point of view it will slow down a little bit the loading of the page.

kemar: Ok, I thought it was more readable. Sometimes, sacrificing a little performance for readability can be worth it!

But I won't block the PR for this :)

kemar (approved):
LGTM

iaso/api/completeness.py
hakifran: @kemar ,for the readability, let's sacrifice the little performance. thank you so much

PR #1357, opened 2024-06-05
IA-3021 Add indexes on deleted_at

bramj:
To my big frustration, this doesn't actually work :( `AddIndexConcurrently` appears to have been added in Django 5, I didn't realize while implementing.

Keeping this PR in suspense while I figure out how to proceed.


bramj:
Closing this, it's not the right time. I'll reconsider when we move to Django 5.

PR #1361, opened 2024-06-06
IA-3021 Remove user profile's location restrictions when searching entities via the mobile app

bramj:
The whole point is to be able to find someone when the data is not on the mobile device.

E.g. for the QR codes for the MILDA distribution: when someone is not found on the device, it means they must be from another org unit, if not the entity would be found instantly on the device.

### DIFF
--- a/iaso/api/mobile/entity.py
+++ b/iaso/api/mobile/entity.py
@@ -32,15 +32,22 @@ def filter_for_mobile_entity(queryset, request):
     return queryset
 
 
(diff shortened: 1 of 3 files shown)


bramj:
Comment from @mestachs 

> note that solution of  dhis2 has 2 access rights on the "encoding form" under a certain zone, and search for tei in another zone (or country level)
image

madewulf (approved):

PR #1365, opened 2024-06-07
POLIO-1562 sub activities bugs

quang-le:
Sub activities UI bugs:
- Vaccines legend had no default value
- Modal title was too vague
- Saving sub activity for a not yet saved round returned an error
- Vaccines and round dates were too restrictive

Related JIRA tickets : POLIO-1562

- [ ] Are there enough tests

## Doc
Comments in the code

## Changes

- Add round number and campaign obr name to modal title
- Remove restrictions on dates and vaccines, i.e they don't have to "match" those of the round
- Save sub-activities via `useSaveCampaigns`:
   - add field `subactivity` to formik state (update TS accordingly)
   - refactor `useSaveCampaigns` to:
       - remove the `subactivity` field and pass the rest of the formik values to the campaigns API
       - make the call to the campaigns API
       - if campaign save is successful: 
           - The snavckbar will iindicate that the campaign was saved
           - the `subactivity` from formik (if it exists) will be sent to the subactivities API.
           - the snack bar will indicate whether it was sucessfully saved or not
       - if campaign save failed:
           - The error snackbar will pop up
           - The subactivity API will not be called 
       - use `useMutation`i.o `useSnackMutation` in order to manage snack bars directly 

## How to test

Go to campaigns page, test all cases with both polio and non-polio campaigns, and for campaigns with and without separate scopes per round:

- Create a new campaign with subactivities. Try using dates outside of roud dates range and different vaccines from the round (eg: round has nOPV2, make sub activity with bOPV)
- Create a new campaign without subactivities
- Edit a campaign without editing subactivities
- Edit a campaign and its subactivities (edit or add activity)

When creating/editing subactivities, test error handling by:
- changing the campaign API url in order to have an error on campaign save
- same with subactivities url

## Print screen / video

https://github.com/BLSQ/iaso/assets/38907762/cb69bdb6-42f9-410e-89ba-cb6d4115ccbc


beygorghor (approved):
i'm stil having strange modal behaviour while saving, we should really consider split the save of sub activities and the whole campaign.

https://github.com/BLSQ/iaso/assets/12494624/d55534a6-a293-48cc-9255-f74b6f97363c

Disabled shapes have still an active cursor:
https://github.com/BLSQ/iaso/assets/12494624/78eac765-ce73-428a-b9a8-2f83760829e7

White space between dates is weird;

Screenshot 2024-06-17 at 11 27 45

Those are not blocking

quang-le:
> i'm stil having strange modal behaviour while saving, we should really consider split the save of sub activities and the whole campaign.
> Screen.Recording.2024-06-17.at.11.27.04.mov
> 
> Disabled shapes have still an active cursor: https://github.com/BLSQ/iaso/assets/12494624/78eac765-ce73-428a-b9a8-2f83760829e7
> 
> White space between dates is weird;
> 
> Screenshot 2024-06-17 at 11 27 45
> 
> Those are not blocking

Yeah, but we have to save both in succession, because if users create a new round and sub-activities for this round, we need to save it in the backend before we can save sub-activities that are attached to it.
I agree it all looks a bit funny though

PR #1372, opened 2024-06-11
IA-3070: registry list select

beygorghor:
We should be able to select from the list of org units to fill the detail.
CleanShot 2024-06-04 at 21 44 20@2x-20240604-194523

Related JIRA tickets : IA-3070

- [ ] Are there enough tests

## Changes

Allow to select an org unit in the list view

## How to test

Open registry on a org unit having children. select list view and clic on one children name. It should be displayed on the right.

## Print screen / video

https://github.com/BLSQ/iaso/assets/12494624/feb445c5-3748-4c39-9fb4-cb57462a7529


quang-le (approved):
I think some color code to highlight the selected org unit in the list would be nice, but it already does the job.

madewulf (approved):

madewulf (approved):
I did write the typescript and react!

PR #1375, opened 2024-06-12
IA-3089: remove create submission button in org unit details page

quang-le:

Related JIRA tickets : IA-3089

- [ ] Are there enough tests

## Doc
left a comment in the code

## Changes

- only show button if no rgUnitId is passed to the columns hook

## How to test

- Go to the details page of an org unit that has submissions
- Go to forms tab
- There should be no action button to add a submission in the forms table

## Print screen / video

Screenshot 2024-06-12 at 09 35 11

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/forms/config/index.js
+++ b/hat/assets/js/apps/Iaso/domains/forms/config/index.js
@@ -238,69 +238,78 @@ export const useFormsTableColumns = ({
                                     {/* If orgUnitId is not undefined, it means the btable is used in the org units details page,
                                     which, in turn, means we shouldn't show the button to create a submission */}
                                     {!orgUnitId && (
-                                        <DisplayIfUserHasPerm
-                                            permissions={[
-                                                Permission.SUBMISSIONS_UPDATE,
-                                            ]}
-                                        >
-                                            <CreateSubmissionModal
-                                                titleMessage={
-                                                    MESSAGES.instanceCreationDialogTitle
-                                                }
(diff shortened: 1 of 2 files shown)


madewulf:
I think this is great, but I'm also understanding, by looking at this, that we should also hide at least the delete buttons (this probably explains why the forms are sometimes deleted while people think that they delete the submission), but I even think that we should hide everything except the list button. 

image

hat/assets/js/apps/Iaso/domains/forms/config/index.js
```diff
@@ -235,40 +235,45 @@ export const useFormsTableColumns = ({
...
-                                        <CreateSubmissionModal
-                                            titleMessage={
-                                                MESSAGES.instanceCreationDialogTitle
-                                            }
-                                            confirmMessage={MESSAGES.ok}
-                                            cancelMessage={MESSAGES.cancel}
-                                            formType={{
-                                                id: settings.row.original.id,
-                                                periodType:
+                                    {/* If orgUnitId is not undefined, it means the btable is used in the org units details page,
```
beygorghor, line 238: btable 😉

beygorghor (approved):

quang-le:
> I think this is great, but I'm also understanding, by looking at this, that we should also hide at least the delete buttons (this probably explains why the forms are sometimes deleted while people think that they delete the submission), but I even think that we should hide everything except the list button.
> 
> image

@madewulf  Should I update the PR to hide all buttons then?

PR #1376, opened 2024-06-12
IA-3056 Change Requests performance improvements

kemar:
Change Requests performance improvements.

Related JIRA tickets : [IA-3056](https://bluesquare.atlassian.net/browse/IA-3056)

## Changes

- added missing `prefetch_related`
- used `m2m.add()` (which uses `bulk_create()`) instead of `m2m.set()`
- used `bulk_create(new_reference_instances)`
- used `id`s instead of `instance`s when possible
- added `write_only` to some serializers fields

## How to test

- create a change request on the mobile application then upload it to the server
- approve a change request on the website

[IA-3056]: https://bluesquare.atlassian.net/browse/IA-3056?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iaso/models/org_unit.py
+++ b/iaso/models/org_unit.py
@@ -737,16 +737,15 @@ class OrgUnitChangeRequest(models.Model):
             if field_name == "new_location_accuracy":
                 continue
             elif field_name == "new_groups":
-                self.org_unit.groups.set(self.new_groups.all())
+                self.org_unit.groups.clear()
+                self.org_unit.groups.add(*self.new_groups.all())
             elif field_name == "new_reference_instances":
-                # Delete old ones.
                 self.org_unit.reference_instances.clear()
-                for instance in self.new_reference_instances.all():
-                    OrgUnitReferenceInstance.objects.create(
-                        org_unit=self.org_unit,
-                        form=instance.form,
-                        instance=instance,
-                    )
+                new_reference_instances = [
+                    OrgUnitReferenceInstance(org_unit_id=self.org_unit.pk, form_id=instance.form_id, instance=instance)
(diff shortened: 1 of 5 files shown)


iaso/models/org_unit.py
```diff
@@ -689,17 +689,17 @@ def save(self, *args, **kwargs):
...
+            self.old_org_unit_type_id = self.org_unit.org_unit_type_id
             self.old_location = self.org_unit.location
             self.old_opening_date = self.org_unit.opening_date
             self.old_closed_date = self.org_unit.closed_date
         super().save(*args, **kwargs)
         if is_new:
             # Wait for the instance to have an ID to save old m2m relations
-            self.old_groups.set(self.org_unit.groups.all())
-            self.old_reference_instances.set(self.org_unit.reference_instances.all())
+            self.old_groups.add(*self.org_unit.groups.all())
```
madewulf, line 701: I'm curious, why is this useful?

kemar: Because using `add()` will not call any `save()` methods, but rather create the relationships using `QuerySet.bulk_create()`.

I wasn't aware of this, see https://docs.djangoproject.com/en/dev/ref/models/relations/#django.db.models.fields.related.RelatedManager.add

madewulf: Ooooh, nifty

madewulf: I'd be curious to see the change in the generated SQL, but we can have that discussion outside of github :-)

madewulf (approved):

PR #1377, opened 2024-06-12
IA-3026: prevent loading map tiles if map is not displayed

quang-le:

Related JIRA tickets : IA-3026

- [ ] Are there enough tests

## Doc
NA

## Changes

- Replaced "hiddenOpacity" class with conditional rendering for map in org unit list and org unit details pages
- Fixed cypress tests

## How to test

- Open browser inspector -> network tab. Make sure to display all, not only XHR
- Go to org units list and search
- Go to map tab: the tiles should only be loaded when switching tabs
- Same operation with org unit detail

## Print screen / video

https://github.com/BLSQ/iaso/assets/38907762/fa61efd1-fdc0-4294-94fa-c61de960f767

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/orgUnits/details.js
+++ b/hat/assets/js/apps/Iaso/domains/orgUnits/details.js
@@ -1,3 +1,4 @@
+/* eslint-disable react/function-component-definition */
 /* eslint-disable camelcase */
 import { Box, Grid, Tab, Tabs } from '@mui/material';
 import { makeStyles } from '@mui/styles';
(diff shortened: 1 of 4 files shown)


madewulf:
Does this also prevent the download of the org units for the map (as mentioned in the ticket:  "and up to 3000 org units to display on the map") ?

beygorghor (approved):
Users will loose zoom, selected layer, selected org unit while changing tab. Is this is what we want then ok

quang-le:
> Does this also prevent the download of the org units for the map (as mentioned in the ticket: "and up to 3000 org units to display on the map") ?

No, I think we need to merge the "SingleTable" PR for that

quang-le:
@madewulf @beygorghor  I suggest we remove this from the release and I'll work out a better solution after we've merge the `SingleTable` PR. What do you think?

quang-le:
closing because stale

PR #1378, opened 2024-06-12
IA-3091 Fix domain name in invitation email

kemar:
Fix domain name in invitation email.

Related JIRA tickets : [IA-3091](https://bluesquare.atlassian.net/browse/IA-3091)

## Changes

Use `DNS_DOMAIN` instead of both:

- `current_site.domain`
- `current_site.name`

## How to test

The unit test should be enough.

[IA-3091]: https://bluesquare.atlassian.net/browse/IA-3091?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iaso/api/profiles.py
+++ b/iaso/api/profiles.py
@@ -4,7 +4,6 @@ from django.conf import settings
 from django.contrib.auth import models, update_session_auth_hash
 from django.contrib.auth.models import Permission, User
 from django.contrib.auth.tokens import PasswordResetTokenGenerator
-from django.contrib.sites.shortcuts import get_current_site
 from django.core.exceptions import ObjectDoesNotExist
 from django.core.mail import send_mail
 from django.core.paginator import Paginator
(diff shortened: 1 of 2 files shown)


iaso/api/profiles.py
```diff
@@ -744,5 +743,5 @@ def delete(self, request, pk=None):
 L'équipe {{site_name}}.</p>
     """
 
-    EMAIL_SUBJECT_FR = "Configurer un mot de passe pour votre nouveau compte sur {dns_domain}"
-    EMAIL_SUBJECT_EN = "Set up a password for your new account on {dns_domain}"
+    EMAIL_SUBJECT_FR = "Configurer un mot de passe pour votre nouveau compte sur {site_name}"
```
madewulf: I think we should use DNS_DOMAIN here too and not use current_site.domain at all

kemar: OK, as you want!

madewulf (approved):

PR #1382, opened 2024-06-13
IA-1164: navigate to correct tab from org unit children

quang-le:
Redirections from OU children table didn't redirect to correct tab

Related JIRA tickets : IA-1164

- [ ] Are there enough tests

## Doc
NA

## Changes

- Remove tab state value and use params.tab instead
- Add redirection to add default tab value when navigation to OU details
- delete outdated duplicate useRedirections hook

## How to test

- Select an org unit with children
- Go to Children tab
- Find a child with location
- click on the map action button: you should be redirected to the child's page, on the map tab
- go back
- click on the eye action button: you should be redirected to the child's page, on the infos tab

## Print screen / video

https://github.com/BLSQ/iaso/assets/38907762/5525ea45-9491-4054-82ff-47265408a6f0

## Notes

Merges into IA-2976


hat/assets/js/apps/Iaso/domains/orgUnits/details.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable react/function-component-definition */
```
beygorghor: this can be removed now no ?

hat/assets/js/apps/Iaso/domains/orgUnits/details.js
beygorghor, line 1: const { tab } = params;  ?

beygorghor (approved):
This is working, just small remarks

PR #1396, opened 2024-06-18
IA-3115: snackbar no redux

beygorghor:

Related JIRA tickets : IA-3115

- [ ] Are there enough tests

## Changes

Make snackbar working with event listener, no redux required

## How to test

On the form creation page, swap tab Version to Attachement before saving, it should display one error, not looping forever


hat/assets/js/apps/Iaso/components/snackBars/EventDispatcher.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    removeEventListener(event, callback) {
+        if (this.listeners[event]) {
+            this.listeners[event] = this.listeners[event].filter(
+                listener => listener !== callback,
+            );
+        }
+    }
+
+    dispatch(event, data) {
```
quang-le, line 24: Is there any gain from the `event` parameter? It's going to be `"snackbar"` in every case, so maybe we could pass the event to the constructor, and save it as a private property of the `EventDispatcher`

That way we could call `dispatch(data)`, and pass only one param

hat/assets/js/apps/Iaso/components/snackBars/EventDispatcher.ts
```diff
@@ -0,0 +1,31 @@
...
+        if (this.listeners[event]) {
+            this.listeners[event].forEach(callback => callback(data));
+        }
+    }
+}
+
+export const dispatcher = new EventDispatcher();
```
quang-le, line 31: If it doesn't break anything, I think it would be neat to export something like 

```typescript
export const openSnackBar = dispatcher.dispatch
```

Then we could import and  re-export `closeSnackBar` from notistack here as well, and manage all our snackbar related methods from the same place. Imports in the codebase wouldn't be split between here and notistack

hat/assets/js/apps/Iaso/domains/forms/components/OrgUnitTypeFilterComponent.js
```diff
@@ -57,7 +55,6 @@ const OrgUnitTypeFilterComponent = props => {
             if (!ot.orgUnits) {
                 promisesArray.push(
                     fetchSubOrgUnitsByType(
-                        dispatch,
```
quang-le, line 60: ❤️

hat/assets/js/apps/Iaso/domains/links/components/LinksDetailsComponent.js
```diff
@@ -46,11 +46,11 @@ class LinksDetails extends Component {
     }
 
     fetchDetail() {
-        const { dispatch, linkId } = this.props;
+        const { linkId } = this.props;
```
quang-le, line 49: You can also remove dispatch from `MapStateToProps`

hat/assets/js/apps/Iaso/redux/snackBarsReducer.js
quang-le, line 1: ❤️

package-lock.json
quang-le, line 1: What changed in the package-lock

plugins/polio/js/src/domains/VaccineModule/SupplyChain/hooks/api/vrf.tsx
```diff
@@ -206,25 +205,22 @@ export const saveVrf = (
 
 export const handleVrfPromiseErrors = (
     data: ParsedSettledPromise<VRF>[],
-    dispatch: Dispatch,
```
quang-le, line 209: You need to update `useSaveSupplyChainForm` where this method is called

quang-le (changes requested):
Works fine, very cool work!

I'm just blocking now on my comments about the re-naming of `dispatcher.dispatch` and grouping the exports of snackbar related methods in the same place . Better to change it now that you're touching this code and enjoy a bit more comfort while using this feature in the long run.

hat/assets/js/apps/Iaso/components/snackBars/EventDispatcher.ts
beygorghor: my guess is that we can use this event dispatcher for other purpose than snackbar, so i would leave it like that

package-lock.json
beygorghor: i have upgraded a library without telling anyone ㊙️

quang-le (approved):
AWESOME!

PR #1398, opened 2024-06-19
IA-3046: Registry read only and reference submission

beygorghor:
On instance staging (account = kqrxutg)

If I have all permission, I see this ==> 

image
 

If I have read only I have this

image

image
 

If Read only, user should still be able to visualise the form

Also on the bottom part

Related JIRA tickets : IA-3046

- [ ] Are there enough tests

## Changes

Remove isReference from api call

## How to test

You need a user with only read permission on registry and an org unit with a reference submission and a regular submission attached to it. While visiting registry on this org unit you should have the same result as somebody with write permission.


bramj (approved):
Works as advertised :+1: 

I do admit that the `onlyReference` query param is not clear to me. I'd expect it being set to true to only return the reference submission, but instead it doesn't return anything? :thinking:

PR #1404, opened 2024-06-19
HOTFIX: Do not download full tree for users without org units in their profile

kemar:
Do not download full tree for users without org units in their profile.


kemar:
Replaced by https://github.com/BLSQ/iaso/pull/1405

PR #1408, opened 2024-06-20
Ia 2993 setuper in openhexa

quang-le:
Refactor setuper to facilitte its use in an openhexa pipeline

Related JIRA tickets : IA-2993

- [ ] Are there enough tests

## Doc

## Changes

- Add arguments to pass credentials as an alternative to using a `credentials` file
- Add a returned value to the create_account function

## How to test

Just run the setuper 🙂


setuper/setuper.py
```diff
@@ -85,5 +83,33 @@ def setup_account(account_name):
...
+    parser.add_argument("-a", "--additional_projects", action="store_true")
+
+    args = parser.parse_args()
+    server_url = args.server_url
+    username = args.username
+    password = args.password
+    additional_projects = args.additional_projects
+
+    if server_url is None or username is None or password is None:
+        from credentials import *
```
bramj, line 105: Maybe we can catch the case were I call the script, but forget one of the params. Right now I get an error because of the credentials file not existing:

```shell
$ python3 setuper.py --username test
Traceback (most recent call last):
  File "/home/bram/Work/iaso/setuper/setuper.py", line 105, in <module>
    from credentials import *
ModuleNotFoundError: No module named 'credentials'
```

I guess it would be better to display the `sys.exit(` message in this case.

bramj (commented):
Works well! :+1:  Only one comment on error-messaging when calling the script with arguments.

PR #1409, opened 2024-06-20
IA-3134 Allow a super user to view the default tree root

kemar:
Allow a super user to view the default tree root even for those having org units linked to their profile.

Related JIRA tickets : [IA-3134](https://bluesquare.atlassian.net/browse/IA-3134)

[IA-3134]: https://bluesquare.atlassian.net/browse/IA-3134?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iaso/api/org_unit_tree/views.py
+++ b/iaso/api/org_unit_tree/views.py
@@ -60,7 +60,8 @@ class OrgUnitTreeViewSet(viewsets.ModelViewSet):
         display_root_level = not parent_id
 
         if display_root_level and self.action == "list":
-            if user.is_authenticated and user.iaso_profile.org_units.exists():
+            force_full_tree = force_full_tree or user.is_superuser
+            if not force_full_tree and user.is_authenticated and user.iaso_profile.org_units.exists():
                 # Root level of the tree for this user (the user may be restricted to a subpart of the tree).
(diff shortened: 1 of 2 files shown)


iaso/api/org_unit_tree/views.py
```diff
@@ -60,7 +60,8 @@ def get_queryset(self):
         display_root_level = not parent_id
 
         if display_root_level and self.action == "list":
-            if user.is_authenticated and user.iaso_profile.org_units.exists():
+            force_full_tree = force_full_tree or user.is_superuser
```
beygorghor, line 63: front-end is never using this param, i don't understand why we have it

kemar: This is currently unused but it was asked by @madewulf and specified in the design doc here https://docs.google.com/document/d/1LczWRYbDCJtbxG2fd_c1WtQ4DmNclRG2ixXnr_C58wo

I thought about removing it because I misimplemented it at first, but now it works and it's well tested.

I can remove it if everyone is OK, or we could use it in the front-end for a new feature.

mathvdh (approved):
Seems OK for me

PR #1416, opened 2024-06-21
POLIO-1583 POLIO-1584 POLIO-1585: Calendar -Non Polio campaigns in orange

beygorghor:
Non polio campaigns should have another specific color.
Also campaign legend is a bit buggy:

https://github.com/BLSQ/iaso/assets/12494624/0e995db4-8c5b-4e25-a4a8-635e3e15a82a

Related JIRA tickets : POLIO-1583 POLIO-1584 POLIO-1585

- [ ] Are there enough tests

## Changes

- remove campaign color for merged shapes as not used anymore. If no vaccine display theme secondary color (orange for Poio)
- apply same logic  on regular shapes
- add Other in map legend
- fix campaigns legend styling issues
- display same color while selecting a scope on non polio campaigns
- adapt celle background color to the same color if a scope exists and no vaccine is used

## How to test

You need a polio campaign with a scope, a non polio one also with scope and a third one without scope.
Make it display at the same time on the calendar for tests purposes.
The polio campaign should use vaccine scope colors on the cells and on the map
The non polio one should use secondary color
The third one should not be displayed on the map but should be in grey.

Make sure you have the same behaviour while zooming on shapes and use non merged shapes, displaying borders according to campaigns legend

## Print screen / video

https://github.com/BLSQ/iaso/assets/12494624/c5ffa54b-a350-492b-8f73-d10444e918e2

Screenshot 2024-06-21 at 13 23 29
Screenshot 2024-06-21 at 13 23 22


plugins/polio/js/src/constants/virus.ts
```diff
@@ -55,4 +58,36 @@ const polioVaccines: PolioVaccine[] = [
     },
 ];
 
-export { polioVaccines, polioViruses };
+export const useOtherVaccineColor = (): string => {
+    const theme = useTheme();
+    return theme.palette.secondary.main;
+};
```
quang-le: Not sure I would do that. I'd rather copy paste the hex value in another constant. There's no reason to tie the map color to the theme

plugins/polio/js/src/constants/virus.ts
```diff
@@ -55,4 +58,36 @@ const polioVaccines: PolioVaccine[] = [
...
+        {
+            value: 'bOPV',
+            label: 'bOPV',
+            color: '#ffff00',
+        },
+        {
+            value: 'other',
+            label: formatMessage(MESSAGES.other),
+            color: otherVaccineColor,
+        },
```
quang-le:
```suggestion
    return [
       ...polioVaccines,
        {
            value: 'other',
            label: formatMessage(MESSAGES.other),
            color: otherVaccineColor,
        },
```

plugins/polio/js/src/domains/Calendar/campaignCalendar/cells/RoundCell.tsx
```diff
@@ -45,7 +45,16 @@ export const RoundCell: FunctionComponent<Props> = ({
 
     const { anchorEl, setAnchorEl } = useContext(RoundPopperContext);
     const [self, setSelf] = useState<HTMLElement | null>(null);
-
+    const otherVaccineColor = useOtherVaccineColor();
+    const getCellColor = (vaccine: string) => {
+        if (!hasScope(campaign)) {
+            return '#bcbcbc';
```
quang-le: This too could be in a constant

plugins/polio/js/src/domains/Calendar/campaignCalendar/map/CalendarMapPanesMerged.tsx
```diff
@@ -1,39 +1,46 @@
...
+    const getColor = useCallback(
+        (vaccine: string | undefined) => {
+            const vaccineColor = polioVaccines.find(
+                v => v.value === vaccine,
+            )?.color;
+            return vaccineColor || otherVaccineColor;
+        },
+        [otherVaccineColor],
+    );
+
```
quang-le: if `otherVaccineColor`was in a `const` we could extract `getColor` and skip `useCallback`

plugins/polio/js/src/domains/Calendar/campaignCalendar/map/CampaignsLegend.tsx
```diff
@@ -1,64 +1,99 @@
...
+        '&.Mui-expanded': {
+            minHeight: theme => theme.spacing(6),
+        },
+    },
+    details: {
+        paddingTop: 0,
+    },
+    mapLegendCampaigns: {
+        width: 165,
+    },
```
quang-le, line 35: This is just copied from `useStyles`. I don't see what we gain from moving  this styling to `sx` if nothing changes. Now if we need to change this styling we have to do it in 2 places

plugins/polio/js/src/domains/Calendar/campaignCalendar/map/CampaignsLegend.tsx
```diff
@@ -1,64 +1,99 @@
...
+        justifyContent: 'flex-start',
+        height: '100%',
+        alignItems: 'center',
+        display: 'flex',
+        wordBreak: 'break-all',
+    },
+    mapLegendCampaignTitle: {
+        fontSize: 14,
+        fontWeight: 'bold',
+    },
```
quang-le, line 61: same

plugins/polio/js/src/domains/Calendar/campaignCalendar/map/VaccinesLegend.tsx
```diff
@@ -1,15 +1,16 @@
...
 import { useSafeIntl } from 'bluesquare-components';
+import React, { FunctionComponent } from 'react';
 import { useStyles } from '../Styles';
 
 import MESSAGES from '../../../../constants/messages';
-import { polioVaccines } from '../../../../constants/virus';
+import { polioVaccines, useMapLegend } from '../../../../constants/virus';
 
-type Vaccine = typeof polioVaccines[number];
+type Vaccine = (typeof polioVaccines)[number];
```
quang-le: Are you sure this is correct?
`typeof polioVaccines` will always be `"object"` so `(typeof polioVaccines)[number]` will be letter, no?

plugins/polio/js/src/domains/Campaigns/Scope/Scopes/MapScope.tsx
```diff
@@ -86,7 +68,7 @@ export const MapScope: FunctionComponent<Props> = ({
                 );
                 return {
                     ...selectedPathOptions,
-                    color: vaccine?.color || theme.palette.primary.main,
+                    color: vaccine?.color || theme.palette.secondary.main,
```
quang-le: I think it would be cleaner to change how we define `vaccine` lines 66-67 with something like

```typescript
    const color = (availableVaccines.find(
                    v => v.value === scope.vaccine,
                ) ?? otherVaccine).color; // where otherVaccine is defined in an exported const
    return { ...selectedPathOptions, color }
```
Or something more legible in the same vein.

again in the spirit of decoupling map colors from theme (or at least theme's secondary color)

quang-le (changes requested):
Nice. I just think we shouldn't couple the shape color to the theme's primary or secondary color

plugins/polio/js/src/domains/Calendar/campaignCalendar/map/CampaignsLegend.tsx
beygorghor: i prefere move to sx has we don't have to use important anymore, sx is adding directly style to mui styilng, avoiding css hierarchy problem. Unfortunately mapLegendCampaigns us used somewhere alse, so i kept it

quang-le (approved):
very orange indeed

PR #1436, opened 2024-06-28
IA-3154 default source version and tree view

beygorghor:
Tree view is not a set on selected version while loading
Remove datasources without versions from the dropdown options.
Preselect the latest version if no default version is available.
Display the description and indicate if a version is the default in the version dropdown options.

Related JIRA tickets : IA-3154

- [ ] Are there enough tests

## Changes

- branch tree view on params to get selected version
- filter data source without version
- preselect first version if no default
- adapt version label to display number, desc and is default

## How to test

You need multiple data source:
- one without version
- one with versions but no default
- one with versions and a default

Go to org unit page and play with those sources, check you don't have the source without version.
Double check tree usage and version label on other pages

## Print screen / video

Screenshot 2024-06-28 at 12 07 47
Screenshot 2024-06-28 at 12 07 29
Screenshot 2024-06-28 at 12 07 21


hat/assets/js/apps/Iaso/domains/orgUnits/hooks/requests/useGetDataSources.ts
```diff
@@ -4,25 +4,28 @@ import { UseQueryResult } from 'react-query';
...
+        `/api/datasources/?filter_empty_versions=${filterEmptyVersions}`,
+    );
 };
 
-export const useGetDataSources = (): UseQueryResult<
-    DropdownOptions<string>[],
-    Error
-> => {
+export const useGetDataSources = (
+    filterEmptyVersions = false,
```
quang-le, line 22: you need to put it in the query key

quang-le (changes requested):
just the query key thing

PR #1455, opened 2024-07-04
IA-278 Add source_updated_at and source_created_at timestamps on Instance and OrgUnit

bramj:
See https://bluesquare.atlassian.net/browse/IA-278

When the mobile app syncs a bunch of forms, the first call to `/instances` will contain these timestamps for the forms. We then create records in `iaso_instance` with these values.

The problem with this is that the `created_at` and `updated_at` are supposed to refer to creation and update in the backend DB. This means that:

- We don't really have the DB `created_at`. Maybe not a big issue, but is very useful for debugging when needed.
- The `updated_at` is not really taken from the form. Since it's a special field, a simple `instance.save()` a bit later on in the import code overwrites it. Right now I would like to compare the _mobile app's_ `updated_at` to know if I need to update a reference form when it already exists on upload and I'm blocked by this.

## The solution (it already partially existed!):

Add new timestamps for the data from the mobile app. To maintain Django convention we add `source_created_at` and `source_updated_at` so that the existing ones refer to the database insertion/update.

There was already an old PR to address this issue: https://github.com/BLSQ/iaso/pull/21

For some now-forgotten reason it was later reverted, potentially because of SQL queries taking to long. To avoid this issue, I take care to run migrations outside db transactions and create indexes concurrently.

Details:

- Add source_updated_at and source_created_at on Instance
- Add source_created_at on orgunit
- Instances API: when creating from mobile use the source_updated_at and source_created_at
- Update references to use these new timestamps on Instances and Org
Units

Notes:

- No source_updated_at on orgunit: In fact we never directly update an orgunit from the mobile app, so this
timestamp is unnecessary. We also won't add this in the future, because
that's what the change requests are for.
- Migrations: Filling data and adding db indexes for timestamps on instances and orgunits

I've discovered the option `state_operations` to add custom SQL while
having Django detect the migration as being created (i.e. it won't show
up again on a makemigrations)!

This is great, it allows the combo of `atomic = False` to remove the db
transaction + `CREATE INDEX CONCURRENTLY` to add these indexes on
potentially very large tables without locking them.


iaso/models/base.py
```diff
@@ -949,6 +954,14 @@ class Instance(models.Model):
     # See public_create_url workflow in enketo/README.md. used to tell we should export immediately
     to_export = models.BooleanField(default=False)
 
+    class Meta:
+        indexes = [
+            models.Index(fields=["created_at"]),
```
kemar, line 959: I've never tested it myself, but I've read that a  `BrinIndex` is very efficient with large tables (see https://hakibenita.com/9-django-tips-for-working-with-databases#brin-indexes).

Could be worth a try?

```django
from django.contrib.postgres.indexes import BrinIndex
…
BrinIndex(fields=["created_at"]),
```

iaso/models/base.py
```diff
@@ -908,12 +908,17 @@ class Instance(models.Model):
     REFERENCE_FLAG_CODE = "flag"
     REFERENCE_UNFLAG_CODE = "unflag"
 
-    created_at = models.DateTimeField(auto_now_add=True)
+    # Previously created_at and update_at where filled by the mobile, now they
```
```suggestion
    # Previously created_at and update_at were filled by the mobile, now they
```

iaso/models/base.py
```diff
@@ -908,12 +908,17 @@ class Instance(models.Model):
...
+    # have been replaced by `source_created_at`` and `update_created_at``.
```
```suggestion
    # have been replaced by `source_created_at` and `update_created_at`.
```

iaso/api/devices.py
```diff
@@ -21,7 +21,7 @@ class Meta:
     def get_synched_at(obj: Device):
         # updated_at field is better compared to created_at
         instance = Instance.objects.filter(device__id=obj.id).order_by("-updated_at").first()
-        return instance.created_at.timestamp() if instance else None
+        return instance.source_created_at.timestamp() if instance else None
```
kemar:
```suggestion
        return instance.source_created_at.timestamp() if instance and instance.source_created_at else None
```

This can now fail because `source_created_at` is nullable.

iaso/dhis2/datavalue_exporter.py
```diff
@@ -32,7 +32,7 @@ def get_event_date(instance, form_mapping):
     event_date_source = MappingVersion.get_event_date_source(form_mapping)
 
     if event_date_source == MappingVersion.EVENT_DATE_SOURCE_FROM_SUBMISSION_CREATED_AT:
-        return instance.created_at.strftime("%Y-%m-%d")
+        return instance.source_created_at.strftime("%Y-%m-%d")
```
kemar: I think this can fail if `instance.source_created_at` is `null`.

kemar (changes requested):
Almost good to me.

But I think we need to be more defensive if `source_created_at` is `null`. What do you think?

iaso/models/base.py
bramj: Interesting, I didn't know this! BRIN index indeed sounds perfect for creation timestamps. However in this case, we've (ab)used the `created_at` timestamp until this PR and set it to the creation in of the form in the mobile app. So I think, the "some natural correlation with their physical location within the table" part of why we should use it doesn't apply here.
I think here it'd be safer to go with a regular B-Tree, what do you think?

bramj:
@kemar You're absolutely right about the potential issues in case of `NULL`.

I wonder if we should just fall back to the db `created_at` upon instance creation (like right now actually) to avoid issues. There are hard-coded SQL queries for stats for example that will be a bit tricky to update.
What do you think?

iaso/models/base.py
kemar: Yup, B-Tree sounds OK by me!

kemar:
> @kemar You're absolutely right about the potential issues in case of `NULL`.
> 
> I wonder if we should just fall back to the db `created_at` upon instance creation (like right now actually) to avoid issues. There are hard-coded SQL queries for stats for example that will be a bit tricky to update. What do you think?

@bramj I don't know if the logic would still be correct but that might be a solution.

Hard-coded SQL queries could look something like that: `SELECT COALESCE(source_created_at, created_at)  as created_at`.

iaso/tests/api/org_unit_change_requests/test_org_unit_change_requests.py
```diff
@@ -330,8 +330,8 @@ def test_export_to_csv(self):
         """
         It tests the CSV export for the org change requests list.
         """
-        m.OrgUnitChangeRequest.objects.create(org_unit=self.org_unit, new_name="Foo")
-        change_request = m.OrgUnitChangeRequest.objects.create(org_unit=self.org_unit, new_name="Bar")
+        change_request = m.OrgUnitChangeRequest.objects.create(org_unit=self.org_unit, new_name="Foo")
+        m.OrgUnitChangeRequest.objects.create(org_unit=self.org_unit, new_name="Bar")
```
bramj, line 334: @kemar I forgot the ask the first time round: The order of the result from this test changed, this change fixed it. Do you think this is okay or did I break something with this?

bramj:
@kemar Okay I've opted to not set the `source_*` timestamp on creation if there isn't any, to keep the data correct.

So I've added defensiveness everywhere to avoid null pointer exceptions. And on statistics and DHIS2 exports I fall back to the regular `created_at` just in case to avoid breaking things. I think the implementation is now reasonable, let me know what you think!

iaso/tests/api/org_unit_change_requests/test_org_unit_change_requests.py
kemar: @bramj that's really strange because the query seems to `ORDER BY "iaso_orgunit"."name" ASC`…

Anyway it's unlikely you broke something. My guess is the test has a bit of flakiness.

kemar (approved):
LGTM

PR #1456, opened 2024-07-04
SLEEP-1448 Bulk upload: support form update and bugfix

bramj:
https://bluesquare.atlassian.net/browse/SLEEP-1448

- Iterate over the instances.json file instead of the folders to avoid files not being found in case of a form that already exists. In this case the instance already exists in the db, but potentially with a different file_name if it was uploaded from a different tablet (nearby share does not preserve filenames).
- If during processing we get a form that already exists, an for which the bulk upload contains a newer form (with a more recent `updated_at` timestamp), we update the form and set the new `source_updated_at`.
- Ignore .xml files when processing attachments to circumvent a bug in the mobile app


bramj:
@madewulf The form update support is only implemented for the bulk upload. I tried to make it work more general for the regular upload as well, but I'm not sure how to pull it off. Because of the separation between the `/instances` call and the `/form_upload`, it's really hard to know if a form_upload should overwrite the file, since the updated_at timestamp is not in that call. One way would be to remove the current file from the form during the `/instances` call, but I find that a bit dangerous if anything goes wrong between the calls.
We should discuss this.

mathvdh:
There are lots on conflicts and I'm not sure how to solve them. Not including this into the release at the moment

bramj:
@mathvdh @madewulf Good for review and merge now :+1:

iaso/tasks/process_mobile_bulk_upload.py
```diff
@@ -96,19 +98,67 @@ def process_mobile_bulk_upload(api_import_id, project_id, task=None):
...
+def get_directory_handlers(zip_ref):
+    result = {}
+    for directory in zipfile.Path(zip_ref).iterdir():
+        if directory.is_dir():
+            result[directory.name] = directory
+    return result
+
+
+def process_instance_xml(uuid, instance_data, zip_ref, user):
+    instance = Instance.objects.get(uuid=uuid)
```
kemar, line 115: It seems that `uuid` can be `None`, so this `.get()` has the potential to fail.

Should we be more defensive and catch the potential error?

kemar (approved):
LGTM

iaso/tasks/process_mobile_bulk_upload.py
bramj: Good catch! Now that I re-read it, `uuid` should not be `None` in the `instances.json` file. So I'm updating the `uuid = instance_data.get("id", None)`, it should fail if the id is None IMO.

PR #1470, opened 2024-07-10
POLIO-1610: im refresh button

quang-le:
Add refresh button to IM pages

Related JIRA tickets :POLIO-1610

- [x] Are there enough tests

## Doc
NA

## Changes

- Backend: 
    - Abstract `RefreshLQASDataViewSet` inorder to be able to reuse it for IM
    - Change the way the serializer is initialized in `ExternbalTaskViewSet` so that `get_serializer_context` is called 

- Front-end:
    - Rename `RefreshLQASData` 
    - Add IM values (endpoint, config slug etc) 
    - Add button text

## How to test

- In Django admin, copy `lqas-pipeline-config` in a new `im-pipeline-config`. The IM pipeline is not entirely set up yet, but we can test the code against the LQAS pipeline
- Go to polio > IM > any IM page
- Select a country
- Click "Refresh IM data"
- Go to tasks
- There should be an IM task visible

## Print screen / video

https://github.com/BLSQ/iaso/assets/38907762/686181a4-8ac7-406f-82f4-cd516643891e

## Notes

Do not merge on main until:
- IM pipeline has been updated to be compatible with our API
- `im-pipeline-config` config has been correctly set up on prod and staging


plugins/polio/tasks/api/refresh_lqas_im_data.py
```diff
@@ -0,0 +1,145 @@
...
+
+class RefreshLQASIMDataGetSerializer(serializers.Serializer):
+    country_id = serializers.IntegerField(required=False)
+
+    def validate(self, attrs):
+        validated_data = super().validate(attrs)
+        request = self.context["request"]
+        country_id = request.query_params.get("country_id", None)
+        # It seems a bit stange to limit the access on country id
+        # but to launch the refresh for all countries if no id is passed
```
madewulf, line 29: I agree with the comment. This lacks a test on the case where the user is limited to some org units and tries to refresh all countries

madewulf: (not a big security risk, on the other hand, refreshes are happening automatically on all countries regularly anyway, this just speeds up the process, so, it's not blocking)

plugins/polio/tasks/api/refresh_lqas_im_data.py
```diff
@@ -0,0 +1,145 @@
...
+    @action(detail=False, methods=["get"], serializer_class=TaskSerializer)
+    def last_run_for_country(self, request):
+        serializer = RefreshLQASIMDataGetSerializer(data=request.data, context={"request": request})
+        serializer.is_valid(raise_exception=True)
+        country_id = request.query_params.get("country_id", None)
+        status_query = Q(status=SUCCESS) | Q(status=RUNNING) | Q(status=ERRORED)
+        queryset = self.get_queryset().filter(status_query).exclude(started_at__isnull=True)
+        # The filter is based on how the task name is generated by ExternalTaskPostSerializer
+        query = (
+            Q(name=self.config_slug) | Q(name=f"{self.config_slug}-{country_id}")
```
madewulf, line 137: It feels like these names of tasks should be generated in a function reused in multiple places and not regenerated (with risks of differences) in multiple place. Not blocking

madewulf (approved):

plugins/polio/tasks/api/refresh_lqas_im_data.py
quang-le: We can't refresh all countries from the polio UI, it's just the pipeline code that behaves this way if I recall correctly

PR #1472, opened 2024-07-11
IA-1715: Convert LinkDetails to function component

quang-le:
One of the last 3 class components to get rid of

Related JIRA tickets : IA-1715

- [ ] Are there enough tests

## Doc
na

## Changes

- Add useGetLinkDetails hook
- change props so the whole `link`is passed from the parent
- thanks to that, make `linkDetailsSubComponent` function that can be passed to the Table (and remove TS error)
- Call `useValidateLinks` directly from `LinkDetails` i.o passing the mutation from the parent
- Delete old class component
- Delete obsolete fetch function
- Fix Cypress test

## How to test

If you don't have links:
- Go to datasources > versions > make a copy of current version
- Go to Org Units > Datasources > Algorithm runs
- Run any algorithm. Select 2 versions of your datasource as origin and destination

Then:
- Go to Links
- Click Search
- Click on the "details" IconButton (the eye):
    - All data should be displayed, including map/geometry
    - you should be able to validate/invalidate any row 

TL;DR: Nothing should have changed

## Print screen / video

https://github.com/BLSQ/iaso/assets/38907762/dff2e30d-a7a1-485a-a291-b0ef8d63f78c

## Notes


hat/assets/js/apps/Iaso/utils/requests.js
beygorghor, line 1: ❤️

beygorghor (approved):
Working nice

PR #1474, opened 2024-07-11
 IA-3215 fixes instance xml to flat json and question_by_path

mestachs:

Related JIRA tickets : IA-3215

- [ ] Are there enough tests

## Changes

- removed the problematic code
- added a test for by_path and by_name questions
- added a command to fix "past" instances 

## How to test

```
docker compose run --rm iaso manage seed_test_data --mode=seed --dhis2version=2.40.4.1
docker-compose run --rm iaso manage fix_submissions_select_multiple
```

you should see that only the `Community Verification Satisfaction` will get fixed.

## Notes

I've used the `query_set.all().iterator()`. I think it's "memory" safe (won't load 100000 instances in the same loop)


iaso/management/commands/fix_submissions_select_multiple.py
```diff
@@ -0,0 +1,36 @@
...
+#
+# testing on a small subset on your dev env
+#
+#    docker compose run --rm iaso manage seed_test_data --mode=seed --dhis2version=2.40.4.1
+#    docker-compose run --rm iaso manage fix_submissions_select_multiple
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **options):
+        date_limite = datetime(2024, 6, 1)
```
bmonjoie, line 17: The bug was merged on April 19th. Are we sure there are no instances with this bug before June 1st?

iaso/management/commands/fix_submissions_select_multiple.py
```diff
@@ -0,0 +1,36 @@
...
+
+        for form in m.Form.objects.all():
```
kemar, line 19: Is the `Form` table a big one?

Because if it's big [using `iterator` with a `chunk_size`](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#iterator) might be a better solution?

kemar (commented):
Looks almost good to me.

Can you please confirm that `date_limite` is the correct date as asked by Benjamin?

kemar (approved):

PR #1477, opened 2024-07-12
IA-278 Improve migrations

bramj:
- Split the creation of the columns into a separate migration (in db transaction, i.e. removed the `atomic=False`)
- Split the "fill" migrations into separate files for easier debugging in case something goes wrong
- Disable triggers before migrating data to avoid old inconsistent data blocking the migration
- Removed the `WHERE` clause of the updates (not needed)
- Also up the work_mem before the concurrent index creation


iaso/migrations/0288_fill_instance_source_created_at.py
```diff
@@ -0,0 +1,32 @@
...
+
+    operations = [
+        # Up the memory to improve performance
+        migrations.RunSQL(
+            sql="SET work_mem = '1GB';",
+            reverse_sql=migrations.RunSQL.noop,
+        ),
+        # Avoid the migration failing because of old incoherent data.
+        migrations.RunSQL(
+            sql="ALTER TABLE iaso_instance DISABLE TRIGGER ALL;",
```
bramj: I hope this works on RDS, I have a feeling this might fail because of not enough user permissions.

iaso/migrations/0288_fill_instance_source_created_at.py
```diff
@@ -0,0 +1,32 @@
...
+    atomic = False
+
+    dependencies = [
+        ("iaso", "0287_instance_source_created_at_and_more"),
+    ]
+
+    operations = [
+        # Up the memory to improve performance
+        migrations.RunSQL(
+            sql="SET work_mem = '1GB';",
```
bramj, line 16: @mestachs Okay for you if we keep it like this for now? Seems a bit complicated to make it smarter and maybe unnecessary.

iaso/migrations/0288_fill_instance_source_created_at.py
bramj: In fact I just tried it in a DB shell on staging and it fails. Disabling a specific one does work, I'm adapting

tdethier (approved):
We'll have a few issues on the instances where the initial 0287 and 0288 migrations were applied, but I'm afraid we don't really have a choice.
Otherwise, LGTM :+1:

iaso/migrations/0288_fill_instance_source_created_at.py
tdethier: Good idea to check that :+1:

PR #1498, opened 2024-07-18
IA-3178 : Have pictures for ALL form submissions (for better demo/Visualization)

butofleury:
Related JIRA tickets : [IA-3178](https://bluesquare.atlassian.net/browse/IA-3178)

- [ ] Are there enough tests

## Changes

Added pictures on submissions for other org unit types than health facilities!

## How to test

From setuper folder, launch `python3 setuper.py` to create data and open submissions linked to the form **Equipment/Pop/Social mob./Microplans** for other type than health facility : http://localhost:8081/dashboard/forms/submissions/accountId/1/formIds/3/tab/files

You should see differents pictures for each org unit type.

## Print screen / video

[Pictures_by_org_unit_type.webm](https://github.com/user-attachments/assets/d414d6a7-b94a-47c6-8610-3f21b72fbb67)

[IA-3178]: https://bluesquare.atlassian.net/browse/IA-3178?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


setuper/submissions.py
```diff
@@ -25,3 +25,21 @@ def org_unit_gps_point(org_unit):
 
 def submission_org_unit_gps_point(org_unit):
     return f"{org_unit['latitude']} {org_unit['longitude']} {org_unit['altitude']}"
+
+
+def picture_by_org_unit_type_name(org_unit_type_name):
+    picture_name = ""
+    match org_unit_type_name:
```
kemar: [Structural Pattern Matching](https://docs.python.org/3/whatsnew/3.10.html#pep-634-structural-pattern-matching) was introduced in Python 3.10 but we're still using Python 3.9 on iaso.

I'm afraid this could fail when launched on AWS machines. What do you think?

setuper/README.md
```diff
@@ -76,4 +76,4 @@ Once the script has run, you can log in to your server using the account name as
 
     You will need to add param `--additionnal_projects`:
 
-        python3 setuper.py --additionnal_projects
\ No newline at end of file
+        python3 setuper.py -a
```
kemar: I tend to think that `--additionnal_projects` is more explicit than `-a`.

What motivated this change?

kemar (changes requested):

setuper/submissions.py
butofleury: To avoid break something, I think it's better to keep the structural pattern if ... else. I will change it to keep it works with Python 3.9.

setuper/README.md
butofleury: I was getting an error when launching the command with `-additionnal_projects` but now it's back. I Will undo the change.

setuper/README.md
```diff
@@ -76,4 +76,4 @@ Once the script has run, you can log in to your server using the account name as
 
     You will need to add param `--additionnal_projects`:
```
kemar:
```suggestion
    You will need to add param `--additional_projects`:
```

Let's fix the typo here too :)

kemar (approved):
LGTM

PR #1500, opened 2024-07-21
IA-3256 Adding a maintenance mode

madewulf:
Adding an environment variable to toggle a maintenance mode: MAINTENANCE_MODE

See the doc for more explanations. 

image

Related JIRA tickets : IA-3256

- [x] Are there enough tests

### DIFF
--- a/hat/settings.py
+++ b/hat/settings.py
@@ -31,6 +31,8 @@ from sentry_sdk.integrations.logging import ignore_logger
 
 from plugins.wfp.wfp_pkce_generator import generate_pkce
 
+MAINTENANCE_MODE = os.environ.get("MAINTENANCE_MODE", "false").lower() == "true"
+
 # security settings
 CSRF_COOKIE_HTTPONLY = os.environ.get("CSRF_COOKIE_HTTPONLY", "false").lower() == "true"
 CSRF_COOKIE_SECURE = os.environ.get("CSRF_COOKIE_SECURE", "false").lower() == "true"
--- a/hat/templates/iaso/maintenance.html
+++ b/hat/templates/iaso/maintenance.html
@@ -1,0 +1,6 @@
+<html>
+<body>
+<h1>This server is under maintenance.</h1>
+<h2>Thanks for your patience!</h2>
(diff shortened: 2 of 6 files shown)


bramj (approved):
LGTM 👍

hat/templates/iaso/maintenance.html
```diff
@@ -0,0 +1,6 @@
+<html>
+<body>
+<h1>This server is under maintenance.</h1>
+<h2>Thanks for your patience!</h2>
+</body>
+</html>
```
kemar, line 6: This is the work of a designer! 👨‍🎨

hat/settings.py
```diff
@@ -31,6 +31,8 @@
 
 from plugins.wfp.wfp_pkce_generator import generate_pkce
 
+MAINTENANCE_MODE = os.environ.get("MAINTENANCE_MODE", "false").lower() == "true"
```
kemar, line 34: I think this should be added to the `docker-compose.yml` in `services > iaso > environment`.

hat/urls.py
```diff
@@ -20,123 +20,130 @@
...
 
 
-if settings.DISABLE_PASSWORD_LOGINS:
-    login_template = "iaso/disabled_password_login.html"
-    urlpatterns = [
-        path("admin/login/", TemplateView.as_view(template_name=login_template), name="admin-login"),
-        path("login/", TemplateView.as_view(template_name=login_template), name="login"),
+if settings.MAINTENANCE_MODE:
+    maintenance_template = "iaso/maintenance.html"
+    urlpatterns = [path("", TemplateView.as_view(template_name=maintenance_template), name="maintenance")]
```
kemar: Should we also use this template as the default one for all 404s?

madewulf: all 404 in maintenance mode, you mean? I think that's a good idea!

kemar: Yup, to let people know that there is an ongoing maintenance instead of just the default 404 template.

We should be able to override `handler404` or something like that.

PR #1507, opened 2024-07-23
IA-3253 Fix: Entity.MultipleObjectsReturned on /api/entities/{pk}/

bramj:
[IA-3253](https://bluesquare.atlassian.net/browse/IA-3253)

Fix: Entity.MultipleObjectsReturned on /api/entities/{pk}/

When a user has geo-restrictions, the `filter_for_user()` will add an
`INNER JOIN ` to the query, resulting in duplicates returned.

Adding the retrieve action explicitly and adding a `distinct` to the query fixes the issue.

Also: no need to add the full instance model in the serializer, it adds a lot of queries and weight to the call and it's not used in the
dashboard. By simply changing to `as_dict()`, the payload is only 1/5th of what it was before.

[IA-3253]: https://bluesquare.atlassian.net/browse/IA-3253?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iaso/api/entity.py
+++ b/iaso/api/entity.py
@@ -56,7 +56,7 @@ class EntitySerializer(serializers.ModelSerializer):
 
     def get_attributes(self, entity: Entity):
         if entity.attributes:
-            return entity.attributes.as_full_model()
+            return entity.attributes.as_dict()
         return None
 
     def get_org_unit(self, entity: Entity):
@@ -213,6 +213,11 @@ class EntityViewSet(ModelViewSet):
         entities = Entity.objects.filter(account=request.user.iaso_profile.account)
         serializer = EntitySerializer(entities, many=True)
         return Response(serializer.data)
+
+    def retrieve(self, request, pk=None):
+        queryset = Entity.objects.filter_for_user(self.request.user).distinct()
+        entity = get_object_or_404(queryset, pk=pk)
+        return Response(EntitySerializer(entity, many=False).data)
 
     def list(self, request: Request, *args, **kwargs):
         queryset = self.filter_queryset(self.get_queryset())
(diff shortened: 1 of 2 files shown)


iaso/api/entity.py
```diff
@@ -56,7 +56,7 @@ class Meta:
 
     def get_attributes(self, entity: Entity):
         if entity.attributes:
-            return entity.attributes.as_full_model()
+            return entity.attributes.as_dict()
```
madewulf, line 59: You're sure that this does not break the entity detail screen?

iaso/api/entity.py
```diff
@@ -214,6 +214,11 @@ def bulk_create(self, request, *args, **kwargs):
         serializer = EntitySerializer(entities, many=True)
         return Response(serializer.data)
 
+    def retrieve(self, request, pk=None):
+        queryset = Entity.objects.filter_for_user(self.request.user).distinct()
```
madewulf, line 218: I'm just wondering if the distinct() should not move to inside filter_for_user. Is there a reason not to do so?

kemar: Also there is a trap with `distinct()` ([quoted from Django doc](https://docs.djangoproject.com/en/5.0/ref/models/querysets/#django.db.models.query.QuerySet.distinct)):

> For a normal `distinct()` call, the database compares each field in each row when determining which rows are distinct.
> For a `distinct()` call with specified field names, the database will only compare the specified field names.

So I would use something like `distinct("id")` to avoid a potential performance issue.

bramj: I'm adding the `"id"` to avoid potential issues. I've tried to add it directly inside the `filter_for_user` method, but that gives me a ` NotImplementedError: annotate() + distinct(fields) is not implemented.` on the list page.

Seems like it doesn't play well with `annotate`, so I'll just leave it on the `retrieve` action.

iaso/api/entity.py
bramj: I've tested and reviewed the code and I can't find any reason to use the `as_full_model()` here

kemar (approved):
LGTM

PR #1508, opened 2024-07-24
IA-3251 Fix: Remove DataSource version description...

bramj:
from label on org unit treeview.

The description is for internal use, only display the version.

See screenshots https://bluesquare.atlassian.net/browse/IA-3251


kemar (approved):
LGTM

PR #1514, opened 2024-07-27
Hotfix csv exports [IA-3267]

madewulf:
Speeding up CSV exports by removing a ton of SQL requests

Related JIRA tickets : IA-3267

- [x] Are there enough tests

## Doc

## Changes

Ensuring that only something like 20 requests are made per 5000 lines of csv exports (but the csv exports are batched by 5k to allow streaming and not overload the database)

## How to test

CleanShot 2024-07-28 at 06 51 48@2x

## Notes

Sorry, a test is not passing, but it's only on github actions for now

### DIFF
--- a/iaso/api/instances.py
+++ b/iaso/api/instances.py
@@ -42,6 +42,7 @@ from . import common
 from .comment import UserSerializerForComment
 from .common import CONTENT_TYPE_CSV, CONTENT_TYPE_XLSX, FileFormatEnum, TimestampField, safe_api_import
 from .instance_filters import get_form_from_instance_filters, parse_instance_filters
+from ..utils.models.common import get_creator_name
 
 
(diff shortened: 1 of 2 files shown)


bramj:
I've made a few changes and fixed the failing test so needs a new review:

- Moved code outside the CSV block so it also applies to the XLSX export
- I've removed the 5th level of org unit parent prefetch, since if I understand correct, we only request 4 levels.
- Brought back a prefetch on the regular non-file endpoint since it's useful for performance there

bramj:
As for Martin's changes: really nice, it greatly improves performance on the file exports! :clap:

kemar (approved):
LGTM

tdethier (approved):
The change is already noticeable with only 50 instances. I can't imagine with millions.

LGTM :+1:

PR #1522, opened 2024-07-30
IA-3258 add payment status filter

quang-le:

Related JIRA tickets : IA-3258

- [x] Are there enough tests

## Doc
JsDoc comments in the code

## Changes
- Backend
    - Add filter in change requests Filterset
    - Move `parse_comma_separated_numeric_values` to common.py and remove duplicate versions
    - Move `PaymentStatuses` outside of `Payment` class for reuse
    - Add paths to LOCALE_PATHS for backend translations
    - Add `locale` folder with translations in `iaso` app folder
    - Add `DropDownOptionsViewSet` and `DropDownOptionsSerializer` for future endpointst that return `TextChoices`
    - Add `PaymentOptionsViewSet`, which extends `DropDownOptionsViewSet`

- Frontend
    - Add translations
    - Add typing for OPTIONS response
    - Add param to change requests route
    - Add `useGetPaymentStatusOptions` (but with GET request) and replace duplicate
    - Add `mapOptions` method in utils to convert OPTIONS response to dropdown list format (+ unit test)
    - Cleanup some old comments
  
## How to test

make sure to run `django-admin compilemessages` when not launching the backend from Docker

- Go change requests page
- Test payments filter
- Change language and check that dropdown is correctly translated

Extra mile:
- Go to polio budget and try to filter by country using multiple values (= testing that `parse_comma_separated_numeric_values` still works as intended)
- Go to Payments, open dialog and try to edit payment status (= test that `useGetPaymentStatusOptions` works as intended)

## Print screen / video

https://github.com/user-attachments/assets/d1dee3b3-e98e-4f8a-8a75-17bcff311368

## Notes

Using the OPTIONS request was causing permission issues, so I made a dedicated endpoint with the `TextChoices`.
I left the `mapOptions` helper in case we figure out how to fix the permissions problem. If not it can be removed.

- Cypress tests are passing: https://github.com/BLSQ/iaso/actions/runs/10214039194


iaso/api/payments/views.py
```diff
@@ -570,14 +570,18 @@ class PaymentsViewSet(ModelViewSet):
 
     """
 
-    http_method_names = ["patch", "get"]
+    http_method_names = ["patch", "get", "options", "post"]
```
quang-le: Authorising POST is necessary for the OPTIONS call to return the choices

iaso/api/payments/views.py
```diff
@@ -570,14 +570,18 @@ class PaymentsViewSet(ModelViewSet):
...
     serializer_class = PaymentSerializer
     permission_classes = [permissions.IsAuthenticated, HasPermission(permission.PAYMENTS)]
 
     def get_queryset(self) -> models.QuerySet:
         return Payment.objects.filter(created_by__iaso_profile__account=self.request.user.iaso_profile.account)
 
+    # Prevent direct POSTing. POST method needs to be authorized to get choices in OPTIONS request
+    def create(self, request):
+        pass
+
```
quang-le: We shouldn't create payments by directly POSTing, so I overrode `create`

iaso/api/payments/views.py
kemar: Yup, I wasn't aware of that, but [this is DRF default behavior](https://github.com/encode/django-rest-framework/blob/8e304e1adbb0f99f91a15ed3abd379104bba3b89/rest_framework/metadata.py#L78)… and I don't know why.

So we have two solutions:

1. do what you've done
1. create a subclass of `SimpleMetadata` and override `determine_actions()`

The only problem with your solution is that it could trick the auto-doc tools like swagger etc.

quang-le: I think if we plan to use that pattern widely in the future, the second solution looks more robust

iaso/tests/api/payments/test_payments.py
```diff
@@ -74,8 +76,34 @@ def test_update_status(self):
...
+        # r = self.assertJSONResponse(response, 200)
+        # for item in r:
+        #     if item['value'] == PaymentStatuses.PENDING:
+        #         self.assertEqual(item['label'], 'En attente')
+        #     elif item['value'] == PaymentStatuses.SENT:
+        #         self.assertEqual(item['label'], 'Envoyé')
+        #     elif item['value'] == PaymentStatuses.REJECTED:
+        #         self.assertEqual(item['label'], 'Rejeté')
+        #     elif item['value'] == PaymentStatuses.PAID:
+        #         self.assertEqual(item['label'], 'Payé')
```
kemar, line 109: This works for me locally.

My guess is that you just need to [reset the active language at the end of the test](https://docs.djangoproject.com/fr/4.2/topics/testing/tools/#setting-the-language) so that other tests don't fail?

One way to do it is:

```
from django.conf import settings
from django.utils import translation


# …

translation.activate(settings.LANGUAGE_CODE)
```

kemar (approved):
LGTM

iaso/tests/api/payments/test_payments.py
kemar: I ran into the same issue: tests working locally but failing in the CI.

It turns out `compilemessages` was not run in CI. This will be fixed in https://github.com/BLSQ/iaso/pull/1538

PR #1524, opened 2024-07-31
IA-3264 Order by created/updated

bramj:
Related PR: https://github.com/BLSQ/bluesquare-components/pull/162

The introduction of the `source_created_at` and `source_updated_at` fields resulted in some issues:
- Filters on creation timestamps
- Columns in the frontend that sort on incorrect db columns

I've decided on the following:
- On the instance list page:
  - timestamps shown are source_* with fallback to db timestamps
  - for this to work correctly, the order by on timestamps now uses `COALESCE` for the fallback
- On the instance details page:
  - Show both the device creation + db creation timestamps + the `updated_at`, which we can call the "last synced". (I decided against showing the `source_updated_at` since that's the least useful one and it will cause more confusion than value)

Related JIRA tickets : https://bluesquare.atlassian.net/browse/IA-3264

Some other fixes and improvements:
- Added the db creation timestamp to the instance details, and make the labels clearer:

- Improved some logic: the frontend should only send the date, e.g.
"2024-07-10", without time. It's the backend that will add the "start of
day" or "end of day" to make sure we fetch all records. The previous
code would not find instances created in the last minute before
midnight.
   **Before:**
   

   **After:**
   

- Refactored to be able to reuse existing code, changed some range
queries for simpler >= and <= date comparisons.
- Remove unused `instance.as_dict_with_parents()`


docker-compose.yml
quang-le, line 1: Why is that part of the PR?

iaso/models/base.py
```diff
@@ -1178,29 +1169,6 @@ def as_dict_with_descriptor(self):
...
-            "org_unit": self.org_unit.as_dict_with_parents() if self.org_unit else None,
-            "latitude": self.location.y if self.location else None,
-            "longitude": self.location.x if self.location else None,
-            "altitude": self.location.z if self.location else None,
-            "accuracy": self.accuracy,
-            "period": self.period,
-            "status": getattr(self, "status", None),
-            "correlation_id": self.correlation_id,
-        }
-
```
quang-le, line 1203: was this not used?

quang-le (commented):
LGTM but I don't know if the changes to docker-compose are part of the PR or should be merged.

I made a commit to remove a (now) unnecessary null check in `DateTimeCell`(related to your PR on blsq-comp)

quang-le (changes requested):
There may be a commit that needs pushing, because `source_created_at` is not in the API response for instances (both list and details)

Screenshot 2024-08-22 at 11 59 19

docker-compose.yml
bramj: :grimacing: sorry, that's a bad oversight

quang-le (approved):

PR #1526, opened 2024-07-31
POLIO-1638 Chronogram: fix the country filter

kemar:
Chronogram: fix the country filter.

Related JIRA tickets : [POLIO-1638](https://bluesquare.atlassian.net/browse/POLIO-1638)

## Video

https://github.com/user-attachments/assets/9be9ded3-b96f-4a05-9749-b1663a0e65d6

[POLIO-1638]: https://bluesquare.atlassian.net/browse/POLIO-1638?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/plugins/polio/js/src/domains/Chronogram/Chronogram/api/useGetChronogram.ts
+++ b/plugins/polio/js/src/domains/Chronogram/Chronogram/api/useGetChronogram.ts
@@ -9,7 +9,11 @@ import { apiBaseUrl } from '../../constants';
 const getChronogram = async (
     params: Partial<ChronogramParams>,
 ): Promise<ChronogramApiResponse> => {
-    const queryString = new URLSearchParams(params).toString();
+    const { country, ...paramsWithoutCountry } = params;
+    let queryString = new URLSearchParams(paramsWithoutCountry).toString();
+    if (country) {
+        queryString += `&country=${country.split(',').join('&country=')}`;
+    }
     return getRequest(`${apiBaseUrl}?${queryString}&fields=:all`);
 };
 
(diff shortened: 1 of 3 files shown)


plugins/polio/js/src/domains/Chronogram/Chronogram/api/useGetChronogram.ts
```diff
@@ -9,7 +9,11 @@ import { apiBaseUrl } from '../../constants';
 const getChronogram = async (
     params: Partial<ChronogramParams>,
 ): Promise<ChronogramApiResponse> => {
-    const queryString = new URLSearchParams(params).toString();
+    const { country, ...paramsWithoutCountry } = params;
+    let queryString = new URLSearchParams(paramsWithoutCountry).toString();
+    if (country) {
+        queryString += `&country=${country.split(',').join('&country=')}`;
```
tdethier, line 15: I see that you used the other format for multiple parameters, it's interesting. 

What happens in the backend if you need to process these parameters/convert them to something else? Do you do that somewhere in the `ChronogramFilter` class?

tdethier (approved):
LGTM :+1:

plugins/polio/js/src/domains/Chronogram/Chronogram/api/useGetChronogram.ts
kemar: > I see that you used the other format for multiple parameters, it's interesting.

Yes, but this is a bit hackish 🤫

> What happens in the backend if you need to process these parameters/convert them to something else? Do you do that somewhere in the `ChronogramFilter` class?

I guess you can. But usually these parameters are mapped to a Django choice and you don't have anything to do.

PR #1527, opened 2024-07-31
[IA-3275] Fix csv instance export when no data

tdethier:
Fixed an error 500 when there is no form data during an instance csv export.

Related JIRA tickets : IA-3275

- [x] Are there enough tests

## Doc
/

## Changes
- Update frontend to block the csv/xlsx buttons to be active if there is no form in the current tab
- Update backend to return a `400` error if there is no `form` nor `form_ids` parameters in the query

## How to test
- Go on the "registry" page
- Select a level which doesn't have any form, and make sure you can't click on the csv/xlsx buttons
- Select a level which has at least one form but no submissions, and make sure you can't click on the csv/xlsx buttons
- (optional) Query the instances API directly and make sure you get a `400` error. For instance: [http://localhost:8081/api/instances/?orgUnitTypeId=404&limit=20&order=org_unit__name&page=1&showDeleted=false&orgUnitParentId=2187066&org_unit_status=VALID&csv=true](http://localhost:8081/api/instances/?orgUnitTypeId=404&limit=20&order=org_unit__name&page=1&showDeleted=false&orgUnitParentId=2187066&org_unit_status=VALID&csv=true)

## Print screen / video

image

## Notes

There were a few comments in the backend code, mentioning some weird logic and the fact that the frontend will always send a `form` or `form_ids` parameter. There were also some mypy annotations to ignore cases where `Form` is `None`. 

Since we somehow ended up in a situation where the frontend didn't this parameter and now there's a check to make sure that there is actually a `form` value, I got rid of these comments. But we might still need to check the logic behind all of this again, because there might still be a few edge cases that will cause issues.

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/registry/components/Instances.tsx
+++ b/hat/assets/js/apps/Iaso/domains/registry/components/Instances.tsx
@@ -191,7 +191,9 @@ export const Instances: FunctionComponent<Props> = ({
                                             xlsxUrl={`${apiUrl}&xlsx=true`}
                                             disabled={
                                                 isFetchingList ||
-                                                data?.count === 0
+                                                data?.count === 0 ||
+                                                formIds === undefined ||
+                                                formIds === null
                                             }
                                         />
                                     </Box>
--- a/hat/assets/js/apps/Iaso/domains/registry/components/Instances.tsx
+++ b/hat/assets/js/apps/Iaso/domains/registry/components/Instances.tsx
@@ -192,8 +192,7 @@ export const Instances: FunctionComponent<Props> = ({
                                             disabled={
                                                 isFetchingList ||
                                                 data?.count === 0 ||
-                                                formIds === undefined ||
-                                                formIds === null
+                                                !formIds
                                             }
                                         />
                                     </Box>
--- a/iaso/tests/api/test_instances.py
+++ b/iaso/tests/api/test_instances.py
@@ -1024,6 +1024,15 @@ class InstancesAPITestCase(APITestCase):
         # Make sure the export is using the default created/updated_at if there is no source
         self.assertEqual(row_to_test, expected_row)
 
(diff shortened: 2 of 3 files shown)


hat/assets/js/apps/Iaso/domains/registry/components/Instances.tsx
```diff
@@ -191,7 +191,9 @@ export const Instances: FunctionComponent<Props> = ({
                                             xlsxUrl={`${apiUrl}&xlsx=true`}
                                             disabled={
                                                 isFetchingList ||
-                                                data?.count === 0
+                                                data?.count === 0 ||
+                                                formIds === undefined ||
```
hakifran: Not very important but I think this `!formIds` can represent the formIds === undefined || formIds === null

iaso/tests/api/test_instances.py
```diff
@@ -1024,6 +1024,15 @@ def test_can_retrieve_submissions_list_in_csv_format_without_source_fields(self)
...
         self.assertEqual(row_to_test, expected_row)
 
+    def test_submissions_list_in_csv_format_error_no_form(self):
+        # Make sure IA-3275 is fixed by sending a 400 instead of letting the backend crash
+        self.client.force_authenticate(self.yoda)
+        response = self.client.get(
+            f"/api/instances/?limit=20&order=org_unit__name&page=1&showDeleted=false&org_unit_status=VALID&csv=true",
+            headers={"Content-Type": "text/csv"},
+        )
+        self.assertJSONResponse(response, 400)
```
hakifran:
```suggestion
        self.assertJSONResponse(response, status.HTTP_400_BAD_REQUEST)
```

hakifran (changes requested):

tdethier:
Thanks @hakifran for the comments!

I made some changes.

hakifran (approved):
LGTM

PR #1533, opened 2024-08-05
IA-3278 Return full metadata on OPTIONS requests

kemar:
Return full metadata for `OPTIONS` requests.

Related JIRA tickets : [IA-3278](https://bluesquare.atlassian.net/browse/IA-3278)

## Changes

I've started to parse the result of `OPTIONS` requests to build front-end UIs elements (e.g. select dropdown options).

This is possible [thanks to DRF exposing some metadata](https://www.django-rest-framework.org/api-guide/metadata/).

But it turns out that it renders options [only for `POST` and `PUT` methods by default](https://github.com/encode/django-rest-framework/blob/8e304e1adbb0f99f91a15ed3abd379104bba3b89/rest_framework/metadata.py#L78).

This is a problem to build filters for API endpoint where `POST` and `PUT` methods are not allowed.

So I suggest to implement a custom metadata class that:

- returns the metadata for `OPTIONS` requests too
- returns the available `choices` even for `read-only` fields (I came across this case when implementing a restricted access to a subset of fields for Chronograms)

[IA-3278]: https://bluesquare.atlassian.net/browse/IA-3278?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bramj (approved):
Nice and creative :+1:  LGTM!

PR #1536, opened 2024-08-06
POLIO-1621 Chronogram template task: make column "Activity" wider by default

kemar:
Chronogram template task: make column "Activity" wider by default

Related JIRA tickets : [POLIO-1621](https://bluesquare.atlassian.net/browse/POLIO-1621)

## How to test

- you must be super admin or have the `iaso_polio_chronogram` permission
- go to `localhost:8081/dashboard/polio/chronogram/templateTask`
- the "Activity" column should be wider by default

## Print screen

wider

[POLIO-1621]: https://bluesquare.atlassian.net/browse/POLIO-1621?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/plugins/polio/js/src/domains/Chronogram/ChronogramTemplateTask/Table/useChronogramTemplateTaskTableColumns.tsx
+++ b/plugins/polio/js/src/domains/Chronogram/ChronogramTemplateTask/Table/useChronogramTemplateTaskTableColumns.tsx
@@ -29,6 +29,7 @@ export const useChronogramTemplateTaskTableColumns = (
                 id: 'description',
                 accessor: 'description',
                 sortable: false,
+                width: 800,
             },
             {
                 Header: formatMessage(MESSAGES.labelStartOffsetInDays),


plugins/polio/js/src/domains/Chronogram/ChronogramTemplateTask/Table/useChronogramTemplateTaskTableColumns.tsx
```diff
@@ -29,6 +29,7 @@ export const useChronogramTemplateTaskTableColumns = (
                 id: 'description',
                 accessor: 'description',
                 sortable: false,
+                width: 800,
```
bmonjoie, line 32: Shouldn't we use an adaptive size here rather than a fixed size? Using a fixed size might behave incorrectly on smaller screen

kemar: I double checked this morning and I can confirm that the layout is still adaptative, even though we're using a fixed size… 

I don't really know why but I guess it could be related to either CSS table layout or some responsive attributes.

In Safari:

width_safari

In Chrome:

width_chrome

bmonjoie (approved):

PR #1540, opened 2024-08-08
IA-3280: Completeness stats: period filter is broken

hakifran:
- Completeness stats: period filter is broken
- This PR is about to display a message which indicates that the selected form has no period type
Related JIRA tickets : [IA-3280](https://bluesquare.atlassian.net/browse/IA-3280)

- [ ] Are there enough tests

## Doc

## Changes

- Display in the periodPicker a different message when the selected form has no period type vs when there is no form selected

## How to test
- Create or use existing form which has no period as period type
- Go to completeness stats and select a form with no period type and select another one with a period type
- Check at each test the period picker behavior

## Print screen / video
[Screencast from 2024-08-08 14-52-40.webm](https://github.com/user-attachments/assets/6635eb86-cfbc-4abd-94af-83856205d843)

## Notes

[IA-3280]: https://bluesquare.atlassian.net/browse/IA-3280?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hakifran:
@cypress

hat/assets/js/apps/Iaso/domains/completenessStats/CompletenessStatsFilters.tsx
```diff
@@ -156,10 +154,22 @@ export const CompletenessStatsFilters: FunctionComponent<Props> = ({
...
+        const messages = {
+            [PERIOD_TYPE_PLACEHOLDER]: MESSAGES.periodPlaceHolder,
+            [NO_PERIOD]: MESSAGES.noPeriodPlaceHolder,
+        };
+
+        return messages[periodType]
+            ? formatMessage(messages[periodType])
+            : undefined;
+    }, [formatMessage, periodType]);
+
```
kemar, line 163: I don't understand why you're using `useMemo` in such a simple computation?

What's the benefit?

For what I know:

- the re-rendering mechanism of React is often fast enough by default
- `useMemo` is best used in case of cost expensive computations which could lead to a re-rendering delay (500ms and more)

kemar (approved):
LGTM

hat/assets/js/apps/Iaso/domains/completenessStats/CompletenessStatsFilters.tsx
hakifran: You are right, there is no need to use useMemo here.

hakifran:
@cypress

PR #1541, opened 2024-08-08
SLEEP-1459 Add retries with exponential backoff to XML file downloads

bramj:
Implemented because of issues with the bulk upload.

The bulk upload downloads a lot of files rapidly from S3 (the XML files) This can result in error:
`HTTP Error 503: Slow Down`

To handle this we add a 3 retries by default. With a 1 sec, 2 sec and 4 sec delay.
For the bulk uploads we can up this to 8 retries (the last one will have a 128 sec delay).

**Note: Already deployed and successfully executed on Trypelim production**

### DIFF
--- a/iaso/models/base.py
+++ b/iaso/models/base.py
@@ -2,11 +2,13 @@ import datetime
 import operator
 import random
 import re
+import time
 import typing
 from copy import copy
 from functools import reduce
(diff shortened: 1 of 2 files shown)


kemar (approved):
LGTM

PR #1542, opened 2024-08-08
SLEEP-1458 Add instance-like filters on entities

bramj:
I'm really excited about this one!

This PR adds JSON logic filters to the entities page to allow searches such as 'give me all entities that have a form instance with "A" == "100"'.

It uses the `react-awesome-query-builder` in the frontend, formatting the data as "subfields", and adds support in the backend to turn this into SQL `EXISTS` subqueries.

PR on `bluesquare-components`: https://github.com/BLSQ/bluesquare-components/pull/164

Related JIRA tickets : https://bluesquare.atlassian.net/browse/SLEEP-1458

**Note: I'm fairly confident that the generated queries are correct. My main concern is performance when the instances/entities tables grow. I plan to use Trypelim as the guinea pig for this. With about 100K entities and 200K instances it should give us an idea of how this will scale.**

- ~My migrations file are include~
- [x] Are there enough tests
- ~Documentation has been included (for new feature)~

## Print screen / video

https://github.com/user-attachments/assets/311c8ad4-db7c-4767-a15d-6666ad4d332f

## Notes

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/entities/entityTypes/hooks/requests/forms.ts
+++ b/hat/assets/js/apps/Iaso/domains/entities/entityTypes/hooks/requests/forms.ts
@@ -35,11 +35,15 @@ export const useGetForm = (
 
 export const useGetForms = (
     enabled: boolean,
+    fields?: string[] | undefined,
 ): UseQueryResult<Form[], Error> => {
+    let url = '/api/forms/?fields=id,name,latest_form_version';
+    if (fields) {
+        url += `,${fields.join(',')}`;
+    }
     return useSnackQuery({
         queryKey: ['forms'],
-        queryFn: () =>
-            getRequest('/api/forms/?fields=id,name,latest_form_version'),
+        queryFn: () => getRequest(url),
         options: {
             staleTime: 60000,
             enabled,
--- a/hat/assets/js/apps/Iaso/domains/entities/entityTypes/hooks/requests/forms.ts
+++ b/hat/assets/js/apps/Iaso/domains/entities/entityTypes/hooks/requests/forms.ts
@@ -6,7 +6,6 @@ import { getRequest } from '../../../../../libs/Api';
 import { Form, PossibleField } from '../../../../forms/types/forms';
 import { usePossibleFields } from '../../../../forms/hooks/useGetPossibleFields';
 
-
 export const useGetForm = (
     formId: number | undefined,
     enabled: boolean,
@@ -37,10 +36,9 @@ export const useGetForms = (
     enabled: boolean,
     fields?: string[] | undefined,
 ): UseQueryResult<Form[], Error> => {
-    let url = '/api/forms/?fields=id,name,latest_form_version';
-    if (fields) {
-        url += `,${fields.join(',')}`;
-    }
+    const apiUrl = '/api/forms/?fields=id,name,latest_form_version';
+    const url = fields ? `${apiUrl},${fields.join(',')}` : apiUrl;
+
     return useSnackQuery({
         queryKey: ['forms'],
         queryFn: () => getRequest(url),
--- a/hat/assets/js/apps/Iaso/domains/forms/fields/hooks/useGetFormDescriptor.ts
+++ b/hat/assets/js/apps/Iaso/domains/forms/fields/hooks/useGetFormDescriptor.ts
@@ -12,7 +12,7 @@ type FormVersionsList = {
 };
 
 const getVersion = (formId: number | undefined): Promise<FormVersionsList> => {
-    return getRequest(`/api/formversions/?form_id=${formId}&fields=descriptor`);
+    return getRequest(`/api/formversions/?form_id=&fields=descriptor`);
 };
 export const useGetFormDescriptor = (
     formId?: number,
@@ -25,7 +25,7 @@ export const useGetFormDescriptor = (
         queryKey,
         queryFn: () => getVersion(formId),
         options: {
-            enabled: Boolean(formId),
+            // enabled: Boolean(formId), TODO
             select: (
                 data: FormVersionsList | undefined,
             ): FormDescriptor[] | undefined => {
--- a/hat/assets/js/apps/Iaso/domains/forms/fields/hooks/useGetFormDescriptor.ts
+++ b/hat/assets/js/apps/Iaso/domains/forms/fields/hooks/useGetFormDescriptor.ts
@@ -12,8 +12,16 @@ type FormVersionsList = {
 };
 
 const getVersion = (formId: number | undefined): Promise<FormVersionsList> => {
-    return getRequest(`/api/formversions/?form_id=&fields=descriptor`);
+    return getRequest(`/api/formversions/?form_id=${formId}&fields=descriptor`);
 };
+
+const processResult = (
+    data: FormVersionsList | undefined,
+): FormDescriptor[] | undefined => {
+    if (!data) return data;
+    return data.form_versions?.map(version => version.descriptor);
+};
+
 export const useGetFormDescriptor = (
     formId?: number,
 ): UseQueryResult<FormDescriptor[] | undefined, Error> => {
@@ -25,13 +33,21 @@ export const useGetFormDescriptor = (
         queryKey,
         queryFn: () => getVersion(formId),
         options: {
-            // enabled: Boolean(formId), TODO
-            select: (
-                data: FormVersionsList | undefined,
-            ): FormDescriptor[] | undefined => {
-                if (!data) return data;
-                return data.form_versions?.map(version => version.descriptor);
-            },
+            enabled: Boolean(formId),
+            select: data => processResult(data),
+        },
+    });
+};
+
+export const useGetAllFormDescriptors = (): UseQueryResult<
+    FormDescriptor[] | undefined,
+    Error
+> => {
+    return useSnackQuery({
+        queryKey: ['instanceDescriptors'],
+        queryFn: () => getRequest('/api/formversions/?fields=descriptor'),
+        options: {
+            select: data => processResult(data),
         },
     });
 };
--- a/hat/assets/js/apps/Iaso/domains/forms/fields/hooks/useGetQueryBuildersFields.ts
+++ b/hat/assets/js/apps/Iaso/domains/forms/fields/hooks/useGetQueryBuildersFields.ts
@@ -3,6 +3,7 @@ import { QueryBuilderFields } from 'bluesquare-components';
 import { formatLabel } from '../../../instances/utils';
 
 import { FormDescriptor, PossibleField } from '../../types/forms';
+import { PossibleFieldsForForm } from '../../hooks/useGetPossibleFields';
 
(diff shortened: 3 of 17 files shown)


hat/assets/js/apps/Iaso/domains/entities/entityTypes/hooks/requests/forms.ts
```diff
@@ -28,11 +28,15 @@ export const useGetForm = (
 
 export const useGetForms = (
     enabled: boolean,
+    fields?: string[] | undefined,
 ): UseQueryResult<Form[], Error> => {
+    let url = '/api/forms/?fields=id,name,latest_form_version';
+    if (fields) {
+        url += `,${fields.join(',')}`;
+    }
```
quang-le: nitpick but I'm not a big fan of this pattern
```suggestion
    const apiUrl = '/api/forms/?fields=id,name,latest_form_version'; // I would even hoist that outside of the hook
    const url = fields ? `${apiUrl},${fields.join(',')}` : apiUrl;
```

hat/assets/js/apps/Iaso/domains/entities/entityTypes/hooks/requests/forms.ts
```diff
@@ -28,11 +28,15 @@ export const useGetForm = (
...
     return useSnackQuery({
         queryKey: ['forms'],
-        queryFn: () =>
-            getRequest('/api/forms/?fields=id,name,latest_form_version'),
+        queryFn: () => getRequest(url),
```
quang-le, line 44: If you merge the previous suggestion:

```suggestion
        queryFn: () => getRequest(apiUrl),
```

hat/assets/js/apps/Iaso/domains/forms/fields/hooks/useGetFormDescriptor.ts
```diff
@@ -12,7 +12,7 @@ type FormVersionsList = {
 };
 
 const getVersion = (formId: number | undefined): Promise<FormVersionsList> => {
-    return getRequest(`/api/formversions/?form_id=${formId}&fields=descriptor`);
+    return getRequest(`/api/formversions/?form_id=&fields=descriptor`);
```
quang-le: If you need to skip the form id for the feature to work, I would split the code or add a parameter to `useGetFormDescriptor` as it needlessly increases the payload of the request for the submission/instances page (where the form id is known).

hat/assets/js/apps/Iaso/domains/forms/fields/hooks/useGetFormDescriptor.ts
```diff
@@ -25,7 +25,7 @@ export const useGetFormDescriptor = (
         queryKey,
         queryFn: () => getVersion(formId),
         options: {
-            enabled: Boolean(formId),
+            // enabled: Boolean(formId), TODO
```
quang-le: Same as above: removing this duplicates the API call on the submissions page

hat/assets/js/apps/Iaso/domains/forms/hooks/useGetPossibleFields.ts
```diff
@@ -41,6 +53,40 @@ export const useGetPossibleFields = (formId?: number): Result => {
...
+                name: form.name,
+                possibleFields,
+            });
+        });
+
+        return {
+            allPossibleFields,
+            isFetchingForms,
+        };
+    }, [isFetchingForms]);
```
quang-le:
```suggestion
    }, [isFetchingForms, allForms]);
```
You need to update the return value if `allForms` change. Normally your linter should warn you if your deps array is incomplete.

hat/assets/js/apps/Iaso/domains/forms/hooks/useGetPossibleFields.ts
```diff
@@ -41,6 +53,40 @@ export const useGetPossibleFields = (formId?: number): Result => {
     return usePossibleFields(isFetchingForm, currentForm);
 };
 
+export const useAllPossibleFields = (
+    isFetchingForms: boolean,
+    allForms: Form[] = [],
+): AllResults => {
+    return useMemo(() => {
+        let allPossibleFields: PossibleFieldsForForm[] = [];
```
quang-le:
```suggestion
        const allPossibleFields: PossibleFieldsForForm[] = [];
```
Pushing values into the array doesn't change its reference, so its safe (and preferred) to declare it with `const`

hat/assets/js/apps/Iaso/domains/forms/hooks/useGetPossibleFields.ts
quang-le: If done on purpose, please leave a comment explaining the reason why, and why it's safe to do so

quang-le (changes requested):
Some code style changes + 2 blocking:
- Please adapt `useGetFormDescriptor` to avoid double fetch + fetching all forms on instances page
- Dependencies array of `useMemo` should not be incomplete

iaso/utils/jsonlogic.py
```diff
@@ -36,28 +36,34 @@ def as_sql(self, compiler, connection):
...
     """
 
+    func = jsonlogic_to_q if recursion_func is None else recursion_func
+
     if "and" in jsonlogic:
         return reduce(
             operator.and_,
-            (jsonlogic_to_q(subquery, field_prefix) for subquery in jsonlogic["and"]),
+            (func(subquery, field_prefix) for subquery in jsonlogic["and"]),
         )
```
kemar: I know that was implemented before and this is largely a personal preference but… every time I see a `reduce()` call, I need to go read the doc to understand what's actually happening.

I tend to think that an accumulation loop is easier to understand, e.g.

```python
sub_query = Q()
for lookup in jsonlogic["and"]:
    sub_query = operator.and_(sub_query, func(lookup, field_prefix))
return sub_query
```

iaso/tests/api/test_entities.py
```diff
@@ -421,12 +525,12 @@ def test_retrieve_entity_only_same_account(self):
...
-            reference_form=self.form_1,
-            fields_detail_info_view=["something", "else"],
-            fields_list_view=["something", "else"],
-        )
+        # entity_type = EntityType.objects.create(
+        #     name="Type 1",
+        #     reference_form=self.form_1,
+        #     fields_detail_info_view=["something", "else"],
+        #     fields_list_view=["something", "else"],
+        # )
```
kemar, line 533: Maybe this could be deleted then?

kemar (commented):
Backend LGTM

Kudos because this is advanced work with the ORM.

I'm also concerned with performance and curious to hear about how Trypelim's going to handle it.

hat/assets/js/apps/Iaso/domains/forms/fields/hooks/useGetFormDescriptor.ts
bramj: Yes absolutely, it needs to be a separate hook, totally forgot this TODO

hat/assets/js/apps/Iaso/domains/entities/entityTypes/hooks/requests/forms.ts
bramj: Not sure I find it more readable, but changed :+1:

hat/assets/js/apps/Iaso/domains/forms/hooks/useGetPossibleFields.ts
bramj: Hm good catch, my linter didn't tell me anything about this, should it come from the `show-lint-problems` task?

hat/assets/js/apps/Iaso/domains/entities/entityTypes/hooks/requests/forms.ts
bramj: I've made the changes you proposed :+1: there it's the `url` we want though

bramj:
> Backend LGTM
> 
> Kudos because this is advanced work with the ORM.
> 
> I'm also concerned with performance and curious to hear about how Trypelim's going to handle it.

So far so good! But it's only on about 100K beneficiaries for now. The good part about the QueryBuilder is that the frontend is already prepared for if/when we want to use something like Elastic search to return the results from the backend.

hat/assets/js/apps/Iaso/domains/forms/fields/hooks/useGetQueryBuildersFields.ts
```diff
@@ -54,3 +55,30 @@ export const useGetQueryBuildersFields = (
...
+        fields[form_id] = {
+            label: name,
+            type: '!group',
+            mode: 'array',
+            conjunctions: ['AND', 'OR'],
+            operators: ['some', 'all', 'none'],
+            defaultOperator: 'some',
+            subfields: subfields,
+        };
+    }
```
beygorghor, line 81:
```suggestion
export const useGetQueryBuilderFieldsForAllForms = (
    formDescriptors?: FormDescriptor[],
    allPossibleFields?: PossibleFieldsForForm[],
): QueryBuilderFields => {
    if (!allPossibleFields || !formDescriptors) return {};
    
    return allPossibleFields.reduce((fields, { form_id, name, possibleFields }) => {
        fields[form_id] = {
            label: name,
            type: '!group',
            mode: 'array',
            conjunctions: ['AND', 'OR'],
            operators: ['some', 'all', 'none'],
            defaultOperator: 'some',
            subfields: useGetQueryBuildersFields(formDescriptors, possibleFields),
        };
        return fields;
    }, {} as QueryBuilderFields);
};
```

beygorghor: not a big fan of iterators

hat/assets/js/apps/Iaso/domains/forms/hooks/useGetPossibleFields.ts
```diff
@@ -42,6 +57,40 @@ export const useGetPossibleFields = (formId?: number, appId?: string): Result =>
...
+                possibleFields,
+            });
+        });
+
+        return {
+            allPossibleFields,
+            isFetchingForms,
+        };
+    }, [isFetchingForms, allForms]);
+};
```
beygorghor, line 84: 😮  Do you really get all possible fields for all forms ? this is a huge query to the backend and can be a big payload. This is one reason why we have to preselect a form to enable querybuilder on instances page

beygorghor (approved):
Small remarks, i'm just wondering if the call to get all forms and their possible field won't crash at some point, do we use some cache in this ?

beygorghor:
can you please move those fields 
Screenshot 2024-09-05 at 14 24 51

hat/assets/js/apps/Iaso/domains/forms/hooks/useGetPossibleFields.ts
quang-le: It's from `react-hooks/exhaustive-deps`

quang-le (approved):

PR #1556, opened 2024-08-16
WC2-475: ETL not recording beneficiary visits data

butofleury:

Related JIRA tickets : [WC2-475](https://bluesquare.atlassian.net/browse/WC2-475)

- [ ] Are there enough tests

## Changes

For some visits the birth date is missing which break the execution of the script. So, added a function to deduce the birth date from registration date and beneficiary age. 

## How to test

From the local instance with wfp coda database, launch with `docker-compose run iaso manage wfp_etl` the ETL script in order to generate data for tableau dashboard. Then check for the beneficiary:

- Children under 5: [http://localhost:8081/wfp/debug/441/](http://localhost:8081/wfp/debug/441/)
- PBWG: [http://localhost:8081/wfp/debug/440/](http://localhost:8081/wfp/debug/440/)

[WC2-475]: https://bluesquare.atlassian.net/browse/WC2-475?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/plugins/wfp/common.py
+++ b/plugins/wfp/common.py
@@ -1,6 +1,7 @@
 from .models import *
 from iaso.models import *
 from datetime import datetime, timedelta, date
+from dateutil.relativedelta import *
 
 
 class ETL:
@@ -326,6 +327,9 @@ class ETL:
             current_journey["weight_difference"] = visit.get("weight_difference", None)
             current_journey["exit_type"] = self.exit_type(visit)
 
+        """ Check if it's first followup visit, in order to calculate the defaulter case based on the number of days defined in the assistance
+        admission form and next visit date in the antropometric followup visit form.
+        When the index is less or equal to 3, it means we still in the admission visit(visit 0). Otherwise, it's a start of first follow up visit(visit 1) """
         if index > 3:
             index = index - 1
             exit = self.exit_by_defaulter(visits, visits[index], followup_forms)
@@ -526,8 +530,7 @@ class ETL:
             beneficiary_age = int(age)
             registered_at = datetime.strptime(registration_date[:10], "%Y-%m-%d").date()
             if age_entry == "years":
-                calculated_date = registered_at - timedelta(days=(beneficiary_age * 365.25) + 1)
+                calculated_date = registered_at - relativedelta(years=beneficiary_age)
             elif age_entry == "months":
-                calculated_date = registered_at - timedelta(days=(beneficiary_age * 30.44) + 1)
-            calculated_date = {"registration_date": registration_date, "birth_date": calculated_date}
+                calculated_date = registered_at - relativedelta(months=beneficiary_age)
         return calculated_date
--- a/plugins/wfp/common.py
+++ b/plugins/wfp/common.py
@@ -301,6 +301,7 @@ class ETL:
         return exit
 
     def journey_Formatter(self, visit, anthropometric_visit_form, followup_forms, current_journey, visits, index):
+        default_anthropometric_followup_forms = followup_forms
         if visit["form_id"] == anthropometric_visit_form:
             current_journey["instance_id"] = visit.get("instance_id", None)
             current_journey["start_date"] = visit.get("start_date", None)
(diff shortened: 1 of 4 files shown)


plugins/wfp/common.py
```diff
@@ -325,9 +326,12 @@ def journey_Formatter(self, visit, anthropometric_visit_form, followup_forms, cu
...
             current_journey["exit_type"] = self.exit_type(visit)
 
-        if index > 0:
+        if index > 3:
             index = index - 1
-        exit = self.exit_by_defaulter(visits, visits[index], followup_forms)
+            exit = self.exit_by_defaulter(visits, visits[index], followup_forms)
+        else:
+            exit = self.exit_by_defaulter(visits, visit, followup_forms)
+
```
bramj: Not sure I understand these changes :thinking: can you explain the change to index `3`? Why the magic number?

plugins/wfp/common.py
```diff
@@ -512,3 +516,18 @@ def missed_followup_visit(self, visits, formIds, next_visit__date__, secondNextV
...
+    def calculate_birth_date(self, current_record):
+        age_entry = current_record.get("age_entry", None)
+        age = current_record.get("age__int__", None)
+        registration_date = current_record.get("registration_date", None)
+        calculated_date = None
+        if age_entry is not None and age_entry != "":
+            beneficiary_age = int(age)
+            registered_at = datetime.strptime(registration_date[:10], "%Y-%m-%d").date()
+            if age_entry == "years":
+                calculated_date = registered_at - timedelta(days=(beneficiary_age * 365.25) + 1)
```
bramj: Not sure why we don't use it yet, but if we'd use `python-dateutil` this would be easier, like
```python
 calculated_date = registered_at - relativedelta(years=beneficiary_age)
```
What do you think, worth it? https://github.com/dateutil/dateutil

plugins/wfp/common.py
```diff
@@ -512,3 +516,18 @@ def missed_followup_visit(self, visits, formIds, next_visit__date__, secondNextV
...
+            elif age_entry == "months":
+                calculated_date = registered_at - timedelta(days=(beneficiary_age * 30.44) + 1)
+            calculated_date = {"registration_date": registration_date, "birth_date": calculated_date}
```
bramj: I think this `registration_date` is not used? Not sure if this method needs to return it

plugins/wfp/common.py
butofleury: When it's index 3 it means It's a first visits. There is admission visit which considered as visit 0 with 3 forms: registration+Anthropometric admission+Medical admission+Assistance Admission) and when it's index 4 it means it's starting the visit 1(Followup visit) after admission visit(visit 0) . 

In some case, there are beneficiaries that never return after first followup visit(visit 1), and we need to get the number of days in previous visits form which is the assistance admission form(with index 3) in order to calculate the 2 next visits dates and flag **as defaulter the beneficiary** if there is no data in the 2 calculated next visits date. When the index is less or equal to 3, it means we still in the admission visit.

Will add a comment to explain more what is index 3.

plugins/wfp/common.py
butofleury: Thanks, I will change and use `dateutil` which is more simple.

plugins/wfp/common.py
butofleury: For example, the following beneficiary missed 2 consecutive visits: https://qa.coda.go.wfp.org/dashboard/entities/details/accountId/1/entityId/254: 
1. Next visit date on: 25/07/2023
2. Second visit date on: 08/08/2023

We need to calculate the 2 next visits dates based on the number of days defined in [assistance admission form](https://qa.coda.go.wfp.org/dashboard/forms/submission/accountId/1/instanceId/1759) **and flag the case as defaulter**.

Coda-2-Admission-Assistance-6_4_db316cad-434b-4082-9751-39dae654e647_2023-07-11_16-54-30

plugins/wfp/common.py
butofleury: will remove it and keep only the calculated birth date.

plugins/wfp/common.py
bramj: :+1: thanks for that info!

butofleury: @bramj , I changed a bit on: https://github.com/BLSQ/iaso/pull/1556/files#diff-adf0bcaae996ccef6e87dcac984ce60a2d7419bba436eeca75ea15fb33153d20R335

It checks if it's followup visit form. This more safer and it also help to know if it was a follow up visit which could be visit 1 or 2 or 3, etc.....

bramj: Nice :+1: looks safer indeed!

bramj (approved):
Nice work, LGTM!

PR #1558, opened 2024-08-19
IA-3102: Permissions : Org unit “view only” permission

hakifran:
- Permissions : Org unit “view only” permission
Related JIRA tickets : [IA-3102](https://bluesquare.atlassian.net/browse/IA-3102)

- [x] Are there enough tests

## Changes
- Add new Organisation units - Read permission
- Check on create_org_unit and edit org unit the Organisation units management permission
- Hide some button and action if the user has not Organisation units management permission on frontend

## How to test
- Create or update a user and remove the Organisation units management and add Organisation units - Read permission to it
- Check all the actions for create and update org units are possible

## Print screen / video
[Screencast from 2024-08-19 14-23-35.webm](https://github.com/user-attachments/assets/781f2a0b-29a3-4287-a6f6-1f022fe8121d)

## Notes

[IA-3102]: https://bluesquare.atlassian.net/browse/IA-3102?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iaso/tests/api/test_orgunits.py
```diff
@@ -934,6 +987,27 @@ def test_edit_org_unit_partial_update(self):
         self.assertEqual(ou.geom.wkt, MultiPolygon(Polygon([(0, 0), (0, 1), (1, 1), (0, 0)])).wkt)
         self.assertEqual(response.data["reference_instances"], [])
 
+    def test_edit_org_unit_partial_update_read_permission(self):
+        """Check tha we can only modify a part of the fille"""
```
kemar:
```suggestion
        """Check that we can only modify a part of the file"""
```

Is this docstring really accurate?

This test seems to ensure that an update is not possible with the read perm.

It looks very much like the test above.

kemar (changes requested):
I've tested it locally by adding just the read-only perm to a user but I can't see any org unit in the list and I've got a bunch of 403.

Is this normal?

https://github.com/user-attachments/assets/47f63fb9-47ae-4f89-a0f5-db70934d9518

hakifran:
> I've tested it locally by adding just the read-only perm to a user but I can't see any org unit in the list and I've got a bunch of 403.
> 
> Is this normal?
> Screen.Recording.2024-08-21.at.15.54.46.mov

That's right, I omitted the read permission on the groups, data source. It is fixed with the [commit](https://github.com/BLSQ/iaso/pull/1558/commits/728fcb1dba3c45c81f931936893a2679fdeb5dff) . Thank you

iaso/tests/api/test_orgunits.py
kemar: @hakifran what do you think about that? Is this test required? Because it seems like a duplicate of the one above?

hakifran: Refactored too, I just pushed the changes

kemar:
Almost good to me.

My last concern is a UI problem:

- if you can just READ org units
- only form buttons are removed
- so you still see inputs elements
- this can make a user believe an input can be changed

I guess it's OK because there are no spec on this in the Jira ticket…

What do you think?

ui

hakifran:
> Almost good to me.
> 
> My last concern is a UI problem:
> 
>     * if you can just READ org units
> 
>     * only form buttons are removed
> 
>     * so you still see inputs elements
> 
>     * this can make a user believe an input can be changed
> 
> 
> I guess it's OK because there are no spec on this in the Jira ticket…
> 
> What do you think?
> 
> ui

I think you are right we should disable the inputs too. Thank you, let me fix it!

hakifran:
> Almost good to me.
> 
> My last concern is a UI problem:
> 
>     * if you can just READ org units
> 
>     * only form buttons are removed
> 
>     * so you still see inputs elements
> 
>     * this can make a user believe an input can be changed
> 
> 
> I guess it's OK because there are no spec on this in the Jira ticket…
> 
> What do you think?
> 
> ui

I'v just pushed the commit about disabling inputs when the use has not org unit managament permission. But there is on input  `<InputComponent
                keyValue="aliases"
                onChange={onChangeInfo}
                value={orgUnitState.aliases.value}
                type="arrayInput"
            />` Which I made a fix but which I think the change(disabled should be added to bluesquare-component)

kemar (approved):
LGTM

PR #1568, opened 2024-08-21
WC2-363 Add django-admin command to delete entities...

bramj:
and form submissions for a given account.

Useful to clean up test/staging environments or prod environments right before going live.

Related JIRA tickets : https://bluesquare.atlassian.net/browse/WC2-363

### DIFF
--- a/iaso/management/commands/delete_all_instances_and_entities.py
+++ b/iaso/management/commands/delete_all_instances_and_entities.py
@@ -1,0 +1,83 @@
+from django.core.management.base import BaseCommand
+from django.db import transaction
+
+from iaso.models import Account, Entity, Instance, InstanceFile, StorageDevice, StorageLogEntry
+
+
+class Command(BaseCommand):
+    """
+    This command deletes all form submissions and entities/beneficiaries for a given account.
+    It's useful to clean up test/staging environments, or to remove test data from
+    production environments right before going live.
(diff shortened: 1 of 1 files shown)


iaso/management/commands/delete_all_instances_and_entities.py
```diff
@@ -0,0 +1,83 @@
...
+
+            print("Deleting entities...")
+            self.delete_resources(
+                "\tEntities",
+                entities_to_delete,
+                options,
+            )
+            print("DONE.")
+
+            print("Deleting storages...")
```
bramj, line 75: @madewulf Should we clean the storages or not really needed?

madewulf: I think we should, indeed

madewulf: Other than that, this LGTM. Did you have the opportunity to test it on a copy of their database and check that it was not failing due to some foreign key constraints?

madewulf (approved):

iaso/management/commands/delete_all_instances_and_entities.py
bramj: Just tested on a copy of their prod database, no issues. Merging!

PR #1569, opened 2024-08-21
SLEEP-1439 Add permission for offline setup zip download

bramj:
Also: Enable task api in case of access to offline setup file.

Without this, the user will see an error when the frontend tries to render the progress bar.

Related JIRA tickets : [SLEEP-1439](https://bluesquare.atlassian.net/browse/SLEEP-1439) (already in Trypelim)

## How to test

- There should be a new permission "Télécharger l'installation hors ligne de l'application mobile" visible in the user management screen
- When you have this new permission, you see this icon appear as a new user action, allowing you to download a mobile app setup-zip as this user:

- If you don't have the permission, you don't see the action on the users

[SLEEP-1439]: https://bluesquare.atlassian.net/browse/SLEEP-1439?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


beygorghor:
@hakifran we need a review here i think please !

hakifran:
> @hakifran we need a review here i think please !

I'm reviewing it

hat/assets/js/apps/Iaso/domains/users/config.js
```diff
@@ -92,7 +92,10 @@ export const usersTableColumns = ({
                             }
                         />
                     )}
-                {currentUser.is_superuser && (
+                {userHasPermission(
+                    Permission.MOBILE_APP_OFFLINE_SETUP,
```
hakifran: This can be achieved by using 
`<DisplayIfUserHasPerm
                    permissions={[Permission.MOBILE_APP_OFFLINE_SETUP]}
                >
                    <ExportMobileAppSetupDialog
                        selectedUser={settings.row.original}
                        titleMessage={MESSAGES.exportMobileAppTitle}
                        params={params}
                        onCreateExport={exportMobileSetup}
                    />
                </DisplayIfUserHasPerm>`
import { DisplayIfUserHasPerm } from '../../components/DisplayIfUserHasPerm.tsx';

iaso/api/tasks/__init__.py
```diff
@@ -72,7 +72,7 @@ class TaskSourceViewSet(ModelViewSet):
     PATCH /api/tasks/<id>
     """
 
-    permission_classes = [permissions.IsAuthenticated, HasPermission(permission.DATA_TASKS)]  # type: ignore
+    permission_classes = [permissions.IsAuthenticated, HasPermission(permission.DATA_TASKS, permission.MOBILE_APP_OFFLINE_SETUP)]  # type: ignore
```
hakifran: @bramj I don't understand why this permission is checked here in TaskSourceVIewSet as this is about:
    """Task API

    GET /api/tasks/
    GET /api/tasks/<id>
    PATCH /api/tasks/<id>
    """

hakifran (changes requested):

hat/assets/js/apps/Iaso/domains/users/config.js
bramj: :+1: adapted

iaso/api/tasks/__init__.py
bramj: This was needed to enable me to display a progress bar on the zip creation (which polled the `GET /api/tasks/:id` that was just created).

However, you're right that it is not clear, and too permissive, so I've changed the behavior of only the `GET /api/tasks/<id>` endpoint.

Basically: keep the tasks endpoint permissions as is, but allow a user to query the state of a task launched by themselves, even if they don't have the `DATA_TASKS` permission.

IMO it makes sense to allow this, but only on the `retrieve` endpoint and only for tasks created by the user themself.

@hakifran what do you think?

hakifran: @bramj I do think that the user who launch the task should have the  DATA_TASKS, if not don't let him launch the mobile app offline setup. Because on my point of view this can lead to a confusion about permissions. @beygorghor or @madewulf can give their point of view on this. thank you

hakifran (approved):

PR #1581, opened 2024-08-24
[WC2-478] Using source_created_at and source_updated_at instead of their server…

madewulf:
See ticket here: https://bluesquare.atlassian.net/browse/WC2-478

The problem is the wrong ordering of visits in this screen
Screenshot_20240824-143247 (1)

As a solution: sending source_created_at and source_updated at to the mobile phones in the mobile entity API because we were sending the update and creation date from the server, which hides the timing between visits at WFP, which is problematic because this timing is essential to find out if a beneficiary is late. 

- [ ] Are there enough tests

## How to test
Upload and download entity on the mobile app

### DIFF
--- a/iaso/api/mobile/entity.py
+++ b/iaso/api/mobile/entity.py
@@ -74,15 +74,8 @@ class MobileEntityAttributesSerializer(serializers.ModelSerializer):
     id = serializers.CharField(read_only=True, source="uuid")
     org_unit_id = serializers.CharField(read_only=True, source="org_unit.id")
     form_version_id = serializers.SerializerMethodField()
-
-    created_at = serializers.SerializerMethodField()
-    updated_at = serializers.SerializerMethodField()
-
-    def get_created_at(self, instance):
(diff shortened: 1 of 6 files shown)


iaso/api/mobile/entity.py
```diff
@@ -75,8 +75,14 @@ class Meta:
     org_unit_id = serializers.CharField(read_only=True, source="org_unit.id")
     form_version_id = serializers.SerializerMethodField()
 
-    created_at = TimestampField()
-    updated_at = TimestampField()
+    created_at = serializers.SerializerMethodField()
+    updated_at = serializers.SerializerMethodField()
+
+    def get_created_at(self, instance):
+        return instance.source_created_at.timestamp() if instance.source_created_at else None
```
bramj: This would do the trick yes, but maybe fallback to created_at just in case would be best

bramj:
@madewulf I've added a fix for when a form is edited with Enketo (creation with Enketo was already okay) and I went further to ensure we have the correct timestamp, while not sending None to the mobile app.
I also noticed we needed the same fix on reference instances.

bramj (approved):

PR #1584, opened 2024-08-26
POLIO-1627 Chronogram: allow to filter by campaign

kemar:
Allow to filter chronograms by campaign.

Related JIRA tickets : [POLIO-1627](https://bluesquare.atlassian.net/browse/POLIO-1627)

## Changes

- allow the backend API to filter chronograms by campaign
- expose the list of campaigns used in chronograms via an OPTIONS endpoint
    - rename `useOptionChronogramTask.ts` to `useOptionChronogram.ts` in the frontend
- allow the frontend to filter chronograms by campaign

## How to test

1. go to `http://localhost:8081/dashboard/polio/chronogram/`
2. ensure you have at least 2 chronograms related to two different campaigns
3. try to filter by campaign

## Video

https://github.com/user-attachments/assets/6d5bb950-b643-4abe-b620-ca37481154b8

[POLIO-1627]: https://bluesquare.atlassian.net/browse/POLIO-1627?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


plugins/polio/api/chronogram/views.py
```diff
@@ -61,6 +61,27 @@ def get_queryset(self) -> QuerySet:
...
+        """
+        Add custom metadata about the API.
+        """
+        if self.metadata_class is None:
+            return self.http_method_not_allowed(request, *args, **kwargs)
+        data = self.metadata_class().determine_metadata(request, self)
+
+        chronogram_campaigns = (
+            self.get_queryset()
+            .order_by("round__campaign__id")
```
quang-le: Wouldn't it make more sense to order by `obr_name` for the front-end?

plugins/polio/js/src/domains/Chronogram/Chronogram/index.tsx
```diff
@@ -22,16 +24,27 @@ export const Chronogram: FunctionComponent = () => {
     const classes: Record<string, string> = useStyles();
     const { formatMessage } = useSafeIntl();
 
+    const { data: chronogramMetaData, isFetching: isFetchingMetaData } =
+        useOptionChronogram();
+
```
quang-le: You could call that hook directly in `ChronogramFilters` since you don't use the data here and just pass it as props

plugins/polio/js/src/domains/Chronogram/Chronogram/Filters/ChronogramFilters.tsx
```diff
@@ -71,7 +76,18 @@ export const ChronogramFilters: FunctionComponent<Props> = ({ params }) => {
...
+                <Grid item xs={12} md={3} lg={3}>
+                    <InputComponent
+                        keyValue="campaign"
+                        clearable
+                        onChange={handleChange}
+                        value={filters.campaign}
+                        options={chronogramMetaData.campaigns}
+                        type="select"
+                        label={MESSAGES.filterLabelCampaign}
+                    />
```
quang-le: you should pass `isFetchingMetaData` as `loading` prop

plugins/polio/js/src/domains/Chronogram/Chronogram/index.tsx
```diff
@@ -22,16 +24,27 @@ export const Chronogram: FunctionComponent = () => {
...
             <TopBar
                 title={formatMessage(MESSAGES.chronogramTitle)}
                 displayBackButton={false}
             />
-            <Box className={classes.containerFullHeightNoTabPadded}>
-                <ChronogramFilters params={paramsNew} />
-                <ChronogramTable params={paramsNew} />
-            </Box>
+            {isFetchingMetaData && <LoadingSpinner />}
+            {!isFetchingMetaData && (
```
quang-le: It would be better to render the filters and pass `isFetchingMetaData` as `loading` to the `InputComponent` in your filters

plugins/polio/js/src/domains/Chronogram/api/useOptionChronogram.ts
```diff
@@ -3,9 +3,33 @@ import { UseQueryResult } from 'react-query';
...
+> =>
+    useSnackQuery({
+        queryKey: ['optionChronogram'],
+        queryFn: () => optionsRequest(`${apiBaseUrl}`),
+        options: {
+            staleTime: 1000 * 60 * 15, // in ms
+            cacheTime: 1000 * 60 * 5,
+            select: data => {
+                return {
+                    campaigns: mapChoices(data.campaigns_filter_choices),
```
quang-le, line 27: I knew this one would be useful 😁

quang-le (changes requested):
1 question and 1 small blocking change in the front-end.

Question: The spec doesn't mention it but wouldn't it be "safer"  to enable multi select on the campaign filter? 

Blocking change: remove conditional rendering of the filters and use loading state for the campaign filter instead

plugins/polio/api/chronogram/views.py
kemar: Seems legit 😄

kemar:
> Question: The spec doesn't mention it but wouldn't it be "safer" to enable multi select on the campaign filter?

It could be a bit of laziness but the spec doesn't mention it 🤣 Anyway it's done here https://github.com/BLSQ/iaso/pull/1584/commits/8d8cacd6457b3e229256c204a004c71b249fd51e

> Blocking change: remove conditional rendering of the filters and use loading state for the campaign filter instead

This has been fixed too.

quang-le (approved):

PR #1586, opened 2024-08-26
IA-3312: User Right - "User management - Geo limited" I can not create user within the accessible part of the pyramide

hakifran:
- User Right - "User management - Geo limited" I can not create user within the accessible part of the pyramide
Related JIRA tickets : [IA-3312](https://bluesquare.atlassian.net/browse/IA-3312)

- [ ] Are there enough tests

## Doc

## Changes
- Let the user with User management - Geo limited to create, delete and bulk-create users

## How to test
- Create or Edit a user and add the **User management - Geo limited permission** without **User management - Admin**
- Check if the user can create, delete and bulk-create users

## Print screen / video
Screenshot from 2024-08-26 19-40-06
Screenshot-from-2024-08-26-19-41-29

## Notes

[IA-3312]: https://bluesquare.atlassian.net/browse/IA-3312?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iaso/api/bulk_create_users.py
```diff
@@ -204,6 +219,13 @@ def create(self, request, *args, **kwargs):
                         if ou.isdigit():
                             try:
                                 ou = OrgUnit.objects.get(id=int(ou))
+                                if self.has_user_managed_permission(request) and ou not in importer_orgunits_hierarchy:
+                                    raise serializers.ValidationError(
+                                        {
+                                            "error": f"Operation aborted. A User with {permission.USERS_MANAGED} permission "
+                                            "has to create users with OrgUnits that are in the its controlled pyramid"
```
kemar, line 227:
```suggestion
                                            "has to create users with OrgUnits that are in its controlled pyramid"
```

Small typo.

iaso/api/bulk_create_users.py
```diff
@@ -196,6 +203,14 @@ def create(self, request, *args, **kwargs):
                         )
 
                     org_units = row[csv_indexes.index("orgunit")].split(value_splitter)
+                    if self.has_user_managed_permission(request) and len(list(filter(None, org_units))) == 0:
```
kemar: The bulk creation is now very slow (more than 20s just to create 1 user):

slow

I think you shouldn't call `has_user_managed_permission` in every sub-loop because it performs database queries to check permissions.

This might be the reason why the bulk creation is now slow.

I recommend checking the rights for the connected user only once at the beginning.

kemar (changes requested):
The bulk creation is now very slow.

I added a comment with what I believe is the reason of the issue.

We should fix this before merging this PR.

kemar (approved):
LGTM

PR #1593, opened 2024-08-28
Fix black in Docker doesn't use the configuration

bmonjoie:
When running `docker compose exec iaso black`, black didn't use the provided configuration (`pyproject.toml`).
This is due to the fact that the file was not exposed to Docker.

Also, the pre-commit hook configuration used a different version of black than the one in `requirements.txt`.

## Changes

- Add `pyproject.toml` in the files exposed to Docker
- Change black's version in the pre-commit hook configuration file.

## How to test

Run `docker compose exec iaso black -v` and check that it doesn't reformat any files.

Also, with -v, it should show that it found the `pyproject.toml` configuration.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,7 +2,7 @@
 # see Black section in README
 repos:
 -   repo: https://github.com/ambv/black
-    rev: 23.3.0
+    rev: 23.12.1
     hooks:
     - id: black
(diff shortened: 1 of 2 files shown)


kemar (approved):
💪

tdethier (approved):
LGTM :+1:

beygorghor (approved):

PR #1594, opened 2024-08-28
WC2-486 Mobile storage logs: Always return 201

bramj:
But log errors to Sentry and create a record in the apiimport table. Adapted the code in the view to let it fail if an org unit or entity is not found.

Related JIRA tickets : https://bluesquare.atlassian.net/browse/WC2-486

## How to test

Hit the endpoint `POST /api/mobile/storage/logs/` with a payload such as:
```json
[
  {
    "entity_id": "f1aa4235-XXXX-XXXX-XXXX-7547c16fb94b",
    "id": "XXXXXXXX-781c-470a-a433-7e14eb3e204a",
    "instances": [],
    "operation_type": "WRITE_PROFILE",
    "org_unit_id": "1",
    "performed_at": 1724313030.741,
    "storage_id": "BEFIuqRwgA==",
    "storage_type": "NFC"
  }
]
```

If the `org_unit_id` or `entity_id` is not known in your db, it will still return a 201, but create a record in the `vector_control_apiimport` table with `has_problem=True`.


hat/api_import/models.py
```diff
@@ -8,6 +8,7 @@
     ("orgUnit", "Org Unit"),
     ("instance", "Form instance"),
     ("bulk", "Bulk Org Units and Instances"),
+    ("storageLog", "Storage logs"),
```
bmonjoie, line 11: What is this for?

bramj: The `@safe_api_import()` decorator will create a new record of `APIImport`. So I just added this new type, it's the first argument of the decorator function.

bmonjoie (approved):
LGTM

PR #1595, opened 2024-08-28
Use `annotate` to avoid relying on `get_serializer_context`

kemar:
I think the code could be simplified and a little bit more performant with `annotate`.

I'm not 100% sure about the logic but the unit tests are OK.

WDYT @bmonjoie?


tdethier:
I like it. To me, it's easier to read

bmonjoie:
I like it as well. You can merge that :)

PR #1598, opened 2024-08-29
IA-3355 add groupsets api

mestachs:
Expose existing GroupSet model via api

Related JIRA tickets : IA-3355

- [ ] Are there enough tests
  
## Doc

## Changes

## How to test

seed data via dhis2 import
```
docker-compose run --rm iaso manage seed_test_data --mode=seed --dhis2version=2.40.4.1
```

http://localhost:8081/api/group_sets/?fields=id,name
http://localhost:8081/api/group_sets/?fields=id,name,groups,source_version

## Print screen / video

## Notes


iaso/tests/api/test_group_sets.py
```diff
@@ -0,0 +1,373 @@
...
+
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        self.assertEqual(GroupSet.objects.count(), 1)
+
+        created_groupset = GroupSet.objects.last()
+        self.assertEqual(created_groupset.name, "New GroupSet")
+        self.assertEqual(created_groupset.source_ref, valid_payload["source_ref"])
+        self.assertEqual(created_groupset.source_version.id, self.source_version_1.id)
+        self.assertEqual(created_groupset.group_belonging, GroupSet.GroupBelonging.SINGLE)
+        self.assertEqual([x.id for x in created_groupset.groups.all()], [self.src_1_group_1.id, self.src_1_group_2.id])
```
tdethier, line 134: Instead of calling `.all()` and doing list comprehension, I used in OUCRC `.values_list("id", flat=True)`. Not sure if that changes a lot since it's run on a test DB without a lot of data

iaso/tests/api/test_group_sets.py
```diff
@@ -0,0 +1,373 @@
...
+        cls.data_source.projects.add(cls.project_1)
+        cls.data_source.save()
+
+        cls.data_source_2.account = star_wars
+        cls.data_source_2.projects.add(cls.project_1)
+        cls.data_source_2.save()
+
+        cls.acccount_1_user_1.iaso_profile.projects.add(cls.project_1)
+
+    def test_authentification_required(self):
```
tdethier: I would maybe add:

- a test for retrieve/POST/PATCH/DELETE without authentication
- a test for list/retrieve/POST/PATCH/DELETE without `org_units` permission

iaso/api/group_sets/serializers.py
```diff
@@ -0,0 +1,185 @@
...
+
+        audit_models.log_modification(original_copy, instance, source=audit_models.GROUP_SET_API, user=user)
+
+        return instance
+
+    source_version = SourceVersionSerializerForGroupset(read_only=True)
+    groups = GroupSerializer(many=True, read_only=True)
+
+    # using none() to avoid leaking other project info
+    # see __init__ to filter based on user access
```
tdethier: Interesting, I'll probably do something like this in OUCRC

iaso/api/group_sets/serializers.py
```diff
@@ -0,0 +1,185 @@
...
+        self.assign_relations(instance)
+
+        request = self.context.get("request")
+        user = request.user
+
+        audit_models.log_modification(original_copy, instance, source=audit_models.GROUP_SET_API, user=user)
+
+        return instance
+
+    source_version = SourceVersionSerializerForGroupset(read_only=True)
```
tdethier: I would put the attributes above, before the `class Meta` so that it's easier to know what's in the serializer, but it's just personal preference

tdethier (commented):
I would add a few more tests to check authentication and permissions, otherwise LGTM :+1:

tdethier (approved):
LGTM :+1:

PR #1600, opened 2024-08-29
IA-3146 orgunit default image

beygorghor:
Be able to choose one of the image uploaded with a submission and set it as default for a specific org unit

Related JIRA tickets : IA-3146

- [x] Are there enough tests

## Changes

- adapt model to allow an instanceFile has default_image + test
- adapt org unit api to return default_image + Tests
- change get attachment method to be able to return only images, returning also mime type
- add a new tab in org unit details to display all possible images
- allow user to select one of the picture to be the default one
- modify update method to save default_image
 
## How to test

Find an existing instance and make sure it is linked to an org unit using the admin
Create a new instance file using an image and link it to this instance
Repeat the process a few times
Visit the detail page of this org unit, click on images and play with the favorite button, reload the page to see if changes has been saved

## Print screen / video

https://github.com/user-attachments/assets/3d798cb6-d179-40a9-be0e-ba81e3a0230e

Screenshot 2024-08-29 at 14 41 37

## Notes
Setuper should probably use a few images while creating org units


quang-le (commented):
Front-end looks OK for me

iaso/api/instances.py
```diff
@@ -165,7 +173,14 @@ def attachments(self, request):
         instances = self.get_queryset()
         filters = parse_instance_filters(request.GET)
         instances = instances.for_filters(**filters)
-        queryset = InstanceFile.objects.filter(instance__in=instances)
+        queryset = InstanceFile.objects.filter(instance__in=instances).annotate(
```
mathvdh, line 176: This is nice but I'm wondering about the performances of this.
I tested locally and it seems you CAN get all the attachments of ALL the instances in 1 single query and filter with only_images = True

http://127.0.0.1:8000/api/instances/attachments/?image_only=true

Do we really need this .. Have all the attachements of all the instances (With or without filtering per image_only)

On my small local DB the perf seemed to be OK but still I'm curious on an server with a lot of instances

mathvdh (commented):
I tested locally and inspected the code and it's all good except for my comment ( But this is actually more general than this PR)

iaso/api/instances.py
beygorghor: Totally agree, this is already the case on staging and production, i'll create another ticket for that
https://iaso-staging.bluesquare.org/api/instances/attachments/

mathvdh (approved):
Ok since we made a separate ticket for the attachment thing

PR #1602, opened 2024-08-29
IA-3375: Export user role and project as name not as id in csv and xlsx user export

hakifran:
- The user csv export on userrole column doesn't fite the import template
Related JIRA tickets : [IA-3375](https://bluesquare.atlassian.net/browse/IA-3375)

- [ ] Are there enough tests

## Doc

## Changes
- Display user role as name not as id
- Display project name instead of id

## How to test
- Go in user management
- Export csv or xlsx file
- Check the user role and project columns

## Print screen / video
[Screencast from 2024-08-29 16-53-08.webm](https://github.com/user-attachments/assets/72d5debb-dfc3-49df-98e7-85454d26f2e5)

## Notes

[IA-3375]: https://bluesquare.atlassian.net/browse/IA-3375?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iaso/tests/api/test_profiles.py
quang-le, line 1: Could you add a test on the value of user roles?

quang-le (changes requested):
It works well, just add the missing test pls

iaso/tests/api/test_profiles.py
hakifran: @quang-le I added the tests on both csv and xlsx

quang-le (approved):

PR #1604, opened 2024-09-02
IA-2293: Fix: Improve performances on bulk create user

madewulf:
Fix performances on user creation

Related JIRA tickets :IA-2293

- [x] Are there enough tests

##
There was a double call to the hierarchy test: you would ask the database to donwload all the org units that are children of the children of the ones you are allowed to see (that's one children too much in this phrase, and that was the bug)


tdethier (approved):
LGTM :+1:

PR #1606, opened 2024-09-02
IA-3338: Add fiter "Project" on Form statistics

hakifran:
- Add fiter "Project" on Form statistics
Related JIRA tickets : [IA-3338](https://bluesquare.atlassian.net/browse/IA-3338)

- [ ] Are there enough tests

## Doc

## Changes
- Added projects filter on front-end form stats page
- Add projects filter in back-end on form stats api

## How to test
- Forms --> Form statistics
- Check if the projects filter works where

## Print screen / video
[Screencast from 2024-09-02 16-59-12.webm](https://github.com/user-attachments/assets/824a9929-712a-4d44-9962-6ea4fc99acf2)
## Notes

[IA-3338]: https://bluesquare.atlassian.net/browse/IA-3338?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/forms/hooks/UseGetFormStats.tsx
+++ b/hat/assets/js/apps/Iaso/domains/forms/hooks/UseGetFormStats.tsx
@@ -1,0 +1,42 @@
+import { UseQueryResult } from 'react-query';
+import { getRequest } from '../../../libs/Api';
+import { useSnackQuery } from '../../../libs/apiHooks';
+import { useParamsObject } from '../../../routing/hooks/useParamsObject';
+import { makeUrlWithParams } from '../../../libs/utils';
+
+const getFormStats = url => {
+    return getRequest(url);
+};
+
+type ApiParams = {
+    project_ids?: string;
+};
+
+const useGetApiParams = (params: {
+    accountId: string;
+    projectIds: string;
+}): ApiParams => ({
+    project_ids: params.projectIds,
+});
+
+export const useGetFormStats = (
+    baseUrl,
+    url,
+    queryKey,
+): UseQueryResult<any, Error> => {
+    const params = useParamsObject(baseUrl) as {
+        accountId: string;
+        projectIds: string;
+    };
+    const apiParams = useGetApiParams(params);
+    const apiUrl = makeUrlWithParams(url, apiParams);
+    return useSnackQuery({
+        queryKey: [queryKey, params],
+        queryFn: () => getFormStats(apiUrl),
+        options: {
+            staleTime: 60000,
+            cacheTime: 60000,
+            keepPreviousData: true,
+        },
+    });
+};
--- a/hat/assets/js/apps/Iaso/domains/forms/hooks/UseGetFormStats.tsx
+++ b/hat/assets/js/apps/Iaso/domains/forms/hooks/UseGetFormStats.tsx
@@ -1,42 +1,43 @@
 import { UseQueryResult } from 'react-query';
 import { getRequest } from '../../../libs/Api';
 import { useSnackQuery } from '../../../libs/apiHooks';
-import { useParamsObject } from '../../../routing/hooks/useParamsObject';
 import { makeUrlWithParams } from '../../../libs/utils';
+import { ApiFormStatsParams } from '../types/formStats';
 
 const getFormStats = url => {
     return getRequest(url);
 };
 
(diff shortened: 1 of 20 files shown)


hat/assets/js/apps/Iaso/domains/forms/stats.tsx
quang-le, line 1: tsx, nice 😀

hat/assets/js/apps/Iaso/domains/forms/hooks/UseGetFormStats.tsx
```diff
@@ -0,0 +1,42 @@
...
+    return useSnackQuery({
+        queryKey: [queryKey, params],
+        queryFn: () => getFormStats(apiUrl),
+        options: {
+            staleTime: 60000,
+            cacheTime: 60000,
+            keepPreviousData: true,
+        },
+    });
+};
```
quang-le: Nice re-usable hook. 
What I would do is make a wrapper on this hook for each stat (stat and stat sum). 
That way we won't have to go to the component to figure out what the query key when debugging.

hat/assets/js/apps/Iaso/domains/forms/hooks/UseGetFormStats.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+const useGetApiParams = (params: {
+    accountId: string;
+    projectIds: string;
+}): ApiParams => ({
+    project_ids: params.projectIds,
+});
+
+export const useGetFormStats = (
+    baseUrl,
```
quang-le: Pass `params`instead, they are already available in the parent

hat/assets/js/apps/Iaso/domains/forms/hooks/UseGetFormStats.tsx
```diff
@@ -0,0 +1,42 @@
...
+    queryKey,
+): UseQueryResult<any, Error> => {
+    const params = useParamsObject(baseUrl) as {
+        accountId: string;
+        projectIds: string;
+    };
+    const apiParams = useGetApiParams(params);
+    const apiUrl = makeUrlWithParams(url, apiParams);
+    return useSnackQuery({
+        queryKey: [queryKey, params],
```
quang-le: Pass `apiUrl` instead of `params`. Not especially blocking, but the idea is that the API call will change when the `apiUrl` changes.
It doesn't make a difference here, but with paginated calls, we sometimes don't get the pagination params from the params and add them when formatting the `apiUrl`, so different sets of params can result in the same `apiUrl`. Using `apiUrl` in the `queryKey` is a small optimization of cache use

hat/assets/js/apps/Iaso/domains/forms/hooks/UseGetFormStats.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+export const useGetFormStats = (
+    baseUrl,
+    url,
+    queryKey,
+): UseQueryResult<any, Error> => {
+    const params = useParamsObject(baseUrl) as {
+        accountId: string;
+        projectIds: string;
+    };
```
quang-le: Cf above: you can save yourself a hook call by passing the `params` as props

hat/assets/js/apps/Iaso/domains/forms/hooks/UseGetFormStats.tsx
quang-le: Something like:
```typescript
export const useGetFormsStatsSum = params => {
    return useGetFormStats({
        params,
        url: '/api/instances/stats_sum/',
        queryKey: ['instances', 'stats_sum'],
    });
};
```

I would also use named arguments, that makes the code more legible when you have more than 2 arguments

quang-le (changes requested):
The alignment of the filter and the button is off.

Screenshot 2024-09-04 at 11 08 02

Code comments are for style/good practice.

hakifran:
> The alignment of the filter and the button is off.
> 
> Screenshot 2024-09-04 at 11 08 02
> 
> Code comments are for style/good practice.
Screenshot from 2024-09-04 14-32-06

quang-le (approved):

PR #1607, opened 2024-09-02
 IA-3394 Feat: improve profile api performance

madewulf:
Feat: improve profile api performance, by removing useless data and prefetching useful data

Related JIRA tickets : IA-3394

- [x] Are there enough tests

### DIFF
--- a/iaso/models/base.py
+++ b/iaso/models/base.py
@@ -1348,7 +1348,7 @@ class Profile(models.Model):
     def __str__(self):
         return "%s -- %s" % (self.user, self.account)
 
-    def as_dict(self):
+    def as_dict(self, small=False):
         user_roles = self.user_roles.all()
         user_group_permissions = list(
             map(lambda permission: permission.split(".")[1], list(self.user.get_group_permissions()))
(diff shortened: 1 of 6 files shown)


iaso/models/base.py
```diff
@@ -1358,27 +1370,50 @@ def as_dict(self):
...
-            "language": self.language,
-            "user_id": self.user.id,
-            "dhis2_id": self.dhis2_id,
-            "home_page": self.home_page,
-            "phone_number": self.phone_number.as_e164 if self.phone_number else None,
-            "country_code": region_code_for_number(self.phone_number).lower() if self.phone_number else None,
-            "projects": [p.as_dict() for p in self.projects.all().order_by("name")],
-        }
+        if not small:
+            print("not small")
```
hakifran: I think this print should be removed

iaso/models/org_unit.py
```diff
@@ -509,6 +509,17 @@ def as_small_dict(self):
             res["search_index"] = self.search_index
         return res
 
+    def as_very_small_dict(self):
+        res = {
```
hakifran: Not blocking but I think this dict should be return directly without passing it to res before to return it

hakifran (changes requested):
I don't know if it's only on my local machine but these changes crushed test_upload_valid_csv and test_upload_valid_csv_with_perms test.

madewulf:
@hakifran there are changes to the test files https://github.com/BLSQ/iaso/pull/1607/files#diff-bc4da97802d782f1e9e50fc072a80ed8b252157a2e01b866130dfada1634e964 but that should not delete the files on your machine

hakifran (approved):

PR #1608, opened 2024-09-03
POLIO-1653 POLIO-1654 Fix for sorting not working on API for Vaccine Stock Management

mathvdh:
API For FormA, Incident Report and Destruction Report ignores the order paramater and doesnt sort properly its results. This fixes it.

Related JIRA tickets : POLIO-1653 POLIO-1654

- [x] Are there enough tests

## Changes

Its a simple change on the API side as the Viewsets for these 3 models inherit from a common base viewset.
Another few changes to make it work on the frontend.
Finally added some tests as this area was under-tested ...

## How to test

When you go in Vaccine Module > Vaccine Stock Management then the details of a Vaccine Stock then Stock variation page, You should be able to order the lists under the 3 tabs Form A, Destruction Reports, Incident Reports


hakifran (approved):
LGTM

PR #1610, opened 2024-09-04
IA-3365 user history backend

quang-le:

Related JIRA tickets : IA-3365

- [x] Are there enough tests

## Doc

## Changes
- update `Modification` model:
    - use SET_NULL i.o DO_NOTHING for user foreign key
    - Add SQL migration to make existing logs consistent
- Add `ProfileAuditLogger, subclassing `AuditLogger`. The shape of the `new_value` and `past_value` entries matches the shape of existing logs for users and adds a few fields. It was kept flat for ease of use with existing front-end components.
- Add `ProfileAuditSerializer` and related nested serializers:
    - In the `new_value` field, we log a boolean `password_updated`.
    - On deletion, we log the time in a `deleted_at` field, in anticipation of soft delete feature (so the logs will be consistent)
- Add audit logging to 'create`method of `ProfileViewset `
- Refactor `partial_update` method of `ProfileViewset` to raise all potential exceptions before writing to the DB. Add audit logging when update is successful
- Refactor `update_single_profile_from_bulk` in the same way. Add  audit log on profile if update successful. Add audit log for teams if teams are updated
- Add audit logging to `BulkCreateUserFromCsvViewSet`.
- Add fixtures and tests

## How to test
- Login as super user to have full access to the `/api/logs/`
Go to iaso>admin>users
- create a single user
- Go to `/api/logs/contentType=iaso.profile&fields=past_value,new_value/`
- The latest log should be the user you created. It should have an empty `past_value`. `new_value.password_updated` should be `True`

- update existing user
- repeat above steps and check that values are correct

- select several users in the table
- With the speed dial, select update
- Update all possible fields and save
- Go to `/api/logs/contentType=iaso.profile&fields=past_value,new_value/`: you should find a log per user
- Go to `/api/logs/contentType=iaso.team&fields=past_value,new_value/`: you should find a log per user and per team modified

- click create from file button and confirm
- Go to `/api/logs/contentType=iaso.profile&fields=past_value,new_value/`: you should find a log per user

## Notes

I haven't refactored the code for user bulk create because we will go through it again for performance and move it in to a task shortly, so it can be done correctly then


iaso/api/profiles/audit.py
```diff
@@ -0,0 +1,91 @@
...
+        return user_role.remove_user_role_name_prefix(user_role.group.name)
+
+
+class NestedUserAuditSerializer(serializers.ModelSerializer):
+    user_permissions = serializers.SerializerMethodField()
+
+    class Meta:
+        model = User
+        # We're serializing the password to be able to determine if it has been changed. but we remove them before saving the log itself
+        # DO NOT reuse this serializer if you're going to send the payload to the front-end
```
hakifran: Why not declare it inline inside the ProfileAuditSerializer to make sure it won't be used any where else

iaso/api/profiles/profiles.py
```diff
@@ -360,55 +362,153 @@ def partial_update(self, request, pk=None):
...
+        return result
+
+    def validate_projects(self, request, profile):
+        result = []
+        request_user = request.user
+        projects = request.data.get("projects", None)
+        if projects is not None:
+            if not request_user.has_perm(permission.USERS_ADMIN):
+                raise PermissionDenied(
+                    f"User with permission {permission.USERS_MANAGED} cannot changed project attributions"
```
hakifran: **Connot change**

iaso/tasks/profiles_bulk_update.py
```diff
@@ -32,21 +39,41 @@ def update_single_profile_from_bulk(
...
+    old_data = audit_logger.serialize_instance(profile)
     account_id = user.iaso_profile.account.id
+    roles_to_be_added = []
+    roles_to_be_removed = []
+    projects_to_be_added = []
+    projects_to_be_removed = []
+    org_units_to_be_added = []
+    org_units_to_be_removed = []
+    # Raise if necessary and prepare values to be updated
+    if teams_id_added is not None:
```
hakifran, line 53: For me it could be more clear if this condition is combine in one condition with the one at 120th line

iaso/tasks/profiles_bulk_update.py
```diff
@@ -32,21 +39,41 @@ def update_single_profile_from_bulk(
...
+        if not user.has_perm(permission.TEAMS):
+            raise PermissionDenied(f"User without the permission {permission.TEAMS} cannot add users to team")
+    if teams_id_removed is not None:
```
hakifran, line 56: same here it could be more clear if this condition is combine in one condition with the one at 132 line

iaso/tasks/profiles_bulk_update.py
```diff
@@ -65,56 +93,80 @@ def update_single_profile_from_bulk(
...
-    if location_ids_added is not None:
-        for location_id in location_ids_added:
-            if managed_org_units and len(managed_org_units) > 0 and location_id not in managed_org_units:
-                raise PermissionDenied(
-                    f"User with permission {permission.USERS_MANAGED} cannot change OrgUnits outside of their own "
-                    f"health pyramid"
-                )
-            org_unit = OrgUnit.objects.get(pk=location_id)
+    if len(roles_to_be_added) > 0:
+        for role in roles_to_be_added:
```
hakifran: As there is no any specific logic to each added user role I think profile.iaso_profile.add(*roles_to_be_added) can work well and avoid the for loop.

iaso/tasks/profiles_bulk_update.py
```diff
@@ -65,56 +93,80 @@ def update_single_profile_from_bulk(
...
+            role.iaso_profile.add(profile)
+    if len(roles_to_be_removed) > 0:
+        for role in roles_to_be_removed:
```
hakifran: Same here  profile.iaso_profile.remove(*roles_to_be_removed)

iaso/tasks/profiles_bulk_update.py
```diff
@@ -65,56 +93,80 @@ def update_single_profile_from_bulk(
...
+            role.iaso_profile.remove(profile)
+    if len(projects_to_be_added) > 0:
```
hakifran, line 154: Same here

iaso/tasks/profiles_bulk_update.py
```diff
@@ -65,56 +93,80 @@ def update_single_profile_from_bulk(
...
+        for project in projects_to_be_added:
+            project.iaso_profile.add(profile)
+    if len(projects_to_be_removed) > 0:
```
hakifran, line 158: Same here

iaso/tasks/profiles_bulk_update.py
```diff
@@ -65,56 +93,80 @@ def update_single_profile_from_bulk(
...
+        for project in projects_to_be_removed:
+            project.iaso_profile.remove(profile)
+    if len(org_units_to_be_added) > 0:
```
hakifran, line 162: Same here

iaso/tasks/profiles_bulk_update.py
```diff
@@ -65,56 +93,80 @@ def update_single_profile_from_bulk(
...
-    if location_ids_removed is not None:
-        for location_id in location_ids_removed:
-            if managed_org_units and len(managed_org_units) > 0 and location_id not in managed_org_units:
-                raise PermissionDenied(
-                    f"User with permission {permission.USERS_MANAGED} cannot change OrgUnits outside of their own "
-                    f"health pyramid"
-                )
-            org_unit = OrgUnit.objects.get(pk=location_id)
+    if len(org_units_to_be_removed) > 0:
+        for org_unit in org_units_to_be_removed:
```
hakifran: Same here

quang-le:
Failing Python tests come from the hotfix on csv user creation performance (on main)

hakifran:
> Failing Python tests come from the hotfix on csv user creation performance (on main)

@madewulf I think this is related to this https://github.com/BLSQ/iaso/pull/1607 and I reported but I thought this was related to my local configuration

hakifran (changes requested):
@quang-le when the save button is triggered without any changes in any input it create a history where the difference is only "password_updated": false, is it the way it should work? for instance if I add the user to a team there is history created on the user with "password_updated": false

iaso/api/profiles/profiles.py
```diff
@@ -372,55 +374,153 @@ def partial_update(self, request, pk=None):
...
+        username = request.data.get("user_name")
+        if not username:
+            raise ProfileError(field="user_name", detail=_("Nom d'utilisateur requis"))
+
+        existing_user = User.objects.filter(username__iexact=username).filter(~Q(pk=user.id))
+        if existing_user:
+            # Prevent from username change with existing username
+            raise ProfileError(field="user_name", detail=_("Nom d'utilisateur existant"))
+
+    def validate_user_permissions(self, request, current_account):
```
hakifran, line 427: another question is that why not filter on permission like this permission = get_object_or_404(Permission, codename__startswith="iaso_", codename=permission_codename) just to let know that a given permission not allowed to be attributed or there is a mistake.

iaso/api/profiles/audit.py
hakifran, line 1: Permissions are not needed here to be able to log the user changes?

quang-le:
> @quang-le when the save button is triggered without any changes in any input it create a history where the difference is only "password_updated": false, is it the way it should work? for instance if I add the user to a team there is history created on the user with "password_updated": false

This is a front-end problem with the "User" interface. You shouldn't be able to save when you made no changes

iaso/tasks/profiles_bulk_update.py
quang-le: The thing is I want to raise all exceptions before writing to the DB to avoid having partial DB updates and no logs for them. That's why the code is split

iaso/tasks/profiles_bulk_update.py
quang-le: Same as above

iaso/api/profiles/audit.py
quang-le: Good catch! I should've thought of that, thanks ! 🙏

hakifran (commented):
@quang-le Screenshot from 2024-09-13 10-39-02

iaso/api/profiles/audit.py
quang-le: No. It's all happening internally, while using other APIs that already control for permissions

hakifran: > No. It's all happening internally, while using other APIs that already control for permissions

You are right!!

iaso/api/profiles/profiles.py
quang-le: I don't know I just moved the existing code in another function

hakifran (approved):

PR #1612, opened 2024-09-04
[IA-3386] Prepare OUCRC API

tdethier:
Adds a web + a mobile API for `OrgUnitChangeRequestConfiguration`

Related JIRA tickets : IA-3386, IA-2896 (epic)

- [x] Are there enough tests

## Doc
Documentation in: 
- the code
- the `docs` folder 
- the Jira Epic 
- Whimsical (see Jira Epic)

## Changes

- Added 2 new endpoints in API
  - Web - `api/orgunits/changes/configs/`
      - GET - list
      - GET - retrieve
      - GET - check_availability
      - POST
      - PATCH
      - DELETE
  - Mobile -  `api/mobile/orgunits/changes/configs/`
    - GET - list
- Added tests for both endpoints

## How to test

Go to the web API, try the various available calls.

(If there are some issues with many-to-many relationships during creation, it's because of how DRF sends m2m data, but it works fine if you create OUCRCs through the Django admin panel, or if you send a payload with curl/postman/....)

## Print screen / video

image

## Notes
/


iaso/api/org_unit_change_request_configurations/views.py
```diff
@@ -0,0 +1,106 @@
...
+    OrgUnitChangeRequestConfigurationWriteSerializer,
+    OrgUnitChangeRequestConfigurationUpdateSerializer,
+    OrgUnitTypeNestedSerializer,
+    OrgUnitChangeRequestConfigurationAuditLogger,
+    ProjectIdSerializer,
+)
+from iaso.models import OrgUnitChangeRequestConfiguration, OrgUnitType, Project
+
+
+class OrgUnitChangeRequestConfigurationViewSet(viewsets.ModelViewSet):
```
mathvdh, line 21: Is there a specific reason you are not deriving from iaso.api.commom.ModelViewSet ?

mathvdh (changes requested):
At the moment there is no way to test the mobile using project's swagger : Can you add a `@swagger_auto_schema(manual_parameters=[app_id_param])` (if you do a search `app_id_param` in the project you will find it)

Screenshot 2024-09-23 at 10 49 25

iaso/api/org_unit_change_request_configurations/views.py
```diff
@@ -0,0 +1,106 @@
...
+        "org_units_editable",
+        "created_at",
+        "updated_at",
+        "created_by__username",
+        "updated_by__username",
+    ]
+    http_method_names = ["get", "post", "patch", "delete"]
+    pagination_class = OrgUnitChangeRequestConfigurationPagination
+
+    def get_queryset(self):
```
mathvdh, line 46: I think we don't want to allow a user to see all the OUCRC of all the accounts from the API.
This should be filtered by the account of the user and project.account and maybe by user.iaso_profile.project ? Not sure about the second one ...

mathvdh, line 46: I think we don't want to allow a user to see all the OUCRC of all the accounts from the API.
This should be filtered by the account of the user and project.account and maybe by user.iaso_profile.project ? Not sure about the second one ...

mathvdh: At the moment when you query `/api/orgunits/changes/configs/ `it returns even OUCRC which I'm not supposed to see as they are link to another project of another account ...

mathvdh: if you would change the `get_queryset` that would correct the fact that at the moment you can delete a OUCRC from another project/account (I think..)

iaso/api/org_unit_change_request_configurations/views.py
tdethier: I was told that the `OrgUnitChangeRequest` API was the most up to date and to do something similar to this one. It doesn't inherit from `iaso.api.common.ModelViewSet` :man_shrugging:

iaso/api/org_unit_change_request_configurations/views.py
tdethier: Yes, there should be some filtering done here. I didn't add a comment there, but I left all permission/authorization checks for another ticket

tdethier:
@mathvdh I added the swagger annotation you wanted

mathvdh (approved):
👍🏼

PR #1617, opened 2024-09-10
IA-3317: Impossible to save the GPS push from a submission while on the OU map

beygorghor:
When viewing submissions from the OU map and clicking on 'Use these coordinates,' the OU does not update. The message says 'Don't forget to save to apply the changes,' but the 'SAVE' button is disabled, making it impossible to save the change.

Related JIRA tickets : IA-3317

- [ ] Are there enough tests

## Changes

Display buttons also on filter tab to allow saving

## How to test

You need to have one instance attached to a specific org unit and having a geolocation.

Open this org unit in the dashboard, click on the ma tab , select the form you use in the the side panel on the map.
Click on the submission marker and select use this location.
Confirm the dialog.
You should now be able to save your change.

## Print screen / video

https://github.com/user-attachments/assets/350405b1-65e2-4e9a-90d5-676bb576c4bb


hakifran (approved):
LGTM

PR #1621, opened 2024-09-11
IA-3377: Fix Form accepts invalid parameter for periods

bmonjoie:
When the mobile application receives a value for the `period_type` and the sum of `periods_before_allowed` and `periods_after_allowed` is less than 1, it doesn't know how to handle that and crashes.

Related JIRA tickets: IA-3377

- [X] Are there enough tests

## Changes

Added some more validation

## How to test

Try to create a form with a period type and set 0 for the period before and after.


beygorghor (approved):
Thx for this ! Just some tests to fix please

PR #1622, opened 2024-09-11
WC2-512 API endpoint for soft deleted entities (mobile)

bramj:
The point is to allow the mobile app to fetch all uuids (and a bit more info) of the soft-deleted entities. That way the mobile app can be aware of outdated data on its end.

Related JIRA tickets : https://bluesquare.atlassian.net/browse/WC2-512

- [x] Are there enough tests

## How to test

- Have some data: entities, some duplicates
- Merge some duplicates
- Visit `http://localhost:8081/api/mobile/entities/deleted/?app_id=XXX`

## Print screen / video

image


iaso/api/mobile/entity.py
```diff
@@ -186,7 +186,7 @@ def get_queryset(self):
         if not app_id:
             raise ParseError("app_id is required")
 
-        queryset = get_queryset_for_user_and_app_id(user, app_id).filter(deleted_at__isnull=True)
+        queryset = filter_on_user_and_app_id(Entity.objects, user, app_id)
```
bmonjoie, line 187: Why remove `.filter(deleted_at__isnull=True)`?

iaso/api/mobile/entity.py
```diff
@@ -197,3 +197,52 @@ def get_queryset(self):
...
+
+    def get_serializer_class(self):
+        return DeletedMobileEntitySerializer
+
+    def get_queryset(self):
+        user = self.request.user
+        app_id = self.request.query_params.get("app_id")
+
+        if not app_id:
+            raise ParseError("app_id is required")
```
bmonjoie: Nitpicking: There is a `AppIdSerializer` that retrieves the `app_id` from the query

iaso/api/mobile/entity.py
```diff
@@ -197,3 +197,52 @@ def get_queryset(self):
...
+
+        queryset = Entity.objects_only_deleted
+        queryset = filter_on_user_and_app_id(queryset, user, app_id)
+
+        return queryset.prefetch_related("merged_to").order_by("id")
```
bmonjoie: Nitpicking: I wonder if we should not specify we only want `uuid`:
```suggestion
        return queryset.select_related("merged_to__uuid").order_by("id")
```

iaso/api/mobile/entity.py
bramj: Entity is a `SoftDeletableModel` so the filter is there by default, this extra one would just add a second `"iaso_entity"."deleted_at" IS NULL` to the generated SQL query.

iaso/api/mobile/entity.py
bramj: You can't do that directly on a field, only on a foreign key relation, you'd have to chain with `only()` to get that. (which is a good idea and I'll do that!)

About the `select_related`, it results in 1 less query since it'll do a `LEFT OUTER JOIN` on the entities table, whereas `prefetch_related` will do a second query on the IDs. However, from what I've seen so far `select_related` scales worse because of this as the amount of data grows, so I prefer to use `prefetch_related` to play safe. In any case this should be fast enough.

bmonjoie (approved):

PR #1627, opened 2024-09-12
IA-3418: HOTFIX empty payment lots

quang-le:
Users who tried to create payment lots from potential payments would sometimes end up with an empty payment lot

Related JIRA tickets : IA-3418

- [ ] Are there enough tests

## Doc
NA

## Changes

-  use `url` i.o params as `queryKey`, because the 'account' redirection would cause a second call to the APi with a race condition. This would sometimes result in 2 potential payments being created, which allowed the creation of empty payment lots down the line

## How to test

You need at least one approved change request with a non-null `created-by` field

- Open the django admin, go to potential payments
- In another tab, open the inspector and go to Potential Payments page

--> The inpsector should show only one call to the potential payments API
--> Refresh the admin: only 1new potential payment should appear

## Print screen / video

https://github.com/user-attachments/assets/d01a4f13-779b-4f3a-bad7-8c5a94323cbc

## Notes

HOTFIX on v1.243

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/payments/hooks/requests/useGetPotentialPayments.tsx
+++ b/hat/assets/js/apps/Iaso/domains/payments/hooks/requests/useGetPotentialPayments.tsx
@@ -8,7 +8,13 @@ import { apiDateFormat, formatDateString } from '../../../../utils/dates';
 
 const apiUrl = '/api/potential_payments/';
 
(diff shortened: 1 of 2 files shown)


beygorghor (approved):

beygorghor:
Deployed in Prod

PR #1632, opened 2024-09-13
IA-3426 show or unit, project and  user role names in user history log

quang-le:
The UI was showing id, which is not so human friendly

Related JIRA tickets : IA-3426

- [x] Are there enough tests

## Doc
NA

## Changes

- Add serializer for `retrieve` method of `/userlogs` endpoint (`ProfileLogsViewset`)
- Make `remove_user_role_name_prefix` method static (`UserRole` model)
- Update API call in front end
- Update `UserLogValue` to extract name from API response
- Add some styling to get links text to display in white if value has been changed (and background is then blue)

## How to test

In iaso> user> management: create or update a user. Change project, location, user role or all 3

Go to users > history and check the latest entry

## Print screen / video

Screenshot 2024-09-13 at 17 08 05


iaso/api/profiles/profile_logs.py
```diff
@@ -137,10 +139,53 @@ def get_fields_modified(self, modification):
...
 class ProfileLogRetrieveSerializer(serializers.ModelSerializer):
+    past_value = serializers.SerializerMethodField(read_only=True)
+    new_value = serializers.SerializerMethodField(read_only=True)
+
     class Meta:
         model = Modification
         fields = ["id", "created_at", "user", "source", "new_value", "past_value", "object_id", "content_type"]
 
+    def get_past_value(self, modification):
+        logged_org_units = modification.past_value[0]["fields"]["org_units"]
```
tdethier, line 159: I created some users, I tried to check the diff and got an error: 

```  File "/home/thibault/dev/iaso/iaso/iaso/api/profiles/profile_logs.py", line 150, in get_past_value
    logged_org_units = modification.past_value[0]["fields"]["org_units"]
                       ~~~~~~~~~~~~~~~~~~~~~~~^^^
IndexError: list index out of range

Comment (tdethier) [reply iaso/api/profiles/profile_logs.py:159]: In the DB: `past_value` = `[]`

Comment (tdethier) [reply iaso/api/profiles/profile_logs.py:159]: So the issue seems to be that, in the diff for creation, `past_value`  is always empty

File: iaso/tests/api/test_profile_logs.py [hunk 2]
```diff
@@ -560,3 +696,68 @@ def test_filters(self):
 
         results = data["results"]
         self.assertEquals(results, [])
+
+    def test_retrieve(self):
```
tdethier, line 704: I would add more tests for retrieve:

- retrieve on created instance without update (to spot the crash I mentioned before)
- retrieve without authentication
- retrieve without permissions
- retrieve for unknown log ID

iaso/api/profiles/profile_logs.py
```diff
@@ -137,10 +139,53 @@ def get_fields_modified(self, modification):
...
+        past_value_copy[0]["fields"]["user_roles"] = user_roles_names_and_ids
+
+        return past_value_copy
+
+    def get_new_value(self, modification):
+        logged_org_units = modification.new_value[0]["fields"]["org_units"]
+        logged_projects = modification.new_value[0]["fields"]["projects"]
+        logged_user_roles = modification.new_value[0]["fields"]["user_roles"]
+
+        org_units_names_and_ids = OrgUnit.objects.filter(id__in=logged_org_units).values("id", "name")
```
tdethier: Since fetching IDs and names is the same for both past and new value, I would put the fetching and transformation in a utility function and call it in both `get_xxx_value()`

hat/assets/js/apps/Iaso/domains/users/history/UserLogValue.tsx
```diff
@@ -38,17 +58,22 @@ export const UserLogValue: FunctionComponent<Props> = ({ fieldKey, value }) => {
             case 'org_units':
```
tdethier, line 58: When you remove a user role, you get a `--` instead in the After column to show that there's nothing. I would do the same with the location I guess 
image

beygorghor: i would display a placeholder in location too, has this is the general behaviour a bit everywhere

beygorghor (approved):

tdethier (approved):
LGTM :+1:

PR #1638, opened 2024-09-17
POLIO-1662 Supply Chain: VRF & DG Approval include options of 'missing' and 'not required'

mathvdh:
Supply Chain: VRF & DG Approval include options of 'missing' and 'not required'

Related JIRA tickets : POLIO-1662

- [x] Are there enough tests

## Changes

- Add a “type” on the VRF model with default “Normal”, with other options “Missing” and “Not Required”.
- When its Not Required, the optional fields of the VRF will be hidden, as well as the the Pre Alerts and VAR tabs.
- Field Quantity Ordered In Doses must become optional (Default to 0)

## How to test

See video below, but basically go to Vaccine Supply Chain and create a new entry or edit an existing entry and try to play with the VRF Type field

## Print screen / video

https://github.com/user-attachments/assets/6bdfa0d1-9d5d-41bd-ba69-178a21e34859


tdethier (commented):
(I didn't understand the context of all of this, so I just looked at the code and tested the feature.)

I managed to update existing data and to change the VRF type.
But I tried to create new data with VRF type `Missing` or `Not Required` and I couldn't do it, because the save button is disabled on both cases. Is this a bug or done on purpose @mathvdh? 

image

mathvdh:
@tdethier it's not normal, I'll have a look into it, thanks

tdethier (approved):
LGTM :+1:

PR #1639, opened 2024-09-17
WC2-527 Add Django admin filters to detect data issues

bramj:
Main additions:
- Show more data on the entities
- New filters on the entities:
	- attributes_id None or not?
	- are the attributes "valid"? This means, if the entity is not deleted, the attributes should not be deleted. Is there a file on the attributes?
	- are there any duplicated uuids?
- New filter on the instances for duplicated uuids
- General performance improvements on several pages, notably on the instances page.

Related JIRA tickets : https://bluesquare.atlassian.net/browse/WC2-527

## How to test

Have some data in your db, look at the Django admin. You can change some things like the uuid in your DB if you want to test the filters.

## Print screen / video

image

### DIFF
--- a/hat/api_import/admin.py
+++ b/hat/api_import/admin.py
@@ -7,4 +7,16 @@ from .models import APIImport
 class APIImportAdmin(admin.GeoModelAdmin):
     date_hierarchy = "created_at"
     search_fields = ("json_body", "headers", "exception")
-    list_filter = ("has_problem", "import_type")
+    list_display = (
+        "id",
+        "import_type",
+        "has_problem",
+        "user",
+    )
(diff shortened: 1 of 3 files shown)


mathvdh (commented):
All good to me,

Except that I have a visibility problem on the merged_to in admin :

Screenshot 2024-09-18 at 11 25 24

I it only because of my local data ??

bramj:
@mathvdh good catch! It's actually because the `name` on the entity is usually blank. In any case it's best to not use a dropdown here for performance reasons. I've just changed it in the last commit:
image

mathvdh (approved):
👍🏼

PR #1640, opened 2024-09-18
WC2-515: Integrate CIAM with multi-account setup

bmonjoie:
When connecting with CIAM, Iaso only connects you to the configured default account.
With the new multi-account setup, we need to be able to authenticate to different accounts based on the app_id

Related JIRA tickets: WC2-515

- [ ] Are there enough tests

## Changes

When exchanging the CIAM token for an Iaso token, Iaso will look for an `app_id` parameter, allowing it to connect to the correct account.

## How to test

We'll have to deploy it on DEV and test it there.

### DIFF
--- a/plugins/wfp_auth/views.py
+++ b/plugins/wfp_auth/views.py
@@ -24,7 +24,8 @@ from requests import RequestException, HTTPError
 from rest_framework.decorators import api_view, authentication_classes, permission_classes
 from rest_framework_simplejwt.tokens import RefreshToken  # type: ignore
 
-from iaso.models import Account, Profile
+from iaso.api.serializers import AppIdSerializer
+from iaso.models import Account, Profile, Project
 from .provider import WFPProvider
 
 logger = getLogger(__name__)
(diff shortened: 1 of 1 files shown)


plugins/wfp_auth/views.py
```diff
@@ -111,7 +112,11 @@ def complete_login(self, request, app, token: str, response) -> SocialAccount:
             email = extra_data["sub"].lower().strip()
         # the sub is the email, wfp verify it so let's trust this
         uid = extra_data["sub"].lower().strip()
-        account = Account.objects.get(name=self.settings["IASO_ACCOUNT_NAME"])
+        app_id = AppIdSerializer(data=self.request.query_params).get_app_id(raise_exception=False)
+        if app_id:
+            account = Project.objects.get(app_id=app_id).account
```
madewulf: In fact, at this point, we should do 
project = get_object_or_404(Project, app_id=app_id)
account = project.account 

so that the API returns a 404 and not a 500 if the provided app_id is wrong

madewulf: (sorry, I know, I gave the wrong code before)

madewulf: The import to do: from django.shortcuts import get_object_or_404

bmonjoie: I have pushed the requested change.

madewulf (approved):

beygorghor (approved):

mathvdh (approved):
👍🏼

PR #1641, opened 2024-09-18
[POLIO-1680] Allowing any user to download the scopes of the campaigns

madewulf:
Related JIRA tickets : POLIO-1680


beygorghor (approved):
I'll remove the import of `DisplayIfUserHasPerm` then

PR #1642, opened 2024-09-18
WC2-511 Improve the handling of soft-deleted and/or merged entities

bramj:
Quite a few changes here:

### Bulk upload: correcly handle incoming data for soft-deleted and merged entities

Upon an incoming entity, there are several options:

1. New entity: regular flow
2. Existing entity:
	a. Non-deleted entity: add new forms, update reference form if timestamp more recent
	b. Soft-deleted entity X:
		- If because of merge:
			1. Find the "active" entity and add new forms on this entity
			2. If X's ref form is updated, also update the "active" entities reference form
		- If not because of merge: Add/update forms with status deleted

*Some clarification with an example:

1. Consider entity A,  B and C
3. We merge A and B into D
4. Now we merge C and D into E
5. When we receive data from the mobile app for entity A, we:
	- Add data to A (everything with status soft-deleted)
	- Add data to D (everything with status soft-deleted)
	- Add new submissions to E
	- Update any forms that have a newer timestamp
	- If a reference form is updated on A, then we also update E's reference form (if the timestamp is newer). (By update we mean the XML file, data such as the UUID)
	
### Audit logging

Because things get complex, it's important we can figure out where data changes came from. To this end:
  - Correctly log instance modification on entity merge. We didn't keep an audit trail on these actions. Refactored a bit to put the logging in the instances API instead of directly on the `soft_delete()` method. There can be multiple sources of soft delete, this allows us to be more granular.
  - Add audit logging to bulk upload form updates (and differentiate between a regular one and a merged entity, can be useful for displaying advance version histories in the future)

Related JIRA tickets : https://bluesquare.atlassian.net/browse/WC2-511

## How to test

Complicated, but to really test this you need to:
1. Get a Iaso app with bulk upload enabled (e.g. the WFP CODA2 app)
2. Enable developer mode and change the server URL to your localhost (e.g. with ngrok)
3. Have your localhost setup with some entity types etc (with setuper it works well)
4. Create 2 entities with the same name on your mobile device
5. Launch the duplicate detection algo
6. Merge the 2 entities, your mobile device will not be aware of this
7. Add a form on your phone's entity (either one of the two) and sync to server
8. You should now see that the file was correctly attached to the merged entity

### DIFF
--- a/hat/assets/js/apps/Iaso/components/stars/StarSvgComponent.js
+++ b/hat/assets/js/apps/Iaso/components/stars/StarSvgComponent.js
@@ -7,6 +7,7 @@ function StarSvg(props) {
         <svg
             style={{
                 width: starWidth,
+                height: starWidth,
             }}
         >
             <g data-name="bg-star" className="bg-star">
--- a/hat/audit/models.py
+++ b/hat/audit/models.py
@@ -25,6 +25,7 @@ PAYMENT_LOT_API = "payment_lot_api"
 ORG_UNIT_CHANGE_REQUEST_API = "org_unit_change_request_api"
 DJANGO_ADMIN = "django_admin"
 BULK_UPLOAD = "bulk_upload"
+BULK_UPLOAD_MERGED_ENTITY = "bulk_upload_merged_entity"
 ENTITY_DUPLICATE_MERGE = "entity_duplicate_merge"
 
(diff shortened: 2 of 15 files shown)


tdethier (approved):
LGTM :+1:

PR #1644, opened 2024-09-20
WC2-503 Fix: Avoid deleted entities in duplicate analysis

bramj:
The algorithm already correctly excludes soft-deleted entities. The issue is when an entity was soft-deleted, its pending duplicate entities were not, resulting in errors upon merging.

**At the moment, the only way to soft-delete an entity is via the Django admin, so that's where we corrected this.**

Related JIRA tickets : https://bluesquare.atlassian.net/browse/WC2-503

## The fix

Override Entity Django admin delete: The Django admin delete simply called `delete()` on the entity. This commit overrides the default Django admin function and:

- soft delete the entity
- soft delete its attached form instances
- delete potential EntityDuplicates

Also: 
- Removed other duplcite algos from dropdown in the frontend. At the moment, only levenshtein is supported.
- Send a Sentry when merging soft-deleted entities. This shouldn't happen.

## How to test

Hit "Delete" button for an entity in the Django admin.

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/entities/constants.ts
+++ b/hat/assets/js/apps/Iaso/domains/entities/constants.ts
@@ -1,5 +1,5 @@
 export const ALGORITHM_DROPDOWN = [
-    { label: 'namesim', value: 'namesim' },
     { label: 'levenshtein', value: 'levenshtein' },
-    { label: 'invert', value: 'invert' },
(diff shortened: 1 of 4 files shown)


tdethier (approved):
LGTM 

Thanks for the small comments in the code :+1:

PR #1646, opened 2024-09-23
IA-3146-bis: adding support for webp image

beygorghor:
webp are filtered out while filtering attachment with image only

Related JIRA tickets : IA-3146

- [ ] Are there enough tests

## Changes

added webp to list of possible extensions

## How to test

You need  a webp image and create one instance files linked to an instance that is linked to an org unit.
YOu can do it using the admin or with enketo and a form using images.


bramj (approved):
LGTM :+1:

PR #1650, opened 2024-09-24
POLIO-1687 Filter by campaign and round date

mathvdh:
The APIs for the Vaccine Dashboard are returning Form A and Destruction Report dates from other campaigns in Same Country/Vaccine. This adds filtering by Campaign and Campaign Round Dates

Related JIRA tickets : POLIO-1687

- [x] Are there enough tests

## Changes

Additional filtering in the APIs per campaign (when available) and campaign round dates.

## How to test

A new campaign in the future should not return the FormA and Destruction reports dates of pre-existing campaigns

### DIFF
--- a/plugins/polio/api/dashboards/supply_chain.py
+++ b/plugins/polio/api/dashboards/supply_chain.py
@@ -1,4 +1,4 @@
-from django.db.models import Q
+from django.db.models import Subquery, OuterRef
 from rest_framework import serializers
 
 from iaso.api.common import ModelViewSet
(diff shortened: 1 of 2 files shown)


plugins/polio/api/dashboards/supply_chain.py
```diff
@@ -54,18 +58,71 @@ def get_stock_in_hand(self, obj, vaccine_stock):
...
+
+        for campaign in campaigns_after_last_round:
+            if vaccine_stock.vaccine in campaign.vaccines:
+                # Do something if the vaccine matches
+                next_campaign_start_date = (
+                    Round.objects.filter(campaign=campaign)
+                    .order_by("started_at")
+                    .values_list("started_at", flat=True)
+                    .first()
+                )
```
quang-le, line 109: This will not always be correct. You need to check if the campaign has separate scopes per round. If so, you need to check the vaccines for the round.

I think the `vaccine_names` method on the `Round` model can be of help

plugins/polio/api/dashboards/supply_chain.py
```diff
@@ -54,18 +58,71 @@ def get_stock_in_hand(self, obj, vaccine_stock):
...
+
+        earliest_round_start_date = (
+            Round.objects.filter(campaign=OuterRef("pk")).order_by("started_at").values("started_at")[:1]
+        )
+
+        distinct_campaign_ids = (
+            Campaign.objects.filter(
+                country=vaccine_stock.country,
+                rounds__started_at__gt=first_round_start_date,
+                rounds__number=1,
```
quang-le: You may have campaigns with a round 0

plugins/polio/api/dashboards/supply_chain.py
```diff
@@ -54,18 +58,71 @@ def get_stock_in_hand(self, obj, vaccine_stock):
...
+
+        campaigns_after_last_round = (
+            Campaign.objects.filter(id__in=Subquery(distinct_campaign_ids))
+            .annotate(earliest_round_start=Subquery(earliest_round_start_date))
+            .order_by("earliest_round_start")
+        )
+
+        next_campaign_start_date = None
+
+        for campaign in campaigns_after_last_round:
```
quang-le, line 101: would it not be better to order `campaigns_after_last_round` by descending order and break the loop once we have a value for `next_campaign_start_date`?

Also, since you already have the `Campaign` instance, you can probably prefetch the `rounds` i.o filtering on all `Round`s (dunno if it's really more performant)

quang-le (changes requested):
Checkout the edge case with Round 0 and the scopes per round

plugins/polio/api/dashboards/supply_chain.py
quang-le: my mistake

quang-le (approved):

PR #1656, opened 2024-09-26
WC2-532 Fix: Emoji parse error on entity merge

bramj:
When entities are merged, the reference form is merged, and the other forms are copied over to the new merged instance.

This PR makes 2 changes:
- When there's a parse error when merging/copying forms, let it blow up in a 500. The catching of the error would result in a merged entity being created regardless, but with missing forms.
- Handle emojis in the XML forms. Reused code from `iaso/models/base.py`.

Related JIRA tickets : https://bluesquare.atlassian.net/browse/WC2-532

## How to test

I'd say a reading of the code is enough. There is a unit test and I'll test this on real form data on QA before moving to prod.


mathvdh (approved):
👍🏼

PR #1661, opened 2024-09-30
IA-3456 improve get or create for orgunit (duplicate on uuid)

mestachs:
To minize duplicate and be "ready" when we will introduce the unique index

Related JIRA tickets : IA-3456

- [ ] Are there enough tests

## Doc

## Changes

see https://github.com/BLSQ/iaso/pull/1652

## How to test

see https://github.com/BLSQ/iaso/pull/1652
but the mobile unit test does that

for a more "real life" reproduction : https://gist.github.com/mestachs/2cc50bf310996d14cc48623a05d42caf

note that we have historical data so we will need an intermediate cleanup
create the a temporary index as in the _notes.md
then really cleanup "old data" and all environments (harder)

## Print screen / video

## Notes


iaso/api/instances.py
```diff
@@ -686,6 +686,7 @@ def instance_logs(self, request, pk=None, logId=None):
 
 def import_data(instances, user, app_id):
     project = Project.objects.get_for_user_and_app_id(user, app_id)
+    default_version = project.account.default_version
```
bramj, line 774: I think it's possible to not have a default version on an account. I think that situation would probably blow up anyways, but not sure if we should consider it here :thinking:

iaso/api/mobile/org_units.py
```diff
@@ -337,11 +337,19 @@ def bbox_merge(a: Optional[tuple], b: Optional[tuple]) -> Optional[tuple]:
...
+    "The uuid is supposed to be unique per source_version"
+    return OrgUnit.objects.filter(version=default_version)
+
+
 def import_data(org_units, user, app_id):
     new_org_units = []
     project = Project.objects.get_for_user_and_app_id(user, app_id)
     org_units = sorted(org_units, key=get_timestamp)
 
+    project = Project.objects.get_for_user_and_app_id(user, app_id)
```
bramj, line 350: Duplicate of line 347?

iaso/api/instances.py
mathvdh: @beygorghor I think we already had to code special cases when there is no default_version so it can happen right ?

mathvdh (commented):
See comments from Bram, otherwise it seems ok to me

iaso/api/instances.py
beygorghor: Yes indeed, the frontend allows no default version

bramj:
@mestachs In order to deblock this PR, do you think it would be a good idea to add a conditional unique index when the `created_at` is greater than today?

bramj:
Conditional indexes to add manually on big servers:

```SQL
CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS iaso_orgunit_uuid_version_idx_unicity_newer ON iaso_orgunit (uuid, version_id)
WHERE
	uuid IS NOT NULL AND version_id IS NOT NULL AND created_at >= '2025-01-14';

CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS iaso_entity_uuidx_unicity_newer ON iaso_entity (uuid)
WHERE
	uuid IS NOT NULL AND created_at >= '2025-01-14';

CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS iaso_instance_uuidx_unicity_newer ON iaso_instance (uuid)
WHERE
	uuid IS NOT NULL AND created_at >= '2025-01-14';
```

bramj:
@mathvdh @madewulf @bmonjoie Do you see any issues with manually adding the indexes in my [above comment](https://github.com/BLSQ/iaso/pull/1661#issuecomment-2587005834)?

Since UUIDs for instances and entities are generated on the mobile app I don't think they should be unique only per project or something like that. Making them real UUIDs feels like the best option to me. For org units we're enforcing uniqueness only per version since @mestachs discovered that import/export/reimport geopackage is putting the same UUID in 2 source versions.

mathvdh:
@bramj I don't see any problem with it (except for duplicate line 350 as commented above). Is the plan to cleanup duplicate uuids and apply the index unconditionally after that ? Or what's the plan ?

bramj:
> @bramj I don't see any problem with it (except for duplicate line 350 as commented above). Is the plan to cleanup duplicate uuids and apply the index unconditionally after that ? Or what's the plan ?

Ideally, yes that would be the plan, but to be honest not sure how doable that is.

PR #1665, opened 2024-10-01
POLIO-1665 broken should have usable out and unusable in

mathvdh:
Broken stock variation should have usable out and unusable in.

Related JIRA tickets : POLIO-1665

- [x] Are there enough tests

## Changes

In Vaccine Stock Management, in stock variation, when you add/edit an incident report. When the Stock Correction field is "broken" it should allow to enter both a Vials In and a Vials Out value (default to 0). Then this should be taken into account by the backend for other screens.

## How to test

Add or edit an incident report and choose "Broken" for the Stock Variation type.

## Print screen / video

Before : 

Screenshot 2024-10-02 at 09 53 12

After : 

Screenshot 2024-10-02 at 09 53 35


quang-le (approved):
Yippie Kay Yay!

PR #1681, opened 2024-10-04
WC2-515: Try to fix "'WSGIRequest' object has no attribute 'query_params'"

bmonjoie:
We cannot use `AppIdSerializer` because we are in a method **before** `dispatch` and therefore `self.request.query_params` is not set yet. Using `request.query_params` results in:
`'WSGIRequest' object has no attribute 'query_params'`.

According to this [answer](https://stackoverflow.com/questions/42892602/django-rest-framework-wsgirequest-object-has-no-attribute-query-params#comment121400693_42892602), calling `request.GET` should work. At least, it shouldn't crash anymore.

Related JIRA tickets : WC2-515

- [ ] Are there enough tests

## Changes

I have replaced `AppIdSerializer` by `request.GET`.

## How to test

🤷‍♂️

### DIFF
--- a/plugins/wfp_auth/views.py
+++ b/plugins/wfp_auth/views.py
@@ -25,7 +25,7 @@ from requests import RequestException, HTTPError
 from rest_framework.decorators import api_view, authentication_classes, permission_classes
 from rest_framework_simplejwt.tokens import RefreshToken  # type: ignore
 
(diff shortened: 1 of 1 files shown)


bramj (approved):
LGTM and working on DEV :+1: :tada:

PR #1684, opened 2024-10-07
WC2-580 Form uploads: don't break over duplicate uuids

bramj:
https://bluesquare.atlassian.net/browse/WC2-580

We don't want the upload to break because of a duplicate UUID. Usually it's just because one of them is deleted. Sometimes we don't have DB access and we can't do a hard delete, or change the UUID of one of them (it's a read-only field).

So this PR tests a few scenarios where there are multiple entities with the same UUID:

- 1 active, 1 deleted -> take the active one
- 0 active, multiple deleted -> take the most "correct" deleted one
- More than 1 active, n deleted -> take the most "correct" one and log Sentry exception

### DIFF
--- a/iaso/api/instances.py
+++ b/iaso/api/instances.py
@@ -805,6 +805,11 @@ def import_data(instances, user, app_id):
 
 
 def _entity_correctness_score(entity):
+    """
+    A small function that allows sorting entities to pick the right one for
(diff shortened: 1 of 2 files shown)


mathvdh (approved):
Don't really know the background of this but it seems ok to me

PR #1685, opened 2024-10-07
POLIO-1685 missing var in supplychain list

quang-le:
Some VAR didn't appear in the supplychain list view, because PO numbers didn't match with pre-alert

Related JIRA tickets :POLIO-1685

- [x ] New translations have been added or updated if new strings have been introduced in the frontend
- [x ] My migrations file are included
- [x] Are there enough tests

## Doc
In the code

## Changes

- Backend:
    - Add validation on both PreAlert and ArrivalReport serializers to check the po_number is not prefixed
    - Add migration to remov "PO" prefix from existing data
    - Change Supplychain list API to show arrival reports that have a PO number that doesn't match a PreAlert (eg: beacuse there's no PreAlert)
    
- Front-end:
    - Add `yup` validation to check that po number is not prefixed and that it is unique for the current VRF + add error messages
    - Remove already selected PO numbers from arrival report PO number dropdown
    - Enable error messages on Select field. Requires  https://github.com/BLSQ/bluesquare-components/pull/165 from bluesquare components

## How to test

- update package.json to use bluesquare-components development branch and run `npm update bluesquare-components`
- Go to polio > Vaccine module > Supply chain
- Create/Edit a supply chain entry so that:
    - There's a vaccine arrival report with a Po number that doesn't match a pre alert --> it should appear when you return to the main list view
    - Try to input a PO number starting with "PO" --> The input should be in error

## Print screen / video

https://github.com/user-attachments/assets/9b8896c6-4be0-42c0-95fd-893d8f00cb5c

## Notes

⚠️  Depends on   https://github.com/BLSQ/bluesquare-components/pull/165


plugins/polio/migrations/0198_remove_po_prefix.py
```diff
@@ -0,0 +1,27 @@
+# Generated by Django 4.2.13 on 2024-10-04 13:18
+
+from django.db import migrations
+
+
+def remove_prefix_from_objects(objects):
+    for object in objects:
```
mathvdh, line 7: Not sure how robust this is. it will only work for PO-XXX nothing else .. Not sure what we have in DB

mathvdh (approved):
Except for my comment it's all good

plugins/polio/migrations/0198_remove_po_prefix.py
quang-le: That's the only thing we have

PR #1687, opened 2024-10-07
IA-3446 Org Unit Type restrictions (frontend)

beygorghor:
In our use case, "admin" users should be able to edit/add new OUs at level 6 for instance (SSC, CAC) but should clearly not be allowed to modify levels 1 to 5 (even within their geographic permission).

Related JIRA tickets : IA-3446

- [ ] Are there enough tests

## Changes

- Added `editable_org_unit_types` to the profile API and frontend.
- Restricted selection to valid org unit types when editing users and setting `editable_org_unit_types`.
- Allowed changes only to writable org unit types when editing individual org units.
- Disabled multi-selection for bulk editing to ensure only writable org unit types can be selected.
- Limited default image edits to writable org unit types.
- Restricted linking of instances to org units based on writable types (requires `SHOW_LINK_INSTANCE_REFERENCE` feature flag, and the instance must be of the org unit's reference form).

## Testing Steps

1. Ensure you have a proper org unit type hierarchy set up in your dataset.
2. Create a new user and assign them write access for org units, along with read and write permissions for forms and submissions.
3. While editing the user, go to the last tab and select some org unit types as editable.
4. Log in with the new user account:
    - Search for all org units and verify that only the selected org unit types are available for bulk updates.
    - Open the bulk update dialog and attempt to change org unit types; you should only be able to select the editable types.
    - Close the dialog.
5. Edit an org unit of an editable type:
    - Confirm that you can modify all details, including the default image (only editable types should be available in the list).
6. Edit an org unit of a non-editable type:
    - Verify that you cannot change any org unit information or update the default image.
7. Find an instance of a reference form for an org unit:
    - Open the instance's details and ensure that the option to link the instance to the org unit is available only if the org unit type is editable.

## Print screen / video

Uploading Screen Recording 2024-10-08 at 14.18.09.mov…

## Notes

- This is going along with this [PR](https://github.com/BLSQ/bluesquare-components/pull/166)
- Should we limit access to org unit types edition/creation for an user that can only create certain org unit types
- Should we filter out non editable types from change config creation/edition
- Stil need to write cypress tests to cover those cases


hat/assets/js/apps/Iaso/utils/usersUtils.ts
```diff
@@ -108,3 +111,32 @@ export const useHasNoAccount = (): boolean => {
     const currentUser = useCurrentUser();
     return Boolean(currentUser && !currentUser.account);
 };
+
+export const useGetUserHasWriteTypePermission = (): ((
+    orgUnitTypeId?: number,
+) => boolean) => {
+    const currentUser = useCurrentUser();
+    return (orgUnitTypeId?: number) => {
```
quang-le, line 119: return a useCallback, so you can use it in your components without having to wrap it

quang-le: also I would rename as `useCheckUserHasWriteTypePermission`,  because `useGet...`is usually an API call

hat/assets/js/apps/Iaso/utils/usersUtils.ts
```diff
@@ -108,3 +111,32 @@ export const useHasNoAccount = (): boolean => {
...
+                    currentUser.editable_org_unit_type_ids?.length === 0 ||
+                    (orgUnitTypeId &&
+                        currentUser.editable_org_unit_type_ids?.includes(
+                            orgUnitTypeId,
+                        ))),
+        );
+    };
+};
+
+export const useGetUserHasWritePermissionOnOrgunit = (
```
quang-le: same re: name

quang-le (changes requested):
Only some nitpicks about the code.

Kind reminder: if the feature is merged on staging but not ready for prod, we can use the `SHOW_DEV_FEATURES` feature flag to keep it away from users

About the feature itself:

- The dropdown in users is confusing if we have more than 1 datasource

Screenshot 2024-10-11 at 12 01 57

- It seems to me some permissions should be mutually exclusive:
    - Org unit types + org unit types restrictions: a user restricted to an org unit type but with the general org unit types permission can overwrite the sub org unit types to display of any org unit type

https://github.com/user-attachments/assets/4622dc3b-ef1a-472a-aec0-d54da885dca0


- Org Unit Change request config: user can edit configs for org unit types they don't have edition rights to



- Registry: links to forbidden types are greyed out but active

https://github.com/user-attachments/assets/2908e99f-bc5c-459a-b635-174f9c309eb5



- Forms: user with type restrictions can assign all types to any form


- Potential Payments/Payments: user can see all potential payments, but cannot check the associated change requests if type/location restricted. The payments should be filtered at API level. [JIRA ticket added](https://bluesquare.atlassian.net/browse/IA-3540)


https://github.com/user-attachments/assets/a7216b31-02b2-4c4a-837d-8b0e46cc08e6


-  User can add org unit type write access to to types of higher level than his location 

https://github.com/user-attachments/assets/8edd71ef-145c-4080-8ba8-40aa7cdec337


- User modal: tab text looks funny in French
Screenshot 2024-10-11 at 15 15 40

PR #1691, opened 2024-10-09
IA-3445 Org Unit Type restrictions in `Profile` (backend)

kemar:
Allow to edit/create org units based on a set of org unit types.

Related JIRA tickets : [IA-3445](https://bluesquare.atlassian.net/browse/IA-3445)

## Changes

- optimise the number of queries in `validate_editable_org_unit_types`
- add `Profile.has_org_unit_write_permission()`
- for users without write perms, prevent:
    - creation of org units
    - update of org units
    - bulk update of org units
- handle `Profile.editable_org_unit_types` in the mobile org unit change request configuration API

## How to test

See unit tests.

## Notes

This PR:

- builds on #1687 to ensure that the business logic is implemented in the backend
- is inspired by #1658 (but do some things differently)

The scope of this PR is limited to the `Profile` model. We'll work on the `UserRole` model in another one.

[IA-3445]: https://bluesquare.atlassian.net/browse/IA-3445?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iaso/tests/api/test_org_units_bulk_update.py
+++ b/iaso/tests/api/test_org_units_bulk_update.py
@@ -260,7 +260,7 @@ class OrgUnitsBulkUpdateAPITestCase(APITestCase):
 
         response = self.client.post(
             f"/api/tasks/create/orgunitsbulkupdate/",
-            data={"select_all": True, "validation_status": m.OrgUnit.VALIDATION_VALID},
+            data={"select_all": True, "validation_status": m.OrgUnit.VALIDATION_REJECTED},
             format="json",
         )
         self.assertJSONResponse(response, 201)
@@ -271,10 +271,6 @@ class OrgUnitsBulkUpdateAPITestCase(APITestCase):
         # Run the task.
         self.runAndValidateTask(task, "SUCCESS")
 
-        for org_unit in [self.jedi_squad_endor_1, self.jedi_squad_endor_2]:
-            org_unit.refresh_from_db()
-            self.assertEqual(org_unit.validation_status, m.OrgUnit.VALIDATION_VALID)
-
         self.assertEqual(2, am.Modification.objects.count())
 
         task.refresh_from_db()
@@ -282,6 +278,14 @@ class OrgUnitsBulkUpdateAPITestCase(APITestCase):
         self.assertIn("3 skipped", task.progress_message)
 
         self.yoda.iaso_profile.editable_org_unit_types.clear()
+
+        for org_unit in [self.jedi_squad_endor_1, self.jedi_squad_endor_2]:
+            org_unit.refresh_from_db()
+            self.assertEqual(org_unit.validation_status, m.OrgUnit.VALIDATION_REJECTED)
+
+        for org_unit in [self.jedi_council_corruscant, self.jedi_council_endor, self.jedi_council_brussels]:
+            org_unit.refresh_from_db()
+            self.assertEqual(org_unit.validation_status, m.OrgUnit.VALIDATION_VALID)
 
     @tag("iaso_only")
     def test_org_unit_bulkupdate_select_all_with_search(self):
--- a/iaso/tests/api/test_org_units_bulk_update.py
+++ b/iaso/tests/api/test_org_units_bulk_update.py
@@ -247,7 +247,7 @@ class OrgUnitsBulkUpdateAPITestCase(APITestCase):
         self.assertEqual(5, am.Modification.objects.count())
 
     @tag("iaso_only")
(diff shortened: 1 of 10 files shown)


iaso/tests/api/test_org_units_bulk_update.py
```diff
@@ -246,6 +246,43 @@ def test_org_unit_bulkupdate_select_all(self):
...
+        data = response.json()
+        task = self.assertValidTaskAndInDB(data["task"], status="QUEUED", name="org_unit_bulk_update")
+        self.assertEqual(task.launcher, self.yoda)
+
+        # Run the task.
+        self.runAndValidateTask(task, "SUCCESS")
+
+        for org_unit in [self.jedi_squad_endor_1, self.jedi_squad_endor_2]:
+            org_unit.refresh_from_db()
+            self.assertEqual(org_unit.validation_status, m.OrgUnit.VALIDATION_VALID)
```
tdethier: They already were in `VALIDATION_VALID` so it's hard to determine if the bulk update happened. I'd make the bulk update to `VALIDATION_NEW` or `VALIDATION_REJECTED`

iaso/tests/api/test_org_units_bulk_update.py
```diff
@@ -246,6 +246,43 @@ def test_org_unit_bulkupdate_select_all(self):
...
+
+        self.assertEqual(2, am.Modification.objects.count())
+
+        task.refresh_from_db()
+        self.assertIn("2 modified", task.progress_message)
+        self.assertIn("3 skipped", task.progress_message)
```
tdethier, line 278: Since there are 5 `OrgUnit` in the test setup, should we check also the status of the other 3?

iaso/tests/api/test_profiles.py
```diff
@@ -296,7 +296,7 @@ def test_profile_list_read_only_permissions(self):
         """GET /profiles/ with auth (user has read only permissions)"""
 
         self.client.force_authenticate(self.jane)
-        with self.assertNumQueries(17):
+        with self.assertNumQueries(11):
```
tdethier, line 299: I'm gonna write this comment here even though it's not related to this specific line: unfortunately I don't see any tests on user creation/update with `editable_org_unit_type_ids` :frowning_face:

tdethier (changes requested):
The main issue is that I feel like some part of the feature is missing. Write access restriction is working on `OrgUnits`, but there is no check during user creation/update process that the given list of `OrgUnitType` IDs is valid for the user that submits the request. So, for instance:
- a restricted user could create another user without restrictions
- a restricted user could remove access to an `OrgUnitType`  from another user, even though they don't have this access themselves

If this feature is urgent, we can go ahead, but we must make another ticket for that and handle it quickly, because right now the restriction is easy to bypass. What do you think @beygorghor?

I also left a few comments about tests.

Otherwise, I'd say good job, as usual :+1:

iaso/tests/api/test_org_units_bulk_update.py
kemar: Fixed in https://github.com/BLSQ/iaso/pull/1691/commits/c42d1cc9237799dbb581cc937bc18eb54c1c8f68

iaso/tests/api/test_org_units_bulk_update.py
kemar: Fied in https://github.com/BLSQ/iaso/pull/1691/commits/c42d1cc9237799dbb581cc937bc18eb54c1c8f68

beygorghor:
> The main issue is that I feel like some part of the feature is missing. Write access restriction is working on `OrgUnits`, but there is no check during user creation/update process that the given list of `OrgUnitType` IDs is valid for the user that submits the request. So, for instance:
> 
> * a restricted user could create another user without restrictions
> * a restricted user could remove access to an `OrgUnitType`  from another user, even though they don't have this access themselves
> 
> If this feature is urgent, we can go ahead, but we must make another ticket for that and handle it quickly, because right now the restriction is easy to bypass. What do you think @beygorghor?
> 
> I also left a few comments about tests.
> 
> Otherwise, I'd say good job, as usual 👍

Regarding this, i made a [ticket](https://bluesquare.atlassian.net/browse/IA-3598) for this and other point that Son discovered . For me this PR can be merged !

beygorghor (approved):

tdethier (approved):
LGTM :+1:

PR #1698, opened 2024-10-10
SLEEP-1499 Support embedding of Superset dashboard in Iaso

bramj:
This PR contains a v1 for embedding Superset dashboards in Iaso, using the `Pages` model.

Specifically:
- Add a new pages template that supports embedding a new type of page "SUPERSET"
- Add an API endpoint POST `/api/superset/token` that creates a Superset "guest token" for a specific dashboard

Related JIRA tickets : https://bluesquare.atlassian.net/browse/SLEEP-1499

**Note that this is a v1. In order to make this properly usable for regular users, I think we should have a place where you can configure the connection to such a dashboard for a given account.**

image

## How to test

1. Create a Superset dashboard, make it embeddable, this will give you a `dashboard_id`.
2. Create an embedded page of type "Superset". Set the `dashboard_id` of the dashboard you want to embed.
3. Visit the page, you should now see your dashboard.
4. In the admin page you can configure the UI config options for your dashboard.
example config:
```json
{
  "hideTab": true,
  "hideTitle": true,
  "hideChartControls": true,
  "filters": {
    "visible": false,
    "expanded": false
  },
}
```
(On Trypelim, I have only the `hideTab` and `hideTitle` set. Filters are useful and chart controls allow things like exporting CSV data)


docker-compose.yml
```diff
@@ -82,7 +82,17 @@ services:
       WFP_EMAIL_RECIPIENTS_NEW_ACCOUNT:
       DISABLE_PASSWORD_LOGINS:
       MAINTENANCE_MODE:
-      SERVER_URL: # Limit logging in dev to not overflow terminal
+      SERVER_URL:
+      SUPERSET_DB_NAME:
+      SUPERSET_DB_HOSTNAME:
+      SUPERSET_DB_PASSWORD:
```
madewulf: Why is that here?

hat/templates/iaso/pages/superset.html
```diff
@@ -0,0 +1,57 @@
+<!doctype html>
+<html>
+    <head>
+        <title>{{title}}</title>
+        {{ analytics_script | safe }}
+        <script src="https://unpkg.com/@superset-ui/embedded-sdk"></script>
```
madewulf: Ideally, we should have this copied in our code base and not depending on an external source

hat/templates/iaso/pages/superset.html
```diff
@@ -0,0 +1,57 @@
...
+                return json_resp.token;
+            }
+
+            const containerRef = document.getElementById(
+                'my-superset-container',
+            );
+
+            supersetEmbeddedSdk.embedDashboard({
+                id: '{{dashboard_id}}',
+                supersetDomain: 'https://superset.trypelim.org',
```
madewulf: Domain name seems hardcoded?

hat/templates/iaso/pages/superset.html
```diff
@@ -0,0 +1,57 @@
+<!doctype html>
+<html>
+    <head>
+        <title>{{title}}</title>
+        {{ analytics_script | safe }}
+        <script src="https://unpkg.com/@superset-ui/embedded-sdk"></script>
+    </head>
+
+    <body>
+        <div id="my-superset-container"></div>
```
madewulf: Not very important, but why "my" ?

bramj:
@beygorghor What do you think of this commit? https://github.com/BLSQ/iaso/pull/1698/commits/0ef8748d57d1b5e7cfde10df7e696f67ff96fc92

@madewulf Okay for re-review!

hat/assets/js/apps/Iaso/domains/pages/components/CreateEditDialog.js
```diff
@@ -233,13 +233,27 @@ const CreateEditDialog = ({ isOpen, onClose, selectedPage }) => {
...
+                                            <>
+                                                <Field
+                                                    label={formatMessage(
+                                                        MESSAGES.superset_dashboard_id,
+                                                    )}
+                                                    name="superset_dashboard_id"
+                                                    component={TextInput}
+                                                    className={classes.input}
+                                                />
+                                            </>
```
beygorghor: this fragment is useless

beygorghor (approved):

hat/assets/js/apps/Iaso/domains/pages/components/CreateEditDialog.js
bramj: Ah yes indeed, I had 2 inputs at some point, forgot to clean this up

madewulf (approved):

PR #1711, opened 2024-10-14
WC2-585: [Web] Duplicate pair details: finish launch analysis modal and rename "duplicate search fields" on entity type

hakifran:
- [Web] Duplicate pair details: finish launch analysis modal and rename "duplicate search fields" on entity type
Related JIRA tickets : [WC2-585](https://bluesquare.atlassian.net/browse/WC2-585)

- [ ] Are there enough tests

## Doc

## Changes
- Add a new field(entity type fields) on analyze modal
- Rename "duplicate search fields" into "Mobile: Config duplicate fields search"

## How to test
- Beneficiaries ---> Duplicates ---> Launch Analysis
- Check if the fields is there and taken in account when the analysis is launched
- Beneficiaries ---> Beneficiaries types
- Check if the duplicate search fields has been renamed

## Print screen / video
[Screencast from 2024-10-14 15-32-14.webm](https://github.com/user-attachments/assets/d77c16e5-2e86-432e-ab73-43d71f13a86a)

[WC2-585]: https://bluesquare.atlassian.net/browse/WC2-585?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hat/assets/js/apps/Iaso/domains/app/translations/en.json
```diff
@@ -202,7 +202,7 @@
     "iaso.entityTypes.dialog.deleteTitle": "Are you sure you want to delete this entity type?",
     "iaso.entityTypes.errors.referenceForm": "Reference form is required",
     "iaso.entityTypes.fieldsDetailInfoView": "Detail info fields",
-    "iaso.entityTypes.fieldsDuplicateSearch": "Duplicate search fields",
+    "iaso.entityTypes.fieldsDuplicateSearch": "Mobile: Config duplicate fields search",
```
bramj: Hm, I didn't realize the naming was this far off...

So actually, the fields you specify here are the fields that the mobile app shows when you search for an existing or create a new entity/beneficiary.

There's actually nothing "duplicate" about this fields, so for me the naming is wrong all the way to the field on the entity `fields_duplicate_search`.
Changing all that would be a lot, so I propose for now to change the label here to just:
```
"Mobile: Entity search fields"
```

hat/assets/js/apps/Iaso/domains/entities/duplicates/list/AnalysisModal.tsx
```diff
@@ -39,31 +43,47 @@ const AnalysisModal: FunctionComponent<Props> = ({ closeDialog, isOpen }) => {
...
-                );
-            } else {
-                setErrorMissingFields('');
-                setFields(entityTypeFields || []);
-            }
-
-            setEntityType(value);
+        } else {
+            setErrorMissingFields('');
+            setDefaultFields(entityTypeDefaultFields || []);
```
bramj: I would remove the default to `filteredEntityType?.original?.fields_duplicate_search`. It doesn't make a lot of sense (even less so with the rename on the entity type) and it's a bit too magical.
I think making the field required and not fall back to any hidden config would be better, what do you think?

bramj (changes requested):
The feature works well :+1:  but a few comments about the usage of the settings on the entity_type. Let me know if you disagree!

bramj (approved):
Great! Looks perfect to me :+1:

PR #1727, opened 2024-10-18
[IA-3571] Added account_name parameter to setuper command

tdethier:
Allows users to choose a name for the account/project/user/... created by the setuper.

Related JIRA tickets : IA-3571

- [ ] Are there enough tests

## Doc
Documentation in the code

## Changes

Added an optional parameter to the setuper command for choosing the name of the account/project/user/...

## How to test

Run the setuper command with the `-n` parameter. 
If you don't add it, the behavior stays the same.

## Print screen / video

image
image

## Notes

During the morning SUM, we discussed the possibility of having random names generated (instead of random string). As nothing specific was decided, I didn't implement this feature.

### DIFF
--- a/setuper/setuper.py
+++ b/setuper/setuper.py
@@ -1,3 +1,5 @@
+import re
+
 from iaso_api_client import IasoClient
 from micro_planning import setup_users_teams_micro_planning
 from data_collection import setup_instances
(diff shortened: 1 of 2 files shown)


setuper/setuper.py
```diff
@@ -55,8 +57,30 @@ def setup_account(account_name, server_url, username, password):
     return iaso_client
 
 
-def create_account(server_url, username, password, additional_projects):
-    account_name = "".join(random.choices(string.ascii_lowercase, k=7))
+def validate_account_name(name: str) -> str:
+    random_name = "".join(random.choices(string.ascii_lowercase, k=7))
+    if not name:  # the user didn't pass a name as parameter
+        return random_name
```
kemar:
```suggestion
    if not name:  # the user didn't pass a name as parameter
        return "".join(random.choices(string.ascii_lowercase, k=7))
```

Possible tiny optimization: this could avoid a small useless computation.

kemar (approved):
LGTM

PR #1733, opened 2024-10-21
IA-3514: Allow to select editable org unit types on user roles

beygorghor:
User should be able to assigne editable org unit types to a user roles, back-end is done.
Alos user in a user roles with editable org unit types should have the restriction from his profile AND user role editable org unit types

Related JIRA tickets : IA-3514

- [ ] Are there enough tests

## Changes

- adapt hook that checks that user has writable permission on a type
- added a field to select editable org unit types on user role dialog

## How to test

- Edit a user roles, check some types
- Connect with a user that have this user role
- Restriction should be the same as on this [PR](https://github.com/BLSQ/iaso/pull/1687)

## Print screen / video

Screenshot 2024-10-21 at 13 11 49


.venv/include/site/python3.9/greenlet/greenlet.h
```diff
@@ -0,0 +1,164 @@
...
+        {                                                                     \
+            _PyGreenlet_API = (void**)PyCapsule_Import("greenlet._C_API", 0); \
+        }
+
+#endif /* GREENLET_MODULE */
+
+#ifdef __cplusplus
+}
+#endif
+#endif /* !Py_GREENLETOBJECT_H */
```
kemar: That's really strange because `.venv/` is in the `.gitignore` file.

Why are we seeing those files?

hat/assets/js/apps/Iaso/domains/userRoles/components/CreateEditUserRole.tsx
```diff
@@ -38,6 +41,7 @@ export const CreateEditUserRole: FunctionComponent<Props> = ({
     isOpen,
     id,
     name,
+    editable_org_unit_type_ids,
```
kemar, line 44: This is always empty for me… Is this a bug? (I hope I didn't break anything during my rebase)

empty

kemar (changes requested):

.venv/include/site/python3.9/greenlet/greenlet.h
beygorghor: my bad, main gitignore as changed and i pushed my venv

hat/assets/js/apps/Iaso/domains/userRoles/components/CreateEditUserRole.tsx
beygorghor: user_roles_editable_org_unit_type_ids=[[]], this should be fixed now

kemar (approved):

PR #1735, opened 2024-10-23
WC2-523: [WEB] NFC card field shows 0 for all beneficiaries even when card is assigned

hakifran:
- [WEB] NFC card field shows 0 for all beneficiaries even when card is assigned
Related JIRA tickets : [WC2-523](https://bluesquare.atlassian.net/browse/WC2-523)

- [ ] Are there enough tests

## Changes
- Add nfc_cards count in the entity dict and serializer
- Link the StorageLogEntry to the right entity if the import is from mobile(entity_uuid)

## How to test
- Create an instance linked to an entity
- Create a StorageDevice linked to that entity with NFC type
- Create one or multiple StorageLogEntry linked the entity, the instance and the created device
- Check the nfc cards on the beneficiary details page and on the instance details page

## Print Screen/video
[Screencast from 2024-10-23 11-18-46.webm](https://github.com/user-attachments/assets/0e4f4e57-a314-4fdd-b97c-5c6d0d571793)

[WC2-523]: https://bluesquare.atlassian.net/browse/WC2-523?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iaso/api/storage.py
```diff
@@ -388,8 +388,9 @@ def create(self, _, request):
                     concerned_orgunit = OrgUnit.objects.get(id=log_data["org_unit_id"])
 
                 concerned_entity = None
-                if "entity_id" in log_data and log_data["entity_id"] is not None:
-                    concerned_entity = Entity.objects.get(uuid=log_data["entity_id"])
+                entity_id = log_data.get("entity_id") or log_data.get("entity_uuid")
+                if entity_id:
+                    concerned_entity = Entity.objects.get(uuid=entity_id)
```
bramj, line 393: :+1:

iaso/api/entity.py
```diff
@@ -80,6 +83,12 @@ def get_submitter(self, entity: Entity):
     def get_duplicates(self, entity: Entity):
         return _get_duplicates(entity)
 
+    def get_nfc_cards(self, entity: Entity):
+        nfc_devices = StorageDevice.objects.filter(entity=entity, type=StorageDevice.NFC)
+        nfc_log_entries_count = StorageLogEntry.objects.filter(device__in=nfc_devices).count()
+
+        return nfc_log_entries_count
+
```
bramj, line 92: I think we should be careful to add this extra query on the EntitySerializer. I also expect the amount of records in `StorageLogEntry` to grow.
 Did you try to do it with a relation and `prefetch_related`? If it's complicated then it's okay for me to go with it like this.

bramj (changes requested):
Nice! The counts work :+1: I have a few doubts about performance issues as the logs table grows, but I think it might be difficult to fix, see comments.

Main thing is I think there's something missing on this screen:
http://localhost:8081/dashboard/storages/
image
I did the setup as per instructions in the PR, but I see no beneficiary on the "General" or "Logs" entries. The status on the log is also missing (I set it to "OK" in my db).

hakifran:
> Nice! The counts work 👍 I have a few doubts about performance issues as the logs table grows, but I think it might be difficult to fix, see comments.
> 
> Main thing is I think there's something missing on this screen: http://localhost:8081/dashboard/storages/ image I did the setup as per instructions in the PR, but I see no beneficiary on the "General" or "Logs" entries. The status on the log is also missing (I set it to "OK" in my db).

Did you link the storageDevice with the entity?
Screenshot from 2024-10-24 17-03-30

The logs status, status_reason and status_comment must also be save in storageLogEntry
Screenshot from 2024-10-24 17-21-37

bramj:
@hakifran Yes I think it's a misalignment between the returned JSON and the frontend

image

Btw, I see the `name` is blank on the Entity, but in reality this will almost always be the case. So better to just display the `id` in the frontend.

iaso/api/entity.py
```diff
@@ -80,6 +83,13 @@ def get_submitter(self, entity: Entity):
     def get_duplicates(self, entity: Entity):
         return _get_duplicates(entity)
 
+    def get_nfc_cards(self, entity: Entity):
+        nfc_log_entries_count = StorageDevice.objects.filter(entity=entity, type=StorageDevice.NFC).aggregate(
+            total_logs=Count("log_entries")
+        )["total_logs"]
```
bramj, line 89: Nice! Looking good to me :+1:

bramj (approved):
Tested again and this looks very good to me :+1:

PR #1739, opened 2024-10-24
IA-3466 Implement permissions for users with org unit types restrictions

kemar:
Implement additional permissions for users with org unit types restrictions.

Related JIRA tickets : [IA-3466](https://bluesquare.atlassian.net/browse/IA-3466)

## Changes

- implement additional permissions for users with org unit types restrictions:

    - in bulk create users (via CSV)
    - in `api/profiles` update and create
    - in profile bulk update

- add validation in `api/profiles` **create** the same way it's implemented in **update**
    - also move methods inside the class so they are logically grouped

## How to test

Use the UI or check the unit tests

[IA-3466]: https://bluesquare.atlassian.net/browse/IA-3466?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iaso/api/profiles/profiles.py
+++ b/iaso/api/profiles/profiles.py
@@ -529,7 +529,9 @@ class ProfilesViewSet(viewsets.ViewSet):
                 invalid_names = ", ".join(
                     name for name in OrgUnitType.objects.filter(pk__in=invalid_ids).values_list("name", flat=True)
                 )
-                raise ValidationError(f"The user does not have rights on the following org unit types: {invalid_names}")
+                raise PermissionDenied(
+                    f"The user does not have rights on the following org unit types: {invalid_names}"
+                )
 
         return editable_org_unit_types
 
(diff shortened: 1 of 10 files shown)


iaso/api/profiles/profiles.py
```diff
@@ -304,6 +336,246 @@ def list(self, request):
...
+            try:
+                if existing_user.iaso_profile and existing_user.iaso_profile.account == current_account:
+                    user_in_same_account = True
+            except User.iaso_profile.RelatedObjectDoesNotExist:
+                # User doesn't have an iaso_profile, so they're not in the same account
+                pass
+
+            if user_in_same_account:
+                return JsonResponse(
+                    {"errorKey": "user_name", "errorMessage": _("Nom d'utilisateur existant")}, status=400
```
hakifran, line 370: I think this message should be adapted to the condition

hakifran (approved):
Seems good to me, apart one comment on exception message

PR #1744, opened 2024-10-28
WC2-589 Fix: Don't fail if correlation_field is blank

bramj:
The explicit check on `is not None` would allow empty strings to pass, resulting in error:
```
TypeError: can only concatenate str (not "NoneType") to str
```

Simply checking on `self.form.correlation_field` will give `False` for both `None` and `""`.

Related JIRA tickets : https://bluesquare.atlassian.net/browse/WC2-589


kemar (approved):
LGTM

PR #1747, opened 2024-10-28
IA-3612: fix duplicate PotentialPayments

quang-le:

Related JIRA tickets : IA-3612

- [ ] Are there enough tests

## Doc

## Changes

- Add unique constraint on `PotentialPayment`'s `user`
- In `calculate_potential_payments`, wrap the block of code calling `get_or_create` and `save` in a `try/except`. Return existing `PotentialPayment` in case of `IntegrityError`
- Test the unicity constraint on the model
- show user in PotentialPayment admin screen
- merge migrations

## How to test

In a DB with at leats one potential payment, go to django admin and try to create a potential payment for a user that already has one. You should get an error

## Print screen / video

https://github.com/user-attachments/assets/bbc9e797-c802-4caf-8e0f-f9384c518f70

## Notes

mergeable version of https://github.com/BLSQ/iaso/pull/1746


iaso/tests/models/test_payments.py
```diff
@@ -72,12 +74,17 @@ class PotentialPaymentModelTestCase(TestCase):
...
+        cls.user = User.objects.create(username="test_user")
         cls.org_unit = m.OrgUnit.objects.create(name="Test OrgUnit")
-        change_request = m.OrgUnitChangeRequest.objects.create(org_unit=cls.org_unit, new_name="Foo")
-        cls.potential_payment = m.PotentialPayment.objects.create(user=user)
-        cls.potential_payment.change_requests.add(change_request)
+        cls.change_request = m.OrgUnitChangeRequest.objects.create(org_unit=cls.org_unit, new_name="Foo")
 
     def test_create(self):
+        self.potential_payment = m.PotentialPayment.objects.create(user=self.user)
+        self.potential_payment.change_requests.add(self.change_request)
```
kemar, line 83: We should use local variables in test methods because assigning instance attributes with `self` make these attributes available in all other test methods and this can pose subtle problems to debug.

kemar (approved):
LGTM

PR #1761, opened 2024-10-31
IA-3609 Avoid `PermissionDenied` when `projects` is an empty list `[]`

kemar:
Avoid `PermissionDenied` when `projects` is an empty list `[]`.

Related JIRA tickets : [IA-3609](https://bluesquare.atlassian.net/browse/IA-3609)

[IA-3609]: https://bluesquare.atlassian.net/browse/IA-3609?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iaso/api/profiles/profiles.py
+++ b/iaso/api/profiles/profiles.py
@@ -695,21 +695,17 @@ class ProfilesViewSet(viewsets.ViewSet):
 
     def validate_projects(self, request, profile):
         result = []
-        request_user = request.user
-        projects = request.data.get("projects", None)
-        if projects is not None:
-            if not request_user.has_perm(permission.USERS_ADMIN):
+        projects = request.data.get("projects")
+        if projects:
+            if not request.user.has_perm(permission.USERS_ADMIN):
                 raise PermissionDenied(
                     f"User with permission {permission.USERS_MANAGED} cannot change project attributions"
                 )
-        # This is bit ugly, but it's to maintain the fetaure's behaviour after adding the check in the permission
-        if projects is None:
-            projects = []
-        for project in projects:
-            item = get_object_or_404(Project, pk=project)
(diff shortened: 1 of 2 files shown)


beygorghor (approved):

quang-le (changes requested):
This doesn't solve the problem if the edited user already has projects in their profile, because then a non-empty list is sent in the PATCH request

Screenshot 2024-10-31 at 23 36 43

kemar:
> This doesn't solve the problem if the edited user already has projects in their profile, because then a non-empty list is sent in the PATCH request

@quang-le It seems to work for me.

Do you have a precise use case to reproduce this issue or the full backend error trace?

https://github.com/user-attachments/assets/65227061-9680-4462-a3a4-d6ceda64ca34

iaso/api/profiles/profiles.py
```diff
@@ -695,21 +695,17 @@ def validate_user_roles(self, request):
 
     def validate_projects(self, request, profile):
         result = []
-        request_user = request.user
-        projects = request.data.get("projects", None)
-        if projects is not None:
-            if not request_user.has_perm(permission.USERS_ADMIN):
+        projects = request.data.get("projects")
+        if projects:
+            if not request.user.has_perm(permission.USERS_ADMIN):
```
quang-le:
```suggestion
       projects = request.data.get("projects", [])
        profile_project_ids = set(profile.projects.values_list("id", flat=True))
        project_ids_from_request = set(map(int, projects))
        if projects:
            if not request.user.has_perm(permission.USERS_ADMIN) and profile_project_ids != project_ids_from_request:
```

quang-le (approved):

PR #1768, opened 2024-11-04
IA-3607 fix payment disappearing after bulk update

quang-le:
Payment lots were not shown in the UI after bulk update because their `task` field was not updated properly

Related JIRA tickets : IA-3607

- [ ] Are there enough tests

## Doc

## Changes

- Backend:
    - bulk update: set `PaymentLot` task field to None when task succeeds or raises
    - mark as read: idem
    - Handle the case where the task is killed, so that the payment lot is still correctly updated (`task` field set to `None`)
- Front-end:
    - disable Payment Lot action butttons in table if PaymentLot task is not null
    - set row background color to yellow to indicate that Payment Lot has a task running

## How to test

setup:  DB with payment lots, and an active worker

- Go to payments lots page
- open payment lot. 
- Select one or several payment with the table checkbox
- Edit and save payment(s)

- The payment lot row should be yellow and the buttons deactivated
- Refresh page: when the change has been processed, the row should be back to normal with active buttons

- repeat operation with the mark as read action button. The behaviour should be the same

## Print screen / video

https://github.com/user-attachments/assets/81b9bd94-c88c-4aa8-9c35-4ee7aa5614ad

## Notes

This ports to main and expands changes made in HOTFIX_IA-3609_IA-3607 for IA-3607

### DIFF
--- a/iaso/tasks/payments_bulk_update.py
+++ b/iaso/tasks/payments_bulk_update.py
@@ -125,14 +125,14 @@ def mark_payments_as_read(
     start = time()
     the_task = task
     the_task.report_progress_and_stop_if_killed(progress_message="Searching for Payments to modify")
-    payment_lot = PaymentLot.objects.get(id=payment_lot_id)
     try:
-        payments = Payment.objects.filter(payment_lot=payment_lot)
+        payment_lot = PaymentLot.objects.get(id=payment_lot_id)
     except ObjectDoesNotExist:
         the_task.status = ERRORED
         the_task.ended_at = timezone.now()
         the_task.result = {"result": ERRORED, "message": "Payment Lot not found"}
         the_task.save()
+    payments = Payment.objects.filter(payment_lot=payment_lot)
     total = payments.count()
 
     # audit stuff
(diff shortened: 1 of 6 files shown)


iaso/tasks/payments_bulk_update.py
```diff
@@ -110,15 +143,20 @@ def mark_payments_as_read(
...
+
     try:
-        payments = Payment.objects.filter(payment_lot=payment_lot)
+        payment_lot = PaymentLot.objects.get(id=payment_lot_id)
     except ObjectDoesNotExist:
         the_task.status = ERRORED
         the_task.ended_at = timezone.now()
         the_task.result = {"result": ERRORED, "message": "Payment Lot not found"}
         the_task.save()
+    # Handling task being killed after we get the payment_lot, so we can set it's task field back to None if the task is killed
```
kemar:
```suggestion
    # Handling task being killed after we get the payment_lot, so we can set its task field back to None if the task is killed
```

kemar (approved):
LGTM

beygorghor (approved):

PR #1773, opened 2024-11-05
IA-3595 Fix profile bulk update bugs

kemar:
Prevent profile bulk update to raise a wrong error when a value is an empty list instead of `None`.

Related JIRA tickets : [IA-3595](https://bluesquare.atlassian.net/browse/IA-3595)

## Changes

- for _list_ values, ensure the `tasks/create/profilesbulkupdate` API treats `None` as an empty list
- adapt tests to reflect this and use the full API payload in some of them

## How to test

The initial bug was:

- on the user admin page
- select a bunch of users to add them a role
- validate this
- then realize that the users were not updated due to this error: `User without the permission menupermissions.iaso_teams cannot add users to team`

This should be fixed and you should be able to perform this operation without any error.

[IA-3595]: https://bluesquare.atlassian.net/browse/IA-3595?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iaso/tests/api/test_profiles_bulk_update.py
```diff
@@ -228,17 +228,28 @@ def test_profile_bulkupdate_select_some(self):
...
+            "unselected_ids": None,
             "projects_ids_added": [
                 self.project.pk,
                 self.project_3.pk,
             ],  # self.project_3 should not be saved, not from the same account
             "projects_ids_removed": [self.project_2.pk],
             "roles_id_added": [self.user_role.pk],
             "roles_id_removed": [self.user_role_2.pk],
+            "location_ids_added": [self.jedi_council_corruscant.pk],
+            "location_ids_removed": [self.jedi_council_endor.pk],
```
quang-le, line 240: Good old jedi council

quang-le (approved):
LGTM

PR #1765, opened 2024-11-04
WC2-591: Integrate Account into beneficiary table

butofleury:
Explain what problem this PR is resolving

Related JIRA tickets : [WC2-591](https://bluesquare.atlassian.net/browse/WC2-591)

## Self proofreading checklist

- [x] Did I use eslint and black formatters
- [ ] Is my code clear enough and well documented
- [ ] Are my typescript files well typed
- [ ] New translations have been added or updated if new strings have been introduced in the frontend
- [ ] My migrations file are included
- [ ] Are there enough tests
- [ ] Documentation has been included (for new feature)

## Changes

As there is data linked to 2 account, adding a new field into beneficiary table is required in order to separate the data linked to 2 country(South Sudan and Nigeria). 

## How to test

From the local instance with wfp coda database, run following tasks:

- creating a new migration for adding new field `account` into `wfp_beneficiary` table: `docker compose run iaso manage makemigrations wfp`
- adding the new field into `wfp_beneficiary` table:  `docker compose run iaso manage migrate wfp`
- generating data for tableau dashboard : `docker compose run iaso manage wfp_etl`

Then login on django admin and check on [beneficiary table](http://localhost:8081/admin/wfp/beneficiary/) . You should see the account field.

## Print screen / video

[Bneeficiary-Administration-de-Iaso-Iaso.webm](https://github.com/user-attachments/assets/7aac3300-8103-4d08-8537-4d87dab6ba4e)

[WC2-591]: https://bluesquare.atlassian.net/browse/WC2-591?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1778, opened 2024-11-06
WC2-592: Integrate Nigeria data to beneficiary table

butofleury:

Related JIRA tickets : [WC2-591](https://bluesquare.atlassian.net/browse/WC2-591),  [WC2-592](https://bluesquare.atlassian.net/browse/WC2-592)

- [ ] Are there enough tests

## Changes

Splitted **`wfp`** plugin into 2 sub plugins **`south_sudan`** and **`nigeria`**, integrated the Nigeria beneficiary data and refactored the code by making it more generic.

## How to test

From the local instance with wfp coda database, run ETL script to generate data for: 
- South Sudan : `docker compose run iaso manage etl_ssd`
- Nigeria :  `docker compose run iaso manage etl_ng`

## Print screen / video

[Administration-de-Wfp-Iaso.webm](https://github.com/user-attachments/assets/2da1e443-913c-4ab3-a506-bf3e6ef8657c)

[WC2-592]: https://bluesquare.atlassian.net/browse/WC2-592?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[WC2-591]: https://bluesquare.atlassian.net/browse/WC2-591?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


butofleury:
This PR depends to https://github.com/BLSQ/iaso/pull/1765.

hat/settings.py
```diff
@@ -176,6 +176,8 @@
     "allauth.account",
     "allauth.socialaccount",
     "storages",
+    "plugins.wfp.south_sudan",
+    "plugins.wfp.nigeria",
```
bramj: We should not load these plugin apps for every Iaso installation. If we really need multiple apps for WFP, we can do it with the `INSTALLED_APPS` ENV as here for Trypelim: https://github.com/BLSQ/trypelim/blob/main/plugin_settings.py

plugins/wfp/common.py
```diff
@@ -16,11 +16,16 @@ def delete_beneficiaries(self):
         print("EXISTING VISITS DELETED", beneficiary[1]["wfp.Visit"])
         print("EXISTING JOURNEY DELETED", beneficiary[1]["wfp.Journey"])
 
+    def account_related_to_entity_type(self):
+        entity_type = EntityType.objects.filter(code__in=self.type)
```
bramj: `self.type` has become a list. I think the naming is confusing like this, we should change it to `types`.

bramj: Actually, why exactly did this change? I see it's being called as:

```python
ETL(["pbwg_1"])
# and
ETL(["child_under_5_1"])
```

plugins/wfp/common.py
```diff
@@ -16,11 +16,16 @@ def delete_beneficiaries(self):
...
+        return account
+
     def retrieve_entities(self):
         steps_id = ETL().steps_to_exclude()
         updated_at = date(2023, 7, 10)
         beneficiaries = (
-            Instance.objects.filter(entity__entity_type__code=self.type)
+            Instance.objects.filter(entity__entity_type__code__in=self.type)
             # .filter(entity__id__in=[1, 42, 46, 49, 58, 77, 90, 111, 322, 323, 330, 196, 226, 254,315, 424, 430, 431, 408, 19, 230, 359])
             # .filter(entity__id__in=[230, 359, 254])
```
bramj: Maybe we should clean up these comments?

plugins/wfp/common.py
```diff
@@ -402,6 +410,26 @@ def map_assistance_step(self, step, given_assistance):
             given_medication = self.split_given_medication(step.get("medication_2"), quantity)
             given_assistance = given_assistance + given_medication
 
+        if step.get("vitamins_given", None) is not None and step.get("vitamins_given", None) == "1":
```
bramj: Just for easier reading, I think this can be simplified into
```python
if step.get("vitamins_given") == "1":
```
Since `get` returns `None` by default when nothing's found (https://docs.python.org/3/library/stdtypes.html#dict.get)
Same for the other conditions.

plugins/wfp/common.py
```diff
@@ -435,6 +463,19 @@ def map_assistance_step(self, step, given_assistance):
                 "quantity": quantity,
             }
             given_assistance.append(assistance)
+        elif step.get("ration_type") is not None:
```
bramj: Maybe irrelevant, but might be safer to do
```python
elif step.get("ration_type"):
```
which will give `False` if ration_type is an empty string?

plugins/wfp/migrations/0013_beneficiary_account.py
```diff
@@ -0,0 +1,21 @@
...
+
+    operations = [
+        migrations.AddField(
+            model_name="beneficiary",
+            name="account",
+            field=models.ForeignKey(
+                blank=True, null=True, on_delete=django.db.models.deletion.CASCADE, to="iaso.account"
+            ),
+        ),
+    ]
```
bramj, line 21: I see no handling of existing beneficiary data (that have account `NULL`), I guess we'll do a delete all beneficiary data and re-run the scripts?

bramj (changes requested):
I don't know the exact details of the data generated, but the scripts run well and appear correctly linked to the correct account :+1: 

General code structure question: is there a reason why you decided to add a new Django app for each country? I think adding the `etl_ssd.py` and `etl_ng.py` scripts directly into `wfp/management/commands` and adding folders there with the classes and helper functions per country would be easier. Then we would also not need anything in `settings.py`. What do you think?

plugins/wfp/common.py
butofleury: I did that change to align South Sudan to Nigeria. As discussed yesterday during the **WFP touchbase** , the entity type for child has been splitted to 2 entity types. 

But as we will merge the 2 types into 1 like South Sudan structure, I will fix it according to the changes will be made once the merge done.

plugins/wfp/migrations/0013_beneficiary_account.py
butofleury: yes, we will do a delete all beneficiary data and re-run the scripts!

bramj (approved):
Looking good! :+1:

PR #1784, opened 2024-11-07
IA-3654: Org Unit Pop up: wrong count for instances

hakifran:
- Org Unit Pop up: wrong count for instances
Related JIRA tickets : [IA-3654](https://bluesquare.atlassian.net/browse/IA-3654)

- [ ] Are there enough tests

## Doc

## Changes
- Add instances_count as a parameter to the retrieve api query
- Get all instances for the orgUnit include the ones for its descendants

## How to test

- Org units --> Org units list
- Filter on parent
- Map tab
- Click on map marker for org Unit and check if the submissions count is correct

## Print screen / video
[Screencast from 2024-11-07 16-24-39.webm](https://github.com/user-attachments/assets/ead61488-d1c4-4ed6-baec-b8ec847c6612)

[IA-3654]: https://bluesquare.atlassian.net/browse/IA-3654?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iaso/api/org_units.py
+++ b/iaso/api/org_units.py
@@ -763,9 +763,14 @@ class OrgUnitViewSet(viewsets.ViewSet):
 
     def retrieve(self, request, pk=None):
         org_unit: OrgUnit = get_object_or_404(
-            self.get_queryset().prefetch_related("reference_instances").annotate(instances_count=Count("instance")),
+            self.get_queryset().prefetch_related("reference_instances"),
             pk=pk,
         )
+
+        if request.query_params.get("instances_count"):
+            instances_count = org_unit.descendants().aggregate(Count("instance"))["instance__count"]
+            org_unit.instances_count = instances_count
+
         self.check_object_permissions(request, org_unit)
         res = org_unit.as_dict_with_parents(light=False, light_parents=False)
         res["geo_json"] = None
--- a/iaso/api/org_units.py
+++ b/iaso/api/org_units.py
@@ -766,10 +766,9 @@ class OrgUnitViewSet(viewsets.ViewSet):
             self.get_queryset().prefetch_related("reference_instances"),
             pk=pk,
         )
-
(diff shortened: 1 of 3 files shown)


iaso/api/org_units.py
```diff
@@ -763,9 +763,14 @@ def create(self, _, request):
 
     def retrieve(self, request, pk=None):
         org_unit: OrgUnit = get_object_or_404(
-            self.get_queryset().prefetch_related("reference_instances").annotate(instances_count=Count("instance")),
+            self.get_queryset().prefetch_related("reference_instances"),
             pk=pk,
         )
+
+        if request.query_params.get("instances_count"):
```
kemar: Why do we need an `instances_count` param?

kemar (changes requested):

iaso/api/org_units.py
hakifran: @kemar  I used that param because the retrieve is used somewhere else(to avoid the performance issue) and as you can see the as_dict_with_parents the instances_count is optianal.
Screenshot from 2024-11-08 12-44-17

kemar: @hakifran what performance issue are you talking about?

hakifran: When it has to count many instances for a parent orgUnit which has many descendants and used in a place where it has other queries but where the instances_count is not needed. But I can remove it if it won't cause any problem

kemar: I think the additional aggregate query is quite performant.

The point is that this API is already hard to understand.

Here, you're adding a new parameter and it's "hidden" somewhere in the `retrieve()` method which adds to the global complexity.

I'd rather remove it until we're certain we have a performance issue.

kemar (approved):
LGTM

PR #1798, opened 2024-11-14
POLIO-1733 creation of vaccine management repository page

mathvdh:

Related JIRA tickets : POLIO-1733

- [ ] Are there enough tests

## Doc

## Changes

## How to test

Explain how to test your PR.

## Print screen / video

## Notes

```
fix: empty instance pop up

Refs: IA-3665
```


plugins/polio/api/vaccines/repository.py
```diff
@@ -0,0 +1,260 @@
...
+            elif campaign_status.upper() == "PREPARING":
+                queryset = queryset.filter(campaign_started_at__gte=today)
+
+        # Filter by country block
+        country_block = request.query_params.get("country_block", None)
+        if country_block:
+            queryset = queryset.filter(campaign__country__groups__in=country_block.split(","))
+
+        # Filter by country (multi)
+        countries = request.query_params.get("countries", None)
```
kemar, line 108: Relying on `request.query_params` is prone to errors because there is no validation of the data.

E.g. this is how to break the endpoint: `http://localhost:8081/api/polio/vaccine/repository/?country_block=foo,bar`.

```
ValueError at /api/polio/vaccine/repository/
Field 'id' expected a number but got 'foo'.
```

Using a [`FilterSet`](https://django-filter.readthedocs.io/en/latest/guide/rest_framework.html#adding-a-filterset-with-filterset-class) is a good way to go to avoid such pitfalls because it comes with built-in validation.

As a quick fix, maybe you can just validate that values are `int`s.

plugins/polio/api/vaccines/repository.py
```diff
@@ -0,0 +1,260 @@
...
+    CampaignType,
+    DestructionReport,
+    IncidentReport,
+    OutgoingStockMovement,
+    Round,
+    VaccinePreAlert,
+    VaccineRequestForm,
+    VaccineRequestFormType,
+)
+
```
kemar, line 23: There is a few unused imports here.

I think I'm gonna add `ruff` to the project because this can be checked and fixed automatically.

plugins/polio/api/vaccines/repository.py
```diff
@@ -0,0 +1,260 @@
...
+        ]
+
+    def get_pre_alert_data(self, obj):
+        pre_alerts = VaccinePreAlert.objects.filter(request_form__campaign=obj.campaign)
+        return [
+            {"date": pa.date_pre_alert_reception, "file": pa.document.url if pa.document else None} for pa in pre_alerts
+        ]
+
+    def get_form_a_data(self, obj):
+        form_as = OutgoingStockMovement.objects.filter(campaign=obj.campaign, round=obj)
```
kemar, line 56: `VaccineRepositorySerializer` is issuing one extra SQL query per `get_*` method.

This results in a N+3 for each result in the list view.

This won't scale well.

Is there an easy way to reduce the number of queries?

kemar (approved):
LGTM as an urgent PR.

But there are some performance and validation issues that should be addressed later.

plugins/polio/api/vaccines/repository.py
mathvdh: Ok I checked it

plugins/polio/api/vaccines/repository.py
mathvdh: I vote FOR ruff

plugins/polio/api/vaccines/repository.py
mathvdh: For this I don't have an easy fix without a major change in the DB structure

PR #1800, opened 2024-11-14
IA-3674 Fix: allow a user with restricted `OrgUnitType` to edit other users

kemar:
Allow a user with restricted `OrgUnitType` to edit another user as long as `org_units` or `editable_org_unit_type_ids` fields are not changed.

Related JIRA tickets : [IA-3674](https://bluesquare.atlassian.net/browse/IA-3674)

## Changes

The validation is too strict for users with the "user management" permission: a user restricted to a given `org_unit_type` can't modify another user if this user has a forbidden `org_unit` or `editable_org_unit_type`.

I'm fixing that by allowing to edit another user as long as `org_unit` or `editable_org_unit_type` fields are not changed.

This will allow to modify phone number, last name, etc. for any user.

## Notes

This complexity arises partly because:

- the API doesn't use a serializer and relies on `request.data`
- the front-end _always_ sends the full payload for each modification
    - i.e. even if you just want to change the phone number, all the `org_unit` and `editable_org_unit_type` will be sent along with the request

## How to test

1. add an org unit type restriction to a user
2. ensure the user has the "user management" permission (`Admin > User management > Geo-limited` in the UI)
3. log out
4. log in with the user you've just modified
5. try to edit users liked to org unit with a type you cannot edit
6. you should be able to edit other fields such as the phone number, the username etc. without erros

[IA-3674]: https://bluesquare.atlassian.net/browse/IA-3674?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iaso/api/profiles/profiles.py
+++ b/iaso/api/profiles/profiles.py
@@ -419,7 +419,7 @@ class ProfilesViewSet(viewsets.ViewSet):
             org_units = self.validate_org_units(request, profile)
             user_roles_data = self.validate_user_roles(request)
             projects = self.validate_projects(request, profile)
-            editable_org_unit_types = self.validate_editable_org_unit_types(request)
+            editable_org_unit_types = self.validate_editable_org_unit_types(request, profile)
         except ProfileError as error:
             # Delete profile if error since we're creating a new user
             profile.delete()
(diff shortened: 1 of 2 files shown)


bramj (approved):
Complicated stuff! :open_mouth:  I tested it locally and it's working, good job :+1:

PR #1811, opened 2024-11-20
POLIO-1738 Remove deprecated fields on Campaign

kemar:
Remove deprecated fields on Campaign.

Related JIRA tickets : [POLIO-1738](https://bluesquare.atlassian.net/browse/POLIO-1738)

## Changes

Remove deprecated fields on Campaign.

[POLIO-1738]: https://bluesquare.atlassian.net/browse/POLIO-1738?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


plugins/polio/api/campaigns/campaigns.py
quang-le, line 1: Unless they are still used in a dashboard, I think `round_one` and `round_two` should be removed as well. That would make the response much smaller (since the info is already in `rounds`)

plugins/polio/models/base.py
quang-le, line 1: Same with the fields moved to the budget model: if no dashboard uses them I think they can be deleted. TBC with @lpontis  maybe

quang-le (commented):
Juts a few fields I think could still be removed, if we can confirm that no dashboard uses them anymore

kemar:
> Juts a few fields I think could still be removed, if we can confirm that no dashboard uses them anymore

This is a good point but I suggest to handle their removal in distinct PRs.

quang-le (approved):

plugins/polio/api/campaigns/campaigns.py
kemar: `round_one` and `round_two` have been removed from the model. Are you talking about the API endpoint?

PR #1818, opened 2024-11-21
[IA-3701] Fixed Django warning

tdethier:
Fixed Django warning about `PotentialPayment`

Related JIRA tickets : IA-3701

- [ ] Are there enough tests

## Doc

/

## Changes

Replaced, as suggested by Django, a `.ForeignKey(unique=True)` by a `OneToOneField()` relation.

## How to test

Start the backend, make sure that there is no more warning (or check in the github actions below :arrow_down:) 

## Print screen / video
image

## Notes
Tests are running fine, but I'm not sure about how to thoroughly test this feature, as I don't know the business use case behind this Model.

```
fix: empty instance pop up

Refs: IA-3665
```


iaso/api/payments/views.py
```diff
@@ -476,19 +475,13 @@ def calculate_new_potential_payments(self):
...
-                    return PotentialPayment.objects.get(
-                        user_id=user["created_by"],
-                    )
+                potential_payment, created = PotentialPayment.objects.get_or_create(
+                    user_id=user["created_by"],
+                )
+                for change_request in change_requests:
+                    change_request.potential_payment = potential_payment
+                    change_request.save()
+                potential_payment.save()
```
kemar, line 484: @tdethier @quang-le I've removed the exception catching here because [`get_or_create`](https://docs.djangoproject.com/en/5.1/ref/models/querysets/#get-or-create) is atomic when the database enforces uniqueness of the keyword arguments (`user` here).

And now, the uniqueness constraint is set by the [`OneToOneField`](https://github.com/django/django/blob/857b1048d53ebf5fc5581c110e85c212b81ca83a/django/db/models/fields/related.py#L1218-L1221).

`IntegrityError` is then [catched in `get_or_create`](https://github.com/django/django/blob/857b1048d53ebf5fc5581c110e85c212b81ca83a/django/db/models/query.py#L945C1-L946C1).

Are you OK with that?

tdethier: It sounds good, but I'd rather let @quang-le decide because I don't really know why it was made this way in the first place

quang-le: @kemar If you're sure, then OK. I remember adding the try/catch because we used to have a race condition, but that may have been before we enforced the uniqueness of the `user`

kemar: Yup I'm sure about that.

kemar (approved):

PR #1832, opened 2024-11-27
IA-3026: In the org unit search, the loading of the map tab should only start when we switch to the map tab

beygorghor:
Currently, if you go to the org unit list, and click search, you’ll see in the requests that it loads tiles and up to 3000 org units to display on the map, while it should only load the org units needed for the list. 

These requests should only start when the map is visible on screen. 

Related JIRA tickets : IA-3026

- [ ] Are there enough tests

## Changes

- refactor the use of react query 
- only fetch map data if on map tab
- fix a bug while deleting a tab (deleting a non selected tab is leading to an inconsistent behaviour)
- fix fit to bound while changing results

## How to test

Play with the org units search and multi tab. 
Make sure it makes only call for the map while on the map.
Order, pagination, filters, try to break it please

## Print screen / video

Nothing to show here as everything should be the same but better

## Notes

⚠️ There is a blsq-component PR [here](https://github.com/BLSQ/bluesquare-components/pull/169) to use with it

```
fix: empty instance pop up

Refs: IA-3665
```


hat/assets/js/apps/Iaso/domains/orgUnits/components/TableList.tsx
```diff
@@ -123,7 +138,7 @@ export const TableList: FunctionComponent<Props> = ({
                 saveMulti={saveMulti}
             />
             <Box mt={-4}>
-                <TableWithDeepLink
+                <Table
```
quang-le, line 141: Didn't we want to deprecate the `resetPageToOne` prop? I think this must be one the last `Table`s that use it.
If I remember correctly, we could just pass the value in `extraProps`.

Not blocking, but a small piece of tech debt that could be paid off

quang-le (changes requested):
When switching datasources and zooming in on the map, the map crashes

https://github.com/user-attachments/assets/76542ee8-aacc-4372-9221-850ddd9293e1

Switching back to list tab will refetch groups and validation status dropdowns. Same on staging, so not blocking. But strange: only these 2 are refetched. The query keys look ok, and playing with staleTime and cacheTime won't fix it.

https://github.com/user-attachments/assets/deb7aa21-e488-4af9-a6ec-396a30e77ef3

Only the map crash is blocking

hat/assets/js/apps/Iaso/domains/orgUnits/components/TableList.tsx
beygorghor: in fact i had to use a dedicated query key to trigger filtering only while clicking on search and not while adding a new search. So using params is not possible. I used this because it was already present, it can be removed from the table for sure but we need to keep it for the query key. 
Also i moved from `TableWithDeepLink` to `Table` to be able to use `onTableParamsChange` and ordering, sorting, page without having to use a `useEffect` on parent

beygorghor:
@quang-le 
i have fixed the zoom issue, for the double call this is a bit weird... maybe should be part of another PR

quang-le (approved):

PR #1836, opened 2024-11-28
IA-3728 Create change requests from differences between two pyramids (backend model)

kemar:
Create change requests from differences between two pyramids (backend model).

Related JIRA tickets : [IA-3728](https://bluesquare.atlassian.net/browse/IA-3728)

## Doc

## Changes

- new `DataSourceSynchronization` model and admin:
    - allows to synchronize two pyramids via "change requests"
    - this model is tightly coupled with the `diffing` module for the logic

- `diffing` module:
    - added tests (this was helpful to understand how it works)
    - fixed `Dumper.dump_as_json()` which was broken
    - fixed `Differ.diff` for groups (only groups from one side were taken into account)
    - fixed `Differ.compare_fields`
        - `if dhis2_value is None and ref_value is not None:` wasn't working when `ref_value = []` (empty list) for empty groups

## How to test

1. go to `/admin/iaso/datasourceversionssynchronization/add/`
2. configure a new "Data source synchronization" with two source versions that you want to synchronize
    - ensure those two source versions have differences
3. save
4. find the ID of the newly created "Data source synchronization"
5. open a Django shell `docker compose exec -ti iaso ./manage.py shell_plus`
6. trigger the synchronization (those steps will be handled by the future API):

    - get the "Data source synchronization" object assuming its ID is `1`:

            sync = DataSourceVersionsSynchronization.objects.get(id=1)

    - generate the diff between the two pyramids (this allows to count how many change requests there will be):

            sync.create_json_diff()

    - synchronize the two pyramids and create "change requests":

            sync.synchronize_source_versions()

7. go to `/admin/iaso/orgunitchangerequest/`
8. filter the list by your "Data source synchronization"
9. you should be able to see the newly created "change requests"

## Print screen

admin1

admin2

[IA-3728]: https://bluesquare.atlassian.net/browse/IA-3728?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iaso/admin.py
```diff
@@ -1077,6 +1081,33 @@ class Media:
...
+        "account",
+        "created_by",
+        "count_create",
+        "count_update",
+    )
+    list_display_links = ("pk", "name")
+    autocomplete_fields = ("account", "created_by", "source_version_to_update", "source_version_to_compare_with")
+    readonly_fields = ("json_diff", "count_create", "count_update", "created_at", "updated_at", "sync_task")
+
+    def get_queryset(self, request):
```
madewulf, line 1105: This is for managing performances in the admin?

iaso/models/org_unit.py
```diff
@@ -132,7 +135,7 @@ class OrgUnitType(models.Model):
     objects = OrgUnitTypeManager()
 
     def __str__(self):
-        return "%s" % self.name
+        return f"#{self.pk} {self.name}"
```
madewulf, line 138: I don't have major objections to these changes, but I'm wondering why you feel it's a better way to format strings

iaso/models/data_source.py
```diff
@@ -161,3 +172,153 @@ def as_report_dict(self):
             group_report[name] = self.orgunit_set.filter(groups__id=ident).count()
         report["groups"] = group_report
         return report
+
+
+class DataSourceVersionsSynchronization(models.Model):
```
madewulf, line 177: This model makes sense to me

iaso/models/org_unit.py
kemar: This format was useful in the admin UI to identify the correct objects.

I'm also relying on this format here https://github.com/BLSQ/iaso/blob/7c294c02e01152969bebe07024a54ae4e28a80fd/iaso/models/data_source.py#L313 but now that I think about it, I may be too brittle… I'm gonna fix this…

iaso/admin.py
kemar: Yes!

iaso/models/org_unit.py
kemar: Fixed in https://github.com/BLSQ/iaso/pull/1836/commits/92379fc661356a01901c198b4da16b123a24ff9b

madewulf (approved):

PR #1857, opened 2024-12-10
WC2-523 Add storage logs to bulk update

bramj:
- Move the storage logs view to a function so we can call it from the bulk upload
- Call said function in the bulk upload
- Add a test for all this

Related JIRA tickets : https://bluesquare.atlassian.net/browse/WC2-523

### Why?

Now that we use the bulk upload on WFP, what happens is (in this order):

1. Mobile app uploads all forms in a zip file
2.  Mobile app uploads the storage logs

Since the first step is executed in an async job, the entity is sometimes not yet created when request 2 hits the server.

The solution is to add the storage logs to the bulk upload zip so it's processed as part of the bulk upload.

### How to test?

Difficult... Honestly I'd say if the code looks reasonable, we deploy to QA and test with a new apk that adds the storage logs correctly.

### DIFF
--- a/iaso/api/storage.py
+++ b/iaso/api/storage.py
@@ -356,61 +356,63 @@ class StorageLogViewSet(CreateModelMixin, viewsets.GenericViewSet):
 
         This will also create a new StorageDevice if the storage_id / storage_type / account combination is not found
         """
-        user = request.user
-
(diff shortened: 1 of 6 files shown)


hakifran (approved):
LGTM

PR #1861, opened 2024-12-11
WC2-615 Delete instances and entities: Add option to filter on entity_type_id

bramj:
Context: When we don't have direct DB access, we cannot remove an `EntityType` when there are already entities associated, because the Django admin only allows soft deletion of entities.

Now on WFP we need to remove some incorrect entity types (on QA). There was already a little script, executable via a simple page, to remove all entities and instances on an account. I just updated this to be able to filter on a specific `entity_type_id`.

Related JIRA tickets : https://bluesquare.atlassian.net/browse/WC2-615

### Beautiful page:

image


mathvdh (approved):
Looks good to me!

PR #1881, opened 2024-12-18
IA-3793 validate xlsform question names uniqueness

mestachs:
A lot of features assume uniqueness of question names

Related JIRA tickets : IA-3793

TODO discuss how to display the validation errors ?

- [ ] Are there enough tests

## Doc

## Changes

Drop support for xls in favor of xlsx (note xls is no more supported by other platform like ona)

Add validations on 
- question name 
  - uniqueness (not if there are some repeat_group)
  - not containing blank
- using in formula known question name (relevant, calculation,...)
- referencing in select_one, select_multiple know list_name in choices

## How to test

launch the seed command 

upload forms from the jira or the fixtures folder see iaso/tests/fixtures/odk_invalid_xlsform.xlsx

## Print screen / video

## Notes

we don't want to support xls only xlsx which is the norm (and not no more working on ona also)

```
fix: validate xlsform when creating form version

Refs: IA-3793
```


iaso/odk/validator.py
```diff
@@ -0,0 +1,177 @@
...
+
+
+def get_list_name_from_select(question):
+    q_type = question.get("type")
+    if "select one from " in q_type:
+        return q_type.split("select one from ")[1]
+    if "select_one" in q_type:
+        return q_type.split("select_one ")[1]
+    if "select one" in q_type:
+        import pdb
```
mathvdh, line 65: Not sure we want to leave that ?

mathvdh (approved):
Except for my comment below, I have another question : When we release this, and people try to edit existing form versions (that are invalid according to this validation) they won't be able to save anymore, right ? We might expect a wave of problems .. Shouldn't we try to make a copy of the prods and run this and try to save all the existing versions and see the amplitude of problems ? Or I'm over stressing for nothing ?
Otherwise very good work !

PR #1883, opened 2024-12-18
IA-3812 Fix: Call to JS bundle on Superset embedding

bramj:
Related JIRA tickets : https://bluesquare.atlassian.net/browse/IA-3812


tdethier (approved):
Was deployed on "file active" project, it works!

LGTM :+1:

PR #1889, opened 2024-12-20
IA-3433 Tasks view: Add extra filters on type, status and username (launcher)

bramj:
This PR improves the tasks. Specifically, it:
- Adds a new column `created_by` on the Task model to differentiate between who did the initial launch, and who did a potential re-launch. 
- Adds the following filters to the tasks page:
  - type (the task's "name")
  - status
  - user (both creator as well as last launcher are take into consideration)
  - start and end date (of any of the timestamps `created_at`, `started_at` and `ended_at`
- General refactoring:
  - File structure of Tasks api endpoint, extracted `filters.py` and`serializers.py`
  - DRY up some Django API filtering code
  - Typescriptify some JS
- Add test for Tasks API filtering

### Screenshot

image


iaso/api/deduplication/filters.py
```diff
@@ -166,23 +163,15 @@ def filter_queryset(self, request, queryset, view):
...
 
-        date_format = "%d-%m-%Y"
-
         if start_date:
             try:
-                start_date_dt = datetime.strptime(start_date, date_format)
-                start_date_dt = timezone.make_aware(start_date_dt, timezone.get_default_timezone())
-                start_date_dt = start_date_dt.replace(hour=0, minute=0, second=0)
-                queryset = queryset.filter(analyze__created_at__gte=start_date_dt)
+                queryset = queryset.filter(analyze__created_at__gte=date_string_to_start_of_day(start_date))
```
kemar, line 168: We may be able to use a string and only filter on the `date` part of the `datetime` field.

E.g. (assuming the date format is correct):

```python
queryset.filter(analyze__created_at__date__gte="2021-05-23")

queryset.filter(analyze__created_at__date__lte="2021-05-23")
```

This may be easier to read and may save some code.

iaso/models/base.py
```diff
@@ -274,11 +275,18 @@ class Task(models.Model):
 
     class Meta:
         ordering = ["-created_at"]
+        indexes = [
+            models.Index(fields=["account"]),
```
```suggestion
```

This is not necessary because a database index is [automatically created](https://docs.djangoproject.com/en/5.1/ref/models/fields/#foreignkey) on the `ForeignKey` (unless `db_index=False` is set).

iaso/models/base.py
```diff
@@ -274,11 +275,18 @@ class Task(models.Model):
...
+            models.Index(fields=["created_at"]),
+            models.Index(fields=["created_by"]),
```
```suggestion
```

Same thing as above.

beanstalk_worker/__init__.py
```diff
@@ -56,6 +56,7 @@ def wrapper(*args, **kwargs):
                 task = Task()
                 user = kwargs.pop("user")
                 task.account = user.iaso_profile.account
+                task.created_by = user
                 task.launcher = user
```
kemar, line 60: What I understand is that we assume that the creator and the launcher are always the same person. Is this correct?

iaso/tests/api/test_tasks.py
```diff
@@ -4,32 +4,34 @@
...
 
 
-@time_machine.travel(DT, tick=False)
+@time_machine.travel(CREATED_AT, tick=False)
 class TaskSerializerSerializerTestCase(TestCase):
     """
     Test Task serializer.
     """
 
+    maxDiff = None
```
kemar, line 21: TBH I'm always tempted to put that in the base class…

iaso/api/tasks/views.py
```diff
@@ -1,92 +1,75 @@
...
     serializer_class = TaskSerializer
     results_key = "tasks"
     queryset = Task.objects.all()
     http_method_names = ["get", "patch", "head", "options", "trace"]
 
     def get_queryset(self):
         profile = self.request.user.iaso_profile
-        order = self.request.query_params.get("order", "created_at").split(",")
-        return Task.objects.select_related("launcher").filter(account=profile.account).order_by(*order)
+        return Task.objects.select_related("created_by", "launcher").filter(account=profile.account)
```
kemar, line 71: We could add `user__iaso_profile` because it's used in `ordering_fields`.

```suggestion
        return Task.objects.select_related("created_by", "launcher", "user__iaso_profile").filter(account=profile.account)
```

iaso/api/tasks/views.py
```diff
@@ -1,92 +1,75 @@
...
         HasPermission(permission.DATA_TASKS),
     ]  # type: ignore
+    filter_backends = [
+        filters.OrderingFilter,
+        django_filters.rest_framework.DjangoFilterBackend,
+        TaskTypeFilterBackend,
+        UsersFilterBackend,
+        StartEndDateFilterBackend,
+        StatusFilterBackend,
+    ]
```
kemar, line 53: Just for you information (no need to rewrite that) I like to do that via [django-filter](https://django-filter.readthedocs.io/en/latest/guide/rest_framework.html).

It has built-in validation and you don't need a bunch of different classes.

You have [various examples](https://github.com/BLSQ/iaso/blob/5c7e90d95dd93f6cfc1b0f8e4414a3916a552800/plugins/polio/api/chronogram/filters.py#L47) in the code base.

hat/assets/js/apps/Iaso/domains/app/translations/fr.json
```diff
@@ -1357,7 +1357,8 @@
     "iaso.tasks.killed": "Arrêtée",
     "iaso.tasks.killSignalSent": "Signal d'arrêt envoyé",
     "iaso.tasks.killTask": "Arrêter la tâche",
-    "iaso.tasks.launcher": "Utilisateur",
+    "iaso.tasks.user": "Utilisateur",
+    "iaso.tasks.last_launch_by": "Dernière execution :",
```
kemar: Excuse my French 😄 

```suggestion
    "iaso.tasks.last_launch_by": "Dernière exécution :",
```

kemar (changes requested):
Good job.

I'd say that you can safely remove 2 redundant indexes and add a relation in the `select_related` method.

The rest is optional.

iaso/api/deduplication/filters.py
bramj: I like the simplicity, but would that take correctly into account the timezone? :thinking:

iaso/models/base.py
bramj: Good point, thanks!

beanstalk_worker/__init__.py
bramj: This is when the task is being created, so here indeed the creator and launcher will be the same person.

It's because of the "relaunch" functionality that I added the creator. When relaunching, only the launcher is updated (see https://github.com/BLSQ/iaso/blob/main/iaso/api/tasks/__init__.py#L135)

iaso/tests/api/test_tasks.py
bramj: ~~Agreed, I'll do it~~

EDIT: Not doing it now, would pollute the PR with a lot of changed files

iaso/api/tasks/views.py
bramj: Oh I like that, thanks! TIL

iaso/api/deduplication/filters.py
kemar: Correct me if I'm wrong but it seems that the incoming format of `request.query_params.get("start_date")` is a date string (i.e. `"%d-%m-%Y"`) without a clearly defined TZ nor any time information.

If that's the case, you'll never be able to get an accurate "TZ aware" date or datetime.

You may as well rely on `__date__`.

kemar (approved):
LGTM

iaso/api/deduplication/filters.py
bramj: @kemar You're right, actually ideally the user should have a timezone on their profile, and then we could use that one to generate the "start of day". If we do it correctly, the current code would be easier to refactor I think. Since it's working as is, okay for you to merge like this?

kemar: @bramj yup OK

PR #1895, opened 2024-12-23
IA-3793 validate xlsform handle or_other and spaces more robustly

mestachs:
From other test : select_one can contains or_other and extra blanks and it shouldn't be considered invalid.

Related JIRA tickets : IA-3793

- [ ] Are there enough tests

## Doc

## Changes

select_one can contains or_other and extra blanks and it shouldn't be considered invalid

- so we strip() (not only rstrip)
- so make get_list_name_from_select more robust against blanks and omit `or_other`

```
fix: validate xlsform handle or_other and spaces more robustly

Refs: IA-3793
```


iaso/odk/validator.py
```diff
@@ -64,13 +63,23 @@ def is_repeat_group(question):
 
 def get_list_name_from_select(question):
     q_type = question.get("type")
```
kemar, line 65: Are we confident enough that `q_type` will never be `None`? Otherwise the code below would fail.

I'd suggest using `q_type  = question.get("type", "")`.

iaso/tests/fixtures/odk_invalid_xlsform_expected_errors.json
```diff
@@ -1006,18 +1006,18 @@
...
+        "message": "choices not found for 'nonexistingchoices' and list_name 'doenst_exist'",
         "question": {
-            "line_number": 98,
-            "type": "calculate",
-            "name": "without_known_ref",
-            "label": "",
+            "line_number": 53,
+            "type": "select_one doenst_exist",
+            "name": "nonexistingchoices",
+            "label": "This should throw a warning there\u2019s no associated choice",
```
kemar, line 1014: I guess `\u2019s` is used on purpose here, or is it a typo?

iaso/tests/fixtures/odk_invalid_xlsform_expected_errors.json
```diff
@@ -1006,18 +1006,18 @@
         "severity": "error"
     },
     {
-        "message": "choices not for 'nonexistingchoices' and list_name 'doenst_exist'",
+        "message": "choices not found for 'nonexistingchoices' and list_name 'doenst_exist'",
```
kemar: I think it's a typo?

```suggestion
        "message": "choices not found for 'nonexistingchoices' and list_name 'doesnt_exist'",
```

kemar (changes requested):

iaso/tests/fixtures/odk_invalid_xlsform_expected_errors.json
mestachs: it's the output of the "json.dumps" so I guess it escape these by default.

iaso/odk/validator.py
mestachs: technically it has survived to `is_select_one(question)` which would have filtered out question without a type.

iaso/tests/fixtures/odk_invalid_xlsform_expected_errors.json
mestachs: fixed the xlsform in fixture

kemar (approved):
LGTM

PR #1899, opened 2025-01-02
IA-3825 API to synchronize two pyramids

kemar:
API to synchronize two pyramids.

Related JIRA tickets : [IA-3825](https://bluesquare.atlassian.net/browse/IA-3825)

## Doc

## Changes

New `/api/datasources/sync/` API endpoint.

See unit tests to check the payload returned in "read" mode.

Note that a synchronization is done in 3 steps:

1. `POST` on `/api/datasources/sync/` to create a new `DataSourceVersionsSynchronization` object
2. `PATCH` on `/api/datasources/sync/{id}/create_json_diff_async/` to compute the differences asynchronously
3. `PATCH` on `/api/datasources/sync/{id}/synchronize_source_versions_async/` to create change requests asynchronously

## How to test

See the unit tests.

## Notes

Check if perms are ok.

[IA-3825]: https://bluesquare.atlassian.net/browse/IA-3825?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tdethier (approved):
Good job :+1: 

I would personally have added a few more tests:

- Testing retrieve & both tasks with an invalid ID
- Testing retrieve & both tasks with a valid ID, but of a `DataSourceVersionsSynchronization` that is in another account
- Testing other permission combos to make sure that you need all 3 permissions to do the operations (but just on a single endpoint, I think that's enough)

Also, I don't know if it makes sense for this use case, but since there are 3 steps for doing a synchronization, should we check that step 2 has already been done at the beginning of the step 3 code?

kemar:
> Also, I don't know if it makes sense for this use case, but since there are 3 steps for doing a synchronization, should we check that step 2 has already been done at the beginning of the step 3 code?

Good point.

You cannot execute step 2 if step 1 hasn't been executed because you would get a 404.

I added a check in step 3.

I also unwrapped step 2 from its backgroud task after a discussion with @beygorghor to make it easier to implement in the UI.

Last but not least, I added a Serializer for the "diff params" which I completely forgot!

iaso/api/data_source_versions_synchronization/serializers.py
```diff
@@ -0,0 +1,133 @@
...
+    """
+    Validate the parameters of DataSourceVersionsSynchronization.create_json_diff().
+    """
+
+    # Version to update.
+    source_version_to_update_validation_status = serializers.ChoiceField(
+        choices=OrgUnit.VALIDATION_STATUS_CHOICES, required=False, default=None
+    )
+    source_version_to_update_top_org_unit = serializers.PrimaryKeyRelatedField(
+        queryset=OrgUnit.objects.all(), required=False, default=None
```
tdethier: Should we filter OrgUnits here with `filter_for_user_and_app_id`? I never know if it's better do to it here or somewhere else

iaso/api/data_source_versions_synchronization/serializers.py
```diff
@@ -0,0 +1,133 @@
...
+    )
+    source_version_to_update_org_unit_types = serializers.PrimaryKeyRelatedField(
+        queryset=OrgUnitType.objects.all(), many=True, required=False, default=None
```
tdethier, line 106: Same comment about filtering :arrow_up:

iaso/api/data_source_versions_synchronization/serializers.py
```diff
@@ -0,0 +1,133 @@
...
+    )
+    # Version to compare with.
+    source_version_to_compare_with_validation_status = serializers.ChoiceField(
+        choices=OrgUnit.VALIDATION_STATUS_CHOICES, required=False, default=None
+    )
+    source_version_to_compare_with_top_org_unit = serializers.PrimaryKeyRelatedField(
+        queryset=OrgUnit.objects.all(), required=False, default=None
```
tdethier: Same comment about filtering :arrow_up:

iaso/api/data_source_versions_synchronization/serializers.py
```diff
@@ -0,0 +1,133 @@
...
+    )
+    source_version_to_compare_with_org_unit_types = serializers.PrimaryKeyRelatedField(
+        queryset=OrgUnitType.objects.all(), many=True, required=False, default=None
```
tdethier, line 114: Same comment about filtering :arrow_up:

iaso/models/data_source.py
```diff
@@ -204,6 +204,8 @@ class DataSourceVersionsSynchronization(models.Model):
 
     """
 
+    SYNCHRONIZABLE_FIELDS = ["name", "parent", "opening_date", "closed_date", "groups"]
```
tdethier, line 207: I would maybe check in `create_json_diff()` that the field names received are all part of this list

tdethier (approved):
Just two comments, but otherwise LGTM :+1:

iaso/api/data_source_versions_synchronization/serializers.py
kemar: That's an idea ;) I'm pushing this to handle that https://github.com/BLSQ/iaso/pull/1899/commits/0555cc8b9ac1746934b6e1c79d3e08ed430922df

iaso/api/data_source_versions_synchronization/serializers.py
kemar: Fixed in https://github.com/BLSQ/iaso/pull/1899/commits/0555cc8b9ac1746934b6e1c79d3e08ed430922df

iaso/api/data_source_versions_synchronization/serializers.py
kemar: That's less of an issue with `OrgUnitType`. It will be filtered anyway later in the code.

iaso/models/data_source.py
kemar: This won't pass the serializer's validation, so I think we're good.

tdethier (approved):
LGTM :+1:

PR #1906, opened 2025-01-08
POLIO-1774: delete rounds without campaign

quang-le:
Polio had rounds without campaign, which shouldn't be possible

Related JIRA tickets :  POLIO-1774

- [x] Are there enough tests

## Doc

NA

## Changes

- Update polio API to delete rounds that are removed from campaign, e.g: remove round 3, keep only rounds 1 & 2.
- Add method to make sure `RoundScope` and associated `Group` are deleted as well when a `Round` is deleted
- Add migration deleting dangling rounds
- Add tests

## How to test

In the UI:

- Add a round to a campaign
- Save
- check the API response and find the new round ID
- find the new round in the django admin
- In the UI: delete the new round and save
- Refresh the django admin: the new round should be gone

Alternately, you can check directly on the DB

## Print screen / video

N/A

```
fix: empty instance pop up

Refs: IA-3665
```


plugins/polio/models/base.py
```diff
@@ -149,6 +151,17 @@ class Meta:
...
 
+# Signal to delete the Group when the Round is deleted
+@receiver(post_delete, sender=RoundScope)
+def delete_group_on_round_delete(sender, instance, **kwargs):
+    """
+    Delete the associated Group when the RoundScope instance is deleted,
+    which happens when a Round is deleted.
+    """
+    if instance.group:
+        instance.group.delete()
```
kemar: There is a rule of thumb with signals: use them as a last resort.

The reason is that after adding a bunch of signals, projects start to turn into knotted hairballs that can't be untangled.

Do you think it would be possible to move this either into `RoundScope.delete()` or in a custom model manager method?

Otherwise, let's move it into a proper `signals.py` file to make it obvious and easy to spot.

plugins/polio/tests/models/test_round_models.py
```diff
@@ -0,0 +1,64 @@
...
+        cls.ou_type_district = cls.create_org_unit_type(name="DISTRICT", projects=[cls.project])
+        cls.obr_name = "DRC-DS-XXXX-TEST"
+        # campaign
+        cls.campaign, cls.rnd1, cls.rnd2, cls.rnd3, cls.country, cls.district = cls.create_campaign(
+            cls.obr_name,
+            cls.account,
+            cls.source_version,
+            cls.ou_type_country,
+            cls.ou_type_district,
+            "GROLAND",
```
kemar, line 31: Is this better than "Hydroponic gardens"? 🤣

plugins/polio/tests/test_api.py
```diff
@@ -26,7 +26,7 @@ def setUpTestData(cls):
         cls.now = now()
         cls.source_version_1 = m.SourceVersion.objects.create(data_source=cls.data_source, number=1)
         cls.account = polio_account = Account.objects.create(name="polio", default_version=cls.source_version_1)
-        cls.yoda = cls.create_user_with_profile(username="yoda", account=polio_account, permissions=["iaso_forms"])
+        cls.user = cls.create_user_with_profile(username="yoda", account=polio_account, permissions=["iaso_forms"])
```
kemar, line 29: 👍

kemar (changes requested):
Almost LGTM.

I have a concern with the usage of signals.

plugins/polio/tests/models/test_round_models.py
quang-le: Totally, and this is a hill I'm ready to die on 😁

More seriously, the name field don't matter that much as long as the variable/property is correctly named , I think

plugins/polio/models/base.py
quang-le: I'll try it

kemar (approved):
LGTM

PR #1912, opened 2025-01-10
IA-3840: Profiles location children filter

beygorghor:
Users : Checkbox "Users with access to children org unit only" work only for direct children, while it should work for all children (N-1 N-2, etc)

Related JIRA tickets : IA-3840

- [x] Are there enough tests

## Doc

## Changes

Adapt profiles filtering to return Users with access to children org unit only on  all children

## How to test

Go to Users page
Link a user to a lower level org unit. 
Click on Users with access to children org unit only
Select a parent of this org unit, not the direct parent.
Search, you should find your user

## Print screen / video
Screenshot 2025-01-10 at 16 30 44
Screenshot 2025-01-10 at 16 30 20

## Notes

```
fix: empty instance pop up

Refs: IA-3665
```

### DIFF
--- a/iaso/api/profiles/profiles.py
+++ b/iaso/api/profiles/profiles.py
@@ -118,7 +118,6 @@ def get_filtered_profiles(
 
             if not parent:
                 queryset = queryset_current
-
             else:
                 queryset = (
                     original_queryset.filter(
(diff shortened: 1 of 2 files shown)


kemar (approved):
LGTM

PR #1928, opened 2025-01-20
WC2-594 + IA-3745 + IA-3841 Improvements and fixes for multi-account users

bramj:
This PR contains several improvements for multi-account users:

- Search on the users page will also search the tenant user's main user if any. Multi-account users are displayed with the main user's first/last/user-name instead of the real account's `auth_user`. E.g. I have a main user `bram`, but I have `bjans_ihp` or `bjans_civ_carte_sanitaire` depending on the account. On the user's page, I am shown as `bram`. With this change, search on `bram`, will also show me in the results.
- For the same reason, we don't allow editing of the personal data of multi-account users (see screenshot)
- Small fix: don't require `SOURCES` permission for account switch

Related JIRA tickets:
- https://bluesquare.atlassian.net/browse/WC2-594
- https://bluesquare.atlassian.net/browse/IA-3745
- https://bluesquare.atlassian.net/browse/IA-3841

## Print screen / video

image

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/app/translations/fr.json
+++ b/hat/assets/js/apps/Iaso/domains/app/translations/fr.json
@@ -1487,7 +1487,7 @@
     "iaso.users.selectedOrgUnits": "Unité d'organisation sélectionnées",
     "iaso.users.update": "Mettre l'utilisateur à jour",
     "iaso.users.userAdminOnly": "Edition pour les administrateurs uniquement",
-    "iaso.users.multiAccountUserInfoDisabledWarning": "Cet utilisateur est un utilisateur multicompte. Pour cet utilisateur, vous ne pouvez modifier que les paramètres spécifiques à ce Iaso compte \"{account}\", tels que ses autorisations.",
+    "iaso.users.multiAccountUserInfoDisabledWarning": "Cet utilisateur est un utilisateur multicompte. Pour cet utilisateur, vous ne pouvez modifier que les paramètres spécifiques à ce compte Iaso \"{account}\", tels que ses autorisations.",
     "iaso.users.userPermissions": "Permissions d'utilisateur",
     "iaso.users.userRoleOrgUnitTypeRestrictionWarning": "Des restrictions d'écriture de rôle d’utilisateur s’appliquent à cet utilisateur. Consulter le rôle d’utilisateur en question pour plus de détails",
     "iaso.users.usersHistory": "Historique des utilisateurs",
--- a/hat/assets/js/apps/Iaso/domains/app/translations/fr.json
+++ b/hat/assets/js/apps/Iaso/domains/app/translations/fr.json
@@ -1487,6 +1487,7 @@
     "iaso.users.selectedOrgUnits": "Unité d'organisation sélectionnées",
     "iaso.users.update": "Mettre l'utilisateur à jour",
     "iaso.users.userAdminOnly": "Edition pour les administrateurs uniquement",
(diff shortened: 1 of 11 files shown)


hat/assets/js/apps/Iaso/domains/app/translations/fr.json
```diff
@@ -1484,6 +1484,7 @@
     "iaso.users.selectedOrgUnits": "Unité d'organisation sélectionnées",
     "iaso.users.update": "Mettre l'utilisateur à jour",
     "iaso.users.userAdminOnly": "Edition pour les administrateurs uniquement",
+    "iaso.users.multiAccountUserInfoDisabledWarning": "Cet utilisateur est un utilisateur multicompte. Pour cet utilisateur, vous ne pouvez modifier que les paramètres spécifiques à ce Iaso compte \"{account}\", tels que ses autorisations.",
```
kemar:
```suggestion
    "iaso.users.multiAccountUserInfoDisabledWarning": "Cet utilisateur est un utilisateur multicompte. Pour cet utilisateur, vous ne pouvez modifier que les paramètres spécifiques à ce compte Iaso \"{account}\", tels que ses autorisations.",
```

iaso/api/accounts.py
```diff
@@ -57,23 +57,30 @@ class AccountViewSet(ModelViewSet):
...
+
+        return [permission() for permission in permission_classes]
+
     @action(detail=False, methods=["patch"], url_path="switch")
     def switch(self, request):
         # TODO: Make sure the account_id is present
         self.permission_classes = [permissions.IsAuthenticated, HasAccountPermission]
         self.check_permissions(request)
-        account_id = request.data.get("account_id", None)
+        account_id = int(request.data.get("account_id", None))
```
kemar:
```suggestion
        account_id = int(request.data["account_id"]) if request.data.get("account_id") else None
```

This could be safer because `int(None)` would raise an error:

```python
>>> int(None)
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```

iaso/api/accounts.py
```diff
@@ -84,6 +91,6 @@ def switch(self, request):
         if user_to_login:
             user_to_login.backend = "django.contrib.auth.backends.ModelBackend"
             login(request, user_to_login)
-            return Response(user_to_login.iaso_profile.account.as_dict())
+            return Response({}, status=status.HTTP_200_OK)
```
kemar, line 95: Are we returning an empty response on purpose?
It seems a bit strange to me.
A comment may help to understand.

iaso/utils/__init__.py
```diff
@@ -127,3 +127,7 @@ def is_mobile_request(request):
     user_agent = request.META.get("HTTP_USER_AGENT", "")
     # Check if the User-Agent string contains keywords typical of mobile devices
     return any(mobile in user_agent.lower() for mobile in ["mobile", "android", "iphone", "ipad", "windows phone"])
+
+
+def is_multi_account_user(user):
+    return hasattr(user, "tenant_user")
```
kemar, line 133: I guess we can't alter the `User` class directly because it belongs to the `django.contrib.auth` application…

[`add_to_class()`](https://code.djangoproject.com/wiki/DevModelCreation#InstallingTheAttributes) could be an option here but it's an internal function and it's a bit of a monkey patch.

In the end, an util function does the job very well 👍

kemar (changes requested):
I think you should fix the potential `int(None)` error.

Also you should ensure that the unit tests are passing.

iaso/api/accounts.py
bramj: Adding a comment, it's on purpose since the frontend doesn't need anything and `account.as_dict()` potentially opens the door to performance issues because of N+1 queries (there are data sources that load versions that load... etc).

kemar (approved):
LGTM

kemar (approved):
LGTM

PR #1936, opened 2025-01-22
IA-3883 try to debug flacky test on github ci

mestachs:
the tests around instances endpoints are flacky.

Related JIRA tickets : IA-3883

- [ ] Are there enough tests

## Doc

## Changes

the default order on updated_at seem to be non deterministic in certain cases
various fixes have implemented across tests generally testing api filters (either using `sorted()` or adding `order=-id`

Here I reuse a shared assertion to apply the same solution (not caring about the order) but also giving more context to actually debug it when assertion are wrong

## How to test

check the CI.

## Print screen / video

image

## Notes

I know I use print but I think it's ok here (only happen on failure so either the test is again flacky, and the additional log will help us, or a modification in the setup changes the results)

Another option would be to throw a new AssertionError with that message but I fear it would make debugging less clear since it will probably "override" the stacktrace or at least hide it more.

```
fix: flacky instances tests

Refs: IA-3883
```

### DIFF
--- a/iaso/tests/api/test_instances.py
+++ b/iaso/tests/api/test_instances.py
@@ -1670,17 +1670,29 @@ class InstancesAPITestCase(APITestCase):
         response = self.client.get(
             "/api/instances/", {"planningIds": planning_1.id}, headers={"Content-Type": "application/json"}
         )
-        self.assertJSONResponse(response, 200)
-        self.assertValidInstanceListData(response.json(), 1)
-        self.assertEqual(response.json()["instances"][0]["id"], instance_1.id)
+
+        self.assertInstanceListContainsStrictly(response, [instance_1])
+
         # it should return all of the instances
         self.client.force_authenticate(self.yoda)
         response = self.client.get("/api/instances/?order=-id", headers={"Content-Type": "application/json"})
-        self.assertJSONResponse(response, 200)
-        self.assertValidInstanceListData(response.json(), 9)
 
-        self.assertEqual(response.json()["instances"][0]["id"], instance_2.id)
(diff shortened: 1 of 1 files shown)


kemar (approved):
This LGTM.

However, here is also an opinionated answer:

1. you're reinventing the wheel
    - because you can use built-in tools like `assertListEqual` or `assertCountEqual` without writing a new method full of `print` statements
2. you are mixing concerns
    - i.e. testing the `status_code` has nothing to do with the title of this method `assertInstanceListContainsStrictly`
3. you lose the ability to expect an ordering
    - the original code seems to be waiting for an explicit order

The same thing can be done with built in tools.

**E.g. when order matters**:

```python

instance_ids = [x.get("id") for x in response.json()["instances"]]
expected_instance_ids = [self.instance_1.pk, self.instance_4.pk, self.instance_5.pk, self.instance_8.pk]
self.assertListEqual(instance_ids, expected_instance_ids)
```

Result:

```python
AssertionError: Lists differ: [497, 500, 504, 501] != [497, 500, 501, 504]

First differing element 2:
504
501

- [497, 500, 504, 501]
?               -----

+ [497, 500, 501, 504]
?            +++++
```


**E.g. when order doesn't matter:**

```python

instance_ids = [x.get("id") for x in response.json()["instances"]]
expected_instance_ids = [self.instance_1.pk, self.instance_4.pk, self.instance_5.pk, self.instance_8.pk]
self.assertCountEqual(instance_ids, expected_instance_ids)
```

When first sequence doesn't contain the same elements as second, regardless of their order, an error message lists the differences:

```python
AssertionError: Element counts were not equal:
First has 1, Second has 0:  512
```

PR #1938, opened 2025-01-22
POLIO-1758 Earmarked stock

mathvdh:
Add the possibility to Earmark Stocks, Use the earmarks or put them back to main stock.

Add the models and APIs for it.

Related JIRA tickets : POLIO-1758

- [x] Are there enough tests

## Changes

Add `EarmarkedStock` model in polio models.
Add new urls
Add new `EarmarkedStockViewSet` and `EarmarkedStockSerializer` in stock_management APIs.
Modifications of the calculations in stock_management API to take it into consideration.
Modified the `OutgoingStockMovementViewSet` to add a create method that tries to see if there is a Earmarked Stock and use it when creating a new Form A.
I had some name clash in `OrgUnitNestedSerializer` in iaso/api/storage, I had to rename it to `StorageOrgUnitNestedSerializer` 

## How to test

Without the frontend the only way is either using django admin or via the APIs to create (see tests)

## Notes

The frontend will be added in another PR

```
fix: empty instance pop up

Refs: IA-3665
```


iaso/test.py
quang-le, line 1: ❤️

plugins/polio/api/vaccines/stock_management.py
quang-le, line 1: There's something missing here with the USED earmarked stocks: when I test in the UI (on the front-end branch), if I add a formA, the vials from the formA are deduced from the earmarked stock AND from the usable vials

plugins/polio/api/vaccines/stock_management.py
```diff
@@ -519,6 +652,31 @@ class OutgoingStockMovementViewSet(VaccineStockSubitemBase):
...
+                    vaccine_stock=movement.vaccine_stock,
+                    campaign=movement.campaign,
+                    round=movement.round,
+                    earmarked_stock_type=EarmarkedStock.EarmarkedStockChoices.USED,
+                    vials_earmarked=vials_earmarked_used,
+                    doses_earmarked=doses_earmarked_used,
+                    comment="Created from Form A submission",
+                )
+
+        return response
```
quang-le: I think we need something similar with incident reports when the number of unused created > number of available usable

plugins/polio/api/vaccines/stock_management.py
```diff
@@ -546,6 +704,44 @@ class DestructionReportViewSet(VaccineStockSubitemBase):
...
+            "earmarked_stock_type",
+            "vials_earmarked",
+            "doses_earmarked",
+            "comment",
+            "created_at",
+            "updated_at",
+        ]
+
+    def get_doses_earmarked(self, obj):
+        return obj.vials_earmarked * DOSES_PER_VIAL[obj.vaccine_stock.vaccine]
```
quang-le: Do we need this, when we have the field on the Model?

quang-le (changes requested):
See the comments. I also get a 500 when trying to edit an existing earmarked stock

Screenshot 2025-01-23 at 18 17 54

Almost there!

quang-le (changes requested):
Used earmarked stock appears twice in unusable table: once for earmarked, once for form A

Screenshot 2025-02-03 at 16 46 01

plugins/polio/models/base.py
```diff
@@ -1548,6 +1548,66 @@ class Meta:
         ]
 
 
+class EarmarkedStock(models.Model):
+    class EarmarkedStockChoices(models.TextChoices):
+        CREATED = "created", _("Created")  #     1. Usable -> Earmark
+        USED = "used", _("Used")  #     2. Earmarked -> Used
+        RETURNED = "returned", _("Returned")  #     3. Used -> Usable
```
kemar: These comments are unclear to me: e.g. what does `# 3. Used -> Usable` mean for `RETURNED`?

Is there a way to make them clearer for somebody new to the topic?

plugins/polio/models/base.py
```diff
@@ -1548,6 +1548,66 @@ class Meta:
...
+
+    earmarked_stock_type = models.CharField(
+        max_length=20, choices=EarmarkedStockChoices.choices, default=EarmarkedStockChoices.CREATED
+    )
+    vaccine_stock = models.ForeignKey(VaccineStock, on_delete=models.CASCADE, related_name="earmarked_stocks")
+    campaign = models.ForeignKey(Campaign, on_delete=models.CASCADE)
+    round = models.ForeignKey(Round, on_delete=models.CASCADE)
+    form_a = models.ForeignKey(
+        OutgoingStockMovement, on_delete=models.CASCADE, null=True, blank=True, related_name="earmarked_stocks"
+    )
```
kemar, line 1555: Why are we calling the field `form_a` when the model is named `OutgoingStockMovement`?

Is this a business rule?

plugins/polio/models/base.py
```diff
@@ -1548,6 +1548,66 @@ class Meta:
...
+
+    vials_earmarked = models.PositiveIntegerField()
+    doses_earmarked = models.PositiveIntegerField()
+
+    comment = models.TextField(blank=True, null=True)
```
kemar:
```suggestion
    comment = models.TextField(blank=True)
```

`blank=True` is enough for text fields, see https://sentry.io/answers/django-difference-between-null-and-blank/

plugins/polio/models/base.py
```diff
@@ -1548,6 +1548,66 @@ class Meta:
...
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+
+    def __str__(self):
+        return f"Earmarked {self.vials_earmarked} vials for {self.campaign.obr_name} Round {self.round.number}"
+
+    @classmethod
+    def get_available_vials_count(cls, vaccine_stock: VaccineStock, _round: Round):
```
kemar, line 1575: This could live in a custom `manager` instead of in a `@classmethod`.

But that point doesn't really matter.

What really matters is that you should have some unit tests on this specific method (`get_available_vials_count`) and maybe separate `matching_earmarks_plus` and `matching_earmarks_minus` to be 100% sure that the outputs are accurate and consistent.

plugins/polio/models/base.py
```diff
@@ -1548,6 +1548,66 @@ class Meta:
...
+        CREATED = "created", _("Created")  #     1. Usable -> Earmark
+        USED = "used", _("Used")  #     2. Earmarked -> Used
+        RETURNED = "returned", _("Returned")  #     3. Used -> Usable
+
+    class Meta:
+        indexes = [
+            models.Index(fields=["vaccine_stock", "campaign"]),
+            models.Index(fields=["created_at"]),
+            models.Index(fields=["round"]),
+        ]
```
kemar: For consistency, `Meta` options should be positioned after the fields in the model.

But since we don't have a convention for that, it doesn't really matter yet.

plugins/polio/api/vaccines/stock_management.py
```diff
@@ -261,6 +297,49 @@ def get_list_of_usable_vials(self, end_date=None):
...
+                #     {
+                #         "date": stock.created_at.date(),
+                #         "action": f"Earmarked used for {stock.campaign.obr_name} Round {stock.round.number}",
+                #         "vials_in": stock.vials_earmarked,
+                #         "doses_in": stock.doses_earmarked,
+                #         "vials_out": None,
+                #         "doses_out": None,
+                #         "type": "earmarked_stock__used",
+                #     }
+                # )
```
kemar: I would remove code that does nothing.

```suggestion
```

plugins/polio/api/vaccines/stock_management.py
```diff
@@ -122,6 +125,24 @@ def get_total_of_unusable_vials(self, end_date=None):
 
         return total_vials_in, total_doses_in
 
+    def get_total_of_earmarked(self, end_date=None):
```
kemar, line 128: Correct me if I'm wrong but I can't find any unit test of the methods in `VaccineStockCalculator`?

Adding unit tests at this specific level (rather than only at the API views level) would certainly ensure robustness.

plugins/polio/api/vaccines/stock_management.py
```diff
@@ -332,6 +420,73 @@ def get_list_of_unusable_vials(self, end_date=None):
...
+                    }
+                )
+
+        return results
+
+    def get_list_of_earmarked(self, end_date=None):
+        earmarked_movements = self.earmarked_stocks
+        if end_date:
+            earmarked_movements = earmarked_movements.filter(created_at__lte=end_date)
+        movements = EarmarkedStockSerializer(earmarked_movements, many=True).data
```
kemar: The use of a serializer here is not obvious. Is there any benefit?

plugins/polio/api/vaccines/stock_management.py
```diff
@@ -546,6 +769,42 @@ class DestructionReportViewSet(VaccineStockSubitemBase):
...
+            campaign = Campaign.objects.get(
+                obr_name=campaign_obr_name, account=self.context["request"].user.iaso_profile.account
+            )
+            return campaign
+        return None
+
+
+class EarmarkedStockViewSet(VaccineStockSubitemEdit):
+    serializer_class = EarmarkedStockSerializer
+    model_class = EarmarkedStock
```
kemar, line 821: You should override `get_queryset` to use `select_related` for relations in the `EarmarkedStockSerializer`.

This could avoid a bunch of queries.

plugins/polio/api/vaccines/stock_management.py
```diff
@@ -712,6 +974,36 @@ def get_unusable_vials(self, request, pk=None):
...
+        try:
+            vaccine_stock = self.get_queryset().get(id=pk)
+        except VaccineStock.DoesNotExist:
+            return Response({"error": "VaccineStock not found"}, status=status.HTTP_404_NOT_FOUND)
+
+        end_date = request.query_params.get("end_date", None)
+        if end_date:
+            parsed_end_date = parse_date(end_date)
+            if not parsed_end_date:
+                raise ValidationError("The 'end_date' query parameter is not a valid date.")
```
kemar, line 1016: If you're not using a `Serializer` for validating the incoming data, you should be consistent in the way you return errors.

Here you're using `raise ValidationError` while the two previous errors are returned via `return Response({"error": "Error"}, status=status.HTTP_400_BAD_REQUEST)`.

`raise ValidationError` everywhere looks like a good way to handle that.

plugins/polio/tests/test_vaccine_stock_earmarked.py
```diff
@@ -0,0 +1,375 @@
...
+        found = False
+        for entry in data["results"]:
+            if entry["vials_in"] == 100 and entry["doses_in"] == 2000 and entry["type"] == "earmarked_stock__used":
+                found = True
+                break
+
+        self.assertTrue(found, "Earmarked stock of type USED should appear in unusable vials list")
+
+    def test_earmarked_stock_creation_not_in_unusable_vials(self):
+        """Test that earmarked stock of type CREATEDdoes not appear in unusable vials list"""
```
kemar:
```suggestion
        """Test that earmarked stock of type CREATED does not appear in unusable vials list"""
```

kemar (approved):
Backend LGTM because the logic seems right to me.

Nothing is blocking because this is urgent.

However, with a little more time, I would:

- add tests for `VaccineStockCalculator`
- optimize some queries

plugins/polio/models/base.py
mathvdh: @kemar, you are right, the comment should actually be : # 3. Earmark -> Usable, I will change it

plugins/polio/models/base.py
mathvdh: It's because they are actually calling it Form A almost everywhere, technically it's a "Formulaire A for Outgoing Stock Movement"

plugins/polio/api/vaccines/stock_management.py
mathvdh: It's not my code, I will have a look

mathvdh: I removed the serializer

quang-le (approved):

PR #1954, opened 2025-02-04
IA-3898: Groups section items overflow container, scrollbar over elements.

beygorghor:
Unless zooming out on my page and making content harder to read, I can't see my dropdown list anymore when there are many groups selected on an OU page
image

Related JIRA tickets : IA-3898

- [ ] Are there enough tests

## Changes

Override body css  overflow to make options available

## How to test

Make sure you have lot's of groups available. assign a lot to an org unit and try to edit the groups option.
You should be able to see the whole list

## Print screen / video

https://github.com/user-attachments/assets/0e9bca0e-f4dc-4441-accb-f20c08dc9579

## Notes

```
fix: empty instance pop up

Refs: IA-3665
```

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/orgUnits/components/OrgUnitInfos.tsx
+++ b/hat/assets/js/apps/Iaso/domains/orgUnits/components/OrgUnitInfos.tsx
@@ -27,6 +27,12 @@ import { OrgUnitMultiReferenceInstances } from './OrgUnitMultiReferenceInstances
 import { useGetOrgUnit } from './TreeView/requests';
 
 const useStyles = makeStyles(theme => ({
+    '@global': {
+        body: {
+            overflowX: 'hidden !important',
+            overflowY: 'auto !important',
+        },
+    },
     speedDialTop: {
         top: theme.spacing(12.5),
     },


hat/assets/js/apps/Iaso/domains/orgUnits/components/OrgUnitInfos.tsx
```diff
@@ -27,6 +27,12 @@ import { OrgUnitMultiReferenceInstances } from './OrgUnitMultiReferenceInstances
 import { useGetOrgUnit } from './TreeView/requests';
 
 const useStyles = makeStyles(theme => ({
+    '@global': {
```
kemar, line 30: I guess this syntax comes from [`jss-plugin-global`](https://cssinjs.org/jss-plugin-global/?v=v10.10.0)?

As far as I understand, this rule will be applied to `body` element on on pages containing this component. Is this correct?

kemar (approved):
LGTM

hat/assets/js/apps/Iaso/domains/orgUnits/components/OrgUnitInfos.tsx
beygorghor: Yes indeed. Main body is limited to the screen size on the whole app to allow overflow on other components. In this specific case the dropdown is going out of the body then not visible. Applying the change on all the page leads to double scroll issues. So the workaround here is to modify the body style only if this component is visible. I'm not really a fan of it, but it's the faster solution.

PR #1963, opened 2025-02-10
POLIO-1722 add endpoint to refresh OpenHexa VRF data

quang-le:
PowerBi can't refresh data handled by OpenHexa pipelines so we need an endpoint in iaso to handle that

Related JIRA tickets : IA-XXX, WC2-XXX, POLIO-1722

- [] Are my typescript files well typed
- [x] Are there enough tests

## Doc
NA

## Changes
 - Add an endpoint using `ExternalTaskViewset`
 - Override list method to behave like POST, because powerBI can't POST

## How to test

- Login to OpenHexa's polio workspace
- Deploy the branch on staging
- go to api/polio/tasks/refreshvrf/:
    - A new task should be visible in the tasks list
    - In openHexa, the pipeline should have run, with the task id ans "staging" as parameters
    - in OpenHExa, the DB table `VaccineRequestFormsStaging` should have been updated

## Print screen / video

Screenshot 2025-02-10 at 12 31 28

## Notes

Once the PR is merged, there are additional steps to perform to switch to prod:
- Update the powerBI dashboard so that the refresh button attacks the new API. Make a copy of the dashbaord that points to staging data first
- Then: 
    - Add the vrf-pipeline-config` to prod. It can be copied from staging, with just the `target` adpated.
    - Update the prod dashboard
    - Delete the old Jupyter notebook, since the code now lives on github  

```
fix: empty instance pop up

Refs: IA-3665
```

### DIFF
--- a/plugins/polio/tasks/api/refresh_vrf_dashboard_data.py
+++ b/plugins/polio/tasks/api/refresh_vrf_dashboard_data.py
@@ -37,3 +37,10 @@ class RefreshVrfDataViewset(ExternalTaskModelViewSet):
         task.status = status
         task.save()
         return Response({"task": TaskSerializer(instance=task).data})
+
+    def get_queryset(self):
+        # user = self.request.user
+        user = User.objects.filter(username="openhexa_iaso_user")
+        account = user.iaso_profile.account
+        queryset = Task.objects.filter(account=account).filter(external=True)
+        return queryset
--- a/plugins/polio/tasks/api/refresh_vrf_dashboard_data.py
+++ b/plugins/polio/tasks/api/refresh_vrf_dashboard_data.py
@@ -22,7 +22,6 @@ class RefreshVrfDataViewset(ExternalTaskModelViewSet):
         # Workaround: we need ro reive a user id an validate it in the serializer
         user = User.objects.get(username="openhexa_iaso_user")
         slug = VRF_CONFIG_SLUG
-        user = request.user
         task = Task.objects.create(
             created_by=user,
             launcher=user,
--- a/plugins/polio/tasks/api/refresh_vrf_dashboard_data.py
+++ b/plugins/polio/tasks/api/refresh_vrf_dashboard_data.py
@@ -19,8 +19,7 @@ class RefreshVrfDataViewset(ExternalTaskModelViewSet):
     # Overriding the list method because powerBI sends a GET request
     # So we have to have a GET that behaves like a POST
     def list(self, request):
-        # Workaround: we need ro reive a user id an validate it in the serializer
-        user = User.objects.get(username="openhexa_iaso_user")
+        user = request.user
         slug = VRF_CONFIG_SLUG
         task = Task.objects.create(
             created_by=user,
@@ -38,8 +37,7 @@ class RefreshVrfDataViewset(ExternalTaskModelViewSet):
         return Response({"task": TaskSerializer(instance=task).data})
 
     def get_queryset(self):
-        # user = self.request.user
-        user = User.objects.get(username="openhexa_iaso_user")
+        user = self.request.user
         account = user.iaso_profile.account
         queryset = Task.objects.filter(account=account).filter(external=True)
         return queryset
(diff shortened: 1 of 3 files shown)


plugins/polio/tasks/api/refresh_vrf_dashboard_data.py
```diff
@@ -0,0 +1,43 @@
...
+
+class RefreshVrfDataViewset(ExternalTaskModelViewSet):
+    http_method_names = ["get", "patch"]
+    model = Task
+    task_name = VRF_TASK_NAME
+    config_slug = VRF_CONFIG_SLUG
+
+    # Overriding the list method because powerBI sends a GET request
+    # So we have to have a GET that behaves like a POST
+    def list(self, request):
```
tdethier, line 21: Is there a reason why you chose to override `list` instead of doing a new action with `GET` (_e.g._  `dropdown` in groups)?

The comment on lines 19-20 is clear, but it might make more sense to use a specific action instead (especially if later we end up needing to use `list`)

plugins/polio/tests/test_refresh_vrf.py
```diff
@@ -0,0 +1,139 @@
...
+        cls.account.default_version = cls.source_version
+        cls.account.save()
+
+        cls.task1 = m.Task.objects.create(
+            status=RUNNING,
+            account=account,
+            launcher=cls.user,
+            name=VRF_CONFIG_SLUG,
+            started_at=datetime.now(),
+            external=True,
```
tdethier, line 37: Based on the names of tasks in the setup, shouldn't this be `False`?

I'm not quite sure what this attribute does

plugins/polio/tests/test_refresh_vrf.py
```diff
@@ -0,0 +1,139 @@
...
+        )
+        cls.task2 = m.Task.objects.create(
+            status=RUNNING, account=account, launcher=cls.user, name="task 2", external=True
```
tdethier, line 40: Same comment as above :arrow_up:

plugins/polio/tests/test_refresh_vrf.py
```diff
@@ -0,0 +1,139 @@
...
+        user_no_perm = self.create_user_with_profile(username="test user2", account=self.account, permissions=[])
+        self.client.force_authenticate(user_no_perm)
+        response = self.client.get(
+            self.url,
+            format="json",
+        )
+        self.assertJSONResponse(response, 200)
+
+    @patch.object(RefreshVrfDataViewset, "launch_task", mock_openhexa_call_running)
+    def test_create_external_task(self):
```
tdethier, line 71: I would add a `create_external_task_without_perms` test

plugins/polio/tests/test_refresh_vrf.py
```diff
@@ -0,0 +1,139 @@
...
+            self.url,
+        )
+        response = self.assertJSONResponse(response, 200)
+        task = response["task"]
+        self.assertEqual(task["status"], SKIPPED)
+        self.assertEqual(task["launcher"]["username"], self.user.username)
+        self.assertEqual(task["name"], VRF_TASK_NAME)
+
+    @patch.object(RefreshVrfDataViewset, "launch_task", mock_openhexa_call_running)
+    def test_patch_external_task(self):
```
tdethier, line 95: I would add a `patch_external_task_without_perms` test

tdethier (approved):
I would change a few things before merging, but if this is urgent, you can go ahead

plugins/polio/tests/test_refresh_vrf.py
quang-le: It indicates that the task is not run by the worker but an external service, in this case OpenHexa

plugins/polio/tasks/api/refresh_vrf_dashboard_data.py
quang-le: I just wanted to keep the endpoint name as simple and straightforward as possible. Given it's entirely dedicated to refreshing from PowerBI, there's no chance we will need the `list` for anything else

PR #1983, opened 2025-02-18
POLIO-1776: subactivity validation

quang-le:

Related JIRA tickets :  POLIO-1776

- [ ] Are there enough tests

## Doc

NA

## Changes

- Add front-end validation for subactivities dates:
    - start and end-date should be after round start date
    - lqas and im dates should be after subactivity end date
- Added round dates to formik values to be able to access them for validation
- Remove round dates before calling subactivities API
- fix some TS typing
- Add backend check on subactivity API to prevent scopes outside of parent round scope to be added
- Add check in `CampaignSerializer` to remove scopes from subactivities when reducing campaign scope

## How to test

- Go to polio campaigns, create a sub-activity with dates that are all wrong
- Reduce the scope of a campaign to be smaller than the subactivity scope and save. Reload and check the API payload: the subactivity scope should not have org units not part of the parent scope (round or campaign)

## Print screen / video

Screenshot 2025-02-18 at 10 44 21

## Notes

- Validation was left front-end only on the basis that we don't validate round dates back-end side

```
fix: empty instance pop up

Refs: IA-3665
```


hakifran (changes requested):
@quang-le  here is video which show a behavior which seems to me not expected.
[Screencast from 2025-03-07 10-38-17.webm](https://github.com/user-attachments/assets/5c077371-4ec2-4465-ab0a-2bb22613767d)
The thing is that when a new subActivity is created for a given round, it doesn't link to orgUnits.

Another thing is that when the campaign is updated at any field, it try to create automatically a new subActivity.
[Screencast from 2025-03-07 10-44-18.webm](https://github.com/user-attachments/assets/94f27e52-fb21-442f-950d-70723b594708)
[Screencast from 2025-03-07 10-44-59.webm](https://github.com/user-attachments/assets/10cd9039-9271-48c6-aa1b-9e7679d85de4)

hakifran (approved):
LGTM

PR #1991, opened 2025-02-21
IA-3967: Org unit search use default version

beygorghor:
Default version is not used while launching an org unit search

Related JIRA tickets : IA-3967

- [ ] Are there enough tests

## Changes

Set default search option to use also default version of data source if set, source if no default version

## How to test

Use a source without version by default, make sure the search uses only the source id
Add a default version to this source, make source ou search is using this version as deafult

## Print screen / video

https://github.com/user-attachments/assets/fae3b85a-1ec6-4bae-aa24-8c9a49878fb1

```
fix: empty instance pop up

Refs: IA-3665
```


kemar (approved):
LGTM

PR #1994, opened 2025-02-24
WC2-680 Django admin: improve management of multi-account users

bramj:
https://bluesquare.atlassian.net/browse/WC2-680

### Improve the Django UserAdmin:

Several things here:
- Customize the `UserAdmin` to override the base Django one. Note that the existing one was not really working. By unregistering and re-registering the Django one, we can override the main Django one.
- Add the "accounts" column, displaying either: the account the user belongs to, or the list of accounts in case of multi-account users.
- Add default filtering on the user list to not display the  "account_users", users that only exist technically to support
multi-account users. To avoid losing total access to these, add a link to the actual user from the tenant user list.
- Add a custom action on the user details page: "Convert to multi-account" or "Edit multi-account" which bring you to a page where you can _add_ access to accounts for said user. This action creates the correct new users + profiles + links to tenant users.
- Extra: A refactoring to start to break up the >1000 lines `admin.py` file.

admin-multi-account-1
image
admin-multi-account

### DIFF
--- a/hat/templates/admin/auth/user/change_form.html
+++ b/hat/templates/admin/auth/user/change_form.html
@@ -1,0 +1,17 @@
+{% extends "admin/change_form.html" %}
+
+{% block object-tools-items %}
+    {{ block.super }}
+
+    {% if not original.tenant_user %}
+        <li>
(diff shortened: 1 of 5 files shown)


mathvdh (approved):
Great !

PR #1995, opened 2025-02-25
IA-3734 New permission to change the default version of a data source

kemar:
Add a new permission to change the default version of a data source.

Related JIRA tickets : [IA-3734](https://bluesquare.atlassian.net/browse/IA-3734)

## Doc

## How to test

https://github.com/user-attachments/assets/5edb311c-75d6-4d5a-9e1c-913796effffb

[IA-3734]: https://bluesquare.atlassian.net/browse/IA-3734?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iaso/tests/api/test_data_sources.py
```diff
@@ -170,6 +174,73 @@ def test_datasource_create_delete_fail_ok(self):
...
+            "description": self.data_source.description,
+            "created_at": None,
+            "updated_at": None,
+            "default_version": None,
+            "tree_config_status_fields": self.data_source.tree_config_status_fields,
+            "projects": None,
+            "versions": None,
+            "url": None,
+            # Non serializer fields… they should have been part of the serializer.
+            "default_version_id": new_default_version_id,
```
tdethier, line 221: In this payload, we have both `default_version` and `default_version_id`. I understand that `default_version_id` is not in the serializer, but why do we still have `default_version` in there?

Is it because we have a single serializer for all operations and we need more than the ID for displaying information from a GET query? In that case, since we don't need it in PUT,  I would probably remove `default_version` from this payload because it's kinda confusing.

tdethier (approved):
Just a small question in the test, otherwise I fully agree with you about the serializer.

LGTM :+1:

iaso/tests/api/test_data_sources.py
kemar: I didn't remove `default_version` because it's been explicitly defined as a `SerializerMethodField`, see https://github.com/BLSQ/iaso/blob/8dda5510c98cf3755242f36cf0a5bd97388e3b16/iaso/api/data_sources.py#L57-L58

So it might be used by the frontend somehow…

PR #2001, opened 2025-03-03
IA-3999: fix datasource read-only edition

quang-le:
User cannot make read-only datasource writable again

Related JIRA tickets : IA-3999

- [ ] Are there enough tests?

## Doc

N/A

## Changes

- check that read-only is not None i.o truthy

## How to test

Explain how to test your PR.

With the appropriate rights, go to datasources, make one read-only, save, then make it writable again and save again

## Print screen / video

https://github.com/user-attachments/assets/e004cd0f-716d-4b68-9439-5bdedb1a631e

## Notes

No tests added since the issue was flagged as blocking and the fix is easily testable and doesn't break existing tests

```
fix: empty instance pop up

Refs: IA-3665
```

### DIFF
--- a/iaso/api/data_sources.py
+++ b/iaso/api/data_sources.py
@@ -56,7 +56,11 @@ class DataSourceSerializer(serializers.ModelSerializer):
 
     @staticmethod
     def get_default_version(obj: DataSource):
-        return obj.default_version.as_dict_without_data_source() if obj.default_version else None
+        return (
+            obj.default_version.as_dict_without_data_source()
+            if obj.default_version
+            else None
+        )
 
     @staticmethod
     def get_projects(obj: DataSource):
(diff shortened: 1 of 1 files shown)


bmonjoie (approved):
AFAICT, the only change is `if read_only is not None:`  which is good to me.

Adding a test for this would have been nice, though.

Otherwise, LGTM

iaso/api/data_sources.py
```diff
@@ -114,7 +125,7 @@ def update(self, data_source, validated_data):
             data_source.name = name
 
         read_only = validated_data.get("read_only")
-        if read_only:
+        if read_only is not None:
```
quang-le, line 128: Here's the fix: `read_only` is a boolean, so it still needs to be updated when it's `False`

PR #2003, opened 2025-03-04
IA-3980 Earmarked stocks make campaign and round selection not mandatory

mathvdh:
Make the campaign and round optional when creating a Earmark, and add a "temporary campaign name" text field on the model and in the front.

Related JIRA tickets : IA-3980, IA-3982

- [x] Are there enough tests?

## Changes

make the "campaign" and "round" field optional on the EarmarkedStock model and add an optional "temporary_campaign_name" charfield on the same model.

Modify the APis for that.
Modify the frontend to reflect that as well.

## How to test

Try to create a earmarked stock without campaign or round using the dashboard.

```
fix: empty instance pop up

Refs: IA-3665
```


plugins/polio/models/base.py
```diff
@@ -1582,13 +1818,24 @@ class EarmarkedStockChoices(models.TextChoices):
...
-    vaccine_stock = models.ForeignKey(VaccineStock, on_delete=models.CASCADE, related_name="earmarked_stocks")
-    campaign = models.ForeignKey(Campaign, on_delete=models.CASCADE)
-    round = models.ForeignKey(Round, on_delete=models.CASCADE)
+    vaccine_stock = models.ForeignKey(
+        VaccineStock, on_delete=models.CASCADE, related_name="earmarked_stocks"
+    )
+    campaign = models.ForeignKey(
+        Campaign, on_delete=models.CASCADE, null=True, blank=True
+    )
+    temporary_campaign_name = models.CharField(max_length=255, null=True, blank=True)
```
quang-le: I have to mention that we shouldn't have both `null=True` and `blank=True` for string fields as it creates 2 values for 'no data'. 

Django doc: https://docs.djangoproject.com/en/5.1/ref/models/fields/#field-options

plugins/polio/js/src/domains/VaccineModule/StockManagement/StockVariation/Modals/validation.ts
quang-le, line 1: We need a custom check to ensure we have either a campaign or a temporary campaign name.

We should probably have such a validation on the back end as well

quang-le (changes requested):
I'm not sure it wouldn't be better to also have a temporary round. Maybe double-check this with PM.

Also there's a mean bug with the validation of the dialog:

https://github.com/user-attachments/assets/34832508-296a-4218-a3de-564384249a11

plugins/polio/js/src/domains/VaccineModule/StockManagement/StockVariation/Modals/validation.ts
mathvdh: @quang-le After discussion with Laure she thinks we should allow them to not choose a campaign AND leave the temporary campaign empty as well ..

quang-le (approved):

PR #2005, opened 2025-03-04
IA-3996: Group filter doesn't seem to work on OU list

butofleury:

Related JIRA tickets : [IA-3996](https://bluesquare.atlassian.net/browse/IA-3996)

- [x] Are there enough tests?

## Doc

## Changes

Added distinct id when filter org unit by groups

## How to test

From org unit list, try to filter org unit by selecting several groups.

## Print screen / video

[Iaso-Org-units-Group_and_Type-filter.webm](https://github.com/user-attachments/assets/d0f64e24-e534-497d-be23-06ae2f0b1e50)

## Notes

```
fix: empty instance pop up

Refs: IA-3665
```

[IA-3996]: https://bluesquare.atlassian.net/browse/IA-3996?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iaso/api/org_unit_search.py
+++ b/iaso/api/org_unit_search.py
@@ -69,7 +69,9 @@ def build_org_units_queryset(queryset, params, profile):
                 queryset = queryset.filter(source_ref__in=[])
                 print("Failed parsing refs in search", search)
         else:
-            queryset = queryset.filter(Q(name__icontains=search) | Q(aliases__contains=[search]))
+            queryset = queryset.filter(
+                Q(name__icontains=search) | Q(aliases__contains=[search])
+            )
 
     if group:
         if isinstance(group, str):
@@ -78,7 +80,7 @@ def build_org_units_queryset(queryset, params, profile):
             group_ids = [group]
         else:
             group_ids = group
-        queryset = queryset.filter(groups__in=group_ids)
+        queryset = queryset.filter(groups__in=group_ids).distinct("id")
 
     if source:
         source = DataSource.objects.get(id=source)
@@ -103,7 +105,9 @@ def build_org_units_queryset(queryset, params, profile):
         queryset = queryset.filter(instance__created_at__lte=date_to)
 
     if date_from is not None and date_to is not None:
-        queryset = queryset.filter(instance__created_at__range=[date_from, date_to]).distinct("id")
+        queryset = queryset.filter(
+            instance__created_at__range=[date_from, date_to]
+        ).distinct("id")
 
     if has_instances is not None:
         if has_instances == "true":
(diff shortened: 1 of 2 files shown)


beygorghor (changes requested):
I don't see what this PR is changing from the actual behavior. I can reprodce the same thing as on your branch on staging without your changes. Maybe @kemar @mathvdh or @tdethier can have a look on this

iaso/api/org_unit_search.py
```diff
@@ -78,7 +80,7 @@ def build_org_units_queryset(queryset, params, profile):
             group_ids = [group]
         else:
             group_ids = group
-        queryset = queryset.filter(groups__in=group_ids)
+        queryset = queryset.filter(groups__in=group_ids).distinct("id")
```
beygorghor, line 83: Is this the only actual change ?

butofleury: yes @beygorghor , yes, it is the only  change I made. 

I added that to allow getting single org unit by id when several groups are selected in the filter. For instance when an org unit belong to 2 groups and those groups are selected in the filter, I've noticed the org unit appears twice in the list.

tdethier:
I took a look at this issue. 

Initially, I struggled to trigger the same error. I eventually managed to get it by putting some org units in multiple groups and selecting these groups in the filters. 

If I do this on `main`, here's what I get:
image


If I do this on this branch, the result is correct (same parameters):
image

So, I think this PR will solve the issue with groups. I don't think that `distinct` will create any issue here, but maybe @kemar might spot something else.

:warning: **However**: kudos for adding a unit test, but unfortunately the test setup is wrong. Since there aren't any org units that have multiple groups, the test will always be successful (I tried running this new test on `main` without this fix and it passed)

butofleury:
> I took a look at this issue.
> 
> Initially, I struggled to trigger the same error. I eventually managed to get it by putting some org units in multiple groups and selecting these groups in the filters.
> 
> If I do this on `main`, here's what I get: image
> 
> If I do this on this branch, the result is correct (same parameters): image
> 
> So, I think this PR will solve the issue with groups. I don't think that `distinct` will create any issue here, but maybe @kemar might spot something else.
> 
> ⚠️ **However**: kudos for adding a unit test, but unfortunately the test setup is wrong. Since there aren't any org units that have multiple groups, the test will always be successful (I tried running this new test on `main` without this fix and it passed)

Thanks @tdethier, I just added in the new test the org unit [to multiple groups](https://github.com/BLSQ/iaso/pull/2005/commits/1e802dc11637dcf2b630f5ad55f4e043409f21d3) and run the test without the fix, It failed and with the fix, it passed!

tdethier (approved):
LGTM :+1:

beygorghor (approved):

PR #2011, opened 2025-03-05
IA-4005 IA-4006: Add a `Type` to OUCRC

bmonjoie:
OUCRC now have a type to differentiate them between OrgUnit edition and creation. 

Related JIRA tickets : IA-4005, IA-4006

- [X] Are there enough tests?

## Changes

The UI has been adapted to take this into consideration and the APIs as well.

In the mobile endpoint, a new `include_creation` parameter has been added not to break with the existing applications which do not expect two kinds of OUCRC.

## How to test

Go to the OUCRC screen and add configuration for creation. You can filter them, delete them, etc.

```
fix: empty instance pop up

Refs: IA-3665
```


hat/assets/js/apps/Iaso/domains/orgUnits/configuration/hooks/useOrgUnitConfigurationTypes.ts
```diff
@@ -0,0 +1,21 @@
...
+    return [
+        {
+            value: orgUnitChangeRequestConfigTypeCreation,
+            label: formatMessage(MESSAGES.creation),
+        },
+        {
+            value: orgUnitChangeRequestConfigTypeEdition,
+            label: formatMessage(MESSAGES.edition),
+        },
+    ];
```
beygorghor: you can `useMemo` here

beygorghor (approved):

iaso/api/org_unit_change_request_configurations/serializers.py
```diff
@@ -185,6 +190,7 @@ class BaseOrgUnitChangeRequestConfigurationWriteUpdateSerializer(serializers.Mod
     class Meta:
         model = OrgUnitChangeRequestConfiguration
         fields = [
+            "type",
```
tdethier:
```suggestion
```

I would not include it in the update serializer, in order to avoid changing the type of an existing OUCRC

iaso/api/org_unit_change_request_configurations/views_mobile.py
```diff
@@ -92,7 +107,10 @@ def list(self, request: Request, *args, **kwargs) -> Response:
             # we have to sort the resulting list manually to keep the pagination working properly.
             queryset = list(
                 chain(
-                    queryset.exclude(org_unit_type__in=non_editable_org_unit_type_ids),
+                    queryset.exclude(
+                        Q(org_unit_type__in=non_editable_org_unit_type_ids)
+                        & Q(type=OrgUnitChangeRequestConfiguration.Type.EDITION)
```
tdethier, line 112: I'm not sure I understand why you need to exclude editions here

iaso/tests/api/org_unit_change_request_configurations/test_filters.py
```diff
@@ -61,17 +62,19 @@ def test_filter_config_on_org_unit_type_id(self):
...
         result = response.json()["results"]
-        self.assertEqual(2, len(result))
+        self.assertEqual(3, len(result))
         # Results should be ordered by ID by default, so the new one will always be last
         self.assertEqual(result[0]["id"], self.oucrc_type_water.id)
-        self.assertEqual(result[1]["id"], new_oucrc.id)
+        self.assertEqual(result[1]["id"], self.oucrc_type_water_creation.id)
+        self.assertEqual(result[2]["id"], new_oucrc.id)
 
         # Filtering on an orgunit type with a single OUCRC
```
tdethier: We should probably remove this comment now :grin:

iaso/tests/api/org_unit_change_request_configurations/test_filters.py
```diff
@@ -37,7 +38,7 @@ def test_filter_config_on_project_id(self):
         response = self.client.get(f"{self.OUCRC_API_URL}?project_id={self.project_johto.id}")
         self.assertJSONResponse(response, 200)
         result = response.json()["results"]
-        self.assertEqual(3, len(result))  # The 3 OUCRCs created in setup
+        self.assertEqual(6, len(result))  # The 6 OUCRCs created in setup
 
         # Making sure that the 3 results do not include the new OUCRC
```
tdethier:
```suggestion
        # Making sure that the 6 results do not include the new OUCRC
```

iaso/tests/api/org_unit_change_request_configurations/test_org_unit_change_request_configs.py
```diff
@@ -799,11 +827,32 @@ def test_check_availability_error_missing_parameter(self):
         )
         self.assertIn("project_id", response.json())
 
+    def test_check_availability_error_missing_parameter_type(self):
```
tdethier, line 830: Thanks for adding this test :+1:

iaso/tests/api/org_unit_change_request_configurations/test_org_unit_change_request_configs_mobile.py
```diff
@@ -34,10 +35,38 @@ def test_list_ok(self):
...
             #   9. PREFETCH OrgUnitChangeRequestConfiguration.other_groups
-            response = self.client.get(f"{self.MOBILE_OUCRC_API_URL}?app_id={self.app_id}")
+            response = self.client.get(
+                f"{self.MOBILE_OUCRC_API_URL}",
+                {
+                    APP_ID: self.app_id,
+                },
+            )
             self.assertJSONResponse(response, status.HTTP_200_OK)
             self.assertEqual(3, len(response.data["results"]))  # the 3 OUCRCs from setup
```
tdethier:
```suggestion
            self.assertEqual(3, len(response.data["results"]))  # the 3 edition OUCRCs from setup
```

tdethier (changes requested):
I only checked the backend since the front was already done.

Overall: good job, as expected :+1: 

I'm requesting changes only for the serializer suggestion. 
Other suggestions/comments in unit tests are nitpicking, but updating these things will make someone's job easier whenever we need to debug something in these tests later.

And also, @kemar's input might be useful for the "fake OUCRC generation" part.

iaso/models/org_unit_change_request_configuration.py
```diff
@@ -33,9 +33,14 @@ class OrgUnitChangeRequestConfiguration(SoftDeletableModel):
     or restrict the list of possible values for each field.
     """
 
+    class Type(models.TextChoices):
+        CREATION = "creation", _("Creation")
+        EDITION = "edition", _("Edition")
```
kemar: The reason why we need this `Type` choice was unclear to me. I had to call @tdethier to understand.

Could you please add a brief comment explaining why this is requires and the problem it fixes?

iaso/api/org_unit_change_request_configurations/filters.py
```diff
@@ -8,6 +8,9 @@
 
 class OrgUnitChangeRequestConfigurationListFilter(django_filters.rest_framework.FilterSet):
     project_id = django_filters.NumberFilter(field_name="project", label=_("Project ID"))
+    type = django_filters.ChoiceFilter(
+        field_name="type", label=_("Type"), choices=OrgUnitChangeRequestConfiguration.Type.choices
+    )
```
kemar: `django_filters` is smart enough to infer this line only by adding `"type"` to the `fields` list in the inner `class Meta`:

```python
    class Meta:
        model = OrgUnitChangeRequestConfiguration
        fields = ["type"]
```

This can save a bunch of typing :)

iaso/api/org_unit_change_request_configurations/views_mobile.py
```diff
@@ -70,7 +79,12 @@ def list(self, request: Request, *args, **kwargs) -> Response:
 
         """
         app_id = AppIdSerializer(data=self.request.query_params).get_app_id(raise_exception=True)
+        include_creation = IncludeCreationSerializer(data=self.request.query_params).get_include_creation(
+            raise_exception=True
+        )
```
kemar, line 84: This could be handled by a custom `filterset_class` instead of a brand new `IncludeCreationSerializer `.

IMO it would be less verbose and just as effective.

iaso/api/org_unit_change_request_configurations/views.py
```diff
@@ -88,13 +90,18 @@ def check_availability(self, request, *args, **kwargs):
             raise serializers.ValidationError("You must be logged in")
 
         project_id = ProjectIdSerializer(data=self.request.query_params).get_project_id(raise_exception=True)
+        oucrc_type = OrgUnitChangeRequestConfigurationTypeSerializer(data=self.request.query_params).get_type(
+            raise_exception=True
+        )
```
kemar, line 95: I would probably have merged those in a single `Serializer` for this custom view.

iaso/api/org_unit_change_request_configurations/views_mobile.py
kemar, line 112: Could things go wrong when `include_creation = true`?

AFAIU line `87` would set `queryset.exclude(type=CREATION)` and this line would set `queryset.exclude(type=EDITION)`.

kemar (changes requested):
Could you just add a comment explaining why we need this new type field? This would be helpful for newcomers.

The rest is just suggestions and questions but none of them are blocking.

iaso/api/org_unit_change_request_configurations/views_mobile.py
bmonjoie: This is the code that removes existing configuration to replace them with the “fake” ones to disable edition. Therefore, I only remove the edition type ones from my request because only those will be replaced, not the edition type ones.

bmonjoie: This is the code that removes existing configuration to replace them with the “fake” ones to disable edition. Therefore, I only remove the edition type ones from my request because only those will be replaced, not the edition type ones.

There is an `&` that removes only the edition ones with the org_unit_type_ids in the given list.

iaso/api/org_unit_change_request_configurations/views_mobile.py
bmonjoie: Could you add a "suggestion" comment to show me how this would be added to the code?

iaso/api/org_unit_change_request_configurations/filters.py
```diff
@@ -8,18 +8,16 @@
...
         method="filter_created_by", label=_("Created by - Users IDs (comma-separated)")
     )
 
     class Meta:
         model = OrgUnitChangeRequestConfiguration
-        fields = []
+        fields = ["type"]
 
-    def filter_created_by(self, queryset, name, value):
+    @staticmethod
```
kemar, line 20: Do you really need a `@staticmethod` here?

iaso/api/org_unit_change_request_configurations/views_mobile.py
kemar: Ok

iaso/api/org_unit_change_request_configurations/views_mobile.py
kemar: You can see an example of what I mean here:

- https://github.com/BLSQ/iaso/blob/413f5c1ef6887bf351bc8f65a9074becb3128c74/iaso/api/org_unit_change_request_configurations/views.py#L39
- https://github.com/BLSQ/iaso/blob/413f5c1ef6887bf351bc8f65a9074becb3128c74/iaso/api/org_unit_change_request_configurations/filters.py#L9

Don't bother if you don't feel comfortable, this could be handled later.

kemar (approved):
LGTM

tdethier (approved):

PR #2017, opened 2025-03-11
IA-4022: Instance upload bug with same OrgUnit UUID

tdethier:
Fix Instance upload bug when there are multiple OrgUnit with the same UUID.
This scenario typically happens when a SourceVersion is copied.

Related JIRA tickets : IA-4022

- [x] Are there enough tests?

## Doc

/

## Changes
/

## How to test

- Make sure that you have at least 2 OrgUnits that have the same UUID (through a copy of SourceVersion)
- Upload a new instance that uses the OrgUnit UUID
- Check if the upload was successful

## Print screen / video
/

## Notes

In the unit tests, I tried calling the function that copies source versions directly, but I couldn't make it work since it has to be triggered by a task. Instead, I made an API call that does it for me. It's a bit weird, but it works :man_shrugging: 

```
fix: empty instance pop up

Refs: IA-3665
```

### DIFF
--- a/iaso/tests/api/test_instances.py
+++ b/iaso/tests/api/test_instances.py
@@ -22,13 +22,12 @@ from iaso import models as m
 from iaso.api import query_params as query
 from iaso.models import FormVersion, Instance, InstanceLock
 from iaso.models.microplanning import Planning, Team
-from iaso.test import APITestCase
-
+from iaso.tests.tasks.task_api_test_case import TaskAPITestCase
 
 MOCK_DATE = datetime.datetime(2020, 2, 2, 2, 2, 2, tzinfo=pytz.utc)
 
 
-class InstancesAPITestCase(APITestCase):
+class InstancesAPITestCase(TaskAPITestCase):
     @classmethod
     @mock.patch("django.utils.timezone.now", lambda: MOCK_DATE)
     def setUpTestData(cls):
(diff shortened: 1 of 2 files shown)


beygorghor (approved):

iaso/tests/api/test_instances.py
```diff
@@ -502,6 +504,60 @@ def test_instance_create_after_sync(self):
         self.assertEqual("RDC Collecte Data DPS_2_2019-08-08_11-54-46.xml", pre_existing_instance.file_name)
         self.assertEqual("Mobile app name", pre_existing_instance.name)
 
+    def test_instance_create_with_org_unit_in_multiple_source_versions(self):
+        """POST /api/instances/ an instance with an orgunit that has been copied in multiple source versions"""
+
+        # First, let's create a user that has permissions to do all of this
+        super_yoda = self.create_user_with_profile(account=self.star_wars, username="super yoda", permissions=["iaso_sources", "iaso_submissions", "iaso_org_units"])
+
+        # Then, let's copy the existing source version through an API call (couldn't call task directly)
```
bramj, line 513: Doing it via API call is fine, but fyi I think you should be able to call it directly with `_immediate=True`, like for example: https://github.com/BLSQ/iaso/blob/main/iaso/tests/tasks/test_process_mobile_bulk_upload.py#L137-L142

bramj (commented):
LGTM :+1:

PR #2026, opened 2025-03-14
[IA-4034] Workflow copy error - name too long

tdethier:
Fixes a 500 error triggered by copying a workflow with a name that is too long.

Related JIRA tickets : IA-4034

- [x] Are there enough tests?

## Doc

/

## Changes

Changes the way copies are named if it exceeds the maximum length.

## How to test
- Have a workflow with a name of 45 to 50 characters
- Copy it
- Notice that there is no crash. Its name is different than the usual "Copy of <name of the previous workflow>"

## Print screen / video

/

## Notes

/

```
fix: empty instance pop up

Refs: IA-3665
```

### DIFF
--- a/iaso/api/workflows/utils.py
+++ b/iaso/api/workflows/utils.py
@@ -5,6 +5,8 @@ from copy import deepcopy
 from django.shortcuts import get_object_or_404
 from rest_framework import serializers
 
+from iaso.models import WorkflowVersion
+
 
 def validate_version_id(version_id, user):
     from iaso.models import EntityType, WorkflowVersion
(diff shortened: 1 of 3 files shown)


quang-le (changes requested):
I think we need to consider:
- increasing `NAME_MAX_LENGTH`. to better fit the names input by the users
- add some validation in the front-end to prevent names that are too long to be saved
- both

I've added a comment to the Jira ticket for further spec

tdethier:
> I think we need to consider increasing NAME_MAX_LENGTH to better fit the names input by the users


Yes, I considered increasing this size, but I remembered that there were some storage constraints with NFC cards and I wasn't sure if it would have an impact on that. So, I just made a quick fix to let users use the feature again.

I agree that there should be some more specification on this topic :+1:

quang-le (approved):

tdethier:
Considering that we need more specs in order to have a proper solution, we agreed that we're going to merge this branch as it is and cover the rest in other tickets.

PR #2032, opened 2025-03-17
IA-3802 enketo erases special values from mobile

mestachs:
Some special values like `current_ou_xxx` and `parent_xxx_yyy` are disappearing once edited in enketo (if in luck prevent the edition, at worst blank out the values to their default values)

Related JIRA tickets : IA-3802

- [ ] Are there enough tests?

## Doc

The doc was already there but not supported when creating/editing submissions via enketo.
see `docs/pages/dev/how_to/iaso_special_question_names/iaso_special_question_names.en.md`

## Changes

The idea is to modify our enketo into something that support these variables.
Sadly this abuse how enketo is supposed to work, most injected values are "with calculate" and calculations that are "trivial" like 0 or empty string/quotes but enketo will always recalculate (and so wiping the values).

So how the change is done
 - change the form xml downloaded by enketo : 
    - change the questions type to look like "hidden" and remove the "calculation" when in the white list
 - change the creation into an edition
    - extract the default submission xml from the form definition xml
    - inject the special values (on create or on edit) with current_ou and others

Note there are 2 "less related" changes
 - a flacky test on instances csv export
 - a fix for the seed command since play.dhis2.org has changed a bit (the login is not more at the same place)

TODO in another PR : implement entity creation/edition ? but it's much bigger (will need to reimplement the workflow, json expresssion evaluation,...

## How to test

**test the "normal" path** (create/edit from the web ui): 
 - launch the seed command
 - see the jira for a xlsform (or the one in the fixtures iaso/tests/fixtures/form_rapide_1666691000_with_injectables.xlsx)
 - make a submission (via enketo)

**to test the "public create"** (mechanism used for pbf to open a given form based on external ids)

```http://localhost:8081/api/enketo/public_create_url/?form_id=injectables&token=944c1d41-a399-456e-b7d8-e12f97c28086&external_org_unit_id=RNGpZqutw3Y```

note 
 - you'll need to gather a lot of ids  
 - the `form_id` is the one in string (not the db id) generally in the xlsform settings page but available in /api/forms/{form_id:int}.
 - the token is on the project and only visible in django admin

**to test the "public launch"** (mechanism where you send links to unauthenticated user so they can fill in for a given orgunit and period a form)
```
http://localhost:8081/api/fill/524c7c9e-97e4-4966-8c96-a435fd31e435/10/202301
```
where 
- the uuid is the one of the form (see django admin, it is not returned by the api http://localhost:8081/admin/iaso/form/xx/change/)
- the second number is the id of the orgunit
- the last number is the period in dhis2 format

## Print screen / video

image

## Notes

```
fix: empty instance pop up

Refs: IA-3665
```

### CODE REVIEW

File: iaso/enketo/enketo_xml.py [hunk 1]
```diff
@@ -100,5 +208,38 @@ def inject_xml_find_uuid(instance_xml, instance_id, version_id, user_id) -> Tupl
...
+    substitutions = {}
+    if instance and instance.org_unit:
+        substitutions = {
+            ".//current_ou_id": deep_getattr(instance, "org_unit.id", ""),
+            ".//current_ou_name": deep_getattr(instance, "org_unit.name", ""),
+            ".//current_ou_type_id": deep_getattr(instance, "org_unit.org_unit_type.id", ""),
+            ".//current_ou_type_name": deep_getattr(instance, "org_unit.org_unit_type.name", ""),
+        }
+
+    if instance and instance.org_unit:
```
tdethier:
```suggestion
```

tdethier: Since it's already on line 213

iaso/api/enketo.py
```diff
@@ -44,6 +44,28 @@ def public_url_for_enketo(request: HttpRequest, path):
...
+    instance_xml = extract_xml_instance_from_form_xml(form.latest_version.file.read(), instance.uuid)
+
+    version_id = instance.form.latest_version.version_id
+    instance_uuid, new_xml = inject_xml_find_uuid(
+        instance_xml, instance_id=instance.id, version_id=version_id, user_id=user_id, instance=instance
+    )
+
+    edit_url = enketo_url_for_edition(
+        public_url_for_enketo(request, "/api/enketo"),
+        form_id_string=instance.uuid,
```
tdethier, line 64: You're passing `instance.uuid` to `form_id_string`, shouldn't it be `instance.form.uuid`?

iaso/enketo/enketo_xml.py
```diff
@@ -79,8 +138,56 @@ def to_xforms_xml(form, download_url, manifest_url, version, md5checksum, new_fo
     return xforms_xml.decode("utf-8")
 
 
+def extract_xml_instance_from_form_xml(form_version_xml, instance_uuid):
```
tdethier, line 150: I don't know anything about XMLs so I guess you know what you are doing in these XML manipulations :sweat_smile:

iaso/api/enketo.py
```diff
@@ -44,6 +44,28 @@ def public_url_for_enketo(request: HttpRequest, path):
     return resolved_path
 
 
+def _enketo_url_for_creation(form, instance, request, return_url):
```
tdethier, line 50: I would maybe leave a comment here to explain why we call the edition function when we want to create an instance

iaso/enketo/enketo_xml.py
```diff
@@ -79,8 +138,56 @@ def to_xforms_xml(form, download_url, manifest_url, version, md5checksum, new_fo
...
+
+    # Create <meta> and <instanceID> nodes
+    instance_id_tag = new_data_node.find(".//meta/instanceID")
+    instance_id_tag.text = f"uuid:{instance_uuid}"
+
+    instance_xml = etree.tostring(new_data_node, encoding="UTF-8", pretty_print=False)
+    return instance_xml
+
+
+def deep_getattr(obj, attr, default=None):
```
tdethier: Nice :+1:

iaso/enketo/enketo_xml.py
```diff
@@ -25,6 +67,23 @@ def inject_instance_id_in_form(xml_str, instance_id):
...
+
+    for bind in bind_nodes:
+        question_name = bind.get("nodeset").split("/")[-1]
+        if question_name in ORG_UNIT_INJECTABLE_QUESTION_NAMES:
+            calculate_expression = bind.get("calculate")
+            if calculate_expression:
+                # if trivial expression like a constant no reference to another variable or function call
+                if "$" not in calculate_expression and "(" not in calculate_expression:
+                    print("deleting ", question_name, bind.get("calculate"))
+                    del bind.attrib["calculate"]
```
tdethier, line 94: This part here is for deleting the default value that Enketo would otherwise place on each calculated field?

tdethier: If so, I would add a comment about that here too

iaso/enketo/enketo_xml.py
```diff
@@ -100,5 +208,38 @@ def inject_xml_find_uuid(instance_xml, instance_id, version_id, user_id) -> Tupl
     if edit_user_id_tag is None:
         edit_user_id_tag = etree.SubElement(root.find(".//meta"), "editUserID")  # type: ignore
     edit_user_id_tag.text = str(user_id)
+
+    substitutions = {}
```
tdethier: If we want to extend this feature to entities, could we instead prepare these substitutions beforehand and pass them as a parameter?

iaso/enketo/enketo_xml.py
```diff
@@ -25,6 +67,23 @@ def inject_instance_id_in_form(xml_str, instance_id):
...
+
+    # Get the default namespace
+    default_ns = root.nsmap[None]  # None gives the default namespace URI
+
+    # Find all <bind> nodes in the default namespace
+    bind_nodes = root.findall(f".//{{{default_ns}}}bind")
+
+    for bind in bind_nodes:
+        question_name = bind.get("nodeset").split("/")[-1]
+        if question_name in ORG_UNIT_INJECTABLE_QUESTION_NAMES:
```
tdethier, line 88: If we want to extend this feature to entities, could we instead prepare this list of fields to remove beforehand and pass them as a parameter?

iaso/tests/api/test_enketo.py
```diff
@@ -336,12 +336,61 @@ def test_public_create_url(self):
...
-        self.assertTrue(
-            responses.assert_call_count("https://enketo_url.host.test/api_v2/survey/single", 1),
+        self.assertTrue(responses.assert_call_count("https://enketo_url.host.test/api_v2/instance", 1))
+
+        self.assertEqual(len(responses.calls), 1)
+        self.assertEqual(old_count + 1, Instance.objects.count())
+
+    @override_settings(ENKETO=enketo_test_settings)
+    @responses.activate
+    def test_public_create_url_injectables(self):
```
tdethier, line 347: Great, but I would definitely add a similar test for the 2 other create endpoints + 1 for instance update

tdethier (changes requested):
The feature works fine, good job :+1: 

I left a few questions.

However:
- I would definitely add more tests
- if possible, I would already try to make it a little more flexible/dynamic because we know that we'll want to use this feature with entities next

beygorghor:
@mestachs when you have time for this please

iaso/api/enketo.py
mestachs: that the whole problem people working on this project not sticking on a "correct" naming. People sometime refer for form where it's submission.

but answering your question iaso is abusing enketo forcing a fresh form each time (where it injects some info also) and it's the same pattern in `_build_url_for_edition` and `inject_instance_id_in_form` where it puts iasoInstanceId

iaso/enketo/enketo_xml.py
mestachs: move to its own functions

iaso/api/enketo.py
tdethier: Ok, thanks for the explanation!

iaso/enketo/enketo_xml.py
```diff
@@ -8,6 +8,56 @@
...
+
+def deep_getattr(obj, attr, default=None):
+    for part in attr.split("."):
+        obj = getattr(obj, part, default)
+        if obj is default:
+            return default
+    return obj
+
+
+def buid_substitutions(instance):
```
tdethier: typo 

```suggestion
def build_substitutions(instance):
```

iaso/enketo/enketo_xml.py
```diff
@@ -100,5 +209,14 @@ def inject_xml_find_uuid(instance_xml, instance_id, version_id, user_id) -> Tupl
     if edit_user_id_tag is None:
         edit_user_id_tag = etree.SubElement(root.find(".//meta"), "editUserID")  # type: ignore
     edit_user_id_tag.text = str(user_id)
+
+    substitutions = buid_substitutions(instance)
```
tdethier:
```suggestion
    substitutions = build_substitutions(instance)
```

iaso/tests/api/test_enketo.py
```diff
@@ -10,6 +10,7 @@
 
 from hat.audit.models import Modification
 from iaso import models as m
+from iaso.enketo.enketo_xml import buid_substitutions
```
tdethier:
```suggestion
from iaso.enketo.enketo_xml import build_substitutions
```

iaso/tests/api/test_enketo.py
```diff
@@ -649,3 +700,79 @@ def test_form_download(self):
...
+    def test_enketo_xml_substitutions_root_ou(self):
+        submission_uuid = "uuid-dup"
+        instance1 = self.create_form_instance(
+            form=self.form_1,
+            period="202001",
+            org_unit=self.jedi_council_corruscant,
+            project=self.project,
+            uuid=submission_uuid,
+        )
+        substitutions = buid_substitutions(instance=instance1)
```
tdethier:
```suggestion
        substitutions = build_substitutions(instance=instance1)
```

iaso/tests/api/test_enketo.py
```diff
@@ -649,3 +700,79 @@ def test_form_download(self):
...
+
+        submission_uuid = "uuid-dup"
+        instance1 = self.create_form_instance(
+            form=self.form_1,
+            period="202001",
+            org_unit=self.jedi_council_corruscant,
+            project=self.project,
+            uuid=submission_uuid,
+        )
+        substitutions = buid_substitutions(instance=instance1)
```
tdethier:
```suggestion
        substitutions = build_substitutions(instance=instance1)
```

iaso/tests/api/test_enketo.py
```diff
@@ -649,3 +700,79 @@ def test_form_download(self):
...
+                ".//parent1_ou_is_root": "1",
+                ".//parent1_ou_name": "Meta Jedi",
+                ".//parent1_ou_type_id": "",
+                ".//parent1_ou_type_name": "",
+            },
+        )
+
+    @override_settings(ENKETO=enketo_test_settings)
+    @responses.activate
+    def test_public_fill(self):
```
tdethier, line 760: :+1:

tdethier (changes requested):
Thanks for the changes!

I would fix the typo (see suggestions) + add a test for the `public_create_url` endpoint. Then I think we're good!

beygorghor (approved):

PR #2037, opened 2025-03-18
[POLIO-1850] Add virus scan v2 on polio pre alerts

tdethier:
Prepare virus scanning v2 and implement it on polio pre alerts.

Related JIRA tickets : POLIO-1850

- [x] Are there enough tests?

## Doc

Now, in v2, file scanning can't make an API call fail. Instead, every file scan result is stored in DB and depending on the result, the file might or might not be available to the users on the web interface.

The default status of a file is `PENDING`.

If file scanning is enabled, the status can become:
- `CLEAN` if there is no virus in the file
- `INFECTED` if a virus was found
-  `ERROR` if there was an issue during the scan 

Infected files cannot be displayed on the web interface, nor downloaded.

## Changes
- Fixed scans in form attachments in v2
- Added uploaded file scans in polio pre alert
- Scans are synchronous, but do not fail. If there is a virus or an error during the file scan, the result is stored.
- Frontend changes - based on the [great designs of Yannick](https://www.figma.com/proto/SknUTK5uSXT8t4UbDZOqvg/IASO-Design?page-id=10%3A4496&node-id=10-4497&viewport=301%2C193%2C0.37&t=wI8qcizHkY7fuDLm-1&scaling=min-zoom&content-scaling=fixed)
  - Infected files are locked and can't be previewed/downloaded by users in the web interface
  - An virus scan header is now displayed on PDF preview
  - An icon show the results before opening the PDF review

## How to test

- Upload files in polio pre alerts
  - First, don't enable file scanning and upload a file:  the scan status will be the default one - `PENDING`
- Start your local ClamAV server (don't forget to activate file scanning through your iaso envs) 
  - Update your document inside the pre alert and notice that its status changed.
  - :warning: there is a known bug (POLIO-1871) - updating a pre alert breaks the document path. After updating pre alerts, you need to manually fix the path in your local DB by removing the extra prefix. **EDIT: it should be fixed now**

- if you want to trigger an `ERROR` status: don't turn on your ClamAV server, but ask iaso to scan files (you can add a short timeout in `settings.py` in order to not wait ~2 minutes)

- if you want to trigger an `INFECTED` status: you can't currently do it through the web interface, since the false positive file is a `txt` file and the pre alert form only accepts `pdf`. You need to manually edit a line in your local DB and change the status to `INFECTED`

## Print screen / video

[Screencast from 2025-03-21 18-11-52.webm](https://github.com/user-attachments/assets/3c695f0d-3cba-4947-b504-fa37f7dc0355)

image

image
image
image
image

## Notes

Other endpoints will be done later, in a separate ticket.

```
fix: empty instance pop up

Refs: IA-3665
```


hat/assets/js/apps/Iaso/components/files/pdf/FileScanHeader.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+    const headerStyle = useMemo(() => {
+        if (scanResult === fileScanResultClean) {
+            return [styles.baseHeader, styles.headerCleanFile];
+        }
+        if (scanResult === fileScanResultInfected) {
+            return [styles.baseHeader, styles.headerInfectedFile];
+        }
+        return [styles.baseHeader, styles.headerPendingFile];
+    }, [scanResult]);
```
quang-le: since `styles` is hoisted outside of the component, it's got a stable reference, so I think `useMemo` can be skipped here

hat/assets/js/apps/Iaso/components/files/pdf/FileScanHeader.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+    const headerIcon = useMemo(() => {
+        if (scanResult === fileScanResultClean) {
+            return <CheckCircleOutline sx={styles.colorCleanFile} />;
+        }
+        if (scanResult === fileScanResultInfected) {
+            return <ErrorOutlineIcon sx={styles.colorInfectedFile} />;
+        }
+        return <WarningAmberIcon sx={styles.colorPendingFile} />;
+    }, [scanResult]);
```
quang-le: This should be a component `HeaderIcon`, that receives a prop `scanResult` i.o `useMemo`

hat/assets/js/apps/Iaso/components/files/pdf/FileScanHeader.tsx
```diff
@@ -0,0 +1,105 @@
...
+            return formatMessage(MESSAGES.fileScanResultInfected);
+        }
+        if (scanResult === fileScanResultClean) {
+            return formatMessage(MESSAGES.fileScanResultSafe);
+        }
+        if (scanResult === fileScanResultError) {
+            return formatMessage(MESSAGES.fileScanResultError);
+        }
+        return formatMessage(MESSAGES.fileScanResultPending);
+    }, [scanResult, formatMessage]);
```
quang-le: I would put that code in a separate hook and declare:

```
const headerText = useHeaderText()
```

The hook:
```typescript
const useHeaderText = (scanResult) :string => {
    const {formatMessage} = useSafeIntl();
     if (scanResult === fileScanResultInfected) {
            return formatMessage(MESSAGES.fileScanResultInfected);
        }
        if (scanResult === fileScanResultClean) {
            return formatMessage(MESSAGES.fileScanResultSafe);
        }
        if (scanResult === fileScanResultError) {
            return formatMessage(MESSAGES.fileScanResultError);
        }
        return formatMessage(MESSAGES.fileScanResultPending);
    
}
```

The point is: since you're returning a string, you don't need to use `useMemo`

hat/assets/js/apps/Iaso/components/files/pdf/FileScanStatusOpenButton.tsx
```diff
@@ -0,0 +1,116 @@
...
+        return (
+            <Warning
+                sx={[
+                    styles.baseIconState,
+                    coloredIcon && styles.coloredStatePending,
+                    !coloredIcon && styles.whiteState,
+                ]}
+            />
+        );
+    }, [coloredIcon, scanResult]);
```
quang-le: Again, I'd make a `FileScanResultIcon` component. A good rule of thumb is that if you're returning jsx, it's probably better to make a component

hat/assets/js/apps/Iaso/components/files/pdf/FileScanStatusOpenButton.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+    const tooltipMessage = useMemo(() => {
+        if (scanResult === fileScanResultClean) {
+            return MESSAGES.fileScanSafeIconTooltip;
+        }
+        if (scanResult === fileScanResultInfected) {
+            return MESSAGES.fileScanInfectedIconTooltip;
+        }
+        return MESSAGES.fileScanPendingIconTooltip;
+    }, [scanResult]);
```
quang-le: cf supra: no `useMemo` needed since you're returning objects with stable reference (not declaring new ones)

hat/assets/js/apps/Iaso/components/files/pdf/FileScanStatusOpenButton.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+    return (
+        <Tooltip title={formatMessage(tooltipMessage)} arrow>
```
quang-le: Actually, you could refactor `tooltipMessage` so it returns a string.
I prefer to avoid function calls in the jsx when possible, in order to separate logic from layout

hat/assets/js/apps/Iaso/components/files/pdf/PdfPreview.tsx
```diff
@@ -103,16 +99,20 @@ export const PdfPreview: FunctionComponent<PdfPreviewProps> = ({
         setOpen(false);
     };
 
+    const isFileSafeToDisplayAndDownload = useMemo(() => {
+        return !scanResult || scanResult !== fileScanResultInfected;
+    }, [scanResult]);
+
```
quang-le: primitives (string, number, boolean) don't require memoïzation, nor do objects with stable reference (i.e: declared outside of a function or component)

iaso/api/form_attachments.py
```diff
@@ -20,14 +20,24 @@
...
+    scan_result = serializers.SerializerMethodField()
+    scan_timestamp = serializers.SerializerMethodField()
+
+    def get_scan_result(self, obj: FormAttachment):
+        return obj.file_scan_status
+
+    def get_scan_timestamp(self, obj: FormAttachment):
+        if obj.file_last_scan:
+            return obj.file_last_scan.timestamp()
+        return obj.file_last_scan
```
quang-le:
```suggestion
        return obj.file_last_scan.timestamp() if obj.file_last_scan else None
```

More legible maybe?

plugins/polio/api/vaccines/supply_chain.py
```diff
@@ -171,6 +174,14 @@ def get_can_edit(self, obj):
...
         )
 
+    def get_scan_result(self, obj):
+        return obj.document_scan_status
+
+    def get_scan_timestamp(self, obj):
+        if obj.document_last_scan:
+            return obj.document_last_scan.timestamp()
+        return obj.document_last_scan
+
```
quang-le, line 219: Could we not make a Mixin for this?

plugins/polio/js/src/domains/VaccineModule/SupplyChain/types.ts
```diff
@@ -49,6 +49,8 @@ export type PreAlert = {
     to_delete?: boolean;
     document?: File;
     can_edit: boolean;
+    scan_result?: string;
```
quang-le: This could be better typed I think:
```typescript
export type ScanResultStatus = 'CLEAN' |  'INFECTED' | 'ERROR' | 'PENDING';
```

plugins/polio/models/base.py
```diff
@@ -1435,6 +1435,10 @@ class VaccinePreAlert(models.Model):
         null=True,
         blank=True,
     )
+    document_last_scan = models.DateTimeField(blank=True, null=True)
+    document_scan_status = models.CharField(
+        max_length=10, choices=VirusScanStatus.choices, default=VirusScanStatus.PENDING
+    )
```
quang-le, line 1446: Could we not make a sub-class of `ModelViewset` and `SoftDeletableModel` with these fields?
It would ensure consistent field naming in the future

plugins/polio/tests/api/vaccine_supply_chain/base.py
quang-le, line 1: ❤️

quang-le (dismissed):
Here are already the code-related comments. I'll run the code on my machine tomorrow

hat/assets/js/apps/Iaso/components/files/pdf/PdfPreview.tsx
tdethier: When I remove `useMemo` from here, I get a warning:

```ESLint: The 'isFileSafeToDisplayAndDownload' function makes the dependencies of useCallback Hook (at line 115) change on every render. To fix this, wrap the definition of 'isFileSafeToDisplayAndDownload' in its own useCallback() Hook. (react-hooks/exhaustive-deps)```

Should I keep `useMemo` then?

iaso/api/form_attachments.py
tdethier: Sure, why not

plugins/polio/api/vaccines/supply_chain.py
tdethier: Probably yeah, I'll do that in the next ticket once I start working on the other endpoints

plugins/polio/models/base.py
tdethier: I considered that, but we already have a different name in polio (all of them are called `document` except `NotificationImport.file`) 😢

Ideally, I would also like to add the document/file field in this class. But I need to think about a proper way of doing that (because we need to pass parameters to that field)

plugins/polio/migrations/0226_merge_20250331_1412.py
kemar, line 1: Nitpick: I would merge migrations files in both `iaso` and `polio` to limit their numbers.

kemar (approved):
LGTM Kudos, this works well locally!

plugins/polio/migrations/0226_merge_20250331_1412.py
tdethier: Good idea, it's getting noisy after all the rebases

tdethier:
Since I made the changes Son requested and Marc approved the backend, I'm going to merge this!

PR #2038, opened 2025-03-19
IA-4032: Embedded links read only permission - should allow to see the list of dashboards per user allowed only

hakifran:
- Embedded links read only permission - should allow to see the list of dashboards per user allowed only
Related JIRA tickets : [IA-4032](https://bluesquare.atlassian.net/browse/IA-4032)

- [x] Are there enough tests?

## Doc

## Changes
- Filter on the connected user when he has only read permission

## How to test
- Create a user on edit one with read only permission for embedded links
- Create or edit two embedded links, one linked to to the created user and the other one not linked to him
- Connect with the user and go to embedded links
- You should see only the one pages linked to the connected user

## Print screen / video
[Screencast from 2025-03-19 09-43-14.webm](https://github.com/user-attachments/assets/85c2b23f-aae6-4b5e-bdb8-41dddca932f7)

## Notes

```
fix: empty instance pop up

Refs: IA-3665
```

[IA-4032]: https://bluesquare.atlassian.net/browse/IA-4032?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iaso/api/pages.py
+++ b/iaso/api/pages.py
@@ -4,7 +4,7 @@ from rest_framework import permissions, serializers
 
 from hat.menupermissions import models as permission
 from iaso.api.common import ModelViewSet
-from iaso.models import Page
+from iaso.models import Page, User
 
 
 class PagesSerializer(serializers.ModelSerializer):
@@ -65,7 +65,8 @@ class PagesViewSet(ModelViewSet):
         user = self.request.user
         order = self.request.query_params.get("order", "created_at").split(",")
         
-        queryset = Page.objects.filter(account__users=user.iaso_profile.account.users)
+        users = User.objects.filter(iaso_profile__account=user.iaso_profile.account)
+        queryset = Page.objects.filter(users__in=users)
         if user.has_perm(permission.PAGES) and not user.has_perm(permission.PAGE_WRITE):
             queryset = queryset.filter(users=user)
 
--- a/iaso/api/pages.py
+++ b/iaso/api/pages.py
@@ -64,8 +64,10 @@ class PagesViewSet(ModelViewSet):
     def get_queryset(self):
         user = self.request.user
         order = self.request.query_params.get("order", "created_at").split(",")
-        if user.has_perm(permission.PAGES) and not user.has_perm(permission.PAGE_WRITE):
-            return Page.objects.filter(users=user).order_by(*order).distinct()
         
-        return Page.objects.filter(users__iaso_profile__account=user.iaso_profile.account).order_by(*order).distinct()
+        queryset = Page.objects.filter(account__users=user.iaso_profile.account.users)
+        if user.has_perm(permission.PAGES) and not user.has_perm(permission.PAGE_WRITE):
+            queryset = queryset.filter(users=user)
+
+        return queryset.order_by(*order).distinct()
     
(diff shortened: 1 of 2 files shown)


iaso/api/pages.py
```diff
@@ -65,5 +64,8 @@ def get_object(self):
     def get_queryset(self):
         user = self.request.user
         order = self.request.query_params.get("order", "created_at").split(",")
-        users = User.objects.filter(iaso_profile__account=user.iaso_profile.account)
-        return Page.objects.filter(users__in=users).order_by(*order).distinct()
+        if user.has_perm(permission.PAGES) and not user.has_perm(permission.PAGE_WRITE):
+            return Page.objects.filter(users=user).order_by(*order).distinct()
+        
+        return Page.objects.filter(users__iaso_profile__account=user.iaso_profile.account).order_by(*order).distinct()
```
bramj: It works as expected, but as a developer it's not really clear to me that the pages are filtered on the `account`. This was already unclear in the code before, and there's actually an `account_id` on the `Page` model, so what do you think about something like this?

```python
        queryset = Page.objects.filter(account=user.iaso_profile.account)

        if user.has_perm(permission.PAGES) and not user.has_perm(permission.PAGE_WRITE):
            queryset = queryset.filter(users=user)

        return queryset.order_by(*order).distinct()
```

iaso/tests/api/test_pages.py
```diff
@@ -62,6 +62,30 @@ def test_pages_list_without_pages_permission(self):
...
         response = self.client.get("/api/pages/")
         self.assertJSONResponse(response, 403)
+    
+    def test_pages_list_linked_to_current_user(self):
+        """Get /pages/ only pages linked to the current user"""
+        self.create_page(name="TEST1", slug="test_1", needs_authentication=False, users=[self.second_user.pk])
+        self.create_page(name="TEST2", slug="test_2", needs_authentication=True, users=[self.second_user.pk])
+
+        # Check when the user has only read permission but not embedded links linked to him
+        self.client.force_login(self.userNoWritePermission)
```
bramj: This was already there before, but we should stick to snakecase

bramj (changes requested):
Tested and works, but I think some small improvements can be made

iaso/api/pages.py
```diff
@@ -65,5 +64,10 @@ def get_object(self):
     def get_queryset(self):
         user = self.request.user
         order = self.request.query_params.get("order", "created_at").split(",")
-        users = User.objects.filter(iaso_profile__account=user.iaso_profile.account)
-        return Page.objects.filter(users__in=users).order_by(*order).distinct()
+        
+        queryset = Page.objects.filter(account__users=user.iaso_profile.account.users)
```
bramj: Why fetch all users from the account instead of directly checking on the account?

hakifran: Because each embedded link should be linked to one or multiple users in the account. So when we filter only on account it will get also embedded links which are not linked to any user. So that why I filtered on users in the account.

bramj (approved):

PR #2047, opened 2025-03-24
POLIO-1871 Fix bug pdf upload on pre alerts

mathvdh:
On the pre-alerts edition, when saving any pre-alerts it would corrupt PDF attachments ( of the other pre-alerts on the same page)

Related JIRA tickets : POLIO-1871

## Changes

On the backend I had to change the validation that detects change to actually inspect the filename and it's size, as otherwise it would ALWAYS re-save everything because it was comparing a FileField with a InMemoryUpload. (It's not an issue for the user but it would create a lot of duplicate files every time you save)

On the frontend there was an issue with how it was constructing the File object being sent to the backend after an initial first save. This is fixed (it was basically using the path of the file as it's content)

## How to test

Go to pre-alerts in Vaccine Supply chain, and try to create a few pre-alerts, change there doses numbers and save a few times. After that the previews of all the documents should still work

```
fix: empty instance pop up

Refs: IA-3665
```


plugins/polio/api/vaccines/supply_chain.py
```diff
@@ -138,27 +167,55 @@ class Meta(NestedVaccinePreAlertSerializerForPost.Meta):
...
         for key in attrs.keys():
-            if hasattr(current_obj, key) and getattr(current_obj, key) != attrs[key]:
-                has_changes = True
+            if key == "document":
+                # Skip if no new document is being uploaded
+                if not attrs[key]:
+                    continue
+
+                # If there's no existing document but we're uploading one
+                if not getattr(current_obj, key):
```
kemar: Is there a reason why you're using `getattr()` here?

Because this would me much more readable:

```suggestion
                if not current_obj.document:
```

plugins/polio/api/vaccines/supply_chain.py
```diff
@@ -138,27 +167,55 @@ class Meta(NestedVaccinePreAlertSerializerForPost.Meta):
...
+                    is_different = True
+                    setattr(current_obj, key, attrs[key])
```
kemar:
```suggestion
                    current_obj.document = attrs[key]
```

I'm not going to point it out everywhere, but it's valid for the whole PR.

plugins/polio/api/vaccines/supply_chain.py
```diff
@@ -138,27 +167,55 @@ class Meta(NestedVaccinePreAlertSerializerForPost.Meta):
...
+                    continue
+
+                # Compare file names and sizes
+                old_file = getattr(current_obj, key)
+                new_file = attrs[key]
```
kemar: You should put that line at the top of the function and use the variable `new_file` instead of `attrs[key]`. This would improve readability.

kemar (changes requested):
I think it's possible to improve the readability by replacing some functions like `getattr()` and `attrs[key]` with meaningful names.

Would it be possible to add at least one unit test covering the changes you made in the backend?

tdethier:
Is comparing file name & size enough when we want to check if the uploaded file was updated? 

I'm thinking about hashing the file and comparing hashes, but I'm not sure if this is overkill or not

plugins/polio/api/vaccines/supply_chain.py
mathvdh: Done

plugins/polio/api/vaccines/supply_chain.py
mathvdh: Done

mathvdh:
> Is comparing file name & size enough when we want to check if the uploaded file was updated?
> 
> I'm thinking about hashing the file and comparing hashes, but I'm not sure if this is overkill or not

@tdethier I was also hesitating with the hashing solution but thought it might be overkill ...

beygorghor (approved):

PR #2061, opened 2025-03-28
POLIO-1845:  Subactivities - calendar visualization 2.0

beygorghor:
We need to have a better way to display sub-activities on polio calendar, see JIRA ticket for client mock-up

Related JIRA tickets : POLIO-1845

- [ ] Are there enough tests?

## Changes

- refactor existing code
- adding the notion of sub activities rows and cell
- make possible to toggle the sub-activities row per campaign
- display a popup with sub-activities  infos

## How to test

You need to add several sub-activities to several rounds on several polio campaigns
Visit both public and non public calendar and check the render of the sub-activities  rows and cells

## Print screen / video

https://github.com/user-attachments/assets/9bc8c312-2460-4454-b590-bedd82bd09e0

## Notes

This doesn't handle the case where two sub-activities are overlapping on the same round. We are still waiting for user specifications


kemar (approved):
LGTM

It works well, kudos.

There is a lot of code so I reviewed it as best as I could.

Note that with such an amount of new code, some unit tests (I'm not talking about Cypress end-to-end tests) could be useful to test the behavior of the new components.

Also don't forget to fix the chevron's orientation!

PR #2084, opened 2025-04-08
IA-4113 Avoid deleting shapes when performing a `PATCH` request on the org unit API

kemar:
Avoid deleting shapes when performing a `PATCH` request on the org unit API.

Related JIRA tickets : [IA-4113](https://bluesquare.atlassian.net/browse/IA-4113)

## How to test

Go to the history of an org unit which has a geo shape.

You should be able to restore a subset of its field from an old version without deleting its shape.

[IA-4113]: https://bluesquare.atlassian.net/browse/IA-4113?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iaso/api/org_units.py
+++ b/iaso/api/org_units.py
@@ -1,4 +1,5 @@
 import json
+import logging
 
 from copy import deepcopy
 from datetime import datetime
@@ -30,6 +31,8 @@ from iaso.utils.gis import simplify_geom
 
 from ..utils.models.common import get_creator_name, get_org_unit_parents_ref
 
+
+logger = logging.getLogger(__name__)
 
 # noinspection PyMethodMayBeStatic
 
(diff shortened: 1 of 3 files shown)


iaso/api/org_units.py
```diff
@@ -474,22 +477,29 @@ def partial_update(self, request, pk=None):
...
-                # keep geom and simplified geom consistent
-                org_unit.geom = multi_polygon
-                org_unit.simplified_geom = simplify_geom(multi_polygon)
+
+        if "geom" in request.data:
+            geom = request.data["geom"]
+            if geom:
+                try:
+                    # Keep geom and simplified geom consistent.
+                    org_unit.geom = GEOSGeometry(geom)
```
tdethier, line 486: Are you sure that the behavior will be the same if you use `GEOSGeometry` instead of `MultiPolygon`?

tdethier (approved):
LGTM :+1:

PR #2092, opened 2025-04-09
IA-4013: SETUPER: Enable payments module & catchment edit functionality

butofleury:

Related JIRA tickets : [IA-4013](https://bluesquare.atlassian.net/browse/IA-4013)

- [ ] Are there enough tests?

## Changes

- Allowed to provide feature flags payload when setting up a new account via `/api/setupaccount/`
- Added new module **PAYMENTS** to account and a list of feature flags

## How to test

From setuper folder, launch setuper via `python3 setuper.py --additional_projects` to generate test data. Then login with admin user from django admin. Check modules and feature linked to the created account, you should see the module **"PAYMENTS"** and feature flags **"ALLOW_CATCHMENT_EDITION"**

## Print screen / video

qufipcm-Modification-de-account-Iaso-04-09-2025_05_05_PM

## Notes

```
fix: empty instance pop up

Refs: IA-3665
```

[IA-4013]: https://bluesquare.atlassian.net/browse/IA-4013?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iaso/api/setup_account.py
+++ b/iaso/api/setup_account.py
@@ -28,6 +28,7 @@ class SetupAccountSerializer(serializers.Serializer):
     user_manual_path = serializers.CharField(required=False)
     modules = serializers.JSONField(required=True, initial=["DEFAULT"])  # type: ignore
     analytics_script = serializers.CharField(required=False)
+    feature_flags = serializers.JSONField(required=False)
 
     def validate_account_name(self, value):
         if Account.objects.filter(name=value).exists():
@@ -75,13 +76,17 @@ class SetupAccountSerializer(serializers.Serializer):
             modules=account_modules,
             analytics_script=validated_data.get("analytics_script", ""),
         )
-        account_feature_flags = [
-            "SHOW_HOME_ONLINE",
-            "SHOW_BENEFICIARY_TYPES_IN_LIST_MENU",
-            "SHOW_PAGES",
-            "SHOW_LINK_INSTANCE_REFERENCE",
-        ]
-        account.feature_flags.set(account_feature_flags)
+
+        if validated_data.get("feature_flags") == None:
+            account_feature_flags = [
+                "SHOW_HOME_ONLINE",
+                "SHOW_BENEFICIARY_TYPES_IN_LIST_MENU",
+                "SHOW_PAGES",
+                "SHOW_LINK_INSTANCE_REFERENCE",
+            ]
+            account.feature_flags.set(account_feature_flags)
+        else:
+            account.feature_flags.set(validated_data.get("feature_flags"))
 
         # Create a setup_account project with an app_id represented by the account name
         app_id = validated_data["account_name"].replace(" ", ".").replace("-", ".")
--- a/iaso/api/setup_account.py
+++ b/iaso/api/setup_account.py
@@ -6,10 +6,17 @@ from rest_framework import permissions, serializers
 from rest_framework.mixins import CreateModelMixin
 from rest_framework.viewsets import GenericViewSet
 
-from hat.menupermissions.constants import MODULES
+from hat.menupermissions.constants import DEFAULT_ACCOUNT_FEATURE_FLAGS, MODULES
 from hat.menupermissions.models import CustomPermissionSupport
 from iaso.api.common import IsAdminOrSuperUser
-from iaso.models import Account, DataSource, Profile, Project, SourceVersion
+from iaso.models import (
+    Account,
+    AccountFeatureFlag,
+    DataSource,
+    Profile,
+    Project,
+    SourceVersion,
+)
 from iaso.utils.module_permissions import account_module_permissions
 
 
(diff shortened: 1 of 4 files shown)


iaso/api/setup_account.py
```diff
@@ -75,13 +76,17 @@ def create(self, validated_data):
...
         )
-        account_feature_flags = [
-            "SHOW_HOME_ONLINE",
-            "SHOW_BENEFICIARY_TYPES_IN_LIST_MENU",
-            "SHOW_PAGES",
-            "SHOW_LINK_INSTANCE_REFERENCE",
-        ]
-        account.feature_flags.set(account_feature_flags)
+
+        if validated_data.get("feature_flags") == None:
```
tdethier: Nitpick: 
I think I would put these values as the default value of that field, just like `modules` in the serializer on line 29:

```python
modules = serializers.JSONField(required=True, initial=["DEFAULT"])
```

This way you don't need to check for `None` in the payload

iaso/api/setup_account.py
```diff
@@ -75,13 +76,17 @@ def create(self, validated_data):
...
+            account_feature_flags = [
+                "SHOW_HOME_ONLINE",
+                "SHOW_BENEFICIARY_TYPES_IN_LIST_MENU",
+                "SHOW_PAGES",
+                "SHOW_LINK_INSTANCE_REFERENCE",
+            ]
+            account.feature_flags.set(account_feature_flags)
+        else:
+            account.feature_flags.set(validated_data.get("feature_flags"))
```
tdethier, line 95: We're not sure that the received feature flags are correct (when happens if I call this endpoint with ["test"] in the feature flags?)

We should have a `validate_feature_flags` method in the serializer (like `SetupAccountSerializer.validate_modules()`).

iaso/tests/api/test_setup_account.py
```diff
@@ -163,3 +170,19 @@ def test_setup_account_project_creation(self):
         data_source = created_data_source.first()
         project_data_sources = project.data_sources.filter(pk=data_source.id)
         self.assertEqual(project_data_sources.first(), data_source)
+
+    def test_setup_account_with_feature_flags(self):
```
tdethier, line 169: I would add more tests:
- One that should raise an exception with an invalid value in feature flags
- One that doesn't send any feature flag (to see if the default 5 flags are there)

iaso/tests/api/test_setup_account.py
```diff
@@ -163,3 +170,19 @@ def test_setup_account_project_creation(self):
...
+        self.client.force_authenticate(self.admin)
+        data = {
+            "account_name": "initial_project_account test-appid",
+            "user_username": "username",
+            "user_first_name": "firstname",
+            "user_last_name": "lastname",
+            "password": "password",
+            "modules": self.MODULES,
+            "feature_flags": self.ACCOUNT_FEATURE_FLAGS,
```
tdethier: And I would send here either more or fewer feature flags in order to see if the account was created with the feature flags you sent or if it has the default feature flags.

setuper/setuper.py
```diff
@@ -64,6 +64,14 @@ def setup_account(account_name, server_url, username, password):
             "DATA_COLLECTION_FORMS",
             "DHIS2_MAPPING",
             "DATA_VALIDATION",
+            "PAYMENTS",
+        ],
+        "feature_flags": [
```
tdethier, line 69: Since you set default feature flags in the API, I guess you don't need to define them here? 

Maybe it's best to define them here and not put a default value in the API instead

iaso/api/setup_account.py
tdethier: But if you set feature flags in the setuper call, I guess you don't need to put any default value here. So you can disregard this comment

tdethier (changes requested):
I would add more tests + validate the received feature flags in the API

setuper/setuper.py
butofleury: I've defined it here to allow provide feature flags that could be differrent to the default one.
But when, it's not provided, a default one will be taken!

iaso/tests/api/test_setup_account.py
butofleury: I've added 2 additional tests:
- one with at least 1 invalid value in the feature flags
- another one without feature flags.

iaso/api/setup_account.py
```diff
@@ -50,6 +58,15 @@ def validate_modules(self, modules):
                 raise serializers.ValidationError("module_not_exist")
         return modules
 
+    def validate_feature_flags(self, feature_flags):
+        default_account_feature_flags = AccountFeatureFlag.objects.all()
+        account_feature_flags = [feature_flag.code for feature_flag in default_account_feature_flags]
+        if len(feature_flags) > 0:
```
tdethier: Nitpick: you don't need to check this if you do a `for x in y` right after

iaso/tests/api/test_setup_account.py
```diff
@@ -163,3 +165,55 @@ def test_setup_account_project_creation(self):
...
+            "user_first_name": "firstname",
+            "user_last_name": "lastname",
+            "password": "password",
+            "modules": self.MODULES,
+        }
+        response = self.client.post("/api/setupaccount/", data=data, format="json")
+        self.assertEqual(response.status_code, 201)
+        created_account = m.Account.objects.filter(name="account with no feature test-featureappid")
+        account_feature_flags = created_account.first().feature_flags.all()
+        feature_flags = [feature_flag.code for feature_flag in account_feature_flags]
```
tdethier: Nitpick:

```suggestion
        feature_flags = created_account.first().feature_flags.values_list("code", flat=True)
```

tdethier (approved):
LGTM :+1:

PR #2093, opened 2025-04-09
[IA-4111] Fix update OrgUnits with default image

tdethier:
Fix the 500 error when OrgUnits with default image are updated

Related JIRA tickets : IA-4111

- [x] Are there enough tests?

## Doc

/

## Changes

- Changed the OrgUnit serializer to return only the default image ID instead of a dict

## How to test

- Choose an OrgUnit that has a default image (click on the :star: button on the `Image` tab; if you don't have any image in OrgUnits, make a new submission with pictures or run the setuper)
- Update a field like opening/closing date, source ref...
- Notice that there was no 500 error

## Print screen / video
[Screencast from 2025-04-09 17-11-00.webm](https://github.com/user-attachments/assets/f2673397-7f22-406f-a5b4-8d5e2fb19d50)

## Notes

Since the `default_image` object in the payload was not used (except for the `id` field), I removed it and I am now sending `default_image_id` instead. 
I did not notice any issue in the frontend, but if something is broken, then we should revert this change and fetch, in the backend, the `id` from the `default_image` object when calling the OrgUnit `PATCH` endpoint

```
fix: empty instance pop up

Refs: IA-3665
```

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/orgUnits/components/OrgUnitImages.tsx
+++ b/hat/assets/js/apps/Iaso/domains/orgUnits/components/OrgUnitImages.tsx
@@ -51,15 +51,15 @@ export const OrgUnitImages: FunctionComponent<Props> = ({
     };
     const isDefaultImage = useCallback(
         (imageId: number) => {
-            return imageId === orgUnit?.default_image?.id;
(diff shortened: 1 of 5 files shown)


beygorghor (approved):
Nice one, there is no reason to use this dict here !

PR #2105, opened 2025-04-14
IA-3961 Restrict user creation and profile edition by projects 

kemar:
Restrict user creation and profile edition by projects.

Related JIRA tickets : [IA-3961](https://bluesquare.atlassian.net/browse/IA-3961)

## Changes

This PR should fix the problem mentioned by Claire in the Jira issue.

It does the following things:

1. it filters the projects API by authorized projects (to limit choices in the UI's dropdowns)
2. it ensures that project restrictions of the current user are applied to edited profiles
3. it implements project restrictions in bulk profile update
4. it implements project restrictions in bulk user creation
5. bonus: it fixes `test_update_user_with_malformed_phone_number()`

[IA-3961]: https://bluesquare.atlassian.net/browse/IA-3961?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iaso/api/profiles/bulk_create_users.py
+++ b/iaso/api/profiles/bulk_create_users.py
@@ -345,29 +345,27 @@ class BulkCreateUserFromCsvViewSet(ModelViewSet):
                     except (IndexError, ValueError):
                         user_roles = None
                     if user_roles:
-                        user_roles = user_roles.split(value_splitter)
+                        user_roles = [role.strip() for role in user_roles.split(value_splitter) if role]
                         # check if the roles exists in the account of the request user
                         # and add it to user_roles_list
                         for role in user_roles:
-                            if role != "":
-                                role = role[1::] if role[:1] == " " else role
-                                try:
-                                    role_instance = UserRole.objects.get(
-                                        account=importer_account,
-                                        group__name=f"{importer_account.id}_{role}",
-                                    )
-                                    user_roles_list.append(role_instance)
-                                    # get the user group linked to the userrole
-                                    user_group_item = Group.objects.get(pk=role_instance.group.id)
-                                    user_groups_list.append(user_group_item)
+                            try:
+                                role_instance = UserRole.objects.get(
+                                    account=importer_account,
+                                    group__name=f"{importer_account.id}_{role}",
+                                )
+                                user_roles_list.append(role_instance)
+                                # get the user group linked to the userrole
+                                user_group_item = Group.objects.get(pk=role_instance.group.id)
+                                user_groups_list.append(user_group_item)
 
-                                except ObjectDoesNotExist:
-                                    raise serializers.ValidationError(
-                                        {
-                                            "error": f"Error. User Role: {role}, at row {i + 1} does not exists: Fix "
-                                            "the error and try again."
-                                        }
-                                    )
+                            except ObjectDoesNotExist:
+                                raise serializers.ValidationError(
+                                    {
+                                        "error": f"Error. User Role: {role}, at row {i + 1} does not exists: Fix "
+                                        "the error and try again."
+                                    }
+                                )
                     try:
                         projects = row[csv_indexes.index("projects")]
                     except (IndexError, ValueError):
@@ -391,12 +389,13 @@ class BulkCreateUserFromCsvViewSet(ModelViewSet):
                             )
 
                     try:
-                        user_permissions = row[csv_indexes.index("permissions")].split(value_splitter)
+                        user_permissions = [
+                            perm.strip() for perm in row[csv_indexes.index("permissions")].split(value_splitter) if perm
+                        ]
                         current_account = request.user.iaso_profile.account
                         module_permissions = self.module_permissions(current_account)
                         for perm in user_permissions:
-                            perm = perm[1::] if perm[:1] == " " else perm
-                            if perm and perm in module_permissions:
+                            if perm in module_permissions:
                                 try:
                                     perm = Permission.objects.get(codename=perm)
                                     user.user_permissions.add(perm)
--- a/iaso/api/profiles/bulk_create_users.py
+++ b/iaso/api/profiles/bulk_create_users.py
@@ -128,6 +128,10 @@ class BulkCreateUserFromCsvViewSet(ModelViewSet):
             user_editable_org_unit_type_ids = request.user.iaso_profile.get_editable_org_unit_type_ids()
             has_geo_limit = True
 
+        user_has_project_restrictions = hasattr(request.user, "iaso_profile") and bool(
+            request.user.iaso_profile.projects_ids
+        )
+
         if request.FILES:
             # Retrieve and check the validity and format of the CSV File
             try:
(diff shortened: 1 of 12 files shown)


iaso/models/base.py
```diff
@@ -1587,6 +1587,15 @@ def has_a_team(self):
 
     @cached_property
     def projects_ids(self) -> set[int]:
+        """
```
tdethier, line 1590: Thanks for the documentation, very good idea :+1:

iaso/tests/test_create_users_from_csv.py
```diff
@@ -510,6 +511,41 @@ def test_create_user_with_projects(self):
...
+            csv_line_2 = csv_reader[2]
+
+            username_index = 0
+            projects_index = 12
+
+            username_1 = csv_line_1[username_index]
+            username_2 = csv_line_2[username_index]
+
+            # Ensure the CSV tries to set `project` as an authorized project.
+            self.assertEqual(csv_line_1[projects_index], self.project.name)
```
tdethier, line 535: Good idea to check things here instead of having to open the csv file :+1:

iaso/api/profiles/bulk_create_users.py
```diff
@@ -369,20 +373,17 @@ def create(self, request, *args, **kwargs):
...
+                        if user_has_project_restrictions:
+                            projects_instance_list = Project.objects.filter(
+                                name__in=project_names,
+                                account=importer_account,
+                            ).filter_on_user_projects(request.user)
+                        else:
+                            projects_instance_list = Project.objects.filter(
+                                name__in=project_names, account=importer_account
+                            )
+
```
tdethier: Did you remove the check on existing project names on purpose? Because you would have to display the distinction between non-existent project and unknown project name?

iaso/tests/test_create_users_from_csv.py
```diff
@@ -510,6 +511,41 @@ def test_create_user_with_projects(self):
...
+        with open("iaso/tests/fixtures/test_user_bulk_create_valid_with_projects.csv") as csv_users:
+            response = self.client.post(f"{BASE_URL}", {"file": csv_users})
+
+        self.assertEqual(response.status_code, 200)
+
+        # Because `user` is restricted to `project2`, `project` should've been skipped.
+        profile_1 = Profile.objects.get(user__username=username_1)
+        self.assertEqual(0, profile_1.projects.count())
+        profile_2 = Profile.objects.get(user__username=username_2)
+        self.assertEqual(0, profile_2.projects.count())
```
tdethier: This means that a restricted user can create users that are not restricted since they don't have any project? Don't we want to do the opposite?

iaso/tests/test_create_users_from_csv.py
```diff
@@ -510,6 +511,41 @@ def test_create_user_with_projects(self):
...
+
```
tdethier, line 551: I was expecting to see another test update - one for checking the error when there's an invalid project name. I guess there wasn't any :sweat_smile:

iaso/tests/models/test_project.py
```diff
@@ -0,0 +1,37 @@
...
+
+        # Restriction on `project_2`: the user shouldn't be able to see `project_1`.
+        self.user.iaso_profile.projects.set([self.project_2])
+        del self.user.iaso_profile.projects_ids
+        filtered_projects = m.Project.objects.filter_on_user_projects(self.user)
+        self.assertEqual(filtered_projects.count(), 1)
+        self.assertEqual(filtered_projects.first(), self.project_2)
+
+        # Restriction on `project_1` and `project_2`: the user should be able to see all projects.
+        self.user.iaso_profile.projects.set([self.project_1, self.project_2])
```
tdethier, line 35: Nitpick: I would maybe add a third project to the account before checking that the user has access to both projects

iaso/tests/api/test_profiles_bulk_update.py
```diff
@@ -27,6 +26,7 @@ def setUpTestData(cls):
         cls.account2 = m.Account.objects.create(name="Comic books")
         cls.account3 = m.Account.objects.create(name="Video game")
         cls.project = m.Project.objects.create(name="Project 1", app_id="account1.app_id", account=cls.account1)
+        cls.project_1 = m.Project.objects.create(name="Project 2", app_id="account1.app_id", account=cls.account1)
```
tdethier: Nitpick: you have 2 projects with the same name here. It doesn't matter because we don't run code that is based on project names like in CSV bulk create, but I guess that's something we should avoid just in case it ends up being used in another way.

(and 3 projects with the same `app_id`)

iaso/tests/api/test_profiles_bulk_update.py
```diff
@@ -176,7 +176,6 @@ def refresh_all_users_from_db(self):
         self.add_users_team_launcher_1.refresh_from_db()
         self.user_with_user_role1.refresh_from_db()
 
-    @tag("iaso_only")
```
tdethier, line 179: Good idea, I don't think it was still relevant

iaso/tests/api/test_profiles_bulk_update.py
```diff
@@ -419,7 +413,58 @@ def test_profile_bulkupdate_should_fail_with_restricted_editable_org_unit_types(
...
+        operation_payload = {
+            "select_all": False,
+            "selected_ids": [user_to_update.iaso_profile.pk],
+            "unselected_ids": None,
+            "projects_ids_added": [
+                self.project.pk,  # `user` has no rights on this project.
+                self.project_1.pk,  # `user` should be able to add this project.
+            ],
+            "projects_ids_removed": [
+                self.project.pk,  # `user` has no rights on this project.
```
tdethier: I would grant this project to `user_to_update` in this test setup to make sure that the project is still there after running the task

tdethier: And then I would use another project on line 431 because the user already has access to it

iaso/api/profiles/profiles.py
```diff
@@ -706,19 +703,40 @@ def validate_user_roles(self, request):
...
-    def validate_projects(self, request, profile) -> QuerySet[Project]:
-        result = []
+    def validate_projects(self, request, profile) -> list:
         project_ids = set([pk for pk in request.data.get("projects", []) if str(pk).isdigit()])
-        if project_ids:
-            profile_project_ids = set(profile.projects.values_list("id", flat=True))
-            if not request.user.has_perm(permission.USERS_ADMIN) and profile_project_ids != project_ids:
+        user_has_project_restrictions = hasattr(request.user, "iaso_profile") and bool(
+            request.user.iaso_profile.projects_ids
+        )
```
tdethier, line 710: Every time I see this, I'm a bit scared of `hasattr(request.user, "iaso_profile") and`. I know it works, but we need to make sure that anonymous users can't reach that code, because it would consider them to not have any restrictions

tdethier (commented):
The feature works fine :+1:

There's just a few things I'd like to discuss before approving: 
- in the bulk create (see comments)
- there might be another issue: if you are restricted by projects and try to edit another user, you'll only see your own projects, which is great. However, if the user you're trying to edit doesn't have any restrictions, then you can impose restrictions on that user, based on your own projects. This doesn't really make sense in my opinion, but I guess we should discuss that together

iaso/api/profiles/bulk_create_users.py
kemar: Yes, I removed the check on purpose.

Non-existent projects are filtered out in the query and simply ignored.

This has the following benefits:

- I replaced a N+1 query in a for loop
- a non-existent project is simply ignored rather than raising an immediate error in the bulk process

I think we can live with it but I could be wrong?

iaso/api/profiles/profiles.py
kemar: Usually (and hopefully), anonymous users are checked at the class level via `permission_classes = [permissions.IsAuthenticated]`.

The `hasattr(request.user, "iaso_profile")` here only ensures that a super user won't trigger a Sentry error.

iaso/tests/api/test_profiles_bulk_update.py
kemar: You're so right! Fixed in https://github.com/BLSQ/iaso/pull/2105/commits/230a7e087a4ea1a59d235d33bc00c6440173262d

iaso/tests/models/test_project.py
kemar: Fixed in https://github.com/BLSQ/iaso/pull/2105/commits/530e65cd1dafff8fd8f2cd4813531dc19420ddc6

iaso/tests/test_create_users_from_csv.py
kemar: You're right, fixed in https://github.com/BLSQ/iaso/pull/2105/commits/3942474ae9b81590580105fc03401c7acdefa7cd

iaso/tests/test_create_users_from_csv.py
kemar: 😢

iaso/api/profiles/bulk_create_users.py
tdethier: It's ok for me, but I wanted to be sure. Maybe @beygorghor can confirm?

tdethier (approved):
LGTM :+1:

PR #2107, opened 2025-04-14
[IA-4114] Filter OrgUnitTypes based on SourceVersion

tdethier:
Allows users to filter OrgUnitTypes based on the selected SourceVersion. 
The goal is to get rid of all OrgUnitTypes that are not currently used in the selected pyramid.

Related JIRA tickets : IA-4114

- [x] Are there enough tests?

## Doc

/

## Changes

- Added a new parameter `source_version_id` to `OrgUnitTypeDropdown`
- Added a new index on OrgUnit, on [`version_id`, `org_unit_type_id`]

## How to test
- Have a setup with at least 2 SourceVersions or 2 DataSources
- Go on the OrgUnit list page
- Check the content of the OrgUnitType dropdown
- Change the SourceVersion or the DataSource
- Check again the content of the OrgUnitType dropdown - it should be different (even if the source version number is the same)

## Print screen / video
[Screencast from 2025-04-14 18-30-42.webm](https://github.com/user-attachments/assets/9864a1b8-f655-45fd-8118-1655c8caddb2)

## Notes

The index should be able to reduce the extra resources needed to filter out OrgUnitTypes. If not, we can always revert this change.

```
fix: empty instance pop up

Refs: IA-3665
```


iaso/api/org_unit_types/serializers.py
```diff
@@ -273,3 +274,28 @@ class Meta:
         model = OrgUnitType
         fields = ["id", "name", "depth"]
         read_only_fields = ["id", "name", "depth"]
+
+
+class SourceVersionIdSerializer(serializers.Serializer):
```
kemar: I would probably have implemented this as a `FilterSet` instead, just like e.g. this https://github.com/BLSQ/iaso/blob/5cedd4d22700883aeb7aec6dc73b1c8711db1ce7/iaso/api/org_unit_change_request_configurations/filters.py#L9

The benefit would be that:

- it could be a little less verbose
- you could rely directly on the output of the `serializer` so you wouldn't have to call `SourceVersionIdSerializer.get_source_version_id()` elsewhere if you need it

hat/assets/js/apps/Iaso/domains/orgUnits/components/OrgUnitsFilters.tsx
```diff
@@ -111,7 +111,11 @@ export const OrgUnitFilters: FunctionComponent<Props> = ({
         sourceVersionId,
     });
     const { data: orgUnitTypes, isFetching: isFetchingOrgUnitTypes } =
-        useGetOrgUnitTypesDropdownOptions({ projectId });
+        useGetOrgUnitTypesDropdownOptions({
+            projectId,
+            dataSourceId,
+            sourceVersionId,
```
kemar: Why do you need both `sourceVersionId` and `dataSourceId`? Isn't `sourceVersionId` enough?

kemar (commented):
Almost LGTM.

To things:

1. a question
2. a suggestion to use a `FilterSet` instead of a new `Serializer` but only if it makes sense to you

hat/assets/js/apps/Iaso/domains/orgUnits/components/OrgUnitsFilters.tsx
tdethier: You're right, I got confused with the `id` and `number` fields in source versions. I wanted to avoid the frontend from caching if the version number was the same but I don't need to do that :sweat_smile:

iaso/api/org_unit_types/serializers.py
tdethier: It was more work than what I had expected, but you should like the new version

kemar (approved):
LGTM 👍

PR #2115, opened 2025-04-16
POLIO-1916: refresh button for VRF dashboard

quang-le:
The refresh button should also refresh "children" dashboards

Related JIRA tickets :  POLIO-1916

- [x] Are there enough tests?

## Doc

[Added a wiki entry](https://wiki.bluesquare.org/e/en/Dev-team/product-management/iaso/Polio/refresh-powerbi-reference)

## Changes

- refactor "openhexa-config" as "powerbi_dataset_configs"
- store openhexa config under "openhexa" key in config dict
- store dataset ids to be synchronized together under "sync-refresh" key

Basically updated the existing config for powerBI updates so that it can take a list of dataset ids (that we retrieve manually from powerBi beforehand) and uses this list to trigger the refresh of the corresponding dashboards

## How to test

Unfortunately it needs to be deployed on prod (because powerBI)

## Print screen / video

N/A

## Notes

I added the new config in the django admin but left the old (=current) in place, so we can transition seamlessly from one to the other

```
fix: empty instance pop up

Refs: IA-3665
```

### DIFF
--- a/iaso/utils/powerbi.py
+++ b/iaso/utils/powerbi.py
@@ -53,11 +53,11 @@ def get_powerbi_report_token(group_id, report_id):
     return get_powerbi_report_token_with_sp(sp_access_token, group_id, report_id)
 
 
-def get_openhexa_config_for_data_set_id(data_set_id):
+def get_extra_config_for_data_set_id(data_set_id):
     from iaso.models.json_config import Config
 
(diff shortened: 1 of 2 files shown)


tdethier (approved):
Impossible to test locally, but LGTM :+1:

PR #2122, opened 2025-04-23
Fix: ensure reimport_failed_imports command works again

madewulf:

Related JIRA tickets : IA-XXX, WC2-XXX, POLIO-XXX

- [ ] Are there enough tests?

## Doc

## Changes

## How to test

Explain how to test your PR.

## Print screen / video

## Notes

```
fix: empty instance pop up

Refs: IA-3665
```


iaso/management/commands/reimport_failed_imports.py
```diff
@@ -1,33 +1,34 @@
...
-                import_units(i.json_body, i.user, i)
+                import_units(i.json_body, i.user, i.headers["QUERY_STRING"][7:].split("&")[0])
                 i.has_problem = False
                 i.save()
                 unit_count = unit_count + 1
             except Exception as e:
                 print("An error happened", e)
 
-        failed_instance_imports = APIImport.objects.filter(has_problem=True).filter(import_type="instance")
+                failed_instance_imports = APIImport.objects.filter(has_problem=True).filter(import_type="instance")
```
quang-le, line 26: Are you sure about this indentation?  failed instances will only be defined if there was an error in the reimport of at least one org unit

quang-le (changes requested):
pls check indentation line 23

PR #2123, opened 2025-04-23
[IA-4146] Fix filters on Instance bulk actions

tdethier:
Added filter handling in Instance bulk actions

Related JIRA tickets : IA-4146

- [x] Are there enough tests?

## Doc

/

## Changes

- Changed the frontend to send filters in `instance_bulk_gps_push` and `instance_reference_bulk_link`
- Changed the various backend endpoints (check functions + task creation) to handle filters before doing anything

## How to test

- Go to the submissions page
- Fill out some filters & click on `select_all` 
  - :warning: `select_all` is very important
  - The goal is to have instances that are all linked to different org units + if possible with a location (so that you can check that the org unit location was updated with the one in the instance)
- Trigger the push bulk GPS action 
- Check results
  - Notice that the `select_all` was not done on all instances, but only the ones that were filtered
  - Notice that the instance location was applied on org units
- Go back to the submissions page
- Fill out some filters & click on `select_all` 
  - :warning: `select_all` is very important
  -  The goal is to not have multiple instances with the same org unit + form  (remember if they were reference instances that you removed or if they became reference instances)
- Trigger the bulk link reference action 
- Refresh the page & check results
  - Notice that the `select_all` was not done on all instances, but only the ones that were filtered

If you don't know how to make the setup for an action, check unit tests (because each action has some impossible scenarios, so you will want to avoid them with your filters)

## Print screen / video

/

## Notes

This only fixed the `select_all` and filters bug. Other known bugs still need to be fixed (like IA-4156).

```
fix: empty instance pop up

Refs: IA-3665
```


iaso/tasks/instance_bulk_gps_push.py
```diff
@@ -45,8 +48,16 @@ def instance_bulk_gps_push(
 
     user = task.launcher
 
-    queryset = Instance.non_deleted_objects.get_queryset().filter_for_user(user)
+    # The QueryDict had to be serialized to be passed to this task, so let's rebuild it
+    query_dict = QueryDict("", mutable=True)
+    query_dict.update(filters)
+    parsed_filters = parse_instance_filters(query_dict)
+
+    # Doing the same operations as in InstancesViewSet.list() because results should be identical
```
kemar, line 56: Hopefully we won't forget to keep these operations in sync between the view and the task 🙏

kemar (approved):
LGTM Kudos for spotting these bugs and fixing them!

PR #2125, opened 2025-04-24
POLIO-1925 preparedness fallback

quang-le:

Related JIRA tickets : IA-XXX, WC2-XXX, POLIO-1925

- [x] Are there enough tests?

## Doc

## Changes

- add `find_snapshot_for_date` method:
     - if no snapshot, use the previous or next one, whichever is closest
     - only use next snapshot if it's less than 2 days away, to avoid using a snapshot for the next milestone 

## How to test

- Create a campaign with round start date at today +29
- generate a preparedness spreadsheet and save
- refresh preparedness data and save
- call the `/api/polio/preparedness_dashboard/?campaign=<your-campaign-obr_name>` 
- Check the "history" value in the response: the prepraredness score for "28 days" should not be null

## Print screen / video

create a spreadsheet and launch the refresh to generate a `SpreadSheetImport`

Screenshot 2025-04-23 at 17 58 20

Use python shell to change `created_at`
Screenshot 2025-04-23 at 22 56 14

check the endpoint

Screenshot 2025-04-24 at 11 29 58

## Notes

```
fix: empty instance pop up

Refs: IA-3665
```


plugins/polio/tests/test_preparednesssheet_dashboard.py
```diff
@@ -4,7 +4,7 @@
 from plugins.polio.models import SpreadSheetImport
 
 
-class SupplyChainDashboardsAPITestCase(APITestCase):
+class PreparednessDashboardsAPITestCase(APITestCase):
```
tdethier, line 7: :+1:

plugins/polio/preparedness/summary.py
```diff
@@ -89,13 +113,13 @@ def score_for_x_day_before(ssi_for_campaign, ref_date: date, n_day: int):
     return ssi.created_at, day, score
 
 
-def history_for_campaign(ssi_qs, round: Round):
+def history_for_campaign(ssi_qs, round: Round, campaign: Campaign):
```
tdethier, line 116: Do you need to pass the campaign as an extra parameter? Can't we get it inside the function based on the round?

plugins/polio/preparedness/summary.py
```diff
@@ -156,7 +180,7 @@ def set_preparedness_cache_for_round(round):
     cache.set(
         f"preparedness-{round.id}",
         value=prep,
-        timeout=TIMEOUT_PREPAREDNESS,
+        timeout=1,
```
tdethier: Did you remove this constant on purpose? If so, wouldn't it be better to keep the constant and update its value instead?

plugins/polio/preparedness/summary.py
```diff
@@ -165,6 +189,6 @@ def get_or_set_preparedness_cache_for_round(campaign, round):
     p = cache.get_or_set(
         f"preparedness-{round.id}",
         default=lambda: _make_prep(campaign, round),
-        timeout=TIMEOUT_PREPAREDNESS,
+        timeout=1,
```
tdethier: Same question about constant

plugins/polio/tests/api/campaigns/test_campaigns_preparedness.py
```diff
@@ -76,3 +88,122 @@ def test_two_campaign_round_error(self):
                 'campaign_id': '6ecda204-2206-4ae2-a38a-5359684dccf6', 'campaign_obr_name': 'campaign A',
                 'indicators': {}, 'round': 'Round3', 'round_id': 45, 'round_start': None, 'round_end': None,
                 'status': 'not_sync', 'details': 'This spreadsheet has not been synchronised yet'}]"""
+
+    def test_find_snapshot_for_date_previous_sheet_only(self):
```
tdethier, line 92: Nitpick: maybe all of these new tests should not be in `api/campaigns` since you're not calling the API

plugins/polio/tests/api/campaigns/test_campaigns_preparedness.py
```diff
@@ -76,3 +88,122 @@ def test_two_campaign_round_error(self):
...
+        previous_sheet.created_at = reference_date - timedelta(days=2)
+        previous_sheet.save()
+        previous_sheet.refresh_from_db()
+
+        ssi_for_campaign = SpreadSheetImport.objects.filter(spread_id=self.campaign.obr_name)
+        snapshot = find_snapshot_for_date(ssi_for_campaign, self.round_1.started_at, 14, self.campaign.obr_name, 1)
+        self.assertEqual(snapshot, previous_sheet)
+
+    def test_find_snapshot_for_date_next_sheet_only(self):
+        # Use the nearest next sheet if no previous sheet and next sheet created_at <= ref date +2
```
tdethier, line 117: Thanks for the comments in all tests, it's very useful :+1:

plugins/polio/tests/api/campaigns/test_campaigns_preparedness.py
```diff
@@ -76,3 +88,122 @@ def test_two_campaign_round_error(self):
...
+        next_sheet.refresh_from_db()
+
+        ssi_for_campaign = SpreadSheetImport.objects.filter(spread_id=self.campaign.obr_name)
+        snapshot = find_snapshot_for_date(ssi_for_campaign, self.round_1.started_at, 14, self.campaign.obr_name, 1)
+        self.assertEqual(snapshot, next_sheet)
+
+    def test_find_snapshot_for_date_select_previous(self):
+        # Use the nearest previous sheet if it's closer to ref date than the next_sheet
+        reference_date = self.round_1.started_at - timedelta(days=14)
+        previous_sheet = SpreadSheetImport.objects.create(
```
tdethier, line 143: I would maybe add a `previous_sheet_2` that's even older to make sure that the right sheet is fetched.

plugins/polio/tests/api/campaigns/test_campaigns_preparedness.py
```diff
@@ -76,3 +88,122 @@ def test_two_campaign_round_error(self):
...
+            url="https://www.example.com",
+            content={"sheet": "previous"},
+            spread_id=self.campaign.obr_name,
+        )
+        previous_sheet.created_at = reference_date - timedelta(days=3)
+        previous_sheet.save()
+
+        next_sheet = SpreadSheetImport.objects.create(
```
tdethier, line 182: I would maybe add a `next_sheet_2` that's even more in the future to make sure that the right sheet is fetched.

tdethier (commented):
The feature works fine :+1: 

I left a few questions

plugins/polio/preparedness/summary.py
quang-le: Well, I have it in the parent already, so I'd rather avoid risking hitting the DB for nothing if there's a prefetch missing somewhere; especially since it's only to format the error message when we throw

plugins/polio/preparedness/summary.py
quang-le: Ooooh, good catch, I shouldn't have committed that

plugins/polio/preparedness/summary.py
quang-le: same same

plugins/polio/tests/api/campaigns/test_campaigns_preparedness.py
quang-le: yeah, but I'm testing a method that is called when the preparedness endpoint of those tests is queried. But the whole file could be moved yeah

tdethier (approved):
LGTM :+1:

PR #2126, opened 2025-04-24
[IA-4159] Round accuracy when there are more than 2 decimal places

bmonjoie:
When the backend receives an accuracy with more than 2 decimal places, it returns a 400 where it could just round it.

Related JIRA tickets : IA-4159

- [X] Are there enough tests?

## Doc

Code comments

## Changes

I have added `RoundingDecimalField` and used it in the proper serializer.

## How to test

Send a `new_location_accuracy` with more than 2 decimal places.

```
fix: empty instance pop up

Refs: IA-3665
```

### DIFF
--- a/iaso/api/common.py
+++ b/iaso/api/common.py
@@ -510,13 +510,3 @@ class DropdownOptionsListViewSet(ViewSet):
         status_choices = self.get_status_choices()
         serializer = self.serializer(status_choices, many=True)
         return Response(serializer.data)
-
-
-class RoundingDecimalField(serializers.DecimalField):
-    """
-    As per the overridden method's documentation, this can be overridden to disable the validation.
-    The quantize method is called afterward and does the rounding.
-    """
-
-    def validate_precision(self, value):
-        return value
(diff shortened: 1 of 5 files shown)


iaso/api/common.py
```diff
@@ -510,3 +510,13 @@ def list(self, request):
...
+
+
+class RoundingDecimalField(serializers.DecimalField):
+    """
+    As per the overridden method's documentation, this can be overridden to disable the validation.
+    The quantize method is called afterward and does the rounding.
+    """
+
+    def validate_precision(self, value):
+        return value
```
kemar: Just wondering if we're not losing some other useful validation by completely overriding this (see https://github.com/encode/django-rest-framework/blob/543996711d323722a1017e376619ae462a726ada/rest_framework/fields.py#L1049-L1082)?

kemar: I suggest implementing `validate_precision` like this instead https://github.com/BLSQ/iaso/pull/2130

Let me know if that's OK for you?

bmonjoie: Sure, I'm fine with that :)

kemar (approved):
LGTM Excellent work!

PR #2060, opened 2025-03-27
IA-4082: Registry - bug when displaying data

hakifran:
What problem is this PR solving? Explain here in one sentence.
- Registry - bug when displaying data
Related JIRA tickets : [IA-4082](https://bluesquare.atlassian.net/browse/IA-4082)

## Self proofreading checklist

- [x] Did I use eslint and ruff formatters?
- [ ] Is my code clear enough and well documented?
- [ ] Are my typescript files well typed?
- [ ] New translations have been added or updated if new strings have been introduced in the frontend
- [ ] My migrations file are included
- [ ] Are there enough tests?
- [ ] Documentation has been included (for new feature)

## Doc

Tell us where the doc can be found (docs folder, wiki, in the code...).

## Changes
- Remove the '-' when the ordering is descending

## How to test
- Go to registry
- Check if the error : `{
  "name": "ApiError",
  "status": 500,
  "message": "Error on GET /api/orgunits/?validation_status=VALID&orgUnitParentId=29682&onlyDirectChildren=true&limit=10&order=-name&page=1&orgUnitTypeId=6"
}` appear again

## Print screen / video
[Screencast from 2025-03-27 18-40-33.webm](https://github.com/user-attachments/assets/56dee2da-e5f1-4576-a68d-5217f56a3ec5)

## Notes

Things that the reviewers should know:

- known bugs that are out of the scope of the PR
- other trade-offs that were made
- does the PR depends on a PR in [bluesquare-components](https://github.com/BLSQ/bluesquare-components)?
- should the PR be merged into another PR?

## Follow the Conventional Commits specification

The **merge message** of a pull request must follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) specification.

This convention helps to automatically generate release notes.

Use lowercase for consistency.

[Example](https://github.com/BLSQ/iaso/commit/8b8d7d3064138c1e57878f17b4eb922516ab0112):

```
fix: empty instance pop up

Refs: IA-3665
```

Note that the Jira reference is preceded by a _line break_.

Both the line break and the Jira reference are entered in the _Add an optional extended description…_ field.

[IA-4082]: https://bluesquare.atlassian.net/browse/IA-4082?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2041, opened 2025-03-19
IA-4041 Fix a `ProgrammingError` exception in `/api/orgunits/` (bis)

kemar:
Fix `SELECT DISTINCT ON expressions must match initial ORDER BY expressions` at another place.

Related JIRA tickets : [IA-4041](https://bluesquare.atlassian.net/browse/IA-4041)

## How to test

Try this URL locally:

```http://localhost:8081/api/orgunits/?limit=1&group&group=846```

It should work without triggering a `ProgrammingError`.

[IA-4041]: https://bluesquare.atlassian.net/browse/IA-4041?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2039, opened 2025-03-19
HOTFIX IA-4041 Fix a `ProgrammingError` exception in `/api/orgunits/`

kemar:
HOTFIX IA-4041 Fix a `ProgrammingError` exception in `/api/orgunits/`

PR #2132, opened 2025-04-25
IA-4155 Fix sorting org units causes a 500 error

kemar:
Fix sorting org units causes a 500 error.

Related JIRA tickets : IA-4155

## Changes

We know that [using `distinct()` is slower](https://docs.djangoproject.com/en/5.0/ref/models/querysets/#django.db.models.query.QuerySet.distinct):

> For a normal `distinct()` call, the database compares each field in each row when determining which rows are distinct. For a `distinct()` call with specified field names, the database will only compare the specified field names.

But because of the way this endpoint is coded, [using `distinct("id")`](https://github.com/BLSQ/iaso/pull/2005/files#diff-a410d2a02a7ac632955a86a162e1e928dcfd747dd4587699e82da331f1a949ecR83) causes a lot more headaches, see:

- https://github.com/BLSQ/iaso/pull/2039
- https://github.com/BLSQ/iaso/pull/2041
- https://github.com/BLSQ/iaso/pull/2060

And now there are new problems:

- https://bluesquareorg.sentry.io/issues/6113446205
- https://bluesquareorg.sentry.io/issues/6559950873

So let's sacrifice a little bit of performance to reduce some of the complexity and use `.distinct()` instead of `.distinct("id")`.

## How to test?

You should be able to order the list of org units by "Number of submissions" without raising 500.

All the unit tests should be OK too.

### DIFF
--- a/iaso/api/org_unit_search.py
+++ b/iaso/api/org_unit_search.py
@@ -90,7 +90,7 @@ def build_org_units_queryset(queryset, params, profile):
             group_ids = [group]
         else:
             group_ids = group
-        queryset = queryset.filter(groups__in=group_ids).distinct("id")
+        queryset = queryset.filter(groups__in=group_ids).distinct()
 
     if source:
         source = DataSource.objects.get(id=source)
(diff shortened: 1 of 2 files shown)


bramj (approved):
Unfortunate of the performance loss, but I think you're making the correct call to remove the complexity. Tested as well and working :+1:

PR #2136, opened 2025-04-28
IA-4173 Fix payments loading takes forever (bis)

kemar:
Fix payments loading takes forever (bis).

Related JIRA tickets : IA-4173

## Notes

This fixes a performance problem in various payments API (see also https://github.com/BLSQ/iaso/pull/2124).

`get_hierarchy_for_user()` is now calculated only once in `get_serializer_context()` instead of evaluating it in a serializer method for each object in a queryset.


mathvdh (approved):
That should speed it up (a bit)
I guess the payments thing seems overly complicated

PR #2137, opened 2025-04-28
IA-4000: Fix setuper and multi account

butofleury:

Related JIRA tickets : [IA-4000](https://bluesquare.atlassian.net/browse/IA-4000)

- [x] Are there enough tests?

## Changes

Fixed setuper muti account 

## How to test

From setuper folder, launch setuper via super user multi account with: `python3 setuper.py --additional_projects -u username -p password`. username and password are respectively the username and password of the super user multi account 

## Print screen / video

[SuperUserMulti-Account.webm](https://github.com/user-attachments/assets/0a5a8a5b-250a-43e6-bb76-bc2d26319e9c)

## Notes

```
fix: empty instance pop up

Refs: IA-3665
```

[IA-4000]: https://bluesquare.atlassian.net/browse/IA-4000?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iaso/serializers.py
```diff
@@ -22,6 +22,12 @@ def validate(self, attrs):
             request = self.context.get("request")
             if request:
                 app_id = request.query_params.get("app_id", None)
+                # For setuper, account name is same as app id for Main Project, so get the first user account with super user status linked to current super user multi account
+                if app_id is None:
+                    account_tenant_user = (
+                        self.user.tenant_users.filter(account_user__is_superuser=True).first().account_user
+                    )
+                    app_id = account_tenant_user
```
bramj: We really should not depend on the coincidence that `str(account_tenant_user) == app_id`.

Thinking about it a bit more, how about we return a token for the actual user if the app_id is None?
What if we changed
```python
if self.user.tenant_users.exists():
```
for
```python
if app_id and self.user.tenant_users.exists():
```
?

In that case we could simply say "if you don't specify an app_id, we'll give you a token for the main user".

bramj (changes requested):
These changes add confusion on an important point in the code, the one that returns the JWT token for the mobile app.

On a general note, I don't think we should really change this code just for the setuper.
Could we not have the setuper send the `app_id` along with the request?

iaso/serializers.py
butofleury: Like, replace line 26 to 30 with ?: 
 ```
if app_id is None:
     return data

```

data on the line 17 contains both refresh and access keys!

tdethier (approved):
I showed Bram the latest changes, we both approve.

LGTM :+1:

bramj (approved):
LGTM :+1: 

@kemar FYI, since you were talking about this serializer yesterday

PR #2139, opened 2025-04-29
IA-4149; Sync version, disable ou tree picker if no version is selected

beygorghor:
Ou tree picker is using default data source version if none is is given. We should disable it on Sync dialog to avoid selecting wrong org unit fro wrong source

Related JIRA tickets : IA-4149

- [ ] Are there enough tests?

## Changes

disable ou tree picker if a version is not selected + helper message

## How to test

Admin > dataSources > sync button
Target org unit tree view should be disable if no source version is selected

## Print screen / video
Screenshot 2025-04-29 at 11 38 53


iaso/tests/api/org_unit_change_requests/test_org_unit_change_requests.py
```diff
@@ -737,10 +737,14 @@ def get_reference_instance_ids(instances):
...
-            expected_row_data,
-        )
+        # Sort instance IDs in both lists before comparison
+        for i, (expected, actual) in enumerate(zip(expected_row_data, first_data_row)):
+            if "," in expected and "," in actual:  # Check if this is a comma-separated list of IDs
+                expected_sorted = ",".join(sorted(expected.split(",")))
+                actual_sorted = ",".join(sorted(actual.split(",")))
+                self.assertEqual(expected_sorted, actual_sorted, f"Difference at index {i}")
+            else:
+                self.assertEqual(expected, actual, f"Difference at index {i}")
```
kemar: I already saw this in another PR and I also fixed it in a different way (see https://github.com/BLSQ/iaso/commit/d6dad372af1fc3b989e3a2189177ea9508909622 and https://github.com/BLSQ/iaso/commit/2712676815720039aae35eb640eed13425bbe21b)

Is this an error?

kemar (changes requested):
Almost good but I think there is one file that shouldn't be here.

Can you confirm that I'm right?

iaso/tests/api/org_unit_change_requests/test_org_unit_change_requests.py
beygorghor: i will remove it thanks

kemar (approved):

PR #2141, opened 2025-04-29
IA-4134: Validate xlsform question type to reject "hidden"

beygorghor:
Validate xlsform question type to reject "hidden"

Related JIRA tickets : IA-4134

- [x] Are there enough tests?

## Doc

## Changes

- improve validator to check hidden types
- add tests

## How to test

See related tests or try to create a form using this XLSX
[odk_invalid_xlsform.xlsx](https://github.com/user-attachments/files/19957851/odk_invalid_xlsform.xlsx)
optional extended description…_ field.


iaso/odk/validator.py
```diff
@@ -99,7 +99,7 @@ def group_by_lambda(collection, field_name_lambda):
 
 
 def validate_xls_form(xls_file):
-    file_extension = os.path.splitext(xls_file.name)[1].lower()
+    file_extension = os.path.splitext(getattr(xls_file, "name", "form.xlsx"))[1].lower()
```
bmonjoie, line 102: Can you explain this change?

beygorghor: The original code xls_file.name assumes that the xls_file object will always have a name attribute. This could fail if:
- The file object doesn't have a name attribute
- The file object is a BytesIO or other in-memory file object
- The file object is a mock or test object
 
The new code getattr(xls_file, "name", "form.xlsx") is safer because:
- It uses getattr() which safely tries to get the name attribute
- If the name attribute doesn't exist, it falls back to the default value "form.xlsx"
- This prevents potential AttributeError exceptions

bmonjoie (approved):
LGTM, but what do I know? :sweat_smile:

PR #2142, opened 2025-04-29
HOTFIX IA-4175 Fix backend returning entities with entity types unknown to the project

kemar:
HOTFIX IA-4175 Fix backend returning entities with entity types unknown to the project

Related JIRA tickets : IA-4175


kemar:
Deployed as a hotfix in https://github.com/BLSQ/iaso/actions/runs/14733986299

PR #2143, opened 2025-04-29
IA-3944: Project QR code

beygorghor:
Make project QR code more visible & easy to share

Related JIRA tickets : IA-3944

- [x] Are there enough tests?

## Changes

- adapt project api to provide a blob of the QR code on paginated list 
- adapt and write some test for it
- custom component to open and share QR code

## How to test

Open projects page, play with the QR codes 

## Print screen / video

https://github.com/user-attachments/assets/44349a1e-0a69-4e34-a59e-b41c28d0afa1


hat/assets/js/apps/Iaso/domains/projects/components/QrCode.tsx
```diff
@@ -0,0 +1,146 @@
...
+    },
+    snackbarPaper: {
+        backgroundColor: 'rgb(50, 50, 50)',
+        color: '#fff',
+        padding: '6px 16px',
+        borderRadius: '4px',
+    },
+};
+
+const QrCode: React.FunctionComponent<Props> = ({ qrCode }) => {
```
quang-le: We usually import `FunctionComponent` directly, so why the change in syntax?

hat/assets/js/apps/Iaso/domains/projects/components/QrCode.tsx
```diff
@@ -0,0 +1,146 @@
...
+        message: '',
+        isError: false,
+    });
+    const { formatMessage } = useSafeIntl();
+
+    const handleOpen = useCallback(() => setOpen(true), []);
+    const handleClose = useCallback(() => setOpen(false), []);
+    const handleSnackbarClose = useCallback(() => {
+        setSnackbar(prev => ({ ...prev, open: false }));
+    }, []);
```
quang-le: If the dependencies array is empty, then there's no need for `useCallback`

hat/assets/js/apps/Iaso/domains/projects/components/QrCode.tsx
```diff
@@ -0,0 +1,146 @@
...
+                        onClose={handleSnackbarClose}
+                        severity="error"
+                        variant="filled"
+                    >
+                        {snackbar.message}
+                    </Alert>
+                ) : (
+                    <Paper sx={styles.snackbarPaper}>{snackbar.message}</Paper>
+                )}
+            </Snackbar>
```
quang-le: What is preventing the use of the normal Snackbars + snack bar context?

hat/assets/js/apps/Iaso/domains/projects/components/QrCode.tsx
```diff
@@ -0,0 +1,146 @@
...
+        </>
+    );
+};
+
+export default memo(QrCode);
```
quang-le: what does that do?

iaso/api/projects/serializers.py
```diff
@@ -42,3 +54,17 @@ def validate_app_id(self, data):
...
+
+        # Only include QR code if the request is paginated
+        if request and (request.query_params.get("page") or request.query_params.get("limit")):
+            qr_image = make_qr_code_image(
+                data='{"url": "' + request.build_absolute_uri("/") + '", "app_id": "' + instance.app_id + '"}',
+                qr_code_options=QRCodeOptions(size="S", image_format="png", error_correction="L"),
+            )
+            data["qr_code"] = f"data:image/png;base64,{base64.b64encode(qr_image).decode('utf-8')}"
+
+        return data
```
quang-le: Would it not be possible to achieve this result with a `SerializerMethodField`?

I don't know what the use cases are for unpaginated calls to the API, but we tend to want to get rid of these, so my instinct would see if I can't use a different serializer for unpaginated requests, for the sake of backwards compatibility.

I would ask @kemar 's opinion here

quang-le (changes requested):
Needs @kemar 's review for the backend code.
Some cleanup + clarifications needed in the front-end code

iaso/api/projects/serializers.py
kemar: There is something I don't understand. Why do we include the QR code only if the request is paginated?

beygorghor: @kemar to avoid the computation of qrcode if we have 5000 projects

kemar: @beygorghor OK, the comment should explain this, otherwise it's almost useless as it's just repeating the code.

Is the computation of the qrcode that slow? If so, `to_representation` is rather appropriate.

Another solution would be to make `qr_code` an "optional" field of the payload, and implement it as a `SerializerMethodField` as suggested by @quang-le

iaso/tests/api/test_projects.py
```diff
@@ -75,34 +95,6 @@ def test_feature_flags_list_except_no_activated_modules(self):
...
-        self.assertJSONResponse(response, 200)
-
-        response_data = response.json()
-        self.assertValidProjectListData(response_data, 1, True)
-        self.assertEqual(response_data["page"], 1)
-        self.assertEqual(response_data["pages"], 2)
-        self.assertEqual(response_data["limit"], 1)
-        self.assertEqual(response_data["count"], 2)
-
-    def test_feature_flags_list_paginated(self):
```
kemar, line 92: Did you remove this test on purpose?

quang-le (approved):
LGTM

PR #2148, opened 2025-05-06
IA-4172 Project field shouldn't be greyed out in user creation/edition

kemar:
Project field shouldn't be greyed out in user creation/edition.

Related JIRA tickets : IA-4172

## Changes

Edition of the `project` field was restricted to admin users, see #1775.

We're changing that following the multi-projects restrictions (IA-2535) implemented in #2105 #2098 #097 #2096.

The project field isn't grayed out anymore in user creation/edition.

Behavior:

- if the `project` field is empty:
    - if the editing user has project restrictions, we set `project` with the same restrictions
    - otherwise we let `project` empty
- if the `project` field is not empty:
    - if the editing user has no project restrictions, we set `project` to the submitted values
    - if the editing user has project restrictions:
        - we set `project` to the submitted values if it's a subset of the user's project restrictions
    - else:
        - we prevent edition

Note that the choices in the project drowpdown are limited tot the project restrictions of the user.

## How to test

Try to reproduce all the scenarios above.

### DIFF
--- a/iaso/api/profiles/profiles.py
+++ b/iaso/api/profiles/profiles.py
@@ -704,21 +704,23 @@ class ProfilesViewSet(viewsets.ViewSet):
 
     def validate_projects(self, request: HttpRequest, profile: Profile) -> list:
         new_project_ids = set([pk for pk in request.data.get("projects", []) if str(pk).isdigit()])
-        user_projects_ids = set(request.user.iaso_profile.projects_ids)
+        user_restricted_projects_ids = set(request.user.iaso_profile.projects_ids)
 
         if not new_project_ids:
-            if user_projects_ids:
+            if user_restricted_projects_ids:
                 # Apply the same project restrictions.
-                return Project.objects.filter(id__in=user_projects_ids, account=profile.account_id)
+                return Project.objects.filter(id__in=user_restricted_projects_ids, account=profile.account_id)
             # No project restrictions.
             return []
 
-        if not user_projects_ids:
-            # The current user has no project restrictions.
+        if not user_restricted_projects_ids:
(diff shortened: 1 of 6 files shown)


iaso/api/profiles/profiles.py
```diff
@@ -703,41 +702,28 @@ def validate_user_roles(self, request):
...
-            unauthorized_projects_ids = [p for p in project_ids if p not in request.user.iaso_profile.projects_ids]
-            unauthorized_projects_names = Project.objects.filter(id__in=unauthorized_projects_ids).values_list(
-                "name", flat=True
-            )
-            if unauthorized_projects_names:
-                raise PermissionDenied(
-                    f"You don't have access to the following projects: {','.join(unauthorized_projects_names)}."
-                )
+        profile_restricted_projects_ids = set(profile.projects_ids)
+        if profile_restricted_projects_ids.issuperset(user_restricted_projects_ids):
```
tdethier, line 720: This doesn't work in the following edge case:
- you have project restrictions
- choose a user that has no project restrictions
- edit this user and restrict them to one of your projects

I know we discussed this during a SUM and Martin didn't agree with me, but this scenario still doesn't make any sense to me :man_shrugging:

tdethier (approved):
LGTM :+1:

tdethier:
Maybe some nitpick in the frontend:

We do want the following error to be triggered
image

But the `Value "3" not found in possible options` error will not make sense to users. I don't know how we could do that, but for the dropdown we should have one list of existing values (even if the current user doesn't have access to these values) + one list of possible values (= what the user can assign). This would allow the project to be properly displayed, but it would probably cause an error

PR #2153, opened 2025-05-08
IA-4180 Add UUID to filename of bulk upload file

bramj:
Maybe this is overkill, but we want to be 100% sure there won't be name clashes.

New result in the Django admin:
image

**To be tested on staging**

Related JIRA tickets : IA-4180


bramj:
@mestachs I'm fairly sure the migration doesn't really do any SQL (wrt. potential issues because of the huge table)

bramj:
As pointed out by @mestachs , the `AWS_S3_FILE_OVERWRITE` ENV in `settings.py` takes care of this (source: https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html). Closing this PR since it's unneeded complexity.

PR #2154, opened 2025-05-12
IA-4172 Ensure that a user with `projects` restrictions can edit a user with the same `projects` restrictions

kemar:
Ensure that a user with `projects` restrictions can edit a user with the same `projects` restrictions.

Related JIRA tickets : IA-4172

Related PR: https://github.com/BLSQ/iaso/pull/2148

## Changes

- ensure that a user with `projects` restrictions can edit a user with the same `projects` restrictions
- ensure that a user with `projects` restrictions cannot create a user without restrictions
   - this was previously done implicitly
   - now an error is triggered and shown in the API when a user with `projects` restrictions tries to create/edit a user without restrictions


tdethier (approved):
LGTM :+1:

PR #2155, opened 2025-05-12
HOTFIX IA-4172 Project field shouldn't be greyed out in user creation/edition

kemar:
HOTFIX IA-4172 Project field shouldn't be greyed out in user creation/edition.

Deploy #2148 and #2154.


kemar:
Branched from wrong tag.

PR #2158, opened 2025-05-13
IA-4199 Make `ProfilesViewSet.create()` and `ProfilesViewSet.partial_update()` atomic

kemar:
Make `ProfilesViewSet.create()` and `ProfilesViewSet.partial_update()` atomic.

Related JIRA tickets : IA-4199

## How to test

Go to user management `/dashboard/settings/users/management/`.

Ensure you are connected with a user having project restrictions, then:

- try creating a user without any project
- you should get an error message telling you that you have to associate a project to it
- add a project to that user

In the past you would get an error message telling you that the user already exists.

This PR fixes this bug and you shouldn't see that message at all.


tdethier (approved):
LGTM :+1:

PR #2165, opened 2025-05-15
POLIO-1939: add endpoint for lqas countries

quang-le:
We need an endpoint that returns dropdown data for countries that have had lqas within a specific month

Related JIRA tickets : POLIO-1939, POLIO-1929

- [x] Are there enough tests?

## Doc

in the code

## Changes

- Add the endpoint + tests 

## How to test

- play with swagger or the django API
- the month should be passed as a `mm-YYYY` string , eg: `12-2022`

## Print screen / video
NA

## Notes

```
fix: empty instance pop up

Refs: IA-3665
```


plugins/polio/api/lqas_im/lqas_im_countries.py
```diff
@@ -0,0 +1,93 @@
...
+            .filter(
+                Q(Q(lqas_ended_at__gte=first_day) & Q(lqas_ended_at__lte=last_day) & Q(lqas_ended_at__isnull=False))
+                | Q(
+                    Q(ended_at__gte=first_day - timedelta(days=10))  # When no lqas end date we use round date +10
+                    & Q(ended_at__lte=last_day - timedelta(days=10))
+                    & Q(lqas_ended_at__isnull=True)
+                )
+            )
+            .values_list("campaign__country__id")
+        )
```
tdethier: I think this is more readable, but it's highly subjective
```suggestion
        with_lqas_end_date = Q(lqas_ended_at__isnull=False, lqas_ended_at__gte=first_day, lqas_ended_at__lte=last_day)
        # When no lqas end date we use round date +10
        without_lqas_end_date = Q(
            lqas_ended_at__isnull=True,
            ended_at__gte=first_day - timedelta(days=10),
            ended_at__lte=last_day - timedelta(days=10),
        )
        countries_with_lqas = (
            Round.objects.filter(campaign__country__in=queryset)
            .filter(with_lqas_end_date | without_lqas_end_date)
            .values_list("campaign__country__id")
        )
```

plugins/polio/api/lqas_im/lqas_im_countries.py
```diff
@@ -0,0 +1,93 @@
...
+
+        return queryset.filter(id__in=countries_with_lqas).distinct("id")
+
+
+class OrgUnitDropDownSerializer(OrgUnitSerializer):
+    label = serializers.SerializerMethodField()
```
tdethier: You can leave your serializer as it is, but FYI if you don't do much processing inside the method field and simply want to change the attribute name in the serializer, you can use the serializer `source` attribute - [like here](https://github.com/BLSQ/iaso/blob/3eb5d0a8efb0c72fdd9ffa433260cf8abe855193/iaso/api/serializers.py#L175)

plugins/polio/api/lqas_im/lqas_im_countries.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+class LqasImCountriesViewset(PolioOrgunitViewSet):
+    http_method_names = ["get"]
+    permission_classes = [HasPolioPermission | HasPolioAdminPermission]
+    filterset_class = LqasImCountryFilter
+    remove_results_key_if_paginated = False
+    results_key = "results"
+
+    def get_serializer_class(self):
```
tdethier, line 71: Here as well, you can leave it as it is, but this method is more useful if you have multiple serializers that need to be used depending on the requested action. Since you always return the same one, you can define it at the viewset class level - [like here](https://github.com/BLSQ/iaso/blob/3eb5d0a8efb0c72fdd9ffa433260cf8abe855193/iaso/api/algorithms.py#L25)

plugins/polio/tests/api/lqas_im/test_lqas_im_countries_options.py
```diff
@@ -0,0 +1,181 @@
...
+        self.assertEqual(len(results), 2)
+        org_unit_ids = [result["value"] for result in results]
+        self.assertFalse(self.india.id in org_unit_ids)
+
+    def test_response_shape(self):
+        self.client.force_authenticate(self.user)
+        response = self.client.get(self.endpoint)
+        response = self.assertJSONResponse(response, 200)
+        res_keys = list(response.keys())
+        print(res_keys)
```
tdethier:
```suggestion
```

plugins/polio/api/campaigns/campaigns.py
```diff
@@ -1438,7 +1438,7 @@ def send_notification_email(self, request, pk, **kwargs):
         return Response({"message": "email sent"})
 
     # We need to authorize PATCH request to enable restore_deleted_campaign endpoint
-    # But Patching the campign directly is very much error prone, so we disable it indirectly
+    # But Patching the campaign directly is very much error prone, so we disable it indirectly
```
tdethier, line 1441: :+1:

plugins/polio/tests/api/lqas_im/test_lqas_im_countries_options.py
```diff
@@ -0,0 +1,181 @@
...
+        results = json_response["results"]
+        self.assertEqual(len(results), 0)
+
+        # Date before lqas dates available
+        response = self.client.get(f"{self.endpoint}?month=12-2020")
+        json_response = self.assertJSONResponse(response, 200)
+        results = json_response["results"]
+        self.assertEqual(len(results), 0)
+
+    def test_lqas_date_used_when_exists(self):
```
tdethier, line 227: Nice one :+1:

plugins/polio/tests/api/lqas_im/test_lqas_im_countries_options.py
```diff
@@ -0,0 +1,181 @@
+import datetime
+
+from iaso.test import APITestCase
+from plugins.polio.tests.api.test import PolioTestCaseMixin
+
+
+class PolioCountriesOptionsTestCase(APITestCase, PolioTestCaseMixin):
```
tdethier, line 8: I would add more tests:
- one with the correct month for India (default values from `create_campaign` is round 1 in 01/2021, round 2 in 02/2021, round 3 in 03/2021), but no results because it's not the right account
-  one - or multiple - that checks all bounds of edge cases. In `test_return_campaign_with_lqasend_date_within_month_param`, the campaign is found because its end round 3 is exactly on the same date as the beginning of the requested month, which is great. We should have similar ones for:
    - end of the requested month = lqas end date
    - beginning of the requested month = 1 day after lqas end date
    - end of the requested month = 1 day before lqas end date
    - all of the above when there's no lqas end date to make sure the 10-day window properly works

tdethier (changes requested):
Overall, it works, nice job :+1: 

I would simply add more tests to make sure the logic regarding end dates is correct

plugins/polio/api/lqas_im/lqas_im_countries.py
quang-le: Even better I'll import the linked serializer 😄

tdethier (approved):
LGTM :+1: 

Nice job splitting this whole LQAS into multiple parts, it will be easier to manage

PR #2183, opened 2025-05-27
POLIO-1941 add more lqas filters

quang-le:

Related JIRA tickets :POLIO-1941

- [x] Are there enough tests?

## Doc
NA

## Changes

- Add endpoints to get Campaigns and rounds for lqas, with filter on dates and country/campaign

## How to test

 play around with django API (no use in UI yet))

## Print screen / video

NA

## Notes

```
fix: empty instance pop up

Refs: IA-3665
```


.gitignore
```diff
@@ -196,3 +196,6 @@ plugins/trypelim
 plugins/snt_malaria
 
 plugins/snt-malaria
+.direnv
+CACHEDIR.TAG
+iaso_3_9_18
```
kemar, line 202: Are those new files linked to SNT-Malaria?

Can you add a comment explaining why we need that?

kemar (changes requested):
The logic is not trivial.

Without tests, it's hard to say what input/output is expected and if the filters and permissions are working as expected.

Can you please add some tests?

.gitignore
quang-le: They're files related to my local setup with direnv, and my venv folder for iaso. I wasn't sure about pushing it, but keeping it local seemed error prone

kemar: Adding these files is fine.

But is it possible to group them with comments? So that the next person looking at this file will know why these files are ignored.

plugins/polio/api/lqas_im/lqas_im_dropdowns.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+class LqasImCountriesOptionsViewset(PolioOrgunitViewSet):
+    http_method_names = ["get"]
+    permission_classes = [HasPolioPermission | HasPolioAdminPermission]
+    filterset_class = LqasImCountryOptionsFilter
+    remove_results_key_if_paginated = False
+    results_key = "results"
+
+    # We need to override this method to override the serializer defined by the pareht viewset
```
kemar:
```suggestion
    # We need to override this method to override the serializer defined by the parent viewset
```

plugins/polio/tests/api/lqas_im/test_lqas_im_options.py
```diff
@@ -0,0 +1,534 @@
+import datetime
+
+from iaso.test import APITestCase
+from plugins.polio.models import Round
+from plugins.polio.tests.api.test import PolioTestCaseMixin
+
+
+class LqasImOptionsTestCase(APITestCase, PolioTestCaseMixin):
+    endpoint: str
+    endpoint = ""
```
kemar, line 10: This looks like a `mixin` to me rather than a test `class`.

Just saying because this kind of code is a bit strange:

```
        if not self.endpoint:
            return
        …
```

kemar: We also usually use an UPPER_CASE naming style for class variables, i.e. `ENDPOINT = ""`.

This makes it easier to understand from where the variable is coming.

plugins/polio/api/lqas_im/lqas_im_dropdowns.py
```diff
@@ -0,0 +1,182 @@
...
+        except:
+            raise ValidationError({"month": [f"Cannot convert {value} to date object"]})
+
+        with_lqas_end_date = Q(lqas_ended_at__isnull=False, lqas_ended_at__gte=first_day, lqas_ended_at__lte=last_day)
+        # When no lqas end date we use round date +10
+        without_lqas_end_date = Q(
+            lqas_ended_at__isnull=True,
+            ended_at__gte=first_day - timedelta(days=10),
+            ended_at__lte=last_day - timedelta(days=10),
+        )
```
kemar: The comment says +10 but the code says -10.

Which one is right?

More importantly, why are we doing this? A comment explaining the "why" would be useful here.

plugins/polio/api/lqas_im/lqas_im_dropdowns.py
```diff
@@ -0,0 +1,182 @@
...
+    def filter_month(self, queryset: QuerySet[OrgUnit], name: str, value: str):
+        try:
+            date_obj = datetime.strptime(value, "%m-%Y")
+            first_day = date_obj.replace(day=1)
+            last_day = date_obj.replace(day=calendar.monthrange(date_obj.year, date_obj.month)[1])
+        except:
+            raise ValidationError({"month": [f"Cannot convert {value} to date object"]})
+
+        with_lqas_end_date = Q(lqas_ended_at__isnull=False, lqas_ended_at__gte=first_day, lqas_ended_at__lte=last_day)
+        # When no lqas end date we use round date +10
```
kemar: Same as above, the comment says +10 but the code says -10.

kemar (changes requested):
Looks almost good to me but I've got a few questions and suggestions.

plugins/polio/tests/api/lqas_im/test_lqas_im_options.py
quang-le: Yeah, but with a mixin the `test_` methods wouldn't be automatically called when running the test, that's why I used a class

kemar (approved):
The shared class used in tests is a bit unconventional, someone might change that one day ;)

PR #2196, opened 2025-06-03
POLIO-1949_supply_chain_filters

quang-le:

Related JIRA tickets :  POLIO-1949

- [ ] Are there enough tests?

## Doc

N/A

## Changes

- Replaced `RoundFilterSet` with a `VaccineFormFilterSet` that handles all filtering behaviour, then remove `filterset_fields` at the viewset level

## How to test

Try the filters on the supplychain page
Try passing a `round_id` via DRF web API (the param is used in OH, but not in the front-end)

## Print screen / video

https://github.com/user-attachments/assets/c28ba23e-09c2-4696-8363-9a2e8e053d69

Screenshot 2025-06-03 at 11 15 51


.gitignore
```diff
@@ -196,3 +196,6 @@ plugins/trypelim
 plugins/snt_malaria
 
 plugins/snt-malaria
+.direnv
+CACHEDIR.TAG
+iaso_3_9_18
```
quang-le, line 201: this is to avoid committing stuff from my local setup

plugins/polio/tests/api/vaccine_supply_chain/test_vaccine_supply_chain.py
```diff
@@ -677,3 +677,113 @@ def test_vrf_cannot_be_created_for_round_without_scope(self):
...
+        test_round = self.vaccine_request_form_rdc_1.rounds.first()
+
+        response = self.client.get(f"{self.BASE_URL}?round_id={test_round.id}")
+        self.assertEqual(response.status_code, 200)
+
+        results = response.data["results"]
+        # Verify each returned form includes the specified round
+        for form in results:
+            round_ids = [r["id"] for r in form["rounds"]]
+            self.assertIn(test_round.id, round_ids)
```
tdethier, line 695: This is a bit weird since the same round won't appear multiple times in the same VRF, right?

Also, I think it might be a little bit better if we test on a specific round (instead of `first()` on line 686) and know exactly how many results we should receive (= raise an error if we get one extra or one fewer)

plugins/polio/tests/api/vaccine_supply_chain/test_vaccine_supply_chain.py
```diff
@@ -677,3 +677,113 @@ def test_vrf_cannot_be_created_for_round_without_scope(self):
...
+
+        # Verify we don't get forms that don't include this round
+        forms_without_round = pm.VaccineRequestForm.objects.exclude(rounds=test_round)
+        for form in forms_without_round:
+            form_ids = [f["id"] for f in results]
+            self.assertNotIn(form.id, form_ids)
```
tdethier, line 701: Testing for a specific number of answers makes this unnecessary

plugins/polio/tests/api/vaccine_supply_chain/test_vaccine_supply_chain.py
```diff
@@ -677,3 +677,113 @@ def test_vrf_cannot_be_created_for_round_without_scope(self):
...
+            f"&rounds__started_at={test_start_date}"
+        )
+
+        self.assertEqual(response.status_code, 200)
+        results = response.data["results"]
+
+        # Verify results match all filter criteria
+        for form in results:
+            self.assertEqual(form["country"]["id"], test_country.id)
+            self.assertEqual(form["vaccine_type"], test_vaccine)
```
tdethier: All VRFs have the same vaccine, so filtering on this is kinda useless without changing the setup a bit

plugins/polio/tests/api/vaccine_supply_chain/test_vaccine_supply_chain.py
```diff
@@ -677,3 +677,113 @@ def test_vrf_cannot_be_created_for_round_without_scope(self):
...
+        test_form = self.vaccine_request_form_rdc_1
+        test_country = test_form.campaign.country
+        test_vaccine = test_form.vaccine_type
+        test_start_date = test_form.rounds.first().started_at
+
+        # Apply multiple filters
+        response = self.client.get(
+            f"{self.BASE_URL}?campaign__country={test_country.id}"
+            f"&vaccine_type={test_vaccine}"
+            f"&rounds__started_at={test_start_date}"
```
tdethier: Here as well, we don't know for sure which round is used (`.first()` on line 710). I would choose a specific one and use a `__lte` or `__gte`filter with it

tdethier (changes requested):
I would change a few tests, first, but otherwise it works :+1:

plugins/polio/tests/api/vaccine_supply_chain/test_vaccine_supply_chain.py
quang-le: Actually, we can have multiple VRFs for one round

plugins/polio/tests/api/vaccine_supply_chain/test_vaccine_supply_chain.py
quang-le: Ok, I removed the filter and test on vaccine type. Since we're testing filter combinbation, and only round_id uses custom code, it's safe to assume that combining 2 filters is as good as combining all (I truist the framework)

plugins/polio/tests/api/vaccine_supply_chain/test_vaccine_supply_chain.py
```diff
@@ -677,3 +677,110 @@ def test_vrf_cannot_be_created_for_round_without_scope(self):
...
+            self.assertNotIn(form.id, form_ids)
+
+    def test_multiple_filters(self):
+        """Test that multiple filters can be combined"""
+        self.client.force_authenticate(user=self.user_rw_perm)
+
+        test_form = self.vaccine_request_form_rdc_1
+        test_country = test_form.campaign.country
+        test_vaccine = test_form.vaccine_type
+        test_start_date = test_form.rounds.filter(number=1).first()
```
tdethier:
```suggestion
        test_start_date = test_form.rounds.filter(number=1).first().started_at
```

tdethier (approved):

PR #2204, opened 2025-06-05
IA-4271 Fix mobile bulk upload

kemar:
Fix mobile bulk upload.

Related JIRA tickets : IA-4271

## Changes

I thought that the background task was tested with `_immediate=False` but that wasn't the case.

I added a proper unit test to reproduce the error seen in prod and then fixed it.


bmonjoie (approved):
LGTM

madewulf (approved):

PR #2207, opened 2025-06-05
[IA-4238] Proper fix for Enketo manifest

tdethier:
Added a signed URL system for enketo to properly fetch form manifests without authentication

Related JIRA tickets : IA-4238

- [x] Are there enough tests?

## Doc

/

## Changes
- Added a system of signed URL
- Added a utility test method for loading a jinja2 template
- Used the signed URLs with Enketo
- Added a new ENV - `ENKETO_SIGNING_SECRET`

## How to test

- Define a new environment variable - `ENKETO_SIGNING_SECRET` - with the value of your choice (this will be used to sign the URLs)
- Have a form that has at least 1 attachment 
   - You can use the form 1027 on staging as an example (see Jira ticket) - https://iaso-staging.bluesquare.org/api/forms/1027
   - Don't forget to download its attachment and set it up in your local installation too :wink: 
- Try to use Enketo to fill out the form
- There shouldn't be any error now 

## Print screen / video
image
image

## Notes

- Thanks @mestachs for the great solution!
- the XML manifest format changed (different number of spaces on various lines), I'm not sure if that's an issue or not

```
fix: empty instance pop up

Refs: IA-3665
```


.gitignore
```diff
@@ -199,3 +199,5 @@ plugins/snt-malaria
 .direnv
 CACHEDIR.TAG
 iaso_3_9_18
+
+iaso-enketo/
```
tdethier, line 205: I added this in order to avoid breaking my local setup and uploading stuff by mistake

kemar: This comment could be more useful in the file itself so that someone can understand why we're ignoring this.

iaso/enketo/enketo_url.py
```diff
@@ -18,7 +23,9 @@ class EnketoError(Exception):
     pass
 
 
-def enketo_settings():
+def enketo_settings(specific_env: str = None):
```
kemar, line 26: The name `specific_env` is a bit misleading because my first assumption was to use `staging` or `prod` :)

But I understand you need this because of the structure of the settings.

kemar (approved):
LGTM

hat/settings.py
```diff
@@ -115,6 +115,7 @@
     "ENKETO_DEV": os.getenv("ENKETO_DEV"),
     "ENKETO_API_TOKEN": os.getenv("ENKETO_API_TOKEN"),
     "ENKETO_URL": os.getenv("ENKETO_URL"),
+    "ENKETO_SIGNING_SECRET": os.getenv("ENKETO_SIGNING_SECRET"),
```
madewulf: What happens on environment that don't have the environment variable set? 

I would tend to use by default the django secret key (if that does not create a security risk) but being able to override it with an environment variable. 

I don't want all deployments to break when upgrading IASO without setting the new environment variable, if we can avoid that

tdethier: Unit tests are running without the environment variable, so it should be ok, but we can already define it on all environments now if you want. 

I was specifically told to not use the default Django secret key

tdethier: I added a fallback 😊

tdethier:
@bmonjoie this might interfere with how the mobile app fetches forms & form attachments, but we're not sure. Could you please tell us if it causes any issue?

iaso/api/forms.py
```diff
@@ -426,34 +428,40 @@ def manifest(self, request, *args, **kwargs):
...
-
-            media_files.append(
-                f"""<mediaFile>
-    <filename>{escape(attachment.name)}</filename>
-    <hash>md5:{attachment.md5}</hash>
-    <downloadUrl>{escape(attachment_file_url)}</downloadUrl>
-</mediaFile>"""
-            )
+        enketo_secret = enketo_settings("ENKETO_SIGNING_SECRET")
+        if verify_signed_url(request, enketo_secret):
```
bmonjoie: Why did you decide to break with the existing? Why can't they live side-by-side?

bmonjoie (changes requested):
This will never work with mobile as this is breaking with the existing and requires the new signing method.

You should not check for the signature in the method itself but add a permission class and check there. You either are authenticated and you have the right to access the file or you have the signing method but not just one.

iaso/api/forms.py
tdethier: I was simply told to use this solution, I fixed it :sweat_smile:

iaso/api/forms.py
```diff
@@ -437,10 +448,10 @@ def manifest(self, request, *args, **kwargs):
...
             media_files.append(
                 f"""<mediaFile>
-    <filename>{escape(attachment.name)}</filename>
-    <hash>md5:{attachment.md5}</hash>
-    <downloadUrl>{escape(attachment_file_url)}</downloadUrl>
-</mediaFile>"""
+                        <filename>{escape(attachment.name)}</filename>
+                        <hash>md5:{attachment.md5}</hash>
+                        <downloadUrl>{escape(attachment_file_url)}</downloadUrl>
+                    </mediaFile>"""
```
bmonjoie, line 454: You should remove those, this makes the payload bigger, and it won't look good when we'll look at the content. Same remark below.

bmonjoie (approved):

PR #2209, opened 2025-06-06
Rebase snt malaria based on main and merge migrations

Bewi:
I create a PR just so you can review the rebase/migration merge from main.
Now, I think to keep a git tree clean, it is best if I push it into the snt-malaria branch directly.
But before I break your local, I think it is better if you confirm that I didn't made a mistake.


bramj:
@Bewi I'd say this is impossible to give a proper review :) but I'd say if

- the rebase runs correctly locally
- you only had to fix some Django migrations (which looks like it's the case)

then you can push it into the `snt-malaria` branch.

PR #2215, opened 2025-06-11
IA-4276, POLIO-1953: add alternative obr field

quang-le:
User needs to be able te enter reference for non SIA related form A

Related JIRA tickets : IA-4276, POLIO-1953

- [ ] Are there enough tests?

## Doc

Comments in the code

## Changes

- Update Form A Modal to switch between Obr name selection and custom name:
    - On create: users can tick a check box to toggle between campaign selection and custom name
    - On edit: users can edit the campaign/name selected but not toggle between the 2 modes
- Add validation rule
- Add translations
- Fix small display bug with Round 0 in Form A table (POLIO-1953)

## How to test

Go to vaccine stock > stock variation > form a

Try to create a form a with campaign name and with custom name
Check that they appear on the table

## Screen capture

https://github.com/user-attachments/assets/3a6a0c7c-1539-44d1-9942-cc6f1c5731cd


plugins/polio/api/vaccines/stock_management.py
```diff
@@ -261,6 +261,7 @@ def validate(self, data):
         # the param name has been changed to the model field name : non_obr_name
         if data.get("campaign", None) is not None and data.get("non_obr_name", None) is not None:
             raise serializers.ValidationError({"error": "campaign and alternative campaign cannot both be defined"})
+        print("data", data)
```
beygorghor: 🗑️

plugins/polio/js/src/domains/VaccineModule/StockManagement/StockVariation/Modals/CreateEditFormA.tsx
```diff
@@ -80,6 +104,27 @@ export const CreateEditFormA: FunctionComponent<Props> = ({
...
+        }
+    }, [
+        withCustomObr,
+        setFieldValue,
+        setValues,
+        formik.values,
+        formik.values.campaign,
+        formik.values.alternative_campaign,
+    ]);
+    console.log('VALUES', formik.values);
```
beygorghor: 🗑️

beygorghor (approved):
Small things to remove. 
Also for the choice between the 2 options, i was more seeing a radio box, with both option visible, enabled while selected.

quang-le:
> Small things to remove. Also for the choice between the 2 options, i was more seeing a radio box, with both option visible, enabled while selected.

Mmmh, why not. But the radio option would have made state a bit more annoying to manage, because if you had 2 values, you would have needed to pass the radio's value to the API hook, then remove it from the payload sent to the API (because the API doesn't tolerate having both).

PR #2218, opened 2025-06-12
IA-4162 New API endpoint to edit the geometric shape of an org unit

kemar:
New API endpoint to edit the geometric shape of an org unit.

Related JIRA tickets : IA-4162

## Changes

New API endpoint to edit the geometric shape of an org unit with 3 methods:

- GET `/api/orgunits/shape/` (list)
- GET `/api/orgunits/shape/{org_unit_id}`
- PATCH `/api/orgunits/shape/{org_unit_id}`

The `PATCH` body consists of a single `geom` value:

```
{"geom": "SRID=4326;MULTIPOLYGON (((-1.3 2.5, -1.7 2.8, -1.1 4.1, -1.3 2.5)))"}
```

This value is the normal (complex) geom. It will be used to regenerate the simplified version.

## How to test

See the unit tests.

Also have a look at the GUI at `http://localhost:8081/api/orgunits/shape/`

## Print screen / video

geo


iaso/api/org_unit_shapes/serializers.py
```diff
@@ -0,0 +1,43 @@
...
+            return GEOSGeometry(value, srid=4326)
+        except (TypeError, ValueError):
+            raise serializers.ValidationError("Invalid geometry.")
+
+    def update(self, instance: OrgUnit, validated_data: dict) -> OrgUnit:
+        original_org_unit = audit_models.serialize_instance(instance)
+
+        geom = validated_data["geom"]
+        instance.geom = geom
+        instance.simplified_geom = simplify_geom(geom)
```
beygorghor, line 42: has discussed with Martin, we should save the normal (complex) geom and then regenerate the simplified version

kemar: this is what I'm doing here…

iaso/api/org_unit_shapes/pagination.py
```diff
@@ -0,0 +1,5 @@
+from iaso.api.common import Paginator
+
+
+class OrgUnitShapePagination(Paginator):
+    page_size = 10
```
quang-le, line 5: I you want to be super careful, there's also a `max_page_size` property

iaso/tests/api/org_unit_shapes/test_serializers.py
```diff
@@ -0,0 +1,77 @@
...
+        cls.user = cls.create_user_with_profile(username="User", account=cls.account, permissions=["iaso_org_units"])
+
+    def test_serialize_ok(self):
+        serializer = OrgUnitShapeSerializer(self.org_unit)
+        json = JSONRenderer().render(serializer.data)
+        expected_json = {
+            "id": self.org_unit.pk,
+            "name": self.org_unit.name,
+            "geom": "SRID=4326;MULTIPOLYGON (((-1.3 2.5, -1.7 2.8, -1.1 4.1, -1.3 2.5)))",
+            "simplified_geom": None,
```
quang-le, line 37: Should we test `simplified_geom` or can we assume that if `geom`is serialized correctly, `simplified_geom` is as well?

iaso/api/org_unit_shapes/pagination.py
kemar: Good point. This makes me think that we should use a common pagination class for most of our endpoints.

iaso/tests/api/org_unit_shapes/test_serializers.py
kemar: Yup we can assume that if `geom` is good, `simplified_geom` will work as well because of the way the function works https://github.com/BLSQ/iaso/blob/27d90bee7f11826697a330c507aebd81fc8c49de/iaso/utils/gis.py#L15

Here I'm only testing that the serialization to JSON is OK. The serializer should ouput `None` when there is no `simplified_geom `.

kemar:
This PR is replaced by #2223

PR #2219, opened 2025-06-13
IA-4273: Fix blank page when 'select-multiple' question is used in form module

beygorghor:
When a form contains a 'select-multiple' question, the request returns a blank page in the form module (specifically in the Advanced Search option on the submissions page). Investigate and resolve the issue so that multiple-choice questions do not cause the page to fail.

Related JIRA tickets : IA-4273

- [x] Are there enough tests?

## Doc

## Changes

- fix mult iselect fiend in `[bluesquare-components](https://github.com/BLSQ/bluesquare-components/pull/182)` to use array of string instead of a pure string as values
- duplicate backend code to be use only with instances and support `some` and `all` operators 
- adding some test to cover those operators

## How to test

You need a form using multi select fields and submissions with this form.

go to instance page, select this form and perform some searches with the query builder on the multi select field 

## Print screen / video

https://github.com/user-attachments/assets/d411f195-79e6-4148-ab75-fcd1a5a12bb2

## Notes

⚠️ this depend on blsq-component [PR](https://github.com/BLSQ/bluesquare-components/pull/182)

```
fix: empty instance pop up

Refs: IA-3665
```

### DIFF
--- a/iaso/utils/jsonlogic.py
+++ b/iaso/utils/jsonlogic.py
@@ -42,9 +42,88 @@ def jsonlogic_to_q(
     field_prefix: str = "",
     recursion_func: Callable = None,
 ) -> Q:
-    """Converts a JsonLogic query to a Django Q object."""
+    """Converts a JsonLogic query to a Django Q object.
+
+    :param jsonlogic: The JsonLogic query to convert, stored in a Python dict. Example: {"and": [{"==": [{"var": "gender"}, "F"]}, {"<": [{"var": "age"}, 25]}]}
+    :param field_prefix: A prefix to add to all fields in the generated query. Useful to follow a relationship or to dig in a JSONField
+    :param recursion_func: Optionally specify a function to call for recursion, allowing this method to be "wrapped". By default, when no function is specified, it calls itself.
+
+    :return: A Django Q object.
+    """
 
     func = jsonlogic_to_q if recursion_func is None else recursion_func
+
+    if "and" in jsonlogic:
+        sub_query = Q()
+        for lookup in jsonlogic["and"]:
+            sub_query = operator.and_(sub_query, func(lookup, field_prefix))
+        return sub_query
+    if "or" in jsonlogic:
+        sub_query = Q()
+        for lookup in jsonlogic["or"]:
+            sub_query = operator.or_(sub_query, func(lookup, field_prefix))
+        return sub_query
+    if "!" in jsonlogic:
+        return ~func(jsonlogic["!"], field_prefix)
+
+    if not jsonlogic.keys():
+        return Q()
+
+    # Binary operators
+    op = list(jsonlogic.keys())[0]
+    params = jsonlogic[op]
+    if len(params) != 2:
+        raise ValueError(f"Unsupported JsonLogic. Operator {op} take exactly two operands: {jsonlogic}")
+
+    lookups = {
+        "==": "exact",
+        "!=": "exact",
+        ">": "gt",
+        ">=": "gte",
+        "<": "lt",
+        "<=": "lte",
+        "in": "icontains",
+    }
+
+    if op not in lookups.keys():
+        raise ValueError(
+            f"Unsupported JsonLogic (unknown operator {op}): {jsonlogic}. Supported operators: f{lookups.keys()}"
+        )
+
+    field_position = 1 if op == "in" else 0
+    field = params[field_position]
+    if "var" not in field:
+        raise ValueError(
+            f"Unsupported JsonLogic. Argument[{field_position}] must contain a variable for given "
+            f"operator : {jsonlogic}"
+        )
+    field_name = field["var"]
+    value = params[0] if op == "in" else params[1]
+
+    extract = ""
+    if isinstance(value, (int, float)) and field_prefix:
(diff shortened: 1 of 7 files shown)


iaso/utils/jsonlogic.py
```diff
@@ -115,6 +116,114 @@ def jsonlogic_to_q(
     return q
 
 
+def instance_jsonlogic_to_q(
+    jsonlogic: Dict[str, Any],
+    field_prefix: str = "",
+    recursion_func: Callable = None,
+) -> Q:
+    """Converts a JsonLogic query to a Django Q object."""
```
Bewi: Je serais pour d'avoir un exemple de JSON ici, ça aiderait à maintenir et re-utiliser sans devoir aller check les tests :D

iaso/utils/jsonlogic.py
```diff
@@ -115,6 +116,114 @@ def jsonlogic_to_q(
...
+
+    func = instance_jsonlogic_to_q if recursion_func is None else recursion_func
```
quang-le: this var needs a better name to make the code clearer

iaso/utils/jsonlogic.py
quang-le: Agreed. Some clarification about what the params are and an example of function call would really help as well. Looking at the code like this, it's hard to figure what `field_prefix` is and the role it plays in the recursion

iaso/utils/jsonlogic.py
```diff
@@ -115,6 +116,114 @@ def jsonlogic_to_q(
...
+        field = var["var"]
+        # Only support: { "in": [ { "var": "" }, [array] ] }
+        if "in" in condition:
+            in_params = condition["in"]
+            if (
+                isinstance(in_params, list)
+                and len(in_params) == 2
+                and isinstance(in_params[0], dict)
+                and in_params[0].get("var") == ""
+            ):
```
quang-le: Shouldn't we raise a `ValueError` if this condition is not met?

iaso/utils/jsonlogic.py
```diff
@@ -115,6 +116,114 @@ def jsonlogic_to_q(
...
+                value_list = in_params[1]
+                if operator_key == "some":
+                    # All values must be present
```
quang-le: I don't understand this comment

iaso/utils/jsonlogic.py
```diff
@@ -115,6 +116,114 @@ def jsonlogic_to_q(
...
+                    q = Q()
+                    for v in value_list:
+                        q = q & Q(**{f"{field_prefix}{field}__icontains": v})
```
quang-le: why not use `operator.and_ ` to combine the queries here, like in the code for group operators?

iaso/utils/jsonlogic.py
```diff
@@ -115,6 +116,114 @@ def jsonlogic_to_q(
...
+                    sorted_values = sorted(value_list)
+                    # Create a pattern that matches the sorted string representation
+                    pattern = r"^" + r" ".join(re.escape(v) for v in sorted_values) + r"$"
+                    q = Q(**{f"{field_prefix}{field}__regex": pattern})
+                    return q
+        raise ValueError(f"Unsupported JsonLogic for '{operator_key}': {jsonlogic}")
+
+    # Handle binary operators
+    if not jsonlogic.keys():
+        return Q()
```
quang-le: Shouldn't that be higher in the code, to return as early as possible?

iaso/utils/jsonlogic.py
```diff
@@ -115,6 +116,114 @@ def jsonlogic_to_q(
...
+        raise ValueError(f"Unsupported 'in' usage: {jsonlogic}")
+
+    lookups = {
+        "==": "exact",
+        "!=": "exact",
+        ">": "gt",
+        ">=": "gte",
+        "<": "lt",
+        "<=": "lte",
+    }
```
quang-le: This should be a `LOOKUPS` constant at the module level

iaso/utils/jsonlogic.py
```diff
@@ -115,6 +116,114 @@ def jsonlogic_to_q(
...
+
+    field = params[0]
+    value = params[1]
+    if "var" not in field:
+        raise ValueError(f"Unsupported JsonLogic. Argument[0] must contain a variable for given operator : {jsonlogic}")
+    field_name = field["var"]
+
+    extract = ""
+    if isinstance(value, (int, float)) and field_prefix:
+        extract = "__forcefloat"
```
quang-le: `def validate_operations` or something

quang-le: A comment explaining why we use forcefloat would be useful

iaso/utils/jsonlogic.py
quang-le, line 1: I think this function needs more comments, and /or to be split. If you're not already quite familiar with jsonlogic, it's hard to make sense of what happens.
An example of function call + arguments and return value in the comments would help.

quang-le (changes requested):
I think some comments, renaming, light refactoring are needed to make teh code easier to understand. Also, I think there's a `raise` missing somewhere (see comments)

quang-le (approved):

PR #2221, opened 2025-06-16
IA-4287 Add MD5 checksum to FormVersions

bmonjoie:
ODK allows us to pass `md5` checksum to avoid downloading known forms. 
This adds the information so that we can avoid downloading forms that are known.

Related JIRA tickets : IA-4287 IA-4268

- [X] Are there enough tests?

## Changes

Added a new `md5` field to `FormVersion`

## How to test

- List a `Project`'s  `FormVersion`s using the API: /api/formversions?app_id=...
- Check that the field `md5` is null for all existing forms
- Add a new `FormVersion` through the UI
- Check that the field `md5` is not null for the new form version.
- Optional, using the local command `md5sum` on your computer, check that the md5 matches the md5 of the xml file.

```
fix: empty instance pop up

Refs: IA-3665
```


iaso/models/forms.py
```diff
@@ -250,11 +252,12 @@ class FormVersionManager(models.Manager):
...
         with transaction.atomic():
             latest_version = self.latest_version(form)  # type: ignore
-
+            file = SimpleUploadedFile(survey.generate_file_name("xml"), survey.to_xml(), content_type="text/xml")
             form_version = super().create(
                 **kwargs,
                 form=form,
-                file=SimpleUploadedFile(survey.generate_file_name("xml"), survey.to_xml(), content_type="text/xml"),
+                file=file,
+                md5=self.md5sum(file),
```
kemar: Do we need to do this only at creation time?

Can the `file` be changed during the lifetime of the `FormVersion`?

iaso/models/forms.py
```diff
@@ -277,6 +287,7 @@ class FormVersion(models.Model):
     form = models.ForeignKey(Form, on_delete=models.CASCADE, related_name="form_versions")
     # xml file representation
     file = models.FileField(upload_to=_form_version_upload_to)
+    md5 = models.CharField(blank=False, null=True, max_length=32)  # Nullable because of legacy
```
kemar: To make a string-based field optional, the Django convention is to avoid using `null=True`.

See https://sentry.io/answers/django-difference-between-null-and-blank/

The absence of value will be represented only as an empty string.

iaso/models/forms.py
```diff
@@ -267,6 +270,13 @@ def create_for_form_and_survey(self, *, form: "Form", survey: parsing.Survey, **
 
         return form_version
 
+    @staticmethod
+    def md5sum(file: File):
```
kemar:
```suggestion
    def md5sum(file: File) -> str:
```

iaso/api/form_versions.py
kemar, line 1: Changes in this file should have broken at least one test… 😓

kemar (changes requested):

iaso/models/forms.py
bmonjoie: AFAIK, `FormVersion`s never change. You can only add a new one.

iaso/models/forms.py
bmonjoie: But I don't want an empty string, I want either a string with a value or null. It's not optional anymore. It's only nullable because we can't go back and update the database for all the files we have.

iaso/api/form_versions.py
bmonjoie: That's what I thought as well but it didn't :upside_down_face:

iaso/migrations/0328_formversion_md5.py
```diff
@@ -0,0 +1,44 @@
...
+
+from django.core.files import File
+from django.db import migrations, models
+
+
+def md5sum(file: File) -> str:
+    md5 = hashlib.md5()
+    for chunk in file.chunks():
+        md5.update(chunk)
+    return md5.hexdigest()
```
tdethier: :warning: this is not relevant to this PR

@beygorghor @kemar shouldn't we use md5 sums in polio endpoints as well to compare files? When we update an object that has a file, file comparisons are made on name and size only... which is "okay", but this would be much better imo

kemar: @tdethier yup this is a good idea, we should do that.

beygorghor: yes totally agree !

beygorghor (approved):
LGTM

kemar (approved):

PR #2222, opened 2025-06-17
WC2-736: another duplicate issue

beygorghor:
The error occurred because the similarity score calculation could produce values outside the valid smallint range (-32768 to 32767). 
Related JIRA tickets :  WC2-736

- [ ] Are there enough tests?

## Doc

## Changes

The fix ensures the score stays within the expected 0-100 range by:

1. Using `COALESCE` to handle NULL values in field comparisons
2. Using `NULLIF` to prevent division by zero
3. Clamping the final score between 0 and 100 using `GREATEST` and `LEAST`

This maintains the existing smallint database schema while ensuring valid score values for the deduplication algorithm.

## How to test

Explain how to test your PR.

## Print screen / video

https://github.com/user-attachments/assets/f113f262-a87e-4f11-ae77-4c7ce9bf5799

## Notes

```
fix: empty instance pop up

Refs: IA-3665
```

### DIFF
--- a/iaso/api/deduplication/algos/levenshtein.py
+++ b/iaso/api/deduplication/algos/levenshtein.py
@@ -29,9 +29,12 @@ def _build_query(params):
             # if field is a number we need to get as a result the difference between the two numbers
             # the final value should be 1 - (abs(number1 - number2) / max(number1, number2))
             fc_arr.append(
-                "(1.0 - ( abs ( (instance1.json->>%s)::double precision - (instance2.json->>%s)::double precision ) / greatest( (instance1.json->>%s)::double precision, (instance2.json->>%s)::double precision )))"
+                "(1.0 - (CASE "
+                "WHEN (instance1.json->>%s) IS NOT NULL AND (instance1.json->>%s) != '' AND (instance2.json->>%s) IS NOT NULL AND (instance2.json->>%s) != '' "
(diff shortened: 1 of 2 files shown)


iaso/api/deduplication/algos/levenshtein.py
```diff
@@ -29,9 +29,12 @@ def _build_query(params):
             # if field is a number we need to get as a result the difference between the two numbers
             # the final value should be 1 - (abs(number1 - number2) / max(number1, number2))
             fc_arr.append(
-                "(1.0 - ( abs ( (instance1.json->>%s)::double precision - (instance2.json->>%s)::double precision ) / greatest( (instance1.json->>%s)::double precision, (instance2.json->>%s)::double precision )))"
+                "(1.0 - (CASE "
+                "WHEN (instance1.json->>%s) IS NOT NULL AND (instance1.json->>%s) != '' AND (instance2.json->>%s) IS NOT NULL AND (instance2.json->>%s) != '' "
+                "THEN abs( (instance1.json->>%s)::double precision - (instance2.json->>%s)::double precision ) / greatest( (instance1.json->>%s)::double precision, (instance2.json->>%s)::double precision ) "
+                "ELSE NULL END))"
```
tdethier, line 35: It's getting harder and harder to read, on top of not being f-strings :face_with_spiral_eyes:

tdethier (approved):
Still not a fan of casting everything in SQL instead of using f-strings/parsing all elements beforehand to reduce the possibilities, but let's test this with real data!

PR #2223, opened 2025-06-18
IA-4162 Allow to edit the geometric shape of an org unit

kemar:
Allow to edit the geometric shape of an org unit.

Related JIRA tickets : IA-4162

---

This PR cancels and replaces #2218.

To restore shape edition as quickly as possible, we decided to rely on the existing API instead of adding a new endpoint.

### DIFF
--- a/iaso/api/org_units.py
+++ b/iaso/api/org_units.py
@@ -821,14 +821,14 @@ class OrgUnitViewSet(viewsets.ViewSet):
             self.get_queryset().prefetch_related("reference_instances"),
             pk=pk,
         )
-        # Get instances count for the Org unit and its descendants
-        instances_count = org_unit.descendants().aggregate(Count("instance"))["instance__count"]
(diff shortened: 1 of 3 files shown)


iaso/api/org_units.py
```diff
@@ -527,11 +528,18 @@ def partial_update(self, request, pk=None):
         elif "simplified_geom" in request.data:
             org_unit.simplified_geom = request.data["simplified_geom"]
 
-        if "geo_json" in request.data and request.data["geo_json"]:
-            logger.warning(
-                "The `geo_json` field is deprecated. Use the `geom` field to modify the geometry.",
-                extra={"request_data": request.data},
-            )
```
quang-le, line 534: So that's why nothing worked 😅

quang-le (approved):
Problem solved

https://github.com/user-attachments/assets/28d52a4a-c690-4dec-b07a-e53af7ac1f97

PR #2224, opened 2025-06-18
[IA-4295] Fix enketo manifest with auth feature flag

tdethier:
Let enketo download a form manifest even when the auth feature flag is enabled.

Related JIRA tickets : IA-4295

- [x] Are there enough tests?

## Doc

/

## Changes

- Split manifest endpoints into 2:
   - one for mobile (already existing) -> revert changes from previous PR back to its original state
   - one for enketo (new one)

## How to test

- Choose a form that has attachments
- Make sure it's linked to a project that has the auth feature flag enabled
- Start enketo
- Make a new submission for your form using enketo - there's shouldn't be any error

## Print screen / video

image

## Notes

The endpoint was split because we need to use a different queryset from the one used in the mobile endpoint (no filtering for forms/projects for anonymous user; direct filtering on `app_id`). 

```
fix: empty instance pop up

Refs: IA-3665
```

### DIFF
--- a/iaso/api/forms.py
+++ b/iaso/api/forms.py
@@ -24,7 +24,7 @@ from ..permissions import IsAuthenticatedOrReadOnlyWhenNoAuthenticationRequired,
 from .common import CONTENT_TYPE_CSV, CONTENT_TYPE_XLSX, DynamicFieldsModelSerializer, ModelViewSet, TimestampField
 from .enketo import public_url_for_enketo
 from .projects import ProjectSerializer
-from .query_params import APP_ID
+from .serializers import AppIdSerializer
 
 
 class HasFormPermission(IsAuthenticatedOrReadOnlyWhenNoAuthenticationRequired):
@@ -442,7 +442,7 @@ class FormsViewSet(ModelViewSet):
         """
         enketo_secret = enketo_settings("ENKETO_SIGNING_SECRET")
         if verify_signed_url(request, enketo_secret):
-            app_id = request.query_params.get(APP_ID, None)
+            app_id = AppIdSerializer(data=request.query_params).get_app_id(raise_exception=True)
             queryset = Form.objects.filter(
                 projects__app_id=app_id
             )  # Not using the default queryset because there's no auth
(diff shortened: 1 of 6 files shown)


iaso/api/forms.py
```diff
@@ -431,44 +423,81 @@ def destroy(self, request, *args, **kwargs):
...
+    @action(detail=True, methods=["get"], permission_classes=[ReadOnly])
+    def manifest_enketo(self, request, pk, *args, **kwargs):
+        """Returns a xml manifest file in the openrosa format for the Form
+
+        This is used by enketo to fetch the list of files attached to the Form
+        see https://docs.getodk.org/openrosa-form-list/#the-manifest-document
+        """
+        enketo_secret = enketo_settings("ENKETO_SIGNING_SECRET")
+        if verify_signed_url(request, enketo_secret):
+            app_id = request.query_params.get(APP_ID, None)
```
kemar: Should we really allow `app_id=None` here?

If the response is no, I'd go for something like that:

```python
app_id = AppIdSerializer(data=request.query_params).get_app_id(raise_exception=True)
```

iaso/api/forms.py
```diff
@@ -431,44 +423,81 @@ def destroy(self, request, *args, **kwargs):
...
+def generate_manifest_media_file(attachment: FormAttachment, url: str) -> str:
+    return f"""<mediaFile>
+        <filename>{escape(attachment.name)}</filename>
+        <hash>md5:{attachment.md5}</hash>
+        <downloadUrl>{escape(url)}</downloadUrl>
+    </mediaFile>"""
+
+
+def generate_manifest_structure(content: list[str]) -> str:
+    nl = "\n"  # Backslashes are not allowed in f-string ¯\_(ツ)_/¯
```
kemar, line 492: TIL

kemar (commented):
Almost good to me.

I just have one question about handling the case where `app_id=None`.

iaso/api/forms.py
tdethier: Yeah this comment was already there, I just refactored this block of code :man_shrugging:

iaso/api/forms.py
tdethier: We shouldn't, but if we get `None` the queryset will be empty on line 446 and a `404` will be raised on line 449

However, I'm still going to use this serializer, it sounds like a good idea

kemar (approved):

PR #2226, opened 2025-06-19
HOTFIX IA-4162 Allow to edit the geometric shape of an org unit

kemar:
Hotfix for #2223


kemar:
Deployed as a hotfix on the demo environment https://github.com/BLSQ/iaso/actions/runs/15753836773

PR #2227, opened 2025-06-19
IA-4293 notif email bug

quang-le:

Related JIRA tickets : IA-4293

- [x] Are there enough tests?

## Doc
NA
## Changes

- Modify code generating mailing list so that it generates a list of emails i.o a list of ids
- Add `try...except` statements to have the task fail when an email fails to be sent
- Add `terminate_with_error_ method on `Task` model for convenience
- Add tests

## How to test

Manually trigger `send_email_vaccine_authorizations_60_days_expiration_alert` and `send_email_expired_vaccine_authorizations_alert` (needs worker). The tasks should not fail.

To further test that exceptions are raised, run  `expired_vaccine_authorizations_email_alert` and `vaccine_authorizations_60_days_expiration_email_alert` with a badly formatted mailing list. It's dubious this will error on localhost though due to the django config in that env. On the other hand this cases are covered by tests

## Print screen / video
NA

## Notes

```
fix: empty instance pop up

Refs: IA-3665
```

### DIFF
--- a/plugins/polio/tasks/vaccine_authorizations_mail_alerts.py
+++ b/plugins/polio/tasks/vaccine_authorizations_mail_alerts.py
@@ -22,8 +22,6 @@ vaccine_dashboard_link = settings.DNS_DOMAIN
 
 
 def vaccine_authorizations_60_days_expiration_email_alert(vaccine_auths, mailing_list):
-    mail_sent = False
-
     for obj in vaccine_auths:
         account_id = obj.account.id
         next_vaccine_auth = (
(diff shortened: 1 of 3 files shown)


plugins/polio/tests/test_vaccine_authorizations.py
```diff
@@ -1127,3 +1124,93 @@ def test_cant_update_to_validate_if_another_validate(self):
 
         self.assertEqual(response.status_code, 403)
         self.assertEqual(response.data["error"], "A vaccine authorization is already validated for this country")
+
+    # Mock send_mail to raise an exception when receiving invalid recipients
+    def mock_send_mail(self, subject, message, from_email, recipient_list, **kwargs):
+        print("---------------------------------")
+        print("recipients list", recipient_list)
+        print("---------------------------------")
```
kemar:
```suggestion
```

plugins/polio/tasks/vaccine_authorizations_mail_alerts.py
```diff
@@ -35,41 +33,47 @@ def vaccine_authorizations_60_days_expiration_email_alert(vaccine_auths, mailing
...
+                    A new authorization is {next_vaccine_auth}, with an expiry date on {next_vaccine_auth.expiration_date}.
+                    Please take appropriate action as needed. 
+                    Link to the platform vaccine authorization page : {vaccine_dashboard_link}/dashboard/polio/vaccinemodule/nopv2authorisation/accountId/{account_id}/order/-current_expiration_date/pageSize/20/page/1"
+                    RRT team
+                    """,
+                    settings.DEFAULT_FROM_EMAIL,
+                    mailing_list,
+                )
+
+            except Exception as e:
```
kemar, line 52: I'd probably have catched a `smtplib.SMTPRecipientsRefused` (https://docs.python.org/3/library/smtplib.html#smtplib.SMTPRecipientsRefused) because it seems that that's what happened in the Sentry error and it has the advantage of being more specific.

kemar (approved):
LGTM

A few comments but nothing blocking.

PR #2229, opened 2025-06-19
IA-4318 Add a migration to fix form_version_id being null where possible

bmonjoie:
Add a migration to fix form_version_id being null where possible

Related JIRA tickets : IA-4318

## How to test

- Check if you have Instance without FormVersion, if not, put the form_version to null directly in the database
- Apply the migration
- Check that the FormVersion is correctly set now

```
fix: empty instance pop up

Refs: IA-3665
```


iaso/migrations/0328_auto_20250619_1436.py
```diff
@@ -0,0 +1,40 @@
...
+from django.db.models.expressions import RawSQL
+
+
+def migrate_data_forward(apps, schema_editor):
+    chunk_size = 10000
+    Instance = apps.get_model("iaso", "Instance")
+
+    instances = (
+        Instance.objects.filter(form_version__isnull=True)
+        .filter(json__isnull=False)
```
tdethier:
```suggestion
        Instance.objects.filter(form_version__isnull=True, json__isnull=False)
```

kemar:
~Do we have a Jira issue for this?~

I created one IA-4318

iaso/migrations/0328_auto_20250619_1436.py
```diff
@@ -0,0 +1,42 @@
...
+        .iterator(chunk_size=chunk_size)
+    )
+
+    instance_to_update = []
+    for instance in instances:
+        if not instance.annotated_form_version_id:
+            continue
+        instance.form_version_id = instance.annotated_form_version_id
+        instance_to_update.append(instance)
+        if len(instance_to_update) > chunk_size:
```
tdethier:
```suggestion
        if len(instance_to_update) >= chunk_size:
```

I'm not sure if this is going to improve anything, but with `>` bulk updates will be triggered once we have 10001 elements, which means fetching 2 chunks from the iterator instead of 1

tdethier (approved):
I don't have any example of this in my local setup, so it's a bit difficult to test. 

The logic and optimization sound good though

PR #2230, opened 2025-06-19
[IA-4301] Add OrgUnit.code + constraint

tdethier:
Added `OrgUnit.code` + unicity constraint + tests

Related JIRA tickets : IA-4301, IA-4030 (epic)

- [x] Are there enough tests?

## Doc

/

## Changes

Added a new attribute to `OrgUnit` + constraint + tests

## How to test

- Go to the admin page
- Add values to the `code` field
- Test using the same `code` value in multiple OrgUnits that are in the same SourceVersion, it should trigger an error

## Print screen / video

`code` in admin panel:
image

Constraint error in admin panel:
image

## Notes

See Jira epic for further developments on this

```
fix: empty instance pop up

Refs: IA-3665
```


iaso/models/org_unit.py
```diff
@@ -319,6 +319,7 @@ class OrgUnit(TreeModel):
     default_image = models.ForeignKey(
         "iaso.InstanceFile", on_delete=models.SET_NULL, null=True, blank=True, related_name="default_for_org_units"
     )
+    code = models.TextField(blank=True, db_index=True)  # DHIS2 code
```
quang-le, line 322: What's the reason for the index?

iaso/tests/models/test_org_unit.py
```diff
@@ -209,6 +209,30 @@ def create_simple_hierarchy(self):
 
         return corrusca, corruscant, first_council, second_council, task_force
 
+    def test_same_code_different_versions(self):
```
quang-le: Why is not in the same test case as `test_same_code_same_version`?

quang-le (commented):
I`m wondering if we should update the API to avoid having a 500 if, e.g a user tries to update an org unit with a code already in use on the current version.

Also (maybe this is handled in another ticket), we might have UX issues for example when changes the validation status of an org unit witrh code to `REJECTED` then wants to assign the same code to a `VALID` org unit. Maybe, the constraint should take the validation status into account? (I honestly don't know)

iaso/models/org_unit.py
tdethier: We will need to do some searches based on this field (in another ticket - IA-4305)

iaso/tests/models/test_org_unit.py
tdethier: I put `test_same_code_same_version` in `OrgUnitModelDbTestCase` because of the class description (testing constraints) and because I wanted to make sure that I raise an exception.

This test was more about making sure that the "standard" behavior works fine, but you're right it's also related to the constraint in a way. I can move it if you want.

tdethier:
> I`m wondering if we should update the API to avoid having a 500 if, e.g a user tries to update an org unit with a code already in use on the current version.

Right now, there's no way for users to update codes, so it's ok. It will be done in another ticket (IA-4304), but it's true - we should make sure that this scenario is covered.

>Also (maybe this is handled in another ticket), we might have UX issues for example when changes the validation status of an org unit witrh code to REJECTED then wants to assign the same code to a VALID org unit. Maybe, the constraint should take the validation status into account? (I honestly don't know)

That's a good point, we had not thought of this scenario before. I will discuss it with Claire.

iaso/tests/models/test_org_unit.py
quang-le: No I was just curious

iaso/tests/models/test_org_unit.py
quang-le, line 1: I would a test that raises the `IntegrityError` where:
- We have a valid and a new org unit on the same version with the same code
- We try to switch the new org unit to valid

I think that's a use case we'll need to handle in the API and the UI as well later on

quang-le (changes requested):
one more little test please 😄

quang-le (approved):

PR #2245, opened 2025-06-25
IA-4328 Expose `InstanceFiles` to mobile small enhancements

kemar:
Small enhancements suggestions for #2239

### DIFF
--- a/iaso/api/instances/instances.py
+++ b/iaso/api/instances/instances.py
@@ -205,7 +205,7 @@ class InstancesViewSet(viewsets.ViewSet):
         queryset = InstanceFile.objects_with_file_extensions.filter(instance__in=instances)
 
         image_only_serializer = ImageOnlySerializer(data=request.query_params)
-        image_only_serializer.is_valid(raise_exception=False)
+        image_only_serializer.is_valid(raise_exception=True)
         image_only = image_only_serializer.validated_data["image_only"]
 
         queryset = queryset.filter_image_only(image_only=image_only)
--- a/iaso/api/instances/instances.py
+++ b/iaso/api/instances/instances.py
@@ -204,7 +204,10 @@ class InstancesViewSet(viewsets.ViewSet):
         instances = instances.for_filters(**filters)
         queryset = InstanceFile.objects_with_file_extensions.filter(instance__in=instances)
 
-        image_only = ImageOnlySerializer(data=request.query_params).get_image_only(raise_exception=False)
+        image_only_serializer = ImageOnlySerializer(data=request.query_params)
+        image_only_serializer.is_valid(raise_exception=False)
+        image_only = image_only_serializer.validated_data["image_only"]
+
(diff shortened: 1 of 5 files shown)


iaso/api/instances/instances.py
```diff
@@ -204,7 +204,10 @@ def attachments(self, request):
         instances = instances.for_filters(**filters)
         queryset = InstanceFile.objects_with_file_extensions.filter(instance__in=instances)
 
-        image_only = ImageOnlySerializer(data=request.query_params).get_image_only(raise_exception=False)
+        image_only_serializer = ImageOnlySerializer(data=request.query_params)
+        image_only_serializer.is_valid(raise_exception=True)
+        image_only = image_only_serializer.validated_data["image_only"]
```
bmonjoie, line 209: I fail to see how this is better than what was there before.
The code is a lot more verbose and you hardcoded `"image_only"`.

If you wanted to raise an exception, you could have done:
```
image_only = ImageOnlySerializer(data=request.query_params).get_image_only(raise_exception=True)
```

iaso/api/instances/views_mobile.py
```diff
@@ -44,10 +43,13 @@ def attachments(self, request, uuid):
         except Instance.DoesNotExist:
             instance = get_object_or_404(self.get_queryset(), pk=uuid)
 
-        queryset = InstanceFile.objects_with_file_extensions.filter(instance=instance)
+        image_only_serializer = ImageOnlySerializer(data=request.query_params)
+        image_only_serializer.is_valid(raise_exception=True)
+        image_only = image_only_serializer.validated_data["image_only"]
```
bmonjoie, line 48: Same remark as before

iaso/api/instances/views_mobile.py
```diff
@@ -44,10 +43,13 @@ def attachments(self, request, uuid):
...
 
-        image_only = ImageOnlySerializer(data=request.query_params).get_image_only(raise_exception=False)
-        queryset = queryset.filter_image_only(image_only=image_only)
+        queryset = instance.instancefile_set(manager="objects_with_file_extensions").filter_image_only(
+            image_only=image_only
+        )
```
bmonjoie, line 52: Nice, I didn't know we could do that!

bmonjoie (commented):
Thanks for making the test more generic and adding the handy shortcut for the manager. As for removing `get_image_only`, I fail to see the added value. IMHO, it makes the code worse

iaso/api/instances/instances.py
kemar: This is not really better but this is something more idiomatic and common in the codebase

bmonjoie (approved):
As discussed, I don't think changing `ImageOnlySerializer` makes the code cleaner. Nevertheless, I am not the one maintaining this code base and, despite my remarks, if you think this is better/more idiomatic/more easily maintainable/whatever, then feel free to merge this :slightly_smiling_face:

PR #2246, opened 2025-06-25
[WC2-778] Try to fix CIAM login for Ethiopia

bmonjoie:
Related JIRA tickets : WC2-778

## Changes

When creating a new user for SocialAccount, add a prefix to avoid unique constraint clashes.

## Notes

This PR was initially an attempt to quick fix the bug by @bmonjoie.

The scope has been extended after @madewulf asked to create a `TenantUser` instead of just hacking the username.

In doing so, I found a bug. The part [where we create a `TenantUser`](https://github.com/BLSQ/iaso/blob/48da9a9f1093a6c4dc74d5c8e6798bf611c3b4c4/iaso/api/profiles/profiles.py#L355-L386) was never reached because of [this condition](https://github.com/BLSQ/iaso/blob/48da9a9f1093a6c4dc74d5c8e6798bf611c3b4c4/iaso/api/profiles/profiles.py#L349-L351).

I fixed that and factorized the code that create a `User` or a `TenantUser`. This should be used every time we put a user in the system.

That PR should fix the CIAM login for Ethiopia in an elegant way.

The code is fully tested and the logic was approved by @bramj.

I also changed the behavior of the admin because we struggled with @beygorghor to understand why we weren't able to see all users.

Now we default on listing all the users and there are new filters:

admin

### DIFF
--- a/iaso/admin/user_admin.py
+++ b/iaso/admin/user_admin.py
@@ -1,9 +1,10 @@
-from django.contrib import messages
+from django.contrib import admin, messages
 from django.contrib.auth.admin import UserAdmin as AuthUserAdmin
 from django.contrib.auth.models import User
 from django.contrib.gis import forms
 from django.db import transaction
-from django.db.models import Exists, OuterRef
+from django.db.models import F
+from django.db.models.lookups import Exact
 from django.shortcuts import get_object_or_404, redirect, render
 from django.urls import path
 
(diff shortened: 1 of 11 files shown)


kemar (dismissed):
LGTM

bramj (dismissed):

iaso/models/tenant_users.py
```diff
@@ -1,5 +1,64 @@
...
+        new_username = f"{username}_{slugify_account(account.name)}"
+        user = User.objects.create(
+            username=new_username,
+            email=email,
+            first_name=first_name,
+            last_name=last_name,
+        )
+        self.create(main_user=main_user, account_user=user)
+
+        return user
```
kemar: @bramj WDYT of this logic?

My goal would be to use this code everywhere and replace the logic here (https://github.com/BLSQ/iaso/blob/48da9a9f1093a6c4dc74d5c8e6798bf611c3b4c4/iaso/api/profiles/profiles.py#L355-L392) by this function.

bramj: It makes my brain hurt but it looks correct to me!  :+1:

kemar: It made mine hurt like hell too but I can't think of something simpler… Good luck to the reviewers 😃

iaso/admin/user_admin.py
```diff
@@ -28,9 +29,67 @@ def __init__(self, *args, **kwargs):
...
+    is_account_user.boolean = True
+    is_account_user.short_description = "multi-account user"
+    is_account_user.admin_order_field = "annotated_is_account_user"
+
+    def is_account_main_user(self, obj):
+        return obj.annotated_is_account_main_user
+
+    is_account_main_user.boolean = True
+    is_account_main_user.short_description = "multi-account main user"
+    is_account_main_user.admin_order_field = "annotated_is_account_main_user"
```
tdethier: Both attributes are not correctly interpreted in the list view when there are users that are not part of a multi account scheme

I guess it's because `tenant_user` is `None` in these cases, which creates some kind of issue with `Exact` ?

image

iaso/models/tenant_users.py
```diff
@@ -1,5 +1,82 @@
...
+    username: str
+
+
+class UsernameAlreadyExistsError(ValidationError):
+    pass
+
+
+def get_unique_username(username: str, account_name: str) -> str:
+    account_name_slug = account_name.lower().replace(" ", "_")
+    return f"{username}_{account_name_slug}"
```
tdethier: In most cases I guess it will not be an issue, but there is a 150-char limit on usernames :warning:

iaso/models/tenant_users.py
```diff
@@ -1,5 +1,82 @@
...
+            main_user = existing_tenant_user.main_user
+
+        # 3) The user doesn't have multiple accounts.
+        elif hasattr(existing_user, "iaso_profile"):
+            existing_account = existing_user.iaso_profile.account
+            new_username = get_unique_username(data.username, existing_account.name)
+            existing_user.username = new_username
+            existing_user.save()
+            main_user = User.objects.create(
+                username=data.username,
```
tdethier: Instinctively, I would have kept the initial username for the main user, and generated a new one for the new user. But I guess it's not an issue

iaso/tests/models/test_tenant_users.py
```diff
@@ -0,0 +1,144 @@
...
+        other_account_user = self.create_user_with_profile(
+            username=self.user_creation_data.username, email="john@doe.com", account=other_account
+        )
+
+        self.assertEqual(m.User.objects.count(), 1)
+        self.assertEqual(m.TenantUser.objects.count(), 0)
+
+        m.TenantUser.objects.create_user_or_tenant_user(data=self.user_creation_data)
+
+        # Users.
```
tdethier, line 147: I would assert here that the existing user had its `username` updated (even though I would not do it, as explained in my previous comment 😁)

plugins/wfp_auth/tests/test_auth.py
```diff
@@ -0,0 +1,134 @@
...
+from plugins.wfp_auth.views import ExtraData
+
+
+class WFPAuthTestCase(APITestCase):
+    @classmethod
+    def setUpTestData(cls):
+        cls.data_source = m.DataSource.objects.create(name="Data source")
+        cls.version = m.SourceVersion.objects.create(number=1, data_source=cls.data_source)
+        cls.account = m.Account.objects.create(name="Account", default_version=cls.version)
+        cls.project = m.Project.objects.create(name="Project", account=cls.account, app_id="test_app_id")
```
tdethier: Nitpick : you could use `cls.create_account_datasource_version_project` (but this might break some tests if other values are hardcoded)
 

```suggestion
        cls.account, cls.data_source, cls.version, cls.project = cls.create_account_datasource_version_project(
            source_name="Data source", account_name="Account", project_name="Project")
```

tdethier (approved):
Good job, this was a tricky one :+1: 

- regarding the WFP auth feature: I didn't test it because I don't know how to setup all of this locally. 
- regarding the tenant feature: works fine

iaso/models/tenant_users.py
kemar: I fixed this in https://github.com/BLSQ/iaso/pull/2246/commits/c32804b4160fdfa7f21bbbfb3f5f806750fef27d but I'm not 100% sure that this is the right call…

We may want that exact scenario to fail explicitly instead.

PR #2250, opened 2025-06-26
IA-4336 Fix: Bulk upload fails when updating an instance without entity

bramj:
- Add fix for edge case of updating an instance when there's no entity attached
- Add 2 tests for missing scenarios of creating/updating instances without entities

Related JIRA tickets : https://bluesquare.atlassian.net/browse/IA-4336


tdethier (approved):
LGTM :+1: 

Thanks for the tests and comments!

PR #2256, opened 2025-07-01
[IA-4341] Make feature flags configurable

bmonjoie:
Some feature flags need a configuration on the mobile side. This will allow us to add a configuration to any feature flag and have it exposed in the frontend.

Related JIRA tickets : IA-4341

- [X] Are there enough tests?

## Doc

The code is the truth, and the tests.

## Changes

Add a `configuration` field to the through table of `Project` and `FeatureFlag`.
Change `ProjectSerializer` to expose the newly added configuration. 
Take into account the configuration when saving a project. 
Validate the configuration and the value types.

## How to test

You need to add a `configuration_schema` to a FeatureFlag in the admin.
Example:
```json
{
  "test": {
    "type": "string",
    "default": "Default",
    "description": "This is a test"
  }
}
```

Then, go to edit a project's feature flags. You should see the configuration form under it.

## Print screen / video

image

## Notes

Due to a limitation in `Bluesquare-Component`, I have had to use a hack to expand the table. We need to fix that before we merge.

```
fix: empty instance pop up

Refs: IA-3665
```


iaso/admin/base.py
```diff
@@ -413,10 +414,19 @@ def feature_flags_list(self, obj):
         return ", ".join(flag.name for flag in flags) if len(flags) > 0 else "-"
 
 
+@admin.register(ProjectFeatureFlags)
+@admin_attr_decorator
+class ProjectAdmin(admin.ModelAdmin):
```
tdethier:
```suggestion
class ProjectFeatureFlagsAdmin(admin.ModelAdmin):
```

iaso/models/feature_flags.py
```diff
@@ -0,0 +1,67 @@
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+
+from .project import Project
+
+
+class FeatureFlag(models.Model):
```
tdethier, line 28: Did you remove `HOME_OFFLINE` on purpose?

iaso/api/apps/serializers.py
```diff
@@ -123,5 +130,63 @@ def set_forms_and_feature_flags(instance, forms, feature_flags):
...
-                f_f_object = FeatureFlag.objects.get(code=f_f["code"])
-                instance.feature_flags.add(f_f_object)
+                f_f_object = FeatureFlag.objects.get(code=f_f["featureflag"]["code"])
+                instance.feature_flags.through.save(
+                    ProjectFeatureFlags(
+                        project=instance,
+                        featureflag=f_f_object,
+                        configuration=f_f.get("configuration", None),
+                    )
+                )
```
tdethier: You're going to like this one, less noise 

```suggestion
                instance.feature_flags.add(f_f_object, through_defaults={"configuration": f_f.get("configuration", None)})
```

iaso/api/projects/serializers.py
```diff
@@ -68,3 +77,13 @@ def get_qr_code(self, instance):
             qr_code_options=QRCodeOptions(size="S", image_format="png", error_correction="L"),
         )
         return f"data:image/png;base64,{base64.b64encode(qr_image).decode('utf-8')}"
+
+    def validate_app_id(self, data):
```
tdethier, line 88: Thanks for moving this :+1:

iaso/api/apps/serializers.py
```diff
@@ -123,5 +130,63 @@ def set_forms_and_feature_flags(instance, forms, feature_flags):
...
+        CONFIG_TYPE_LONG: int,
+        CONFIG_TYPE_NUMBER: int,
+        CONFIG_TYPE_FLOAT: float,
+        CONFIG_TYPE_DOUBLE: float,
+        CONFIG_TYPE_DECIMAL: float,
+        CONFIG_TYPE_URL: str,
+        CONFIG_TYPE_TEXT: str,
+        CONFIG_TYPE_STR: str,
+        CONFIG_TYPE_STRING: str,
+    }
```
tdethier: Do we need all these types if we end up with only 3?

tdethier: Nitpick: I would move that above with the other constant values rather than in between methods

iaso/api/apps/serializers.py
```diff
@@ -110,11 +117,11 @@ def update(self, instance, validated_data):
     @staticmethod
     def needs_authentication_based_on_feature_flags(feature_flags):
         if feature_flags:
-            return REQUIRE_AUTHENTICATION_FEATURE_FLAG in list(f_f["code"] for f_f in feature_flags)
+            return FeatureFlag.REQUIRE_AUTHENTICATION in list(f_f["featureflag"]["code"] for f_f in feature_flags)
         return False
 
     @staticmethod
-    def set_forms_and_feature_flags(instance, forms, feature_flags):
+    def set_forms_and_feature_flags(instance: Project, forms, feature_flags):
```
tdethier, line 124: This reminds me of why I really hate the `Instance` model and want to rename it to `Submission`

tdethier (commented):
I only checked the backend.

I left a few questions and suggestions. Overall nice job, python is fun :innocent:

iaso/models/feature_flags.py
bmonjoie: I had not, but it seems unused so I won't add it back.

iaso/api/apps/serializers.py
bmonjoie: Good question, I suppose I wanted to let people be able to be more explicit about the type. Though, we do have a distinction between `str` and `url` (there is an extra check). So we actually end-up with 4 different types. I can restrict it if the team thinks it's more confusing.

bmonjoie: I had in the beginning, but it seemed to me that having all those variables at the top, even though they only concern one method, was too much noise for the rest of the class.

iaso/api/apps/serializers.py
```diff
@@ -1,20 +1,25 @@
...
 
-REQUIRE_AUTHENTICATION_FEATURE_FLAG = "REQUIRE_AUTHENTICATION"
-
 
 class AppSerializer(ProjectSerializer):
+    APP_ID = "app_id"
+    COLOR = "color"
+    FEATURE_FLAGS = "projectfeatureflags_set"
+    FORMS = "forms"
+    NAME = "name"
```
kemar, line 21: Sometimes constants are great and beautifull, e.g. when `403` means nothing compared to `ERROR_FORBIDDEN`.

But in this case, they are IMO not much useful and somewhat redundant. I mean, what is the value of e.g. `FORMS = "forms"`?

I can only see disadvantages:

- the mental cost of navigating through files to discover that the meaning is the same
- the cost of storing them
- sometimes the cost of importing them

At least, put them under the docstring ;)

iaso/api/apps/serializers.py
```diff
@@ -123,5 +130,59 @@ def set_forms_and_feature_flags(instance, forms, feature_flags):
...
+        CONFIG_TYPE_LONG: int,
+        CONFIG_TYPE_NUMBER: int,
+        CONFIG_TYPE_FLOAT: float,
+        CONFIG_TYPE_DOUBLE: float,
+        CONFIG_TYPE_DECIMAL: float,
+        CONFIG_TYPE_URL: str,
+        CONFIG_TYPE_TEXT: str,
+        CONFIG_TYPE_STR: str,
+        CONFIG_TYPE_STRING: str,
+    }
```
kemar: Class variables are, by convention, typically placed right below the class header.

If this raises a concern, then this may be a sign that it would be better defined as a local variable of the `validate_configuration()` method.

I know you love constants, but once again the value here is very low (IMO).

iaso/migrations/0331_featureflag_configuration_projectfeatureflags_and_more.py
```diff
@@ -0,0 +1,56 @@
...
+            field=models.ManyToManyField(
+                blank=True,
+                related_name="_iaso_project_feature_flags_+",
+                through="iaso.ProjectFeatureFlags",
+                to="iaso.FeatureFlag",
+            ),
+        ),
+    ]
+    operations = [
+        migrations.SeparateDatabaseAndState(state_operations=state_operations),
```
kemar: Believe it or not but this is the first time I've seen `SeparateDatabaseAndState`.

I understand it's used to keep the existing intermediate table while adding a `through`.

I was a bit nervous about a data loss possibility but it seems to do the job well.

iaso/api/projects/serializers.py
```diff
@@ -68,3 +77,13 @@ def get_qr_code(self, instance):
...
+        data = data.strip()
+        request = self.context["request"]
+        if not data:
+            raise serializers.ValidationError("An App id is required")
+        if data != "" and data != request.data.get("id"):
+            if Project.objects.filter(app_id=data).count() > 0:
+                raise serializers.ValidationError("App id already used")
+        return data
```
kemar, line 96: I know this code has just been moved.

Yet it's a bit hard to understand…

Don't we need to do this check only at creation time? If so the check could be moved and simplified in the `create()` method.

Otherwise I think you can make `app_id` a requirement by adding this line to the `Meta` block:

```python
extra_kwargs = {"app_id": {"required": True, "allow_null": False, "allow_blank": False}}
```

kemar (commented):
I fixed the migration file which caused tests to fail. Django was complaining about a missing migration.

I added some comments but nothing is blocking.

This PR adds complexity but the tests are helpful to understand.

I don't approve yet because the PR is still marked as a WIP.

iaso/api/apps/serializers.py
bmonjoie: I did that because of:
```python
    FEATURE_FLAGS = "projectfeatureflags_set"
```
Having the string like this one hardcoded made it harder to read. Also, I almost did not change it in `update`. Having constants made it easier to read, consistent, and foolproof if we have to update another one in the future not to forget to change it in `update` as well.

> At least, put them under the docstring ;)

I don't know what that means :sweat_smile:

iaso/migrations/0331_featureflag_configuration_projectfeatureflags_and_more.py
bmonjoie: Believe it or not, I have no idea what I'm doing :grin:

iaso/api/projects/serializers.py
bmonjoie: I believe this is out of scope for this PR. I don't want to change logic that is working. This PR is complex enough as it is to add side changes to it.

hat/assets/js/apps/Iaso/domains/projects/components/FeatureFlagsSwitches.tsx
```diff
@@ -24,52 +27,164 @@ const styles = theme => ({
...
+            return 'text';
+    }
+};
+
+export const useFeatureFlagConfiguration = (
+    ff: FeatureFlag,
+    onFeatureFlagChanged: (changed: ProjectFeatureFlag) => void,
+    pff?: ProjectFeatureFlag,
+) => {
+    if (ff.configuration_schema == null) {
```
beygorghor, line 74: In this case, ff.configuration_schema == null would work the same as ff.configuration_schema === null because null is a primitive value, but using === is the preferred practice. This is why eslint is not complaining here 
```suggestion
    if (ff.configuration_schema === null) {
```

hat/assets/js/apps/Iaso/domains/projects/components/FeatureFlagsSwitches.tsx
```diff
@@ -24,52 +27,164 @@ const styles = theme => ({
...
+        return null;
+    }
+    const onConfigurationChanged = (key: string, value: any) => {
+        let f: ProjectFeatureFlag;
+        if (pff != null) {
```
beygorghor, line 79:
```suggestion
        if (pff !== null) {
```

hat/assets/js/apps/Iaso/domains/projects/components/FeatureFlagsSwitches.tsx
```diff
@@ -24,52 +27,164 @@ const styles = theme => ({
...
+
+export const useFeatureFlagConfiguration = (
+    ff: FeatureFlag,
+    onFeatureFlagChanged: (changed: ProjectFeatureFlag) => void,
+    pff?: ProjectFeatureFlag,
+) => {
+    if (ff.configuration_schema == null) {
+        return null;
+    }
+    const onConfigurationChanged = (key: string, value: any) => {
```
beygorghor, line 77: UseCallback ?

hat/assets/js/apps/Iaso/domains/projects/components/FeatureFlagsSwitches.tsx
```diff
@@ -24,52 +27,164 @@ const styles = theme => ({
...
-    const columns = useFeatureFlagColumns(
-        setFeatureFlag,
-        projectFeatureFlagsValues,
-    );
+    const columns = useFeatureFlagColumns(setFeatureFlag, projectFeatureFlags);
+    const expanded = useMemo(() => {
+        const expandedFf = {};
+        featureFlags.forEach(featureFlag => {
+            expandedFf[featureFlag.code] =
+                featureFlag.configuration_schema != null &&
```
beygorghor, line 191:
```suggestion
                featureFlag.configuration_schema !== null &&
```

hat/assets/js/apps/Iaso/domains/projects/components/FeatureFlagsSwitches.tsx
```diff
@@ -24,52 +27,164 @@ const styles = theme => ({
...
+        case 'url':
+        case 'str':
+        case 'string':
+            return 'text';
+        default:
+            return 'text';
+    }
+};
+
+export const useFeatureFlagConfiguration = (
```
beygorghor: We should rename this, it looks like a hook but it's a component in fact

beygorghor (changes requested):
some small changes

hat/assets/js/apps/Iaso/domains/projects/components/FeatureFlagsSwitches.tsx
bmonjoie: If I use `useCallback` here, when something changes, React crashes, saying there is not the same number of hooks. I have no idea why.

hat/assets/js/apps/Iaso/domains/projects/components/FeatureFlagsSwitches.tsx
bmonjoie: When I do this, I get a warning from TypeScript that  `f` could be `underfined`.  It seems that doing `!= null` is better because it guards against both `null` and `undefined`, which is what I want.
Therefore, should I put: `pff !== null && pff !== undefined` everywhere or leave the `!= null` everywhere?

hat/assets/js/apps/Iaso/domains/projects/components/FeatureFlagsSwitches.tsx
quang-le: Because you return `null` early. Using a hook (like `useCallback`) after an early return will crash because it breaks the rule of hooks. Basically,  hook calls are evaluated as a list, and the list's length should be stable across renders. If you call a hook after an early return, the length will vary across renders, hence the crash.

The solution here would be to decalre  `onConfigurationChanged` before your early return.
You lose some benefit of the early return, but it's the most idiomatic way to do this I think

bmonjoie: No, it's because of that, because when I had used `useCallback`, I put it above the early return (lint already warned me about this).

beygorghor (approved):
You have failing test now

iaso/api/apps/serializers.py
```diff
@@ -123,5 +130,50 @@ def set_forms_and_feature_flags(instance, forms, feature_flags):
...
+            "int": int,
+            "long": int,
+            "number": int,
+            "float": float,
+            "double": float,
+            "decimal": float,
+            "url": str,
+            "text": str,
+            "str": str,
+            "string": str,
```
tdethier, line 158: I still don't see the point of all of this :sweat_smile:

tdethier (approved):
Backend looks good enough! 

I'm still not convinced by all the possible configuration types since there aren't as many types in python, so why bother :man_shrugging: 

This PR indeed adds complexity to an already complex concept, but if that's required for the mobile app, let's go

PR #2257, opened 2025-07-01
Polio 1965 add antivirus mixin

quang-le:
Add virus scan in for all polio models and API that require it

Related JIRA tickets : POLIO-1965

- [x] Are there enough tests?

## Doc

docstring in the code

## Changes

- Move`clamav.py`'s code to `virus_scan.py` and add Model and Serializer abstarct classes
- Rename `document` field to `file`. same with result and timestamp fields
- Adapt API code to use abstract classes and renamed fields
- Add migration
- Updater and add test

## How to test

see [front-end PR](https://github.com/BLSQ/iaso/pull/2258), or test directly API endpoints

## Print screen / video

N/A see [front-end PR](https://github.com/BLSQ/iaso/pull/2258)


tdethier (dismissed):
I encountered some issues while manually testing:
- When I'm trying to update an existing pre alert document, or create a new pre alert with a document, I get a 400 bad request. I don't see any stack trace though

image

- I got another error when trying to update an existing Form A

image


I didn't go through the code yet, I'm going to do this now

iaso/utils/virus_scan.py
```diff
@@ -0,0 +1,398 @@
...
+            The updated model instance.
+
+        Example:
+            # The file will be automatically scanned if it's different
+            doc = DocumentSerializer().update(existing_doc, {'file': new_file})
+            # doc.file_scan_status will be updated if file changed
+        """
+        has_updated = scan_file_and_update(instance, validated_data)
+        if has_updated:
+            instance.save()
```
tdethier: Calling `super().update()` will save for you anyway

```suggestion
```

iaso/utils/virus_scan.py
```diff
@@ -0,0 +1,398 @@
...
+            The created model instance.
+
+        Example:
+            # The file will be automatically scanned during creation
+            doc = DocumentSerializer().create({'title': 'Test', 'file': uploaded_file})
+            # doc.file_scan_status will contain the scan result
+        """
+        model_class = self.Meta.model
+        scan_file(validated_data)
+        return model_class.objects.create(**validated_data)
```
tdethier: I would call `super().create()` here, because it will take take of m2m relations (which will not work this way)

plugins/polio/api/vaccines/stock_management.py
```diff
@@ -319,13 +322,18 @@ def create(self, validated_data):
...
     def update(self, instance, validated_data):
         campaign = self.extract_campaign_data(validated_data)
         if campaign:
             instance.campaign = campaign
-        return super().update(instance, validated_data)
+        instance.save()
+        super().update(instance, validated_data)
+        scan_file_and_update(instance, validated_data)
+        instance.save()
+        return instance
```
tdethier:
```suggestion
        scan_file_and_update(instance, validated_data)
        return super().update(instance, validated_data)
```

iaso/utils/virus_scan.py
```diff
@@ -0,0 +1,398 @@
...
+                has_updated = True
+                result, timestamp = scan_uploaded_file_for_virus(new_file)
+                obj.file = new_file
+                obj.file_scan_status = result
+                obj.file_last_scan = timestamp
+                print("-" * 50)
+                print(result, timestamp)
+        elif hasattr(obj, key) and getattr(obj, key) != validated_data[key]:
+            has_updated = True
+            setattr(obj, key, validated_data[key])
```
tdethier: I would not reimplement `update()`, I would simply scan the file, add the new fields and call `super().update()` later (= have a function that only scans, not scans and updates)

iaso/utils/virus_scan.py
```diff
@@ -0,0 +1,398 @@
+import logging
```
tdethier: I think I'd rather split the model-level classes/functions and the API-level classes/functions in two different files :sweat_smile: 

Having generic file scanning concepts split from clamav-specific concepts would also be interesting in case we want/need to use another antivirus

iaso/utils/virus_scan.py
```diff
@@ -0,0 +1,398 @@
...
+
+            new_file = validated_data[key]
+            old_file = obj.file
+
+            if not old_file or (
+                os.path.basename(old_file.name) != os.path.basename(new_file.name) or old_file.size != new_file.size
+            ):
+                print("+" * 50)
+                print("UPDATING")
+                print("+" * 50)
```
tdethier:
```suggestion
```

iaso/utils/virus_scan.py
```diff
@@ -0,0 +1,398 @@
...
+                has_updated = True
+                result, timestamp = scan_uploaded_file_for_virus(new_file)
+                obj.file = new_file
+                obj.file_scan_status = result
+                obj.file_last_scan = timestamp
+                print("-" * 50)
+                print(result, timestamp)
```
tdethier:
```suggestion
```

iaso/utils/virus_scan.py
```diff
@@ -0,0 +1,398 @@
...
+        # This will only scan the file if it's different from the existing one
+        # and will update the instance's file_scan_status and file_last_scan fields
+    """
+    has_updated = False
+    for key in validated_data.keys():
+        if key == "file":
+            if not validated_data[key]:
+                print("+" * 50)
+                print("NO FILE FOUND")
+                print("+" * 50)
```
tdethier: It could be interesting to log something instead

tdethier (commented):
A few things:
- Nice job implementing these mixins in both the model and the api level :+1: 
- The feature seems to be working now, but I still need to make one last thorough test to make sure everything is ok
- I would split the `virus_scan.py` file in multiple ones
- I would not manually perform `create` and `update` in serializers, but call their `super()` instead
- I would not couple file scanning with updating in serializers
- I did not look at tests yet, I'll do it after taking care of any necessary change

tdethier:
After discussing this PR with @kemar, we agree that we need to make some changes. Since @quang-le is on holiday, I'm going to continue this PR

iaso/utils/virus_scan/model.py
```diff
@@ -0,0 +1,59 @@
+from django.db import models
+from django.utils.translation import gettext as _
+
+
+class VirusScanStatus(models.TextChoices):
```
kemar, line 5: I'd have put the `VirusScanStatus` enumeration type *inside* the `ModelWithFile` model because [that's what we usually do in Django](https://docs.djangoproject.com/en/5.2/ref/models/fields/#enumeration-types).

**But** we can live with it outside as it's now because we'd have to change a lot of code otherwise.

iaso/utils/virus_scan/serializers.py
```diff
@@ -0,0 +1,123 @@
...
+        """
+        needs_scanning = self._check_if_file_exists_and_needs_scanning(validated_data, obj)
+        if needs_scanning:
+            result, timestamp = scan_uploaded_file_for_virus(validated_data["file"])
+            validated_data["file_scan_status"] = result
+            validated_data["file_last_scan"] = timestamp
+        return needs_scanning
+
+    def _check_if_file_exists_and_needs_scanning(self, validated_data, obj):
+        """ "
```
kemar:
```suggestion
        """
```

kemar (approved):
LGTM Good job 👏 

The code is now easier to read and understand.

It's true that there is a bit of repetition (e.g., `self.scan_file_if_exists`) but I firmly believe that this will save us a lot of headaches related to serializers inheritance.

---

Also I can't help but say that there is a bit too much useless documentation which smells LLM-written doc. Proper typing would've been the alternative.

PR #2259, opened 2025-07-01
[IA-4337] Filter out mobile entities with deleted reference instances

tdethier:
Entities whose reference instance is soft deleted are no longer returned to the mobile API endpoints.

Related JIRA tickets : IA-4337

- [x] Are there enough tests?

## Doc

/

## Changes

- Filtered out queryset used in mobile endpoints
- Refactored some tests + added some new ones

## How to test

This might be a bit tricky to manually test, but the tests explain what to do:
- Define a project, a reference form, an entity type that uses the reference form
- Fill out the form and create entities with these submissions
- Soft delete the instances
- The mobile endpoints should no longer return the entities because their reference instances were deleted
    - `/api/mobile/entitytypes/{entity_type.pk}/entities/?app_id={app_id}`
    - `/api/mobile/entities/`

## Print screen / video

/

## Notes
/

```
fix: empty instance pop up

Refs: IA-3665
```

### DIFF
--- a/iaso/api/mobile/entity.py
+++ b/iaso/api/mobile/entity.py
@@ -17,7 +17,6 @@ def filter_for_mobile_entity(queryset, request):
             queryset = queryset.filter_for_mobile_entity(
                 request.query_params.get("limit_date"), request.query_params.get("json_content")
             )
-            queryset = queryset.filter(attributes__deleted=False)
         except InvalidLimitDateError as e:
             raise ParseError(e.message)
         except InvalidJsonContentError as e:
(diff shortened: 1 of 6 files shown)


iaso/api/mobile/entity.py
```diff
@@ -17,6 +17,7 @@ def filter_for_mobile_entity(queryset, request):
             queryset = queryset.filter_for_mobile_entity(
                 request.query_params.get("limit_date"), request.query_params.get("json_content")
             )
+            queryset = queryset.filter(attributes__deleted=False)
```
kemar: I would probably have put that filter directly in `EntityQuerySet.filter_for_mobile_entity()` instead of adding an inline filter here.

WDYT?

kemar (approved):
LGTM

Just a comment with a suggestion but nothing blocking.

iaso/api/mobile/entity.py
tdethier: It looks like it's only called once (here), but sure, I can do that

PR #2260, opened 2025-07-03
WFP changes to settings.py

beygorghor:
Davis from WFP made some changes on settings.py to be sure all errors are send to sentry

- [ ] Are there enough tests?

## Doc

## Changes

 formating settings.py and some more changes

## How to test

Make sure everything is still working properly

### DIFF
--- a/hat/settings.py
+++ b/hat/settings.py
@@ -23,8 +23,10 @@ import sentry_sdk
 from django.core.exceptions import ImproperlyConfigured
 from django.utils.translation import gettext_lazy as _
 from requests.exceptions import HTTPError
+from sentry_sdk.integrations.celery import CeleryIntegration
 from sentry_sdk.integrations.django import DjangoIntegration
 from sentry_sdk.integrations.logging import ignore_logger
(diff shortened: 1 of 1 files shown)


hat/settings.py
```diff
@@ -551,12 +561,23 @@ def sentry_error_sampler(_, hint):
...
 
+    # Build integrations list conditionally
+    integrations = [DjangoIntegration()]
+
+    if USE_CELERY:
+        integrations.append(CeleryIntegration(monitor_beat_tasks=True))
+
+    # Add Redis integration if using Redis for cache or Celery
+    if CACHE_BACKEND == "django_redis.cache.RedisCache" or USE_CELERY:
+        integrations.append(RedisIntegration())
```
tdethier, line 572: This is the only change that could have some effect on our instances (only on sentry), provided that they use celery and/or redis.
I'm not sure if we have an instance that has sentry configured that uses any of these.

tdethier (approved):
It's mostly some formatting, except for the sentry/redis/celery thing (and some sentry sample rate).

I don't think it will cause an issue

PR #2266, opened 2025-07-04
IA-4331: Dropdown of the FORM field show duplicate forms by the number of project associated to the form

beygorghor:
Every form that is associated to multiple projects appears in duplicate (by the number of the projects associated to the form.
image

Related JIRA tickets : IA-4331

- [x] Are there enough tests?

## Doc
/

## Changes

Added distinct on get_queryset, added a test

## How to test

Link form to multiple project, 
open submissions page, make sure this form is present only once in the dropdown result

## Print screen / video

Screenshot 2025-07-04 at 12 07 13
Screenshot 2025-07-04 at 12 07 03

## Notes

```
fix: empty instance pop up

Refs: IA-3665
```

### DIFF
--- a/iaso/api/forms.py
+++ b/iaso/api/forms.py
@@ -392,6 +392,9 @@ class FormsViewSet(ModelViewSet):
 
         queryset = queryset.order_by(*order)
 
+        # Ensure duplicates are removed after all joins and annotations
+        queryset = queryset.distinct()
+
         return queryset
 
     def list(self, request: Request, *args, **kwargs):
--- a/iaso/tests/api/test_forms.py
+++ b/iaso/tests/api/test_forms.py
@@ -776,3 +776,18 @@ class FormsAPITestCase(APITestCase):
 
         for form_data in response.json()["forms"]:
             self.assertIn("instances_count", form_data)
+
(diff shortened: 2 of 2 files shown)


iaso/api/forms.py
```diff
@@ -392,6 +392,9 @@ def get_queryset(self, mobile=False):
 
         queryset = queryset.order_by(*order)
 
+        # Ensure duplicates are removed after all joins and annotations
+        queryset = queryset.distinct()
```
kemar, line 396: I just wanted to point out that [using `distinct()` without a specific field is slower](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#django.db.models.query.QuerySet.distinct) in general:

> For a normal `distinct()` call, the database compares each field in each row when determining which rows are distinct. For a `distinct()` call with specified field names, the database will only compare the specified field names.

However, we are mixing aggregation (`annotate`) and `distinct` here… So calling `distinct` with specific field names [would cause a lot more headaches](https://code.djangoproject.com/ticket/32663).

Let's stick with `distinct()`.

kemar (approved):

PR #2271, opened 2025-07-07
IA-4315: Org unit details map fit to bounds

beygorghor:
Map should be centered on where there is data (even if children data only)

Related JIRA tickets : IA-4315

- [ ] Are there enough tests?

## Doc

## Changes

- fix bounds calculation to handle not valid bounds
- added two button to allow user to fit map to children or submissions bounds

## How to test

You'll need an org unit with children (with geolocations or shapes) and submissions (also with geolocation)
Open this org unit, oopen the map, select children and submissions on the right, play with the eye button

## Print screen / video

https://github.com/user-attachments/assets/324a087b-0023-4a27-850c-e09a007e28a7

## Notes
-


hat/assets/js/apps/Iaso/domains/orgUnits/components/orgUnitMap/OrgUnitMap/useGetBounds.ts
```diff
@@ -15,37 +15,71 @@ export const useGetBounds = ({
...
+    if (locations.length > 0) {
+        const locationsBounds = L.latLngBounds(locations);
+        if (locationsBounds && locationsBounds.isValid()) {
+            if (finalBounds) {
+                finalBounds = finalBounds.extend(locationsBounds);
+            } else {
+                finalBounds = locationsBounds;
+            }
+        }
+    }
```
Phil-V: This block might not be reachable, as `locations` is initialized to an empty array but never populated.

Phil-V: Small nitpick: `useGetBounds` could also be renamed without the use- prefix if it is not meant to be used as a React hook.

hat/assets/js/apps/Iaso/domains/forms/components/FormsFilterComponent.js
```diff
@@ -49,32 +55,70 @@ export const FormsFilterComponent = ({
...
+
+    const canFitToBounds = computedBounds && map && map.current;
+    const triggerFitBounds = () => {
+        if (canFitToBounds) {
+            map.current.fitBounds(computedBounds, {
+                padding: [10, 10],
+                maxZoom: 18,
+            });
+        }
+    };
```
Phil-V: This could make use of `useCallback` for consistency with the similar function in `OrgUnitTypeFilterComponent.tsx` (I think this is the one you pointed out to me in our meeting, apologies if it's not).

Phil-V (approved):

PR #2274, opened 2025-07-08
[IA-3801] Fix auto-OUCR creates invalid OUCR

bmonjoie:
When an OUCR is created automatically, it adds all existing reference instances. This is not supposed to happen.

Related JIRA tickets: IA-3801

- [ ] Are there enough tests?

## Changes

Only add the newly created instance to the automatically created OUCR

## How to test

Submit an instance of a reference form with the auto-OUCR flag activated.
This needs to be submitted to an OU with several validated reference instances.
Check that the change request contains only the newly created submission.

```
fix: empty instance pop up

Refs: IA-3665
```

### DIFF
--- a/iaso/models/org_unit.py
+++ b/iaso/models/org_unit.py
@@ -873,19 +873,20 @@ class OrgUnitChangeRequest(SoftDeletableModel):
                 self.org_unit.groups.clear()
                 self.org_unit.groups.add(*self.new_groups.all())
             elif field_name == "new_reference_instances":
-                previous_reference_instances = list(self.org_unit.reference_instances.all())
                 self.org_unit.reference_instances.clear()
-                new_reference_instances = self.new_reference_instances.all()
-                # We need to merge the old reference instances with the new ones. To do so, we need to check which
-                # forms do the new reference instances point to, to filter out the old ones that point to the same forms.
-                new_reference_forms = list(map(lambda i: i.form, new_reference_instances))
-                new_reference_forms.append(
-                    list(filter(lambda i: i.form not in new_reference_forms, previous_reference_instances))
-                )
+
+                new_reference_instances = list(self.new_reference_instances.all())
+                new_reference_forms_ids = self.new_reference_instances.values_list("form_id", flat=True)
+
+                for instance in self.old_reference_instances.all():
+                    if instance.form_id not in new_reference_forms_ids:
+                        new_reference_instances.append(instance)
+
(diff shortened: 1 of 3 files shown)


iaso/models/org_unit.py
```diff
@@ -860,19 +860,20 @@ def __apply_changes(self, user: User, approved_fields: typing.List[str]) -> None
...
-                new_reference_forms.append(
-                    list(filter(lambda i: i.form not in new_reference_forms, previous_reference_instances))
-                )
+
+                new_reference_instances = list(self.new_reference_instances.all())
+                new_reference_forms_ids = self.new_reference_instances.values_list("form_id", flat=True)
+
+                for instance in self.old_reference_instances.all():
+                    if instance.form_id not in new_reference_forms_ids:
+                        new_reference_instances.append(instance)
```
bmonjoie: This is not correct because `self.old_reference_instances` is a snapshot in time. It will override changes that were made between the moment the OUCR was created and it is approved.

kemar: A business decision has to be made.

Change requests do not handle concurrency because neither the org unit nor the change request are locked while waiting for a review.

We based all the UI on the `self.old_*` values because that's what the user see when he creates the change request.

The rationale behind this decision is that we should'nt have too many conflicting changes.

I think it makes sense to stick to `self.old_reference_instances` to stay coherent.

We could change this behavior, but then we should do it everywhere. And this requires some thought.

WDYT @beygorghor @madewulf ?

iaso/models/org_unit.py
```diff
@@ -861,11 +861,20 @@ def __apply_changes(self, user: User, approved_fields: typing.List[str]) -> None
                 self.org_unit.groups.add(*self.new_groups.all())
             elif field_name == "new_reference_instances":
                 self.org_unit.reference_instances.clear()
-                new_reference_instances = [
+
+                new_reference_instances = list(self.new_reference_instances.all())
+                new_reference_forms_ids = self.new_reference_instances.values_list("form_id", flat=True)
+
+                for instance in self.old_reference_instances.all():
```
madewulf: This is going to resurface old instances that have been changed in the mean time

iaso/models/org_unit.py
kemar: I'm marking this as resolved since we agreed on using the current state of the reference instances.

kemar (approved):

PR #2284, opened 2025-07-10
IA-4354 Update the main user's password if the user is a multi tenant user

kemar:
Update the main user's password if the user is a multi tenant user.

Related JIRA tickets : IA-4354

## Changes

- updated the `/api/profiles/` API to update the right user's password
- removed dead code
- moved some logic to keep it all at the same place
- added unit tests

## How to test

- connect the web UI with a multi-tenant user
- go to the profile UI
- change your password
- validate
- log out
- you should now be able to log in with your new password

### DIFF
--- a/iaso/api/profiles/profiles.py
+++ b/iaso/api/profiles/profiles.py
@@ -353,6 +353,9 @@ class ProfilesViewSet(viewsets.ViewSet):
             return JsonResponse({"errorKey": "password", "errorMessage": _("Mot de passe requis")}, status=400)
 
         try:
+            # Currently, the `account` is always the same in the UI.
+            # This means that we'll never get back a `tenant_main_user` here - at least for the moment.
+            # Yet we keep `create_user_or_tenant_user()` here to avoid repeating part of its logic.
             new_user, tenant_main_user, tenant_account_user = TenantUser.objects.create_user_or_tenant_user(
                 data=UserCreationData(
                     username=username,
--- a/iaso/api/profiles/profiles.py
+++ b/iaso/api/profiles/profiles.py
@@ -365,14 +365,13 @@ class ProfilesViewSet(viewsets.ViewSet):
         except UsernameAlreadyExistsError as e:
             return JsonResponse({"errorKey": "user_name", "errorMessage": e.message}, status=400)
 
+        user_who_logs_in = new_user or tenant_main_user
+        if password != "":
+            user_who_logs_in.set_password(password)
+            user_who_logs_in.save()
+
         user = new_user or tenant_account_user
 
-        if new_user and password != "":
-            user.set_password(password)
-            user.save()
-
-        # Create an Iaso profile for the new user and attach it to the same account
-        # as the currently authenticated user
         user.profile = Profile.objects.create(
             user=user,
             account=current_account,
@@ -381,17 +380,15 @@ class ProfilesViewSet(viewsets.ViewSet):
             organization=request.data.get("organization", None),
         )
 
-        profile = get_object_or_404(Profile, id=user.profile.pk)
-
         try:
             user_permissions = self.validate_user_permissions(request, current_account)
-            org_units = self.validate_org_units(request, profile)
+            org_units = self.validate_org_units(request, user.profile)
             user_roles_data = self.validate_user_roles(request)
-            projects = self.validate_projects(request, profile)
-            editable_org_unit_types = self.validate_editable_org_unit_types(request, profile)
+            projects = self.validate_projects(request, user.profile)
+            editable_org_unit_types = self.validate_editable_org_unit_types(request, user.profile)
         except ProfileError as error:
             # Delete profile if error since we're creating a new user
-            profile.delete()
+            user.profile.delete()
             return JsonResponse(
                 {"errorKey": error.field, "errorMessage": error.detail},
                 status=status.HTTP_400_BAD_REQUEST,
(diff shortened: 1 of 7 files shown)


iaso/models/tenant_users.py
```diff
@@ -131,36 +131,18 @@ class Meta:
...
+            account_name = "Unknown"
         return f"{self.main_user} -- {self.account_user} ({account_name})"
 
-    @property
-    def account(self):
-        try:
-            return self.account_user.iaso_profile.account if self.account_user.iaso_profile else None
-        except User.iaso_profile.RelatedObjectDoesNotExist:
-            return None
-
```
kemar, line 148: I'm removing this dead code.

iaso/models/tenant_users.py
```diff
@@ -131,36 +131,18 @@ class Meta:
...
-        try:
-            if self.account_user.iaso_profile:
-                account_dict = self.account_user.iaso_profile.account.as_dict()
-        except User.iaso_profile.RelatedObjectDoesNotExist:
-            pass
-        return {
-            "id": self.id,
-            "main_user_id": self.main_user_id,
-            "account": account_dict,
-        }
```
kemar, line 166: I'm removing this dead code.

iaso/utils/__init__.py
```diff
@@ -128,7 +128,3 @@ def is_mobile_request(request):
     user_agent = request.META.get("HTTP_USER_AGENT", "")
     # Check if the User-Agent string contains keywords typical of mobile devices
     return any(mobile in user_agent.lower() for mobile in ["mobile", "android", "iphone", "ipad", "windows phone"])
-
-
-def is_multi_account_user(user):
-    return hasattr(user, "tenant_user")
```
kemar, line 134: I'm moving this method to keep the logic in the same place.

iaso/api/profiles/profiles.py
```diff
@@ -365,14 +365,13 @@ def create(self, request):
         except UsernameAlreadyExistsError as e:
             return JsonResponse({"errorKey": "user_name", "errorMessage": e.message}, status=400)
 
-        user = new_user or tenant_account_user
+        user_who_logs_in = new_user or tenant_main_user
+        if password != "":
+            user_who_logs_in.set_password(password)
+            user_who_logs_in.save()
```
kemar, line 374: Although we used `create_user_or_tenant_user()`, we'll **never** get back a `tenant_main_user` here because the `account` is always the same in the UI.

Tenant users can currently only be created in the admin or via the `wfp_auth` login.

I'm now wondering if we shouldn't just remove `create_user_or_tenant_user()` here and instead just create a single user?

iaso/api/profiles/profiles.py
```diff
@@ -381,25 +380,23 @@ def create(self, request):
             organization=request.data.get("organization", None),
         )
 
-        profile = get_object_or_404(Profile, id=user.profile.pk)
```
kemar, line 384: This call was useless, let's save one query.

iaso/models/tenant_users.py
```diff
@@ -131,36 +131,18 @@ class Meta:
...
-        try:
-            return self.account_user.iaso_profile.account if self.account_user.iaso_profile else None
-        except User.iaso_profile.RelatedObjectDoesNotExist:
-            return None
-
     def get_all_account_users(self):
         return [tu.account_user for tu in self.main_user.tenant_users.all()]
 
     def get_other_accounts(self):
         return [tu.account for tu in self.main_user.tenant_users.exclude(pk=self.pk) if tu.account]
```
tdethier: It's not caused by this PR, but I got an error while trying to login as a main multi tenant user:

```  File "/home/thibault/dev/iaso/iaso/iaso/models/tenant_users.py", line 144, in get_other_accounts
    return [tu.account for tu in self.main_user.tenant_users.exclude(pk=self.pk) if tu.account]
                                                                                    ^^^^^^^^^^
AttributeError: 'TenantUser' object has no attribute 'account'
``` 

I guess it was already there, but we couldn't reach this code before and now we can? :man_shrugging:

tdethier: Not 100% sure, but:

```suggestion
        return [tu.account_user.iaso_profile.account for tu in self.main_user.tenant_users.exclude(pk=self.pk) if tu.account_user.iaso_profile]
```

iaso/api/profiles/profiles.py
tdethier: I also believe we'll never get a `tenant_main_user` there - at least for the moment. But since `create_user_or_tenant_user()` takes care of some extra logic, I would keep it. Otherwise we'll need to repeat part of this logic here.

tdethier (commented):
After fixing the issue in my comments below and logging in with my main user, I never was able to edit one of my account users through the web interface:
image


The logic sounds good, but I'm not sure why I couldn't make this work. Let's discuss this next week!

iaso/api/profiles/profiles.py
kemar: All right! I added a comment explaining just that.

iaso/models/tenant_users.py
kemar: Turns out I was a little to fast while removing code I thought was dead 😨

Your issue should be fixed by https://github.com/BLSQ/iaso/pull/2284/commits/b834d26c8dc66a065bd05c2361566ec4c5e38391

tdethier (approved):
I was able to change my password through the web interface, good job :+1:

PR #2297, opened 2025-07-15
fix: bulk uploads remove error from APIImport on successfull run

bramj:
Small fix: When an errored task is relaunched, we want the corresponding `api_import` to no longer show up as errored in the Django admin.


tdethier (approved):
LGTM :+1:

PR #2300, opened 2025-07-16
IA-4377 Form versions speedup

mestachs:
the load of the "self.get_objects" trigger the whole get_queryset with heavy count and max(instance_updated_at)

Related JIRA tickets : IA-4377

- [ ] Are there enough tests?

## Doc

## Changes

I fixed the test that didn't work in docker (known issue by thibault)

I tried various way to skip the get_object but the http status behaviour was never the one expected by the test. So I ended up check if is_request_from_manifest to skip all the heavy queries.

## How to test

seed a project
login
edit the PCA form 
  - add an attachement

then test by accessing the url (change the id with your seed form)

`view-source:http://localhost:8081/api/forms/7/manifest/`

it should display an xml like this one
```
<?xml version="1.0" encoding="UTF-8"?>
<manifest xmlns="http://openrosa.org/xforms/xformsManifest">
    <mediaFile>
        <filename>annif.jpeg</filename>
        <hash>md5:f2da2c4d60f2ec309accf178f785a852</hash>
        <downloadUrl>http://iaso:8081/media/form_attachments/7/annif.jpeg</downloadUrl>
    </mediaFile>
</manifest>
```
and the logs should says 

```
iaso-1     | http://localhost:8081/api/forms/7/manifest/
iaso-1     | |------|-----------|----------|----------|----------|------------|
iaso-1     | | Type | Database  |   Reads  |  Writes  |  Totals  | Duplicates |
iaso-1     | |------|-----------|----------|----------|----------|------------|
iaso-1     | | RESP |  default  |    9     |    0     |    9     |     0      |
iaso-1     | |------|-----------|----------|----------|----------|------------|
iaso-1     | Total queries: 9 in 0.1167s 
```

previous it was 17 and the main sql was with condition like
```
  MAX("iaso_instance"."updated_at") AS "instance_updated_at",
                COUNT(DISTINCT "iaso_instance"."id") FILTER (
                                                             WHERE (NOT ("iaso_instance"."file" = ''
                                                                         AND "iaso_instance"."file" IS NOT NULL)
                                                                    AND NOT ("iaso_device"."test_device"
                                                                             AND "iaso_device"."test_device" IS NOT NULL)
                                                                    AND NOT ("iaso_instance"."deleted"
                                                                             AND "iaso_instance"."deleted" IS NOT NULL))) AS "instances_count",
```

now there's no more joins on iaso_instance.

```
fix: empty instance pop up

Refs: IA-3665
```

### DIFF
--- a/iaso/api/forms.py
+++ b/iaso/api/forms.py
@@ -298,7 +298,11 @@ class FormsViewSet(ModelViewSet):
             queryset = queryset.filter(projects__id__in=projects_ids.split(","))
 
         requested_fields = self.request.query_params.get("fields")
-        order = self.request.query_params.get("order", "instance_updated_at").split(",")
+
+        is_request_from_manifest = self.request.path.endswith("/manifest/")
+        default_order = "id" if is_request_from_manifest else "instance_updated_at"
+
+        order = self.request.query_params.get("order", default_order).split(",")
 
         if is_field_referenced("has_mappings", requested_fields, order):
             queryset = queryset.annotate(
@@ -315,10 +319,10 @@ class FormsViewSet(ModelViewSet):
         else:
             profile = False
 
-        if is_field_referenced("instance_updated_at", requested_fields, order):
+        if is_field_referenced("instance_updated_at", requested_fields, order) and not is_request_from_manifest:
             queryset = queryset.annotate(instance_updated_at=Max("instances__updated_at"))
 
-        enable_count = is_field_referenced("instances_count", requested_fields, order)
+        enable_count = is_field_referenced("instances_count", requested_fields, order) and not is_request_from_manifest
 
         if not mobile and enable_count:
             if profile and profile.org_units.exists():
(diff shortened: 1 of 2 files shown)


iaso/api/forms.py
```diff
@@ -298,7 +298,11 @@ def get_queryset(self, mobile=False):
             queryset = queryset.filter(projects__id__in=projects_ids.split(","))
 
         requested_fields = self.request.query_params.get("fields")
-        order = self.request.query_params.get("order", "instance_updated_at").split(",")
+
+        is_request_from_manifest = self.request.path.endswith("/manifest/")
```
kemar, line 302: You can also test `self.action`, like:

```python
self.action == "manifest"
```

iaso/tests/api/test_forms_attachment.py
```diff
@@ -373,6 +391,13 @@ def test_manifest_anonymous_app_id(self):
         )
         self.assertXMLResponse(response, 200)
 
+        # to ensure performance we clearly don't want to hit iaso_instance
+        sql_queries = [q["sql"] for q in connection.queries]
+        self.assertEqual(len(sql_queries), 8, f"should have collected queries {sql_queries}")
+        matching = [q for q in sql_queries if "iaso_instance" in q.lower()]
+
+        self.assertEqual(len(matching), 0, f"'iaso_instance' found in queries: {matching}")
```
kemar, line 399: Tests are always executed with `DEBUG=False` and [the docs explain the rationale for this](https://docs.djangoproject.com/es/5.2//topics/testing/overview/#other-test-conditions). I think this a good thing to stay as close as possible to production.

I guess this is why you used `@override_settings`, because Django tracks all queries in the connection object only if `DEBUG=True`. Yet this is a bit hard to understand if you don't know that.

It turns out there is another option which does not require to override settings.

Use `CaptureQueriesContext` ([credit StackOverflow](https://stackoverflow.com/a/60299392)) to temporarily enable query tracking on a connection.

E.g.

```python
    def test_manifest_anonymous_app_id(self):
        f"""GET {BASE_URL} via app id"""

        from django.test.utils import CaptureQueriesContext

        with CaptureQueriesContext(connection):

            response = self.client.get(
                MANIFEST_MOBILE_URL.format(form_id=self.form_2.id),
                headers={"Content-Type": "application/json"},
                data={"app_id": self.project_1.app_id},
            )
            self.assertXMLResponse(response, 200)

            # to ensure performance we clearly don't want to hit iaso_instance
            sql_queries = [q["sql"] for q in connection.queries]
            self.assertEqual(len(sql_queries), 8, f"should have collected queries {sql_queries}")
            matching = [q for q in sql_queries if "iaso_instance" in q.lower()]

            self.assertEqual(len(matching), 0, f"'iaso_instance' found in queries: {matching}")

```

Of course, you should move the import at the top of the file, should you decide to accept the suggestion.

kemar (approved):
LGTM

I have a few suggestions that I think could be worth it but it's not blocking.

PR #2302, opened 2025-07-16
IA-4363 Fix flaky tests (bis)

kemar:
Fix flaky tests.

Related JIRA tickets : [IA-4363](https://bluesquare.atlassian.net/browse/IA-4363)

## Changes

- Fix `ModificationModelTestCase.test_create()`
    - there was an error and we were just lucky that the test succeeded with a fresh DB
- Fix `PaymentLotsViewSet.create()`
    - there was a bug where a number made up of multiple digits was parsed as multiple integers
    - e.g., `56` was parsed as `5` and `6`
    - this bug was just real in tests since `format="json"` was missing from the HTTP requests

[IA-4363]: https://bluesquare.atlassian.net/browse/IA-4363?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iaso/api/payments/views.py
+++ b/iaso/api/payments/views.py
@@ -227,16 +227,19 @@ class PaymentLotsViewSet(ModelViewSet):
         responses={status.HTTP_201_CREATED: PaymentLotSerializer()},
     )
     def create(self, request):
-        # with transaction.atomic():
-        # Extract user, name, comment, and potential_payments IDs from request data
         user = self.request.user
         name = request.data.get("name")
         comment = request.data.get("comment")
-        potential_payment_ids = request.data.get("potential_payments", [])  # Expecting a list of IDs
-        potential_payment_ids = [int(pp_id) for pp_id in potential_payment_ids]
-        # TODO move this in valdate method
+        potential_payment_ids = request.data.getlist("potential_payments")  # Expecting a list of IDs
+
+        try:
+            potential_payment_ids = [int(pp_id) for pp_id in potential_payment_ids]
+        except ValueError:
+            raise ValidationError("Expecting `potential_payments` to be a list of IDs.")
+
         if not potential_payment_ids:
-            raise ValidationError("At least one potential payment required")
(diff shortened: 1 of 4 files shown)


iaso/api/payments/views.py
```diff
@@ -227,16 +227,19 @@ def update(self, request, *args, **kwargs):
...
     def create(self, request):
-        # with transaction.atomic():
-        # Extract user, name, comment, and potential_payments IDs from request data
         user = self.request.user
         name = request.data.get("name")
         comment = request.data.get("comment")
-        potential_payment_ids = request.data.get("potential_payments", [])  # Expecting a list of IDs
-        potential_payment_ids = [int(pp_id) for pp_id in potential_payment_ids]
-        # TODO move this in valdate method
+        potential_payment_ids = request.data.getlist("potential_payments")  # Expecting a list of IDs
```
tdethier: I tried this fix, but got an error:

```  File "/home/thibault/dev/iaso/iaso/iaso/api/payments/views.py", line 234, in create
    potential_payment_ids = request.data.getlist("potential_payments")  # Expecting a list of IDs
                            ^^^^^^^^^^^^^^^^^^^^
AttributeError: 'dict' object has no attribute 'getlist'
```

tdethier (changes requested):
The audit test is fixed, but the other fix didn't work

iaso/api/payments/views.py
kemar: This is because DRF uses its [own `Request` class](https://github.com/encode/django-rest-framework/blob/master/rest_framework/request.py#L140) and `data` does not return a `QueryDict` when you access `Request.data`, but a `dict` instead.

I forgot this point. My last commit should fix this.

tdethier (approved):
LGTM :+1:

PR #2303, opened 2025-07-16
[IA-3928] Split core permissions from hat.menupermissions

tdethier:
Split core permissions from hat.menupermissions

Related JIRA tickets : IA-3928, IA-3893 (same ticket with polio permissions)

- [x] Are there enough tests?

## Doc

/

## Changes

- Removed core IASO permissions from `hat.menupermissions` and placed them in `iaso.permissions`
- Created a `CorePermissionSupport` class to manage them
- Replaced all calls to `hat.menupermissions` with `iaso.permissions`

## How to test

Nothing should change for end users.

Log into the system with a super user, visit some pages. 
Try to do the same with a normal user, make sure that their access is correctly restricted to their permissions

## Print screen / video

/

## Notes
/

```
fix: empty instance pop up

Refs: IA-3665
```


iaso/migrations/0333_split_core_permissions_from_hat.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+class Migration(migrations.Migration):
+    """
+    This is the equivalent of the 0235 migration in the polio plugin.
+    """
+
+    dependencies = [
+        ("iaso", "0332_corepermissionsupport"),
+        ("menupermissions", "0074_alter_custompermissionsupport_options"),
```
tdethier: I had to add this `menupermissions` dependency, otherwise the migrations would fail on an empty DB. I'm not sure if that's a good idea

hat/locale/fr/LC_MESSAGES/django.po
```diff
@@ -455,6 +295,126 @@ msgstr "L'équipe %(site_name)s."
...
+#~ msgstr "Écrire source de données"
+
+#~ msgid "Write page"
+#~ msgstr "Écrire page"
+
+#~ msgid "Payments page"
+#~ msgstr "Page des paiements"
+
 #~ msgid "Polio"
 #~ msgstr "Polio"
```
kemar, line 419: Commented translations can be deleted safely.

hat/menupermissions/constants.py
```diff
@@ -229,60 +147,29 @@
...
-        "read": "iaso_planning_read",
-        "write": "iaso_planning_write",
-    },
-    "iaso_page_permissions": {"read": "iaso_pages", "write": "iaso_page_write"},
-    "iaso_user_permissions": {"geo_limited": "iaso_users_managed", "all": "iaso_users"},
-}
+READ_EDIT_PERMISSIONS = {}
+
+
+def load_permissions_from_permissions_module(permissions_module: ModuleType):
```
kemar, line 155: I'd say that this could be worth a test.

kemar (commented):
LGTM but I don't approve yet, because I think it would be better if people working on dedicated plugins could have a look.

iaso/migrations/0334_split_core_permissions_from_hat.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+class Migration(migrations.Migration):
+    """
+    This is the equivalent of the 0235 migration in the polio plugin.
+    """
+
+    dependencies = [
+        ("iaso", "0333_corepermissionsupport"),
+        ("menupermissions", "0074_alter_custompermissionsupport_options"),
```
tdethier, line 121: I had to add this `menupermissions` dependency, otherwise the migrations would fail on an empty DB. I'm not sure if that's a good idea, it feels like it's going to remove the permissions first and this migration will not be able to switch them to the new ones :thinking:

tdethier: I just tested on a new DB: the 0074 `menupermissions` migration does not remove the permissions inside the `auth_permission` table, even if core permissions are no longer in the migration file. 

This migration file here (0334 `iaso`) is the one that removes permissions from the table, so the order is fine and there shouldn't be any issue with existing instances.

The only questions I have left are what are we going to do with this (and its equivalents in plugins) when:
- we want to squash migrations in the future?
- we completely get rid of `menupermissions`?

iaso/migrations/0334_split_core_permissions_from_hat.py
```diff
@@ -0,0 +1,119 @@
...
+from django.contrib.contenttypes.models import ContentType
+from django.db import migrations
+
+
+def migrate_permissions(apps, schema_editor):
+    print("Starting permission migration...")
+
+    # Get content types
+    old_ct = ContentType.objects.get(app_label="menupermissions", model="custompermissionsupport")
+    new_ct, _ = ContentType.objects.get_or_create(app_label="iaso", model="corepermissionsupport")
```
tdethier: In some cases, Django can try to apply this migration while `0333`'s transaction is not completely done yet. This can create an issue since this `ContentType` does not fully exist yet. 
This is why Mathieu - in his first polio PR - had initially used a `get_or_create` here as well as when fetching individual permissions from this content type below.

iaso/migrations/0334_split_core_permissions_from_hat.py
```diff
@@ -0,0 +1,119 @@
...
+    for old_codename, new_codename in permission_mapping.items():
+        print(f"\nProcessing permission: {old_codename} -> {new_codename}")
+        try:
+            old_perm = Permission.objects.get(codename=old_codename, content_type=old_ct)
+            print(f"Found old permission: {old_perm}")
+
+            # Create new permission if it doesn't exist
+            new_perm, created = Permission.objects.get_or_create(
+                codename=new_codename, content_type=new_ct, defaults={"name": old_perm.name}
+            )
```
tdethier: Same comment about `get_or_create`, see above

iaso/migrations/0334_split_core_permissions_from_hat.py
kemar: Okay, I think I've figured out what's going on.

It seems to be a known problem:

- [permissions](https://code.djangoproject.com/ticket/29843)
- [content types](https://groups.google.com/g/django-developers/c/HDKVHrmSMdE)

These two commands are executed via the `post_migrate` signal:

- [`create_permissions`](https://github.com/django/django/blob/591b23a11b099081fc39d00171f5dc9efbd67e13/django/contrib/auth/apps.py#L19-L22)
- [`create_contenttypes`](https://github.com/django/django/blob/stable/4.2.x/django/contrib/contenttypes/apps.py#L20)

But `post_migrate` is triggered [only after migrate](https://docs.djangoproject.com/en/5.2/ref/signals/#post-migrate).

In other words, only after all migrations.

Which explains why you can't find the right content type.

A workaround would be to call `create_contenttypes` at the start of the migration operation https://stackoverflow.com/a/40092780

This would probably guarantee a more stable state than `get_or_create`.

tdethier: Thanks for the research! 

I tested locally these changes and it works. However, I already did not encounter any issues with the previous versions, so I guess we'll see during the deployment if it causes any issue.

kemar (approved):

PR #2304, opened 2025-07-16
IA-4401 Simplify duplicates code

kemar:
Simplify duplicates code.

Related JIRA tickets : IA-4401

## Changes

While reading the code to understand it, I made some simplifications that I think are worth it:

- use `ModelViewSet` instead of `viewsets.GenericViewSet ` (this allows to save a bunch of code)
- make the model the single source of truth for `POSSIBLE_ALGORITHMS`
    - and use an enum for `PossibleAlgorithms`
    - I also removed the class method that registered a subclass in a class variable since this was causing `import` headaches
- deleted unused code
- use ` bulk_create` for performance
- rewrite the SQL query with Common Table Expressions and avoid relying on `created_at`
    - note: this is solely for readability, I'll need to check that the performance is not worse than using a query without CTE

## How to test

Everything is covered by the existing unit tests plus a bunch that I added.


docs/pages/dev/reference/entities_in_iaso/entity_deduplication.en.md
quang-le, line 1: ❤️

iaso/api/deduplication/algos/finalize.py
```diff
@@ -1,29 +1,40 @@
...
+
+    task.report_progress_and_stop_if_killed(progress_message="Started creation of entity duplicates")
 
-    ed_model = apps.get_model("iaso", "EntityDuplicate")
+    batch_size = 100
 
-    for pot_dup in potential_duplicates:
-        ed, _ = ed_model.objects.get_or_create(
+    entity_duplicates = (
+        EntityDuplicate(
```
quang-le, line 22: Do we not risk creating duplicate duplicates (ent 1 + ent 2 --> ent 2 + ent 1) if we remove the `get_or_create`?

iaso/api/deduplication/algos/levenshtein.py
```diff
@@ -15,17 +17,18 @@
 
 
 def _build_query(params):
-    the_fields = params.get("fields", [])
+    reference_form_fields = params.get("fields", [])
```
quang-le, line 20: ❤️

quang-le: Although I'm not sure something like `selected_fields` or `fields_to_compare` wouldn't be more accurate, since users can select the fields they want to compare. Not blocking, ofc

iaso/api/deduplication/entity_duplicate_analyzis.py
```diff
@@ -266,10 +186,9 @@ def create(self, request, *args, **kwargs):
             "fields": data["fields"],
             "parameters": {param["name"]: param["value"] for param in data["parameters"]},
         }
-        the_task = run_deduplication_algo(algo_name=algo_name, algo_params=algo_params, user=request.user)
+        task = run_deduplication_algo(algo_name=algo_name, algo_params=algo_params, user=request.user)
```
quang-le, line 215: I know that in other parts of the code we avoid using the `task` variable name, I think because it can mess with the decorator (but that's just a wild guess/hazy recollection)

iaso/tasks/run_deduplication_algo.py
```diff
@@ -14,10 +17,29 @@ def run_deduplication_algo(algo_name=None, algo_params=None, task=None):
...
+    # =>
+    # [
+    #     {'name': 'Prenom', 'type': 'text', 'label': "{'English': 'Prenom'}"},
+    #     {'name': 'Nom', 'type': 'text', 'label': "{'English': 'Nom'}"},
+    #     {'name': 'age__int__', 'type': 'calculate', 'label': "{'English': 'Age (Months)'}"},
+    # ]
+    possible_fields = EntityType.objects.values_list("reference_form__possible_fields", flat=True).get(
+        id=algo_params["entity_type_id"]
+    )
+    reference_form_fields = [find_question_by_name(field, possible_fields) for field in algo_params.get("fields", [])]
```
quang-le, line 37: OK, I get why you named it like that now

quang-le (commented):
Here are already my comments about the code. It's much easier to read that way indeed.
I'll test the feature later today and approve by then

iaso/api/deduplication/algos/finalize.py
kemar: Nope, because we have a unicity constraint on ent 1 and ent 2 -> https://github.com/BLSQ/iaso/blob/b70434aa7000635c884ed1be26d6d1de2caa747b/iaso/models/deduplication.py#L60

Any duplicate will result in a "unique constraint error" which will be ignored by the `ignore_conflicts` argument in `bulk_create()` a little further down.

This will be much faster than a loop and a get or create for each object.

iaso/api/deduplication/entity_duplicate_analyzis.py
kemar: That's what I thought too but it seems `task` is not shadowing any other local variables so I changed its name.

If you think that `the_task` is safer, I can revert this change.

quang-le (approved):

PR #2306, opened 2025-07-17
POLIO-1948: azure storage

beygorghor:
Currently IASO is programmed to run either on local file system or on s3 compatible backends. 

For polio, we now have to run IASO on azure, so, we need make that possible through the options in settings.py, leveraging what django-storages provides. 

We need to test then if it works well on Azure by running the setuper on an instance connected to azure storage to ensure that everything is working well?) 

I’m expecting a bit of problems around the fact that on s3, we have static files and media files in the same bucket, except that the media files are available without authentication. 

Related JIRA tickets : POLIO-1948

- [ ] Are there enough tests?

## Doc

docs/Azure-Storage-Configuration.md

## Changes

- adapt settings.py to be able to use S3 OR Azure storage
- Wrote some docs regarding this
- I have added `AzureStaticStorage` and `AzureMediaStorage`

## How to test

- Mbayang made some test on azure and static files seems to work properly
- I also deployed this branch on staging, files are accessible and nothing seems to be broken

## Print screen / video

-

## Notes

```
fix: empty instance pop up

Refs: IA-3665
```


scripts/test_azure_storage.py
quang-le, line 1: Please remove emojis

quang-le (changes requested):

quang-le (approved):

PR #2309, opened 2025-07-18
IA-4392 improve forms  / submissions pages

beygorghor:
- Counting submissions for each form is very heavy especially on production. My advice here will be to disable the count of submissions by default and have a switch on the front to enable it if requested

- Also launching search on the forms by default while loading the page is not very efficient too. We should wait that the user hit the search button first like on the org units page. Same thing can be done on instances pages (something also requested by Caroline on Trypelim project

Related JIRA tickets : IA-4392

- [x] Are there enough tests?

## Doc

-

## Changes

- adding display count checkbox to hide / display instances count on form list
- disable auto-search while loading page on forms and instances
- fix cypress tests

## How to test

- visit forms page, click on search, show/hide count column, play with filters
- visit instances page, click on search, play with filters

## Print screen / video

https://github.com/user-attachments/assets/a230d7c0-5455-4fe4-a2dc-95f41023cea4

## Notes

-

```
fix: empty instance pop up

Refs: IA-3665
```


hat/assets/js/apps/Iaso/components/tables/ColumnSelectDrawer/index.tsx
```diff
@@ -130,7 +130,7 @@ export const ColumnsSelectDrawer: FunctionComponent<Props> = ({
             {(!disabled || (disabled && !disabledMessage)) && (
                 <Button
                     disabled={disabled}
-                    variant="contained"
+                    variant="outlined"
```
quang-le, line 133: Why the switch to outlined? If it's a design/style decision, it should be documented somewhere and we should have a ticket to update the style everywhere (incl. polio)

hat/assets/js/apps/Iaso/domains/app/translations/en.json
```diff
@@ -349,6 +349,7 @@
     "iaso.forms.viewForm": "View form",
     "iaso.forms.viewInstance": "View submission",
     "iaso.forms.viewInstances": "View submission(s)",
+    "iaso.forms.showInstancesCount": "Show instances count",
```
quang-le:
```suggestion
    "iaso.forms.showInstancesCount": "Show instance count",
```

hat/assets/js/apps/Iaso/domains/forms/components/Filters.tsx
```diff
@@ -1,53 +1,60 @@
...
     const handleShowDeleted = useCallback(
         (key, value) => {
-            // converting false to undefined to be able to compute `filtersUpdated` correctly
-            const valueForParam = value || undefined;
+            const valueForParam = value ? 'true' : undefined;
+            handleChange(key, valueForParam);
+        },
+        [handleChange],
+    );
+    const handleShowInstancesCount = useCallback(
```
quang-le, line 54: There's a `useCheckboxState` hook in useFilterState.ts that could be used (FYI), but I'm not sure the code would gain much clarity in this particular case

hat/assets/js/apps/Iaso/domains/forms/hooks/useGetForms.tsx
```diff
@@ -1,11 +1,48 @@
...
+    'possible_fields',
+    'legend_threshold',
+    'has_mappings',
+];
 
 const getForms = params => {
-    const queryString = new URLSearchParams(params).toString();
+    const fieldsParams = {
+        fields: FIELDS_PARAMS.join(','),
+    };
```
quang-le: this could be moved outside of `getForms` as well

hat/assets/js/apps/Iaso/domains/forms/hooks/useGetForms.tsx
```diff
@@ -1,11 +1,48 @@
...
+
+    if (params.showInstancesCount === 'true') {
+        fieldsParams.fields += ',instances_count';
+    }
+
+    const queryString = new URLSearchParams({
+        ...params,
+        ...fieldsParams,
+    }).toString();
```
quang-le:
```suggestion
    const fields = params.showInstancesCount === 'true'
        ? FIELDS_PARAMS.join(',') + ',instances_count' 
        : FIELDS_PARAMS.join(',');
        
       const queryString = new URLSearchParams({
        ...params,
        fields,
    }).toString();
```

This would avoid creating an object with just one field and spread it immediately after

hat/assets/js/apps/Iaso/domains/instances/components/InstancesFiltersComponent.js
```diff
@@ -563,7 +564,11 @@ const InstancesFiltersComponent = ({
                             </Box>
                         )}
                         <Button
-                            disabled={textSearchError || isSearchDisabled}
+                            disabled={
+                                textSearchError ||
+                                (params?.isSearchActive === 'true' &&
+                                    isSearchDisabled)
```
quang-le: SearchActive is true and Search is Disabled at the same time. The naming is a bit confusing here. If I understand well, the check on SearchActive is to "counter" the check on `!isInstancesFilterUpdated`, but the result is a bit confusing to read 😅

quang-le (changes requested):
Code review without feature test, mostly code style remarks.
Really blocking is documenting the change on the button style change and the name confusion around the boolean.
I'll test the feature and add a review later today

quang-le:
Tested the feature. Works OK, so no further comments

hat/assets/js/apps/Iaso/components/tables/ColumnSelectDrawer/index.tsx
beygorghor: indeed that was a remark from Martin. Search is the primary action here, this is mainly to force user to focus on the search button because search is not enabled by default. I can create a ticket for that, but the purpose here is mainly on those pages where the search is different

beygorghor: I change it to be coherent on orgunit page too, for me the rest can stay like that for now

hat/assets/js/apps/Iaso/domains/forms/components/Filters.tsx
beygorghor: yes indeed, this is adding a state that i just removed, this part of the code is tricky and works like that, i'll leave it like that

quang-le (approved):

PR #2319, opened 2025-07-22
SLEEP-1622: Entity csv/xlsx high memory usage

Phil-V:
Fix out-of-memory error when exporting large amounts of entities.

Related JIRA tickets : [SLEEP-1622](https://bluesquare.atlassian.net/browse/SLEEP-1622)

- [x] Are there enough tests?

## Doc

/

## Changes

- Fixed high memory usage when using the entity csv/xlsx export. 
- On the entity details page, removed the csv/xlsx export buttons.
- Fixed N+1 queries in the entity list view.
- When extra columns from the entity type are displayed on the entities list page , the csv/xlsx export will now also include those extra columns.

## How to test

Visit the entity list page, with a single id in the `entityTypeIds` parameter specified (`/dashboard/entities/list/accountId/1/entityTypeIds/1/...`). The EntityType should have a non-empty `fields list view` attribute (See `test_export_entities_extra_columns` for an example setup). 

Memory usage should stay relatively constant when exporting large amounts of entities csv/xlsx (previously about 1GB/100k entities). Can keep an eye on memory usage with `docker stats`. 

## Print screen / video

/

## Notes

- The EntityViewSet.list view doesn’t rely on the function `_get_duplicates` anymore, to get rid of the N+1 queries. But it might have introduced a subtle shift in behavior (see the if/elif logic in `_get_duplicates` in `/api/entity.py`). I don’t think it matters but that logic seemed odd and I didn't replicate it fully so it is an area to double-check. 
- The csv export will stream a response to the user immediately, but the xlsx export will still assemble the final xlsx file in-memory before sending it as a response (this should be only a few megabytes even with hundreds of thousands of rows). It is not great UX however and only a stop-gap fix. 

[SLEEP-1622]: https://bluesquare.atlassian.net/browse/SLEEP-1622?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iaso/api/entity.py
```diff
@@ -336,55 +343,67 @@ def list(self, request: Request, *args, **kwargs):
...
-                columns_list = []
-                possible_fields_list = entity.entity_type.reference_form.possible_fields or []
-                for items in possible_fields_list:
-                    for k, v in items.items():
-                        if k == "name" and v in entity.entity_type.fields_list_view:
-                            columns_list.append(items)
+        # if exactly one type of entity specified, add columns for that type of entity
+        columns_list = []
+        if entity_type_ids is not None and len(entity_type_ids.split(",")) == 1:
+            entity_type = EntityType.objects.get(id=int(entity_type_ids))
```
Phil-V: Missed a select_related here!

kemar (approved):
LGTM

PR #2328, opened 2025-07-25
POLIO-1982: country block dropdown api

quang-le:
Add API for a country block drop down in lqas

Related JIRA tickets : POLIO-1982, POLIO-1971

- [x] Are there enough tests?

## Doc

NA.

## Changes

- Add endpoint for country block dropdown
    - applies the same logic as the existing dropdowns in the same module (re: dates)
    - also handles subactivities

## How to test
 No interface, so play around with the API (the tests should cover all possible sceanrios already)


plugins/polio/tests/api/lqas_im/test_lqas_im_options.py
```diff
@@ -731,3 +733,443 @@ def test_lqas_date_used_when_exists(self):
         result = results[0]
         self.assertEqual(result["label"], f"Round {self.benin_round_3.number}")
         self.assertEqual(result["value"], str(self.benin_round_3.number))
+
+
+class PolioLqasImCountryBlockOptionsTestCase(LqasImOptionsTestCase):
```
tdethier: This file is getting huge, I would maybe make a new directory for all LqasImOptionsTestCase, one base file and one file for each endpoint

plugins/polio/api/lqas_im/lqas_im_dropdowns.py
```diff
@@ -195,3 +197,80 @@ class LqasImRoundOptionsViewset(ModelViewSet):
...
+                days=10
+            ),  # since we can't use ended_at +10 >= first_day, we compare ended_at with first_day -10
+            end_date__lte=last_day - timedelta(days=10),  # same logic as above
+        )
+
+        # Find subactivities within the date-month period
+        sub_activities_with_lqas = SubActivity.objects.filter(
+            subactivity_with_lqas_end_date | subactivity_without_lqas_end_date
+        )
+        with_subactivity = Q(sub_activities__in=sub_activities_with_lqas)
```
tdethier: I think you can optimize it a bit more and fetch only IDs since you don't need whole `SubActivity` objects (like what you did below)

plugins/polio/api/lqas_im/lqas_im_dropdowns.py
```diff
@@ -195,3 +197,80 @@ class LqasImRoundOptionsViewset(ModelViewSet):
...
+            raise ValidationError({"month": [f"Cannot convert {value} to date object"]})
+
+        with_lqas_end_date = Q(lqas_ended_at__isnull=False, lqas_ended_at__gte=first_day, lqas_ended_at__lte=last_day)
+        # When no lqas end date we use round date +10 as fallback option (same convention used in Openhexa pipeline)
+        without_lqas_end_date = Q(
+            lqas_ended_at__isnull=True,
+            ended_at__gte=first_day
+            - timedelta(
+                days=10
+            ),  # since we can't use ended_at +10 >= first_day, we compare ended_at with first_day -10
```
tdethier, line 233: Thanks for the comments, this 10-day period is confusing :sweat_smile:

plugins/polio/tests/api/lqas_im/test_lqas_im_options.py
```diff
@@ -1,8 +1,10 @@
 import datetime
 
+from iaso.models.base import GROUP_DOMAIN, Group
```
tdethier: `GROUP_DOMAIN` no longer exist, I still don't understand why these tests don't fail :face_with_spiral_eyes: 
Are we sure that they were actually run?

tdethier: Ok I see that `plugins.polio.tests.api.lqas_im` doesn't have any `__init__.py` file, so these tests were not run. I added this file locally and when I run these tests, they fail because of the invalid import

plugins/polio/tests/api/lqas_im/test_lqas_im_options.py
```diff
@@ -731,3 +733,443 @@ def test_lqas_date_used_when_exists(self):
...
+        json_response = self.assertJSONResponse(response, 200)
+        results = json_response["results"]
+        self.assertEqual(len(results), 0)
+
+    def test_get_without_params(self):
+        self.client.force_authenticate(self.user)
+        response = self.client.get(self.endpoint)
+        json_response = self.assertJSONResponse(response, 200)
+        results = json_response["results"]
+        self.assertEqual(len(results), 2)
```
tdethier: I would maybe add an assert with the expected IDs

plugins/polio/tests/api/lqas_im/test_lqas_im_options.py
```diff
@@ -731,3 +733,443 @@ def test_lqas_date_used_when_exists(self):
...
+
+    def test_return_blocks_with_lqasend_date_within_month_param(self):
```
tdethier:
```suggestion
    def test_return_blocks_with_lqas_end_date_within_month_param(self):
```

plugins/polio/tests/api/lqas_im/test_lqas_im_options.py
```diff
@@ -731,3 +733,443 @@ def test_lqas_date_used_when_exists(self):
...
+        self.assertNotEqual(result["label"], new_eligible_country_block.name)
+        self.assertNotEqual(result["value"], new_eligible_country_block.id)
+        self.assertEqual(result["label"], self.eligible_country_block.name)
+        self.assertEqual(result["value"], self.eligible_country_block.id)
+
+        test_campaign.is_test = False
+        test_campaign.on_hold = True
+        test_campaign.save()
+
+        #  ingnore campaign on hold
```
tdethier:
```suggestion
        #  ignore campaign on hold
```

plugins/polio/tests/api/lqas_im/test_lqas_im_options.py
```diff
@@ -731,3 +733,443 @@ def test_lqas_date_used_when_exists(self):
...
+        self.assertNotEqual(result["value"], new_eligible_country_block.id)
+        self.assertEqual(result["label"], self.eligible_country_block.name)
+        self.assertEqual(result["value"], self.eligible_country_block.id)
+
+        test_campaign.on_hold = False
+        test_campaign.save()
+        test_rnd3.on_hold = True
+        test_rnd3.save()
+
+        #  ingnore if round for selected period is on hold
```
tdethier:
```suggestion
        #  ignore if round for selected period is on hold
```

plugins/polio/tests/api/lqas_im/test_lqas_im_options.py
tdethier: I would also add the following tests:
- calling the endpoint with the anon user
- calling the endpoint with the user without perms
- calling the endpoint with an invalid value for `month`
  - wrong format
  - wrong values (like `?month=13-2025`)

tdethier (changes requested):
I think I found why the tests didn't fail :sweat_smile: 

See my comments below, there should be a few things to fix

plugins/polio/tests/api/lqas_im/test_lqas_im_options.py
quang-le: anon user and user without perms is handled: it's inherited from the parent class

tdethier (approved):
LGTM :+1:

PR #2330, opened 2025-07-28
POLIO-1988: remove "domain" field in Groups

quang-le:
`Group`model has a legacy `domain` field that is deprecated but could still cause bugs and does create unexpected errors in tests

Related JIRA tickets :  POLIO-1988

- [ ] Are there enough tests?

## Changes

- Add a migration to enable `objects` to return all groups
- Remove all references to `domain` in the codebase
- Add a migration setting the value of `domain` to None where it exists
- Remove extra `Managers` for `Group` and move the `filter_for_user` method to a custom queryset
- Add a migration removing the field from the model

- 
## How to test

- Try importing or exporting groups via gpkg
- Try using a group dropdown anywhere in the UI

Nothing should break

## Print screen / video

N/A


iaso/migrations/0336_merge_20250731_0902.py
```diff
@@ -0,0 +1,12 @@
+# Generated by Django 4.2.23 on 2025-07-31 09:02
```
tdethier, line 1: Nitpick: I would have edited all files in order to avoid an extra 4th migration, but it's alright

tdethier (approved):
LGTM :+1: 

Be careful when this is released: I guess that some hidden groups might reappear and disturb some screens/dashboards

PR #2331, opened 2025-07-28
[IA-4405] New nginx config to compress requests

madewulf:
Please SQUASH this messy PR when merging

The backend currently does not gzip its answers. The best place to do that is in nginx for performance reasons. 

This PR just do a replacement of gzip off to gzip on in the default nginx config. 

I tried to do that by adding config file, but that does not work because you can't have the gzip command twice in a nginx config (I was first looking at overriding the default setting). 

This is basically impossible to test outside of beanstalk, so I tested by deploying the branch. 
Here is the proof that it works
CleanShot 2025-07-28 at 17 20 27@2x

As a bonus, I realized while looking at this that some of our additional nginx configuration for higher timeouts and allowing bigger file uploads were not active because not in the right directory. I moved them to the correct directory to be picked up by the nginx.configuration. 

The default nginx beanstalk configuration looks like this: (the crucial part being for these two files that additional configuration is picked in conf.d/elasticbeanstalk/*.conf;) 

``` 
http {
    server_tokens off;

    include       /etc/nginx/mime.types;
    default_type  application/octet-stream;

    log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
                      '$status $body_bytes_sent "$http_referer" '
                      '"$http_user_agent" "$http_x_forwarded_for"';

    include       conf.d/*.conf;

    map $http_upgrade $connection_upgrade {
        default     "upgrade";
    }

    server {
        listen        80 default_server;
        access_log    /var/log/nginx/access.log main;

        client_header_timeout 60;
        client_body_timeout   60;
        keepalive_timeout     60;
        gzip                  off;
        gzip_comp_level       4;
        gzip_types text/plain text/css application/json application/javascript application/x-javascript text/xml application/xml application/xml+rss text/javascript;

        # Include the Elastic Beanstalk generated locations
        include conf.d/elasticbeanstalk/*.conf;
    }
}
```
Related JIRA tickets : IA-4405

- [x ] Is my code clear enough and well documented?

```
fix: empty instance pop up

Refs: IA-3665
```


tdethier (approved):
LGTM :+1:

PR #2336, opened 2025-07-30
SLEEP-1623: Mail alert to coordinations

Phil-V:
Note: also see PR in the trypelim plugin repo https://github.com/BLSQ/trypelim/pull/189

We should notify the coordination (based on the uploader) whenever a mobile app upload contains confirmed cases. We already do this for the uploads from the old mobile app, and we want the same for uploads from IASO.

- [x] Are there enough tests?

## Doc

/

## Changes

The process_mobile_bulk_upload task was modified to keep track of the import instances ids.
At the end of the task, a query is run to check for new positive confirmations.
If confirmations are found, an email with an URL pointing to the new instances in the dashboard is generated and sent to the coordination the task user belongs to. 

## How to test

You need a user that has a coordination with an email address. 
With this user, try a bulk import task containing submissions that are considered confirmation tests (see `plugins/trypelim/constants.py` `CONFIRMATION_TEST_FORM_IDS`). 
An email should be sent to the coordination. 

## Print screen / video

/ 

## Notes

- Don't hesitate to nitpick file locations, I wasn't sure of the best place to put new fixtures for instances.
- Is the IASO entities page the correct destination for the URL generated in the email?
- Haven't tested 'manually' the import of new cases and that the generated URL works.

### DIFF
--- a/iaso/tasks/process_mobile_bulk_upload.py
+++ b/iaso/tasks/process_mobile_bulk_upload.py
@@ -16,6 +16,7 @@ import time
 import uuid
 import zipfile
 
+from collections import defaultdict
 from copy import copy
 from datetime import datetime
 from traceback import format_exc
@@ -39,6 +40,7 @@ from plugins.trypelim.common.form_utils import (
     get_population_instances,
 )
 from plugins.trypelim.common.utils import sns_notify
+from plugins.trypelim.import_export.bulk_upload import notify_coordinations, positive_instance_qs
 
 
 INSTANCES_JSON = "instances.json"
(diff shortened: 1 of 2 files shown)


bramj (approved):
Looking good Philippe! 👍 

> Don't hesitate to nitpick file locations, I wasn't sure of the best place to put new fixtures for instances.

It's also not super clear to me where what should live. If we manage to generalize this for IASO some day it'll become clearer I suppose. For now, I think what you did is reasonable.

> Is the IASO entities page the correct destination for the URL generated in the email?

That would have been okay as well, but I see you ended up picking the submissions search. Both are find and I like how you put it into 1 URL!

> Haven't tested 'manually' the import of new cases and that the generated URL works.

You should be able to test this on staging, you can ask Benjamin to get you an APK for this. (FYI I did not run it locally)

PR #2340, opened 2025-07-31
IA-4280 redesign of the feature flag modal UI for project web

Bewi:

Related JIRA tickets : IA-4280

- [ ] Are there enough tests?

## Changes

DB: 
- Adding an order property to feature flag to sort them by that property instead of by name
- Adding a category to be able to group them when needed
- Adding a is_dangerous flag to know which one are risky when enabled / disabled 
- Migration to initialize those properties based on excel file provided I the story

FE:
- Grouping feature flags per category
- Toggle feature flag category
- Display a warning icon instead of question mark for dangerous feature flags

## How to test

Going to Projects (dashboard/settings/projects/accountId/X)
When editing an existing project, go to Options tab. 
To validate: 
 - Feature flags order is the same as specified in excel sheet from the story
 - New lines are added for categories and following feature flags are matching excel file
 - Activating / deactivating a feature flag still works and save applies them properly
 - A Warning message is shown for Specific Options feature flags

## Print screen / video

https://github.com/user-attachments/assets/e032e875-1ac5-4f69-8764-308e70809833

```
fix: empty instance pop up

Refs: IA-3665
```


iaso/models/base.py
```diff
@@ -1817,6 +1827,9 @@ class FeatureFlag(models.Model):
     name = models.CharField(max_length=100, null=False, blank=False)
     requires_authentication = models.BooleanField(default=False)
     description = models.TextField(blank=True)
+    category = models.TextField(choices=FeatureFlagCategory.choices, default=FeatureFlagCategory.NA)
+    order = models.IntegerField(default=0)
+    is_dangerous = models.BooleanField(default=False)
```
quang-le, line 1833: super reassuring filed name 😆

iaso/migrations/0334_set_feature_flag_category.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def reverse_set_feature_flag_category_and_is_dangerous(apps, schema_editor):
+        # We can ignore this, not sure how to say that there is nothing to revert.
+        pass
+
+    operations = [
+        migrations.RunPython(
+            set_feature_flag_category_and_is_dangerous, reverse_set_feature_flag_category_and_is_dangerous
+        )
+    ]
```
quang-le:
```suggestion
    operations = [
        migrations.RunPython(
            set_feature_flag_category_and_is_dangerous, migrations.RunPython.noop,elidable=True
        )
    ]
```

iaso/locale/fr/LC_MESSAGES/django.po
quang-le, line 1: There is a way to send these translations to the front-end, so that you don't have to duplicate them in the json files

iaso/migrations/0336_featureflag_category_featureflag_is_dangerous_and_more.py
quang-le, line 1: You could merge the 2 migration in 1 file. Not blocking but only because I also always forget to do it 😁

hat/assets/js/apps/Iaso/domains/projects/config.tsx
```diff
@@ -116,9 +119,26 @@ export const useFeatureFlagColumns = (
...
+                            onClick={() =>
+                                toggleFeatureGroup(settings.row.original.code)
+                            }
+                        >
+                            {settings.row.original.collapsed ? (
+                                <ExpandMore />
+                            ) : (
+                                <ExpandLess />
+                            )}
+                        </Button>
```
quang-le: That's a lot of ternaries. I would move that to a separate component and make the code easier to read, even if it gets more verbose

hat/assets/js/apps/Iaso/domains/projects/components/FeatureFlagsSwitches.tsx
```diff
@@ -53,17 +53,60 @@ export const FeatureFlagsSwitches: React.FunctionComponent<Props> = ({
...
     const columns = useFeatureFlagColumns(
         setFeatureFlag,
+        toggleFeatureGroup,
         projectFeatureFlagsValues,
     );
 
+    const featureFlagsData = useMemo(() => {
+        const groupRows: string[] = [];
+        // Assuming order from Back end is related to groups as well ..
+        // Maybe we should make sure of that.
```
quang-le: you could write a test in the backend for that 😁

quang-le: and/or make the api accept an `order`queryParam and pass it when fetching data for this page

hat/assets/js/apps/Iaso/domains/app/translations/fr.json
```diff
@@ -1279,6 +1279,14 @@
     "iaso.projets.featureflag.tooltip.write_on_nfc_cards": "Active la possibilité d'enregistrer les données d'entités sur une carte NFC qui se scanne contre l'appareil",
     "iaso.projets.featureflag.transport_tracking": "GPS des trajets",
     "iaso.projets.featureflag.write_on_nfc_cards": "Utiliser la carte NFC pour enregistrer l'entité",
+    "iaso.projets.featureflag.category.DCO": "Options de collection de données",
```
tdethier:
```suggestion
    "iaso.projets.featureflag.category.DCO": "Options de collecte de données",
```

quang-le (changes requested):
The translation thing we can check together.

- Please update the API and the front-end to enforce ordering + add test in the backend
- The confirmation for activation of dangerous flags is missing
- Some tooltip messages and translations appear to be missing, pls document them in a separate ticket
- I would open the modal with the groups collapsed, so users can see all categories

iaso/migrations/0336_featureflag_category_featureflag_is_dangerous_and_more.py
Bewi: I had a chat with Thibault about it where it could happens that fields are not yet fully accessible if I put this logic in the same migration file.
Which is why I putted aside :)

iaso/models/base.py
Bewi: I think they really want them to pay attention :D

iaso/models/base.py
```diff
@@ -1817,6 +1827,9 @@ class FeatureFlag(models.Model):
     name = models.CharField(max_length=100, null=False, blank=False)
     requires_authentication = models.BooleanField(default=False)
     description = models.TextField(blank=True)
+    category = models.TextField(choices=FeatureFlagCategory.choices, default=FeatureFlagCategory.NA)
+    order = models.IntegerField(default=0)
```
tdethier: Nitpick: I don't think it makes a big difference, but I would have used
```suggestion
    order = models.PositiveSmallIntegerField(default=0)
```

iaso/api/feature_flags.py
```diff
@@ -49,7 +59,7 @@ def get_queryset(self):
             if "SHOW_MOBILE_NO_ORGUNIT_PROJECT_FEATURE_FLAG" not in account_feature_flags:
                 featureflags = featureflags.exclude(code="MOBILE_NO_ORG_UNIT")
 
-        return featureflags.order_by("name")
+        return featureflags.order_by("order")
```
tdethier: I would not return all objects simply based on `order` because it could break `category` ordering, which is more important than `order`.

hat/assets/js/apps/Iaso/domains/app/translations/fr.json
```diff
@@ -1279,6 +1279,14 @@
...
+    "iaso.projets.featureflag.category.REO": "Options de rafraichissement",
+    "iaso.projets.featureflag.category.GEO": "Options géographiques",
+    "iaso.projets.featureflag.category.DAV": "Validation de données",
```
tdethier, line 1285: I would either make all of them start with `Options` or none of them in order to be consistent. Either of them works for me

iaso/migrations/0337_set_feature_flag_category.py
```diff
@@ -0,0 +1,98 @@
...
+            "MOBILE_ENTITY_LIMITED_SEARCH": "ENT",
+            "WRITE_ON_NFC_CARDS": "ENT",
+            "PLANNING": "PLA",
+            "TRANSPORT_TRACKING": "SPO",
+            "GPS_TRACKING": "SPO",
+            "MOBILE_HIDE_CLOSED_ORG_UNIT": "SPO",
+            "MOBILE_USE_ETHIOPIC_CALENDAR": "SPO",
+            "MOBILE_NO_ORG_UNIT": "SPO",
+            "REPORTS": "SPO",
+            "MOBILE_ORG_UNIT_DEEP_SEARCH": "NA",
```
tdethier: It's strange that this one was left out 😅

Maybe it didn't exist when the spreadsheet was made

iaso/api/feature_flags.py
tdethier: I will show you a proposal to fix this

tdethier (changes requested):
First of all, the feature works, so that's nice :+1: 

I only checked the backend. I left a few optional comments, but I would definitely:
-  change how ordering is managed, because it can break categories right now
- add tests

iaso/migrations/0336_featureflag_category_featureflag_is_dangerous_and_more.py
tdethier: I think that since you're not creating a new model it should work. I would try, but of course you need to put the `RunPython` operation at the end of the list

hat/assets/js/apps/Iaso/domains/app/translations/fr.json
Bewi: That's not up to me, comes from their excell stuff

iaso/migrations/0337_set_feature_flag_category.py
Bewi: Nice catch, I'll move it as we know it is a geographic option now

Bewi:
> The translation thing we can check together.
> 
> * Please update the API and the front-end to enforce ordering + add test in the backend
> * The confirmation for activation of dangerous flags is missing
> * Some tooltip messages and translations appear to be missing, pls document them in a separate ticket
> * I would open the modal with the groups collapsed, so users can see all categories

* API have been updated to order properly and a test is there to verify that the order keeps also item grouped.
* Confirmation modal will be done in another stage, the most important was to have a visual hint that it is a dangerous flag
* I will document missing tooltips
* Group are now collapsed by default

iaso/locale/fr/LC_MESSAGES/django.po
Bewi: I'll move this to technical debt as it requires quite some changes and this has taken me too long already

hat/assets/js/apps/Iaso/domains/app/translations/fr.json
tdethier: I think it's more of a proposition than a mandatory list to implement

tdethier (approved):
Backend looks good :+1:

quang-le (approved):

PR #2341, opened 2025-07-31
[IA-4409] Add Predefined Filters to Web UI 

bmonjoie:
- Add an endpoint to expose `FormPredefinedFilter`s
 - Add a new tab in the `Form details` page to manage `Predefined filters`

Related JIRA tickets : IA-4409

- [X] Are there enough tests?

## Changes

 - Add an endpoint to expose `FormPredefinedFilter`s
 - Add a new tab in the `Form details` page to manage `Predefined filters`
 - Also, I added a checkbox to surface the `prevent_add_if_duplicate` in `EntityType`

## How to test

Go into the details of a form, select the "Predefined filters" tab and go for it!

## Notes

There is a bug that the JSON logic doesn't display on first load, I don't know why :slightly_frowning_face: 

```
fix: empty instance pop up

Refs: IA-3665
```


plugins/polio/js/src/constants/translations/en.json
```diff
@@ -280,10 +280,10 @@
     "iaso.polio.forms.options.approval_ongoing": "Approval ongoing",
     "iaso.polio.hideTestCampaigns": "Hide test campaigns",
     "iaso.polio.import_file.label": "Excel Line File",
+    "iaso.polio.label.12months": "Last 12 months",
```
tdethier, line 283: I always have this issue too, it feels like the command that checks translations uses a tool that behaves differently in Linux and MacOS :face_with_spiral_eyes:

iaso/api/form_predefined_filters/serializers.py
```diff
@@ -0,0 +1,30 @@
...
+        Serializer for `{FORM_ID}` when passed in query_params.
+
+        Used to handle parsing and errors:
+
+            serializer = FormIdSerializer(data=self.request.query_params)
+            serializer.is_valid(raise_exception=True)
+            form_id = serializer.validated_data[FORM_ID]
+        """
+
+    form_id = serializers.CharField(allow_blank=False)
```
tdethier: Shouldn't it be a `serializers.IntegerField`?

iaso/api/form_predefined_filters/views.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def get_queryset(self, mobile=False):
+        orders = self.request.query_params.get("order", "name").split(",")
+        queryset = FormPredefinedFilter.objects
+        serializer = FormIdSerializer(data=self.request.query_params)
+        form_id = serializer.validated_data[FORM_ID] if serializer.is_valid(raise_exception=False) else None
+        if form_id:
+            form = get_object_or_404(Form.objects, id=form_id)
+            queryset = queryset.filter(form=form)
+        return queryset.order_by(*orders)
```
tdethier: 2 issues here:
- from what I understand, if the `FormIdSerializer` doesn't consider the ID as valid, all the `FormPredefinedFilter` objects are going to be returned, is that the expected behavior? It sounds strange
- I don't see any notion of tenancy, I guess it should be included somehow through `Form`

iaso/api/entity_types.py
```diff
@@ -25,6 +25,7 @@ class Meta:
             "fields_detail_info_view",
             "fields_list_view",
             "fields_duplicate_search",
+            "prevent_add_if_duplicate_found",
```
tdethier, line 28: Are the `iaso/api/entity_types` changes related to this PR?

iaso/tests/api/form_predefined_filters/test_form_predefined_filters_serializers.py
```diff
@@ -0,0 +1,32 @@
...
+class FormPredefinedFilterSerializersTestCase(TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        form = m.Form.objects.create(name="Form", legend_threshold=10)
+        form_predefined_filter = m.FormPredefinedFilter.objects.create(
+            form=form, name="filter", short_name="short", json_logic="{}"
+        )
+
+        cls.form = form
+        cls.form_predefined_filter = form_predefined_filter
```
tdethier: Nitpick: you could do it in one line above

iaso/tests/api/form_predefined_filters/test_form_predefined_filters_serializers.py
```diff
@@ -0,0 +1,32 @@
...
+            "id": self.form_predefined_filter.id,
+            "form_id": self.form.id,
+            "name": "filter",
+            "short_name": "short",
+            "json_logic": "{}",
+            "created_at": self.form_predefined_filter.created_at.timestamp(),
+            "updated_at": self.form_predefined_filter.updated_at.timestamp(),
+        }
+
+        with self.assertNumQueries(0):
```
tdethier: `assertNumQueries` checks how many calls were made to the DB, I don't think it's relevant here

iaso/tests/api/form_predefined_filters/test_form_predefined_filters_views.py
```diff
@@ -0,0 +1,150 @@
...
+        cls.user_with_rights = cls.create_user_with_profile(
+            username="Authorized",
+            account=account,
+            permissions=["iaso_forms"],
+        )
+        cls.user_without_rights = cls.create_user_with_profile(
+            username="Unauthorized",
+            account=account,
+            permissions=[],
+        )
```
tdethier: Nitpick: there's a utility method for that - `create_base_users()`

iaso/tests/api/form_predefined_filters/test_form_predefined_filters_views.py
```diff
@@ -0,0 +1,150 @@
...
+        cls.form1 = form1 = m.Form.objects.create(name="Form1", legend_threshold=10)
+        cls.form2 = form2 = m.Form.objects.create(name="Form2", legend_threshold=10)
+
+        unauthenticated_project.forms.add(form1)
+        unauthenticated_project.forms.add(form2)
+        unauthenticated_project.save()
+
+        authenticated_project.forms.add(form1)
+        authenticated_project.forms.add(form2)
+        authenticated_project.save()
```
tdethier:
```suggestion
        authenticated_project.forms.set([form1, form2])
```

iaso/tests/api/form_predefined_filters/test_form_predefined_filters_views.py
```diff
@@ -0,0 +1,150 @@
...
+            app_id="authenticated",
+            account=account,
+            needs_authentication=True,
+        )
+        cls.form1 = form1 = m.Form.objects.create(name="Form1", legend_threshold=10)
+        cls.form2 = form2 = m.Form.objects.create(name="Form2", legend_threshold=10)
+
+        unauthenticated_project.forms.add(form1)
+        unauthenticated_project.forms.add(form2)
+        unauthenticated_project.save()
```
tdethier:
```suggestion
        unauthenticated_project.forms.set([form1, form2])
```

iaso/tests/api/form_predefined_filters/test_form_predefined_filters_views.py
```diff
@@ -0,0 +1,150 @@
...
+        self.assertJSONResponse(response, 401)
+
+    def test_formpredefinedfilters_list_without_auth_for_project_not_requiring_auth(self):
+        f"""GET {BASE_URL} without auth for project which doesn't requires it: 200"""
+
+        response = self.client.get(BASE_URL, {APP_ID: self.unauthenticated_project.app_id})
+        self.assertJSONResponse(response, 200)
+        self.assertValidFiltersListData(response.json(), 0)
+
+    def test_formpredefinedfilters_list_with_wrong_form_id(self):
```
tdethier: What about an invalid value? (i.e. `form_id="pikachu"`)

iaso/tests/api/form_predefined_filters/test_form_predefined_filters_views.py
```diff
@@ -0,0 +1,150 @@
...
+
+        response = self.client.get(
+            path=BASE_URL,
+            data={"form_id": self.form1.id},
+            format="json",
+        )
+        self.assertJSONResponse(response, 200)
+        self.assertValidFiltersListData(response.json(), 1)
+        self.assertEqual(response.json().get("form_predefined_filters")[0].get("name"), "test")
+        id = response.json().get("form_predefined_filters")[0].get("id")
```
tdethier: Nitpick: I would probably have created intermediary variables to make it easier to read

iaso/tests/api/form_predefined_filters/test_form_predefined_filters_views.py
```diff
@@ -0,0 +1,150 @@
...
+        self.assertJSONResponse(response, 200)
+
+        response = self.client.get(
+            path=BASE_URL,
+            data={"form_id": self.form1.id},
+            format="json",
+        )
+        self.assertJSONResponse(response, 200)
+        self.assertValidFiltersListData(response.json(), 1)
+        self.assertEqual(response.json().get("form_predefined_filters")[0].get("name"), "test2")
```
tdethier: Here, I would also check that all the order fields are still the same

iaso/tests/api/form_predefined_filters/test_form_predefined_filters_views.py
```diff
@@ -0,0 +1,150 @@
...
+
+from iaso import models as m
+from iaso.api.query_params import APP_ID, FORM_ID
+from iaso.test import APITestCase
+
+
+BASE_URL = "/api/formpredefinedfilters/"
+
+
+class FormPredefinedFilterViewsTestCase(APITestCase):
```
tdethier: What about querying stuff that is linked to another account?

tdethier (changes requested):
Feature-wise:
- I'm not quite sure I understand this feature, but it works
- there's a regression: forms list page no longer requires the user to trigger the search before trying to load stuff.
- it's not really human readable - especially dates - but maybe it's not an issue
image

On the backend side:
- I think something is missing tenancy-wise (+ tests), so I'm requesting changes
- there are a few nitpicks here and there

iaso/api/entity_types.py
bmonjoie: It's a small change a smuggled into this PR because it bothered me not to have this

iaso/tests/api/form_predefined_filters/test_form_predefined_filters_serializers.py
bmonjoie: I know. I wanted to have this in the tests so that if we ever change that (let's say, by adding a relationship) the test could potentially fail and we would have to make a conscious choice about it.

tdethier: I understand the reason why, but still, this is not the place to do it (I don't think `assertEqual` is going to trigger this anyway). This issue is not related to (de)serializing stuff.

Doing this in the `views` tests makes more sense. There are already some examples of that:
- `OrgUnitChangeRequestAPITestCase.test_list_ok`
- `DataSourceVersionsSynchronizationModelTestCase.test_create_change_requests`

iaso/tests/api/form_predefined_filters/test_form_predefined_filters_views.py
```diff
@@ -0,0 +1,164 @@
...
+
+from iaso import models as m
+from iaso.api.query_params import APP_ID, FORM_ID
+from iaso.tests.tasks.task_api_test_case import TaskAPITestCase
+
+
+BASE_URL = "/api/formpredefinedfilters/"
+
+
+class FormPredefinedFilterViewsTestCase(TaskAPITestCase):
```
tdethier, line 11: I would add a test for fetching things from the wrong account. For instance user of account 1 queries with a valid form ID that is inside account 2.

iaso/api/form_predefined_filters/views.py
```diff
@@ -0,0 +1,78 @@
...
+                raise exceptions.NotFound(f"Project not found for {app_id}")
+            queryset = queryset.filter(form__projects__app_id=app_id)
+        elif self.request.user.is_anonymous:
+            raise exceptions.NotAuthenticated
+        else:
+            profile = self.request.user.iaso_profile
+            queryset = queryset.filter(form__projects__account=profile.account)
+
+        # We don't send filters for deleted forms
+        queryset = queryset.filter(form__deleted_at=None)
```
tdethier: Nitpick: you can already put that filter when you declare the queryset on line 37

iaso/api/form_predefined_filters/views.py
```diff
@@ -0,0 +1,78 @@
...
+
+        serializer = FormIdSerializer(data=self.request.query_params)
+        form_id = serializer.validated_data[FORM_ID] if serializer.is_valid(raise_exception=True) else None
+        if form_id:
+            form = get_object_or_404(Form.objects, id=form_id)
```
tdethier: I think that passing `Form.objects` instead of a tenancy-filtered queryset might cause some issue, especially with the anonymous user stuff (hence the test I'm asking)

iaso/api/form_predefined_filters/views.py
```diff
@@ -0,0 +1,78 @@
...
+        if app_id is not None:
+            try:
+                Project.objects.get_for_user_and_app_id(user=self.request.user, app_id=app_id)
+            except Project.DoesNotExist:
+                if self.request.user.is_anonymous:
+                    raise exceptions.NotAuthenticated
+                raise exceptions.NotFound(f"Project not found for {app_id}")
+            queryset = queryset.filter(form__projects__app_id=app_id)
+        elif self.request.user.is_anonymous:
+            raise exceptions.NotAuthenticated
```
tdethier, line 50: Should anonymous users be able to fetch form predefined filters if they provide a correct `app_id`?

tdethier (changes requested):
There are a few things I would still change in tests & `get_queryset`

iaso/api/form_predefined_filters/views.py
bmonjoie: I got confused by this as well when I copied it from `FormVersion` but this is in the `elif` of `if app_id is not None` so, it won't be triggered if an `app_id` is provided.

hat/assets/js/apps/Iaso/domains/forms/components/PredefinedFilterModal.tsx
```diff
@@ -0,0 +1,195 @@
...
+                            onChange={(result: JsonLogicResult) => {
+                                let parsedValue;
+                                if (result?.logic && fields)
+                                    parsedValue = parseJson({
+                                        value: result.logic,
+                                        fields,
+                                    });
+                                setFieldTouched('json_logic');
+                                setFieldValue('json_logic', parsedValue);
+                            }}
```
beygorghor: This should be part of a useCallback

hat/assets/js/apps/Iaso/domains/forms/hooks/useSaveFormPredefinedFilter.ts
```diff
@@ -0,0 +1,34 @@
...
+    }
+    return patchRequest(`/api/formpredefinedfilters/${filter.id}/`, {
+        name: filter.name,
+        short_name: filter.short_name,
+        json_logic: filter.json_logic,
+    });
+};
+
+export const useSaveFormPredefinedFilter = (): UseMutationResult => {
+    return useSnackMutation<any, any, any, any>({
```
beygorghor, line 27: not mandatory but we have some example of correct typing here...
```
export const useLaunchDiff = () => {
    return useSnackMutation<SyncResponse, Error, SyncParams>({
        mutationFn: ({ id }) => launchDiff({ id }),
        showSucessSnackBar: false,
    });
};

```

iaso/api/form_predefined_filters/views.py
tdethier: I'm asking this because there's an identical check for that inside the `except`  when the project can't be found.

If it was me, I would put a single check for anonymous users at the very beginning of this method and get this out of the way. But doing this prevents them from accessing data if try provide a correct `app_id` - and I'm not sure if we want that behavior and if it's applicable to any use cases. I'd say no auth -> you get nothing, but maybe that's not what we want :sweat_smile:

bmonjoie: It's because some projects do not require authentication. The check for that flag is done in `Project.objects.get_for_user_and_app_id`:
https://github.com/BLSQ/iaso/blob/main/iaso/models/project.py#L28

tdethier: Alright :man_shrugging:

tdethier (approved):
Backend looks good, thanks for the changes :+1: 

@beygorghor up to you for the frontend checks

beygorghor (approved):
I did some changes to make it work properly
- title was not correct while adding a new item (id was not correctly passed to dialog)
- validation was always true, due to empty string or object
- change dialog width regarding used tab
- added a third tab "general" to allow dialog resize

PR #2342, opened 2025-08-01
[SNT-106] Snt malaria data models

Bewi:
Data model from branch ant-malaria to merge into main.

Related JIRA tickets : SNT-106

- [ ] Are there enough tests?

## Doc

## Changes

New entities for snt-malaria
Json logic updates to cover filters from snt malaria
Open hexa config

## How to test

Migrate the data and make sure Iaso is running properly (that nothing breaks).
Those changes should only have impact on ant-malaria at the moment.

For SNT-Malaria, you can try a config from scratch (new db) and check that everything works properly.

## Notes

```
fix: empty instance pop up

Refs: IA-3665
```


iaso/api/metrics/views.py
```diff
@@ -0,0 +1,72 @@
...
+        group_by_field_name = "org_unit_id"
+        if not json_filter:
+            return queryset.values(group_by_field_name).distinct().values_list(group_by_field_name, flat=True)
+
+        q = jsonlogic_to_exists_q_clauses(
+            json.loads(json_filter), MetricValue.objects, "metric_type_id", group_by_field_name
+        )
+
+        qs = queryset.filter(q).values_list(group_by_field_name, flat=True).distinct()
+        return qs
```
tdethier: You can also put them in a dedicated `filters.py` file, like you did with serializers

iaso/models/metric.py
```diff
@@ -0,0 +1,46 @@
...
+    account = models.ForeignKey("iaso.Account", on_delete=models.CASCADE)
+    name = models.CharField(max_length=255)
+    code = models.CharField(max_length=255)
+    description = models.TextField(blank=True)
+    source = models.CharField(max_length=255, blank=True)
+    units = models.CharField(max_length=255, blank=True)
+    unit_symbol = models.CharField(max_length=255, blank=True)
+    category = models.CharField(max_length=255, blank=True)
+    comments = models.TextField(blank=True)
+    legend_type = models.CharField(choices=[(t, t) for t in LEGEND_TYPES], max_length=40, default="threshold")
```
tdethier: Nitpick: this works, but if the backend ever needs to prepare a translation of all legend types, it's going to be difficult. Using a nested enum class will make it easier. You don't need to change anything if translations and not planned

iaso/models/metric.py
```diff
@@ -0,0 +1,46 @@
...
+
+LEGEND_TYPES = ["threshold", "linear", "ordinal"]
+
+
+class MetricType(models.Model):
+    class Meta:
+        ordering = ["id"]  # force ordering in order of creation (for demo)
+        unique_together = [
+            ["account", "code"],
+        ]
```
tdethier, line 17: Nitpick: we usually put the `class Meta` below the attributes and above methods. You don't need to move this, but I'm surprised that `ruff` didn't complain about it

iaso/api/metrics/views.py
```diff
@@ -0,0 +1,72 @@
...
+    It is used in the frontend to display the org units for a given metric type.
+    """
+
+    serializer_class = OrgUnitIdSerializer
+    queryset = MetricValue.objects.all()
+    filter_backends = [DjangoFilterBackend, ValueAndTypeFilterBackend]
+    filterset_fields = ["metric_type_id"]
+
+    def get_queryset(self):
+        return MetricValue.objects.filter(metric_type__account=self.request.user.iaso_profile.account)
```
tdethier, line 44: Both of these feel very similar. I don't know the business logic in SNT, but is there a reason why there are 2 viewsets instead of 1?

iaso/models/metric.py
```diff
@@ -0,0 +1,46 @@
...
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+
+    def __str__(self):
+        return "%s %s" % (self.name, self.id)
+
+
+class MetricValue(models.Model):
+    class Meta:
+        unique_together = [["metric_type", "org_unit", "year"]]
```
tdethier, line 44: Same comment about `Meta`

iaso/models/metric.py
```diff
@@ -0,0 +1,46 @@
...
+
+    metric_type = models.ForeignKey(MetricType, on_delete=models.PROTECT)
+    org_unit = models.ForeignKey(OrgUnit, null=True, blank=True, on_delete=models.PROTECT)
+    year = models.IntegerField(null=True, blank=True)
+    value = models.FloatField(null=True)
+    string_value = models.TextField(null=True)
```
tdethier: The convention in Django is to not have `None` when a `str` attribute does not have any value, but rather use `""` (empty string).

```suggestion
    string_value = models.TextField(blank=True)
```

iaso/models/metric.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+class MetricValue(models.Model):
+    class Meta:
+        unique_together = [["metric_type", "org_unit", "year"]]
+
+    metric_type = models.ForeignKey(MetricType, on_delete=models.PROTECT)
+    org_unit = models.ForeignKey(OrgUnit, null=True, blank=True, on_delete=models.PROTECT)
+    year = models.IntegerField(null=True, blank=True)
+    value = models.FloatField(null=True)
```
tdethier: Otherwise the admin panel will require you to enter a value here

```suggestion
    value = models.FloatField(null=True, blank=True)
```

iaso/utils/jsonlogic.py
```diff
@@ -388,6 +388,75 @@ def _handle_comparison_operators(op: str, params: list, field_prefix: str) -> Q:
...
+            {"<=":[{"var":"23"},1500]},
+            {"==":[{"var":"24"},1000]}
+        ]}
+    ]
+    :param entities: The Django model manager for the entities to filter. Should be a QuerySet or Manager.
+    :param id_field_name: The name of the field in the entities that corresponds to the "var" in the JsonLogic query.
+    :param group_by_field_name: The name of the field used to group the entities in the Exists clause.
+    :return: A Django Q object.
+    """
+    print(f"jsonlogic_to_exists_q_clauses: {jsonlogic}")
```
tdethier: Or use a logger if it's really important

```suggestion
```

iaso/utils/jsonlogic.py
```diff
@@ -388,6 +388,75 @@ def _handle_comparison_operators(op: str, params: list, field_prefix: str) -> Q:
...
+        return Q()
+
+    op = list(jsonlogic.keys())[0]
+    params = jsonlogic[op]
+
+    field_position = 1 if op == "in" else 0
+    field = params[field_position]
+    value = params[0] if op == "in" else params[1]
+    # This break flexibility, maybe we can two parameters "value_field_name" and "string_value_field_name"
+    value_field_name = "value" if floatTryParse(value) else "string_value"
```
tdethier: What about `int` or other data types?

iaso/tests/utils/test_jsonlogic.py
```diff
@@ -1,8 +1,105 @@
...
+            'EXISTS(SELECT 1 AS "a" FROM "iaso_metricvalue" U0 WHERE (U0."metric_type_id" = 23 AND U0."org_unit_id" = ("iaso_metricvalue"."org_unit_id") AND U0."value" <= 1500.0) LIMIT 1) '
+            "AND "
+            'EXISTS(SELECT 1 AS "a" FROM "iaso_metricvalue" U0 WHERE (U0."metric_type_id" = 24 AND U0."org_unit_id" = ("iaso_metricvalue"."org_unit_id") AND U0."value" = 1000.0) LIMIT 1)'
+            ")"
+            ")"
+        )
+        q = jsonlogic_to_exists_q_clauses(filters, MetricValue.objects, "metric_type_id", "org_unit_id")
+        querySet = MetricValue.objects.filter(q)
+        self.assertEqual(str(querySet.query), expectedQuerySet)
+
```
tdethier, line 118: Honestly, I didn't fully understand the whole `jsonlogic_to_exists_q_clauses` function, but I would add more tests because some cases do not seem to be covered:
- one with a `!` at the beginning of the `jsonlogic`
- one with a `in` operator
- one with `string_value`

iaso/utils/jsonlogic.py
```diff
@@ -388,6 +388,75 @@ def _handle_comparison_operators(op: str, params: list, field_prefix: str) -> Q:
...
+    :param entities: The Django model manager for the entities to filter. Should be a QuerySet or Manager.
+    :param id_field_name: The name of the field in the entities that corresponds to the "var" in the JsonLogic query.
+    :param group_by_field_name: The name of the field used to group the entities in the Exists clause.
+    :return: A Django Q object.
+    """
+    print(f"jsonlogic_to_exists_q_clauses: {jsonlogic}")
+    if "and" in jsonlogic:
+        sub_query = Q()
+        for lookup in jsonlogic["and"]:
+            sub_query = operator.and_(
```
tdethier, line 415: TIL `operator.and_()`

tdethier (changes requested):
A few comments:
- it looks like this doesn't break the standard app :+1: 
- I didn't test that on SNT as I don't have a local SNT setup - but I imagine that you have a similar PR in SNT (+ more changes to replace old SNT imports by imports from iaso classic)
- you will need to rebase `main` because of usual migrations issues
- the whole `jsonlogic_to_exists_q_clauses()` function was not clear to me, but maybe it's because I lack some context
- the other comments are optional, but I'm requesting changes for:
  - a few things at the model level
  - more tests

iaso/api/metrics/views.py
Bewi: Yes, it is using a different filter_backends, one of them accept a json logic that will be applied on value only
Other one will accept a json logic that will be applied to value and type

iaso/models/metric.py
Bewi: Allright, this has been changed and migration aligned

iaso/utils/jsonlogic.py
Bewi: Not covered atm

iaso/tests/utils/test_jsonlogic.py
Bewi: Added a small test with !.
But the thing is that we don't cover in and not (at root level) operators for current usage.
This should be enriched later, ain't gonna be a 2 minute job.

And for the string_value, there already is one: 
test_jsonlogic_to_exists_q_clauses__simple_string_value

iaso/utils/jsonlogic.py
```diff
@@ -449,3 +517,11 @@ def entities_jsonlogic_to_q(jsonlogic: Dict[str, Any], field_prefix: str = "") -
 
 def matches_all(field_value, expected_list):
     return sorted(field_value.split()) == sorted(expected_list)
+
+
+def floatTryParse(value):
```
tdethier:
```suggestion
def float_try_parse(value):
```

iaso/models/metric.py
```diff
@@ -0,0 +1,55 @@
...
+    source = models.CharField(max_length=255, blank=True)
+    units = models.CharField(max_length=255, blank=True)
+    unit_symbol = models.CharField(max_length=255, blank=True)
+    category = models.CharField(max_length=255, blank=True)
+    comments = models.TextField(blank=True)
+    legend_type = models.CharField(
+        choices=[(tag.value, tag.value) for tag in LegendType],
+        max_length=40,
+        default=LegendType.THRESHOLD.value,
+    )
```
tdethier: Sorry my previous comment wasn't clear enough. I meant we should do something like `OrgUnitChangeRequest.kind` or `OrgUnitChangeRequest.status`. This way we can have a translation for each option (see the `_("something"`) part of each possible value.

This issue with this current version is that again we can't properly provide a translation

tdethier: If you change this, you'll also need to run the translation script, because it will add new translations

tdethier (changes requested):
I would just refactor the `MetricType.legend_type` (sorry that my previous message was not clear enough), otherwise it's good

tdethier (approved):
LGTM :+1:

PR #2344, opened 2025-08-01
WC2-581 Improve performance of the Levenshtein algo for entity duplicates

kemar:
Improve performance of the Levenshtein algo for entity duplicates.

Related JIRA tickets : WC2-581

## Changes

I tried a lot of different ways to improve the performance of this code.

But Levenshtein operations are not index-optimisable:

- Levenshtein distance requires full string comparison between every pair
- no index can pre-compute similarity scores between arbitrary string pairs
- GIN/GiST indexes on JSON don't support fuzzy string matching

At the end:

- I used batches to prevent the query to consume an enormous amount of memory
- I added an index to avoid a full scan of the `Entity` table

The query should now both perform a (little) better and NOT consume an enormous amount of memory.

Here is what the query is doing:

- CTE 1: filtered_entities

    - gets all non-deleted entities of a specific type
    - excludes soft-deleted records (`deleted_at IS NULL`)
    - there is a new index to speed up this step

- CTE 2: entity1_batch

    - creates a batch of entities for comparison
    - uses ID range filtering (`id BETWEEN %s AND %s`) for performance optimization in large datasets

- CTE 3: entity_pairs

    - creates pairs of entities to compare for duplicates
    - `entity1.id` > `entity2.id` ensures each pair is only compared once (avoids duplicate comparisons)
    - joins with `iaso_instance` tables to access the actual attribute data (JSON) for comparison
    - calculates a similarity score using {fields_comparison} (Levenshtein distance calculations)
    - excludes pairs that are already marked as duplicates in `iaso_entityduplicate`

- CTE 4: scored_pairs

    - normalizes the raw score to a 0-100 percentage scale
    - uses `GREATEST(LEAST(...), 0)` to clamp scores between 0-100

- final SELECT

    - returns entity pairs with similarity scores above a threshold to focus on likely duplicates


tdethier (approved):
I tested this on an account that has 125k entities (3 types split evenly). It took a while to search for duplicates based on 4 fields, but it did not crash, so that's a clear improvement. 

Kudos on this PR, it was a difficult one :+1:

PR #2347, opened 2025-08-05
IA-4388 Fix n+1 query on `/api/microplanning/plannings`

kemar:
Fix n+1 query on `/api/microplanning/plannings`.

Related JIRA tickets : IA-4388

[Jira issue](https://bluesquareorg.sentry.io/issues/6746773431/?project=5530884).

### DIFF
--- a/iaso/api/microplanning.py
+++ b/iaso/api/microplanning.py
@@ -381,7 +381,9 @@ class PlanningViewSet(AuditMixin, ModelViewSet):
 
     def get_queryset(self):
         user = self.request.user
-        return self.queryset.filter_for_user(user)
+        return (
+            self.queryset.filter_for_user(user).select_related("project", "org_unit", "team").prefetch_related("forms")
+        )
 
 
 class AssignmentSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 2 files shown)


tdethier (commented):
This is definitely better, but I still see repeated queries in django debug toolbar, do you think we could get rid of them? I'm not sure why this check on org units in the user profile is called 3 times, I don't really see it in the code

image

kemar:
> This is definitely better, but I still see repeated queries in django debug toolbar, do you think we could get rid of them? I'm not sure why this check on org units in the user profile is called 3 times, I don't really see it in the code

@tdethier I tried a lot of things to get rid of those extra queries but none worked.

This is due to the way the serializer has been implemented:

https://github.com/BLSQ/iaso/blob/36af8b0d11d48a973698c2616aacdff4cf84fb7d/iaso/api/microplanning.py#L264-L267

There must be a good reason for this.

Anyway, even with a few extra queries, the global execution time is largely acceptable so I would suggest to merge the PR as-is.

tdethier (approved):
Fine by me, it's not worth - _at the moment_ - refactoring the whole endpoint in order to change the serializer

PR #2349, opened 2025-08-05
IA-4390 Fix n+1 query on `/api/instances`

kemar:
Fix n+1 query on `/api/instances`.

Related JIRA tickets : IA-4390

[Jira issue](https://bluesquareorg.sentry.io/issues/6748372064/?project=5530884).


tdethier (approved):
This one is so much better, we should definitely see an improvement :+1:

PR #2353, opened 2025-08-05
POLIO-1983: Add LQAS models

quang-le:
Add models and commands to convert lqas json data to DB entries

Related JIRA tickets :  POLIO-1983, POLIO-1971

- [x] Are there enough tests?

## Doc
Doc strings

## Changes

- Add models to store lqas json data in DB
- Add LQasDataManager to handle json to DB conversion
- Add commands to
    - manually "upload" a json file to DB
    - convert data from existing `JsonDataStore` to DB (without erasing existing data)

## How to test

- Get one or several lqas JSonDataStore from polio staging
- Add them to your local DB
- run the conversion command

## Print screen / video

N/A


plugins/polio/models/lqas_im.py
```diff
@@ -0,0 +1,203 @@
...
+                fields=["round", "subactivity"],
+                name="unique_lqasabsencestats_round_subactivity",
+                condition=models.Q(subactivity__isnull=False),
+            ),
+            models.UniqueConstraint(
+                fields=["round"],
+                name="unique_lqasabsencestats_round_no_subactivity",
+                condition=models.Q(subactivity__isnull=True),
+            ),
+        ]
```
kemar: Very roughly, this is what I understand and I think the models could be simpler by getting rid of the abstract ones:

model_720

Also merging some models would allow simpler queries by avoiding unnecessary joins, especially when gathering stats.

iaso/models/org_unit.py
```diff
@@ -215,14 +222,15 @@ def filter_for_user_and_app_id(
...
-                .values_list("id", flat=True)
-                .distinct()
-            )
-            queryset = queryset.filter(version_id__in=version_ids)
+            # version_ids = (
+            #     SourceVersion.objects.filter(data_source__projects__account=account)
+            #     .values_list("id", flat=True)
+            #     .distinct()
+            # )
+            # queryset = queryset.filter(version_id__in=version_ids)
```
kemar:
```suggestion
```

plugins/polio/api/lqas_im/lqas_data_manager.py
```diff
@@ -0,0 +1,477 @@
...
+    def parse_json_and_create_lqas_activities(self, lqas_json_data):
+        data_to_parse = lqas_json_data["stats"]
+        rounds, districts = self._prefetch_rounds_and_districts(lqas_json_data)
+        total_created = 0
+        total_failed = 0
+        total_parsed = 0
+        districts_not_found = []
+        rounds_not_found = []
+        for obr_name, campaign_data in data_to_parse.items():
+            for round_data in campaign_data.get("rounds", []) or []:
```
kemar, line 117: The last `or []` is not necessary since you already fallback to a list in the `get` method:

```suggestion
            for round_data in campaign_data.get("rounds", []):
```

This appears elsewhere, but since it's a nitpick, I'm only mentioning it once here.

plugins/polio/api/lqas_im/lqas_data_manager.py
```diff
@@ -0,0 +1,477 @@
...
+                except IntegrityError as e:
+                    self.logger.warning(e)
+                    total_failed += 1
+                    total_parsed += 1
+                    continue
+                except ValidationError as e:
+                    self.logger.warning(e)
+                    total_failed += 1
+                    total_parsed += 1
+                    continue
```
kemar: You can save some lines here (nitpick, only mentioned once):

```suggestion
                except (IntegrityError, ValidationError) as e:
                    self.logger.warning(e)
                    total_failed += 1
                    total_parsed += 1
                    continue
```

plugins/polio/api/lqas_im/lqas_data_manager.py
```diff
@@ -0,0 +1,477 @@
...
+            raise e
+
+    def _update_caregiver_stats(self, lqas_entry: LqasEntry, district_data):
+        caregiver_data = district_data.get(LqasEntry.JSON_KEYS["care_giver_stats"], None)
+        if caregiver_data:
+            best_info_source_data = [
+                item
+                for item in caregiver_data.items()
+                if item[0] not in ["ratio", "caregivers_informed", "caregivers_informed_ratio"]
+            ]  # TODO Test thoroughly
```
kemar: So, did you test it thoroughly? 😃

kemar (changes requested):
The code that converts the JSON to Django model is a bit verbose but there isn't really any other solution. It looks defensive enough and well covered by tests.

My main concern is about the models.

I would advice using a simpler model structure because it would require less joins in a couple of use cases (e.g., gathering stats).

Of course this is not mandatory as it would work as-is, but if you too think that's it worth it, let's simplify the models.

plugins/polio/api/lqas_im/lqas_data_manager.py
quang-le: Actually I need this to check for both the presence of the field, then its value. The json sometimes has `field:null`

quang-le:
@kemar you'll need to migrate back to polio 0237 and apply the updated migration

Thanks for the review, the code is a lot simpler and less bloated now

plugins/polio/api/lqas_im/lqas_data_manager.py
quang-le: of course 😛

plugins/polio/tests/models/test_lqas_district_data.py
```diff
@@ -0,0 +1,314 @@
...
+            total_sites_visited=8,
+            cg_ratio=0.80,
+            cg_best_info_source="Community Worker",
+            cg_best_info_ratio=0.85,
+            cg_caregivers_informed=60,
+            cg_caregivers_informed_ratio=0.80,
+        )
+        self.assertEqual(district_data2.district, district2)
+
+    def test_lqas_district_data_unique_constraint_round_district(self):
```
kemar, line 201: Nice job testing unicity constraints 👍

kemar (approved):
LGTM, these models are way simpler to understand!

PR #2372, opened 2025-08-14
IA-4426 Fix n+1 query on `/api/orgunittypes`

kemar:
Fix n+1 query on `/api/orgunittypes`.

Related JIRA tickets: IA-4426

[Jira issue](https://bluesquareorg.sentry.io/issues/6748465481/?project=5530884).

This one was triggered when calling e.g.:

```
/api/orgunittypes/?app_id=my_app_id&fields=id,name,short_name,sub_unit_types,created_at,updated_at
```

Note that this will not completely fix the problem because the serializer method is operating on a recursive data structure.


tdethier (approved):
I wish we could simply remove this deprecrated endpoint instead of spending time on it :sweat_smile: 

LGTM

PR #2373, opened 2025-08-14
IA-3684: Create OU from the children page of an existing OU

beygorghor:
At the moment, in the App SSC usecase we regularly need to manually create new villages or SSC within the pyramid. In order to avoid creating duplicates with slightly different names, it is systematically necessary to start by searching existing OUs to make sure that the OU-to-be-created doesn’t already exist. Once this has been checked, the user currently has to go back to the first page of the OU list search in order to access the create new OU button. It would be very helpful and save a lot of time to managers if there could be a “create” button visible when on the children tab of an existing OU. 

Creating a new OU from that place would by default create a new OU having the existing one as its parent OU, and then the user will have to specify the name/status/OUT before being able to save.

Related JIRA tickets : IA-3684

- [ ] Are there enough tests?

## Doc

-

## Changes

- add a button to open a new tab to create a child org unit on children tab
- new param parentOrgUnitId to pref ill org unit tree while creating an org unit

## How to test

Open the detail of an org unit. 
Go to children tab
Click on add child button, it should open a new tab to create an org unit with previous org unit as parent

## Print screen / video

https://github.com/user-attachments/assets/838f1169-8a51-4302-a815-9281c14016b0

## Notes

-

```
fix: empty instance pop up

Refs: IA-3665
```


Bewi (approved):
Tested and looks good, maybe a small icon for the button, feels weird that it is the only one "icon less", feels like an orphan :D 

You'll need to rebase if you merge the other PR first, as it contains translations changes as well.

Anyway, LGTM

PR #2379, opened 2025-08-19
Set GDAL environment variables for building Fiona

kemar:
Set GDAL environment variables for building Fiona


kemar:
This was an attempt to fix a problem with fiona but it wasn't successful.

PR #2383, opened 2025-08-19
[IA-4141] Fix file paths in local setup

tdethier:
Fix file paths in local setup

Related JIRA tickets : IA-4141

- [ ] Are there enough tests?

## Doc

/

## Changes

- update to the ngrok documentation
- replaced some `as_dict` by a serializer, to have DRF build the proper file URL
- added some context to manually called serializers, for the same reason

- impacted endpoints:
  - GET /api/formversions/
  - GET /api/mobile/forms/
  - GET /api/mobile/instances/<instance_id>/attachments/  
  - GET /api/mobile/orgunits/<orgunit_id>/reference_instances/?app_id=<app_id>

## How to test

1. Web tests:
  - Start your local instance without ngrok
  - Go to the forms page and try to download any as XML/XLSX :arrow_right: you should be able to download them
  - Start ngrok
  - Go to the forms page while browsing through the ngrok url (not localhost) :arrow_right: you should be able to download files as well

2. Mobile tests: 
  - Start ngrok
  - Connect your mobile app through ngrok :arrow_right: you should be able to download master data (including form definitions) without errors

## Print screen / video

/

## Notes

:warning: there are still some places where `as_dict` is called for returning file URLs (see `Instance`). However, they shouldn't be called by the mobile app and these calls work in the frontend, so I did not replace them

```
fix: empty instance pop up

Refs: IA-3665
```

### DIFF
--- a/iaso/models/forms.py
+++ b/iaso/models/forms.py
@@ -3,6 +3,7 @@ import typing
 
 from uuid import uuid4
 
+from django.conf import settings
 from django.contrib.auth.models import AnonymousUser, User
 from django.contrib.postgres.fields import ArrayField
 from django.core.files.uploadedfile import SimpleUploadedFile
@@ -334,8 +335,8 @@ class FormVersion(models.Model):
         return {
             "id": self.id,
             "version_id": self.version_id,
-            "file": self.file.url,
-            "xls_file": self.xls_file.url if self.xls_file else None,
+            "file": settings.FILE_SERVER_URL + self.file.url,
+            "xls_file": settings.FILE_SERVER_URL + self.xls_file.url if self.xls_file else None,
             "created_at": self.created_at.timestamp() if self.created_at else None,
             "updated_at": self.updated_at.timestamp() if self.updated_at else None,
         }
--- a/iaso/models/forms.py
+++ b/iaso/models/forms.py
@@ -3,7 +3,6 @@ import typing
 
 from uuid import uuid4
 
-from django.conf import settings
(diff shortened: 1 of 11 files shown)


iaso/models/forms.py
```diff
@@ -328,8 +329,8 @@ def as_dict(self):
         return {
             "id": self.id,
             "version_id": self.version_id,
-            "file": self.file.url,
-            "xls_file": self.xls_file.url if self.xls_file else None,
+            "file": settings.FILE_SERVER_URL + self.file.url,
+            "xls_file": settings.FILE_SERVER_URL + self.xls_file.url if self.xls_file else None,
```
bmonjoie: There are other places where this is needed, like the attachments.

bmonjoie:
As discussed, according to the [documentation](https://docs.djangoproject.com/en/dev/topics/files/), `file.url` **should** return the full url:
> ```python
>car = Car.objects.get(name="57 Chevy")
>car.photo.path
>'/media/cars/chevy.jpg'
> car.photo.url
>'https://media.example.com/cars/chevy.jpg'
>```

 So, this is not normal we have to do this.

Form the forms' manifests, we use `request.build_absolute_uri(path)` which gives the full URL.

I think we should fix the root cause of why `.url` doesn't provide the full URL instead of adding this workaround everywhere

beygorghor (approved):
This is very cool !

bmonjoie (approved):
LGTM! :+1:

PR #2384, opened 2025-08-19
[IA-3736] Clean up tmp files after running tests

tdethier:
Clean up tmp files after running tests

Related JIRA tickets : IA-3736

- [x] Are there enough tests?

## Doc

/

## Changes

- Tmp polio files are no longer staying on your disc after running tests

## How to test

- Run polio tests
- The `calendar_2022-10-01` file will no longer appear and bother you :smiling_imp: 

## Print screen / video

/

## Notes

/

```
fix: empty instance pop up

Refs: IA-3665
```

### DIFF
--- a/plugins/polio/tests/api/campaigns/test_campaigns_2.py
+++ b/plugins/polio/tests/api/campaigns/test_campaigns_2.py
@@ -1,4 +1,5 @@
 import datetime
+import os
 
 from typing import List
 from unittest import mock, skip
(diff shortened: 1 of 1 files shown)


kemar (approved):
This works so I approve the PR.

Note that you can also use `tearDown` or `teardownClass` ([like this](https://github.com/BLSQ/iaso/blob/27d90bee7f11826697a330c507aebd81fc8c49de/plugins/polio/tests/models/test_notification_models.py#L29-L35)) so that your files are removed even if a test is failing.

Another way to test files is to not write them on disk, [via e.g. `InMemoryStorage`](https://docs.djangoproject.com/en/5.2/ref/files/storage/#django.core.files.storage.InMemoryStorage).

tdethier:
> Note that you can also use tearDown or teardownClass ([like this](https://github.com/BLSQ/iaso/blob/27d90bee7f11826697a330c507aebd81fc8c49de/plugins/polio/tests/models/test_notification_models.py#L29-L35)) so that your files are removed even if a test is failing.

I initially used that, but since the file has a different name depending on round dates, I switched to something else

> Another way to test files is to not write them on disk, [via e.g. InMemoryStorage](https://docs.djangoproject.com/en/5.2/ref/files/storage/#django.core.files.storage.InMemoryStorage).

I agree, I think that's what should have been used in the API endpoint (it's not the test that's writing files), but I wasn't sure about rewriting it (strange polio things, you know) :sweat_smile:

PR #2385, opened 2025-08-20
IA-3147 Org unit types deletion

beygorghor:
Frontend allows to delete a org unit types with new and rejected org units as the backend is not. 

Related JIRA tickets : IA-3147

- [ ] Are there enough tests?

## Doc

## Changes

- count also new and rejected types
- change wording

## How to test

-  Create an org unit types without org unit, you should be able to delete it
- make another one with a new or rejected org unit, it should not be possible to delete it

## Print screen / video

-

## Notes

```
fix: empty instance pop up

Refs: IA-3665
```


iaso/tests/api/test_org_unit_types_v2.py
```diff
@@ -105,7 +105,7 @@ def test_org_unit_types_list_with_auth(self):
             self.assertEqual(len(org_unit_type_data["projects"]), 1)
 
     def test_org_unit_types_list_count_valid_orgunits(self):
-        """GET /orgunittypes/ with checks on the count of valid org units"""
+        """GET /orgunittypes/ with checks on the count of org units (all validation statuses)"""
```
tdethier, line 108: Nitpick: I would add an extra test that checks what happens when you try to delete an org unit type that has some org units

tdethier (approved):
LGTM :+1: 

I think it would be worth preparing another ticket for all 401 results that shouldn't be a 401 (like the one here)

PR #2387, opened 2025-08-21
IA-4437 Fix n+1 query on `/api/orgunits/{pk}/`

kemar:
Fix n+1 query on `/api/orgunits/{pk}/`.

Related JIRA tickets : IA-4437

[Jira issue](https://bluesquareorg.sentry.io/issues/6749129513/?project=5530884).

This one was triggered when calling the detail of an org unit, especially when there are a lot of ancestors:

```
/api/orgunits/{pk}/
```


tdethier (approved):
The way this API is made creates a lot of n+1, it's annoying we can't go further because there are still lots of repetitions.

But this is clearly an improvement :+1:

PR #2388, opened 2025-08-21
IA-4438 Fix n+1 query on `/api/instances/{pk}/`

kemar:
Fix n+1 query on `/api/instances/{pk}/`.

Related JIRA tickets : IA-4438

[Jira issue](https://bluesquareorg.sentry.io/issues/6748906815/?project=5530884).

This one was triggered when calling the detail of an instances, especially when its org unit has a lot of ancestors:

```
/api/instances/{pk}/
```

## Note

This should be reviewed and merged after https://github.com/BLSQ/iaso/pull/2387 or directly merged into https://github.com/BLSQ/iaso/pull/2387.

### DIFF
--- a/iaso/api/instances/instances.py
+++ b/iaso/api/instances/instances.py
@@ -568,14 +568,32 @@ class InstancesViewSet(viewsets.ViewSet):
                 "instancelock_set__top_org_unit",
                 "instancelock_set__locked_by",
                 "instancelock_set__unlocked_by",
+                "org_unit__groups",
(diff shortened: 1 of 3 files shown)


tdethier (approved):
LGTM :+1: 

I wanted to get rid of n+1 on `InstanceFile` as well, but we'll always get it because we fetch each org unit's default image in the hierarchy. We should definitely refactor both the instance and org unit APIs...

PR #2389, opened 2025-08-21
[IA-4244] Change path of uploaded files

tdethier:
Change path of uploaded files

Related JIRA tickets : IA-4244

- [x] Are there enough tests?

## Doc

/

## Changes

**Feature:** 
- Updated the `upload_to` value of all core `models.FileField()` to sort data by account/type/...
- Added `Account.short_sanitized_name` (used in `upload_to`)

**Refactoring:**
- Split the following models from `iaso.models.base` and placed them into a separate file:
  - `Instance`
  - `InstanceFile`
  - `InstanceLock`
  - `BulkCreateUserCsvFile`
- Split `iaso.models.base.generate_id_for_dhis_2()` and placed it into a separate utility file.

**Other:**
- Changed settings to use `InMemoryFileStorage` for all tests instead of default disk storage

## How to test

- Using your preferred tool (web app/mobile app/admin panel), create some data - including a file! - for the following models:
  - `APIImport`
  - `Instance`
  - `InstanceFile`
  - `BulkCreateUserCsvFile`
  - `FormVersion`
  - `FormAttachment`
  - `ImportGPKG`
  - `ReportVersion`
- :information_source: in your file browser, check your `settings.MEDIA_ROOT` directory and make sure that files are now ordered by account/type/.... (otherwise you can also check the path in your admin panel or in the DB)

## Print screen / video

image

## Notes
- The `ReportVersion` might not be useful since there is no mechanism to fill the new `created_by` field - I didn't find in the codebase how these objects were created, unless through manual action from the admin panel :man_shrugging: 

```
fix: empty instance pop up

Refs: IA-3665
```


hat/api_import/models.py
```diff
@@ -11,6 +17,19 @@
 )
 
 
+def api_import_upload_to(api_import: "APIImport", filename: str):
+    today = date.today()
```
kemar, line 21:
```suggestion
    today = timezone.now().date()
```

Nitpick but this is an old habit of mine: always using UTC (`from django.utils import timezone`) inside the codebase.

`timezone.now().date()` ensures we align with our configured timezone rather than the server's physical location.

hat/settings.py
```diff
@@ -803,3 +803,15 @@ def sentry_error_sampler(_, hint):
             f"\tno plugin_settings.py file found for plugin {plugin_name}, appending plugins.{plugin_name} to INSTALLED_APPS"
         )
         INSTALLED_APPS.append(f"plugins.{plugin_name}")
+
+# Making sure that files are not stored on disk while running tests
+# This allows faster tests and easier clean up of test files
```
kemar, line 808: 👍

iaso/models/base.py
```diff
@@ -149,6 +120,19 @@ class Account(models.Model):
...
+        """
+        Short sanitized name mainly used in file path storage
+        """
+        text = self.name.lower()
+        text = re.sub(r"\s+", "_", text)
+        text = re.sub(r"[^a-z0-9_]", "", text)
+        text = re.sub(r"_+", "_", text)
+        text = text[:30]
+        text = text.strip("_")
+        return text if text else "invalid_name"
```
```suggestion
    @property
    def short_sanitized_name(self):
        """
        Short sanitized name mainly used in file path storage
        """
        if not self.name or not self.name.strip():
            return "invalid_name"

        text = slugify(self.name, allow_unicode=False)
        text = text.replace('-', '_')[:30]

        return text if len(text) >= 2 else "invalid_name"
```

I would go for Django's slugify (`from django.utils.text import slugify`) because it's battle tested and it handles Unicode and special characters gracefully:

- it converts rather than strips: "café" → "cafe" (not "caf")
- another benefit: less custom regex logic to maintain

kemar (approved):
LGTM

Kudos for:

- the test coverage for upload scenarios
- the clean code (re)organization
- the performance improvement with `InMemoryStorage`, this will result in faster tests and less garbage on disk!

I've left a few suggestions that you can implement if they make sense to you.

PR #2394, opened 2025-08-25
[WC2-807] Adding Stock Management

bmonjoie:

Related JIRA tickets : WC2-807

- [X] Are there enough tests?

## Changes

This PR adds the new models and endpoint which is the base for Stock Management.

## How to test

Try the endpoints or through the admin.

```
fix: empty instance pop up

Refs: IA-3665
```


entrypoint.sh
```diff
@@ -100,6 +100,10 @@ case "$1" in
     python "${@:2}"
   ;;
   * )
-    show_help
+    if [[ $2 == /opt/.pycharm_helpers/* ]]; then
+      ${@}
+    else
+      show_help
+    fi
```
bmonjoie, line 107: This allows PyCharm to use the Python interpreter from Iaso's Docker Compose. It's not related to the PR but is very useful to me.

iaso/models/forms.py
```diff
@@ -87,13 +87,18 @@ def filter_for_user_and_app_id(self, user: typing.Union[User, AnonymousUser], ap
...
-                queryset = queryset.exclude(derived=True)  # do not include derived instances for the mobile app
+                queryset = self.filter_for_project(project=project, queryset=queryset)
             except Project.DoesNotExist:
                 return self.none()
 
         return queryset
 
+    def filter_for_project(self, project: Project, queryset=None):
+        if queryset is None:
+            queryset = self.all()
```
tdethier: I'm not sure if this is actually going to change anything performance-wise, but it should be better

```suggestion
            queryset = self
```

iaso/models/org_unit.py
```diff
@@ -100,12 +100,17 @@ def filter_for_user_and_app_id(
...
-                queryset = queryset.filter(projects__in=[project])
+                queryset = self.filter_for_project(project, queryset)
             except Project.DoesNotExist:
                 return self.none()
 
         return queryset
 
+    def filter_for_project(self, project: Project, queryset=None):
+        if queryset is None:
+            queryset = self.all()
```
tdethier: I'm not sure if this is actually going to change anything performance-wise, but it should be better
```suggestion
            queryset = self
```

iaso/models/stocks.py
```diff
@@ -0,0 +1,306 @@
...
+from iaso.models.entity import UserNotAuthError
+from iaso.utils.models.soft_deletable import SoftDeletableModel
+
+
+class StockImpacts(models.TextChoices):
+    """StockImpacts is an Enum-like class representing the impact of a rule on the stock."""
+
+    ADD = "ADD", "Add"
+    SUBTRACT = "SUBTRACT", "Subtract"
+    RESET = "RESET", "Reset"
```
tdethier, line 20: Nitpick: we could translate the displayed values (e.g. `ADD = "ADD", _("Add")` with an import of `from django.utils.translation import gettext as _`)

iaso/models/stocks.py
```diff
@@ -0,0 +1,306 @@
...
+
+class StockKeepingUnit(SoftDeletableModel):
+    """
+    A distinct type of item for tracking in inventory. This is often referred as SKU.
+    """
+
+    name = models.CharField(max_length=255)
+    short_name = models.CharField(max_length=255)
+    account = models.ForeignKey("iaso.Account", on_delete=models.CASCADE, null=False, blank=False)
+    projects = models.ManyToManyField("Project", related_name="sku_projects", blank=True)
```
tdethier: Nitpick: 

```suggestion
    projects = models.ManyToManyField("Project", related_name="skus", blank=True)
```

iaso/models/stocks.py
```diff
@@ -0,0 +1,306 @@
...
+    org_unit_types = models.ManyToManyField("OrgUnitType", related_name="sku_org_unit_types", blank=True)
```
tdethier: Nitpick:
```suggestion
    org_unit_types = models.ManyToManyField("OrgUnitType", related_name="skus", blank=True)
```

iaso/models/stocks.py
```diff
@@ -0,0 +1,306 @@
...
+    forms = models.ManyToManyField("Form", related_name="sku_forms", blank=True)
```
tdethier: Nitpick: 
```suggestion
    forms = models.ManyToManyField("Form", related_name="skus", blank=True)
```

iaso/models/stocks.py
```diff
@@ -0,0 +1,306 @@
...
+    children = models.ManyToManyField("self", blank=True, through="StockKeepingUnitChildren")
+    display_units = models.CharField(max_length=25, null=True, blank=True)
```
tdethier: Django considers [a bad practice](https://docs.djangoproject.com/en/5.2/ref/models/fields/#django.db.models.Field.null) to have both `null=True` and `blank=True` on a `CharField`

```suggestion
    display_units = models.CharField(max_length=25, blank=True)
```

iaso/models/stocks.py
```diff
@@ -0,0 +1,306 @@
...
+    """
+    The display_precision should be a multiple of 10 by which the value will be divided.
+    E.g.: 
+    - if you want to display kilograms, the value should be in grams but the display_precision is going to be 1000
+    """
+    display_precision = models.IntegerField(null=False, blank=False, default=1)
```
tdethier: If this needs to be a multiple of 10, I would add some validation on that field at the model level.

- Good way: use [validators](https://docs.djangoproject.com/en/5.2/ref/validators/#writing-validators)
- Ok way: override `.clean()` and check it there

iaso/models/stocks.py
tdethier: Is `account` really useful when you already have `projects`? I'm not sure why it's needed

iaso/models/stocks.py
```diff
@@ -0,0 +1,306 @@
...
+        return self.filter(sku__account=profile.account)
+
+
+class StockItem(models.Model):
+    """
+    The number of items in a stock for a particular SKU and OrgUnit.
+    """
+
+    org_unit = models.ForeignKey(
+        "OrgUnit", related_name="stock_item_org_unit", null=False, blank=False, on_delete=models.CASCADE
```
tdethier: Nitpick: 
```suggestion
        "OrgUnit", related_name="stock_items", null=False, blank=False, on_delete=models.CASCADE
```

iaso/models/stocks.py
```diff
@@ -0,0 +1,306 @@
...
+    )
+    sku = models.ForeignKey(
+        "StockKeepingUnit", related_name="stock_item_sku", null=False, blank=False, on_delete=models.CASCADE
```
tdethier: Nitpick: 
```suggestion
        "StockKeepingUnit", related_name="stock_items", null=False, blank=False, on_delete=models.CASCADE
```

iaso/models/stocks.py
```diff
@@ -0,0 +1,306 @@
...
+            sku__account=project.account,
+            sku__projects__in=[project],
+        )
+
+
+class StockLedgerItem(models.Model):
+    """
+    An action that changed the stock item value based on a StockItemRule.
+
+    Note: We are not referencing StockItemRule as they can change over time but a StockLedgerItem is immutable.
```
tdethier: Two comments:
- I don't understand the referencing StockItemRule part because there's a FK to that model below
- If this has to be immutable, I would probably enforce it at the model level to be sure: maybe by overriding `.save()`, checking if there's a primary key/ID and throwing an error if that's the case

iaso/models/stocks.py
```diff
@@ -0,0 +1,306 @@
...
+    """
+
+    id = models.UUIDField(primary_key=True, auto_created=True, default=uuid.uuid4, editable=False)
+    sku = models.ForeignKey(
+        "StockKeepingUnit", related_name="ledger_item_sku", null=False, blank=False, on_delete=models.CASCADE
```
tdethier: Nitpick: 
```suggestion
        "StockKeepingUnit", related_name="ledger_items", null=False, blank=False, on_delete=models.CASCADE
```

iaso/models/stocks.py
```diff
@@ -0,0 +1,306 @@
...
+    )
+    org_unit = models.ForeignKey(
+        "OrgUnit", related_name="ledger_item_org_unit", null=False, blank=False, on_delete=models.CASCADE
```
tdethier: Nitpick:
```suggestion
        "OrgUnit", related_name="ledger_items", null=False, blank=False, on_delete=models.CASCADE
```

iaso/models/stocks.py
```diff
@@ -0,0 +1,306 @@
...
+    )
+    rule = models.ForeignKey(
+        "StockItemRule", related_name="ledger_item_rule", null=True, blank=True, on_delete=models.CASCADE
```
tdethier: Nitpick:
```suggestion
        "StockItemRule", related_name="ledger_items", null=True, blank=True, on_delete=models.CASCADE
```

iaso/models/stocks.py
```diff
@@ -0,0 +1,306 @@
...
+    )
+    submission = models.ForeignKey(
+        "Instance", related_name="ledger_item_instance", null=True, blank=True, on_delete=models.CASCADE
```
tdethier: Nitpick:
```suggestion
        "Instance", related_name="ledger_items", null=True, blank=True, on_delete=models.CASCADE
```

iaso/models/stocks.py
```diff
@@ -0,0 +1,306 @@
...
+    )
+    question = models.CharField(max_length=255, null=True, blank=True)
```
tdethier: Same comment about `null=True` and `blank=True` in `CharField`
```suggestion
    question = models.CharField(max_length=255, blank=True)
```

iaso/models/stocks.py
```diff
@@ -0,0 +1,306 @@
...
+    value = models.IntegerField()
+    impact = models.CharField(max_length=8, choices=StockImpacts.choices, null=True, blank=True)
```
tdethier: Same comment about `null=True` and `blank=True` in `CharField`
```suggestion
    impact = models.CharField(max_length=8, choices=StockImpacts.choices, blank=True)
```

iaso/models/stocks.py
```diff
@@ -0,0 +1,306 @@
...
+    created_by = models.ForeignKey(
+        User, related_name="ledger_item_created_by", null=False, blank=False, on_delete=models.CASCADE
```
tdethier: Nitpick:
```suggestion
        User, related_name="ledger_item", null=False, blank=False, on_delete=models.CASCADE
```

iaso/models/stocks.py
```diff
@@ -0,0 +1,306 @@
...
+    )
+    created_at = models.DateTimeField()
+    received_at = models.DateTimeField(auto_now_add=True)
```
tdethier, line 169: Is there a difference between the two of them? It looks like the same thing to me

iaso/models/stocks.py
```diff
@@ -0,0 +1,306 @@
...
+        )
+
+
+class StockItemRule(models.Model):
+    """
+    Defines how a form impacts stock items.
+    """
+
+    sku = models.ForeignKey(
+        "StockKeepingUnit", related_name="stock_item_rule_sku", null=False, blank=False, on_delete=models.CASCADE
```
tdethier: Nitpick: 
```suggestion
        "StockKeepingUnit", related_name="stock_item_rules", null=False, blank=False, on_delete=models.CASCADE
```

iaso/models/stocks.py
```diff
@@ -0,0 +1,306 @@
...
+    )
+    form = models.ForeignKey(
+        "Form", related_name="stock_item_rule_form", null=False, blank=False, on_delete=models.CASCADE
```
tdethier: Nitpick:
```suggestion
        "Form", related_name="stock_item_rules", null=False, blank=False, on_delete=models.CASCADE
```

iaso/models/stocks.py
```diff
@@ -0,0 +1,306 @@
...
+    )
+    question = models.CharField(max_length=255, null=False, blank=False)
```
tdethier: Same comment about `null=True` and `blank=True` in `CharField`
```suggestion
    question = models.CharField(max_length=255, blank=False)
```

iaso/models/stocks.py
```diff
@@ -0,0 +1,306 @@
...
+    impact = models.CharField(max_length=8, choices=StockImpacts.choices, null=True, blank=True)
```
tdethier: Same comment about `null=True` and `blank=True` in `CharField`
```suggestion
    impact = models.CharField(max_length=8, choices=StockImpacts.choices, blank=True)
```

iaso/models/stocks.py
```diff
@@ -0,0 +1,306 @@
...
+    def filter_published(self):
+        return self.filter(status=StockRulesVersionsStatus.PUBLISHED)
+
+
+class StockRulesVersionsStatus(models.TextChoices):
+    """StockRulesVersionsStatus is an Enum-like class for the Status of a StockRulesVersion."""
+
+    DRAFT = "DRAFT", "Draft"
+    UNPUBLISHED = "UNPUBLISHED", "Unpublished"
+    PUBLISHED = "PUBLISHED", "Published"
```
tdethier, line 249: Same comment about translations

iaso/models/stocks.py
```diff
@@ -0,0 +1,306 @@
...
+
+class StockRulesVersion(SoftDeletableModel):
+    """
+    A version of the stock item rules.
+    """
+
+    NAME_MAX_LENGTH = 50
+
+    account = models.ForeignKey(
+        "iaso.Account", related_name="stock_rule_version_account", on_delete=models.CASCADE, null=False, blank=False
```
tdethier: Nitpick:
```suggestion
        "iaso.Account", related_name="stock_rule_versions", on_delete=models.CASCADE, null=False, blank=False
```

iaso/models/stocks.py
```diff
@@ -0,0 +1,306 @@
...
+StockRulesVersionStatusAllowedTransitions = {
+    StockRulesVersionsStatus.DRAFT: {StockRulesVersionsStatus.UNPUBLISHED, StockRulesVersionsStatus.PUBLISHED},
+    StockRulesVersionsStatus.UNPUBLISHED: {StockRulesVersionsStatus.PUBLISHED},
+    StockRulesVersionsStatus.PUBLISHED: {StockRulesVersionsStatus.UNPUBLISHED},
+}
+
+
+def is_transition_allowed(self: "StockRulesVersion", new_status: str):
+    allowed_set: typing.Set[str] = StockRulesVersionStatusAllowedTransitions.get(self.status, set())
+    return new_status in allowed_set
```
tdethier, line 261: I like that 👍

iaso/models/stocks.py
```diff
@@ -0,0 +1,306 @@
...
+                StockRulesVersion.objects.filter(
+                    account=self.account, status=StockRulesVersionsStatus.PUBLISHED
+                ).update(status=StockRulesVersionsStatus.UNPUBLISHED)
+
+            return {"success": True}
+
+        return {
+            "success": False,
+            "error": f"Transition from {old_status_str} to {new_status_str} is not allowed",
+        }
```
tdethier, line 310: It's interesting, I'm not sure I would have done it like that, but it works 👍

iaso/admin/base.py
```diff
@@ -743,6 +749,68 @@ class StorageLogEntryInline(admin.TabularInline):
...
+    fields = ("parent", "child", "created_at", "updated_at", "created_by", "updated_by")
+    readonly_fields = ("created_at", "updated_at", "created_by", "updated_by")
+    list_display = ("parent", "child", "value")
+    list_filter = ("parent", "child")
+
+
+@admin.register(StockLedgerItem)
+class StockLedgerItemAdmin(admin.ModelAdmin):
+    fields = ("rule", "sku", "org_unit", "submission", "question", "impact", "value", "created_at", "created_by")
+    readonly_fields = ("created_at", "created_by")
```
tdethier, line 810: Wasn't this model immutable? If so, shouldn't all of the fields be in this list?

iaso/permissions.py
```diff
@@ -39,6 +39,7 @@
 _SOURCE_WRITE = "iaso_write_sources"
 _SOURCES = "iaso_sources"
 _SOURCES_CAN_CHANGE_DEFAULT_VERSION = "iaso_sources_can_change_default_version"
+_STOCK_MANAGEMENT = "iaso_stock_management"
```
tdethier: Depending on which PR gets merged first (this one or the permissions 2.0 one), a new permission is going to make more work for one of us 😅

iaso/api/stocks/serializers.py
```diff
@@ -0,0 +1,397 @@
...
+            "name",
+            "short_name",
+            "projects",
+            "org_unit_types",
+            "forms",
+            "display_units",
+            "display_precision",
+        ]
+
+    def validate(self, validated_data):
```
tdethier, line 137: I guess we should also check if:
- the user has access to these org unit types
- the user has access to these forms
- the forms are linked to the org unit types? I'm not sure about this one

iaso/api/stocks/serializers.py
```diff
@@ -0,0 +1,397 @@
...
+
+    class Meta:
+        model = StockItem
+        fields = [
+            "sku",
+            "org_unit",
+            "value",
+        ]
+
+    def validate(self, validated_data):
```
tdethier, line 199: Should we check that the received `org_unit` is part of the `sku.org_unit_types`?

iaso/api/stocks/serializers.py
```diff
@@ -0,0 +1,397 @@
...
+    )
+    org_unit = PrimaryKeyRelatedField(
+        queryset=OrgUnit.objects.all(),
+        required=True,
+    )
+    rule = PrimaryKeyRelatedField(
+        queryset=StockItemRule.objects.all(),
+        required=False,
+    )
+    submission_id = IdOrUuidRelatedField(source="submission", queryset=Instance.objects.all(), required=True)
```
tdethier, line 261: Same question here, should we check that the user has access to these objects (tenancy-wise) and should we check that they are actually linked together (= is the sku in that rule the same as the received sku)?

iaso/tests/api/stocks/test_serializers.py
```diff
@@ -0,0 +1,670 @@
...
+    StockRulesVersionSerializer,
+    StockRulesVersionWriteSerializer,
+)
+from iaso.test import TestCase
+
+
+@time_machine.travel("2025-08-19T13:00:00.000Z", tick=False)
+class StockKeepingUnitSerializerTestCase(TestCase):
+    @classmethod
+    def setUpTestData(cls):
```
tdethier, line 31: Most of the test cases in this file seem to have a very similar setup, should we maybe have a common base for them?

iaso/tests/api/stocks/test_serializers_mobile.py
```diff
@@ -0,0 +1,123 @@
+import time_machine
+
+from iaso import models as m
+from iaso.api.stocks.serializers_mobile import StockItemRuleMobileSerializer, StockKeepingUnitMobileSerializer
+from iaso.test import TestCase
+
+
```
tdethier, line 16: What about tests for the other 2 serializers? (`StockRulesVersionMobileSerializer` and `StockLedgerMobileItemSerializer`)

iaso/api/stocks/utils.py
```diff
@@ -0,0 +1,58 @@
...
+    new_version.id = None
+
+    new_version.name = COPY_OF_PREFIX + orig_version.name
+    if len(new_version.name) > StockRulesVersion.NAME_MAX_LENGTH:
+        new_version.name = f"Copy of version {new_version.id}"
+    new_version.status = StockRulesVersionsStatus.DRAFT
+    new_version.updated_by = request.user
+    new_version.save()
+
+    for rule in StockItemRule.objects.filter(version=orig_version):
```
tdethier:
```suggestion
    for rule in orig_version.rules.all():
```

iaso/api/stocks/utils.py
```diff
@@ -0,0 +1,58 @@
...
+        serializer = StockLedgerItemWriteSerializer(data=item)
+        serializer.is_valid(raise_exception=True)
+        sku = serializer.validated_data.get("sku")
+        if sku.account != user.iaso_profile.account:
+            raise serializers.ValidationError("User doesn't have access to this SKU")
+        submission = serializer.validated_data["submission"]
+        if submission.project != project:
+            raise serializers.ValidationError(
+                "You can only create StockLedgerItem on the same project as the submission"
+            )
```
tdethier, line 48: This could be part of the serializer's `.validate()`

iaso/api/stocks/utils.py
```diff
@@ -0,0 +1,58 @@
...
+        serializer.validated_data["created_at"] = submission.created_at
+        serializer.validated_data["created_by"] = user
+        serializer.save()
+
+        org_unit = serializer.validated_data["org_unit"]
+        sku = serializer.validated_data["sku"]
+        stock_item, _ = StockItem.objects.get_or_create(org_unit=org_unit, sku=sku)
+        stock_item.value = compute_new_stock_item_value(stock_item)
+        stock_item.updated_by = user
```
tdethier: I don't see `updated_by` on the `StockItem` model

iaso/api/stocks/utils.py
```diff
@@ -0,0 +1,58 @@
+from copy import deepcopy
```
tdethier, line 1: Maybe we should have some unit tests for these 2 util functions

iaso/models/stocks.py
```diff
@@ -0,0 +1,306 @@
...
+        return f"{self.name} ({self.short_name})"
+
+
+class StockKeepingUnitChildren(models.Model):
+    parent = models.ForeignKey(
+        "StockKeepingUnit", related_name="sku_children_parent", on_delete=models.CASCADE, null=False, blank=False
+    )
+    child = models.ForeignKey(
+        "StockKeepingUnit", related_name="sku_children_child", on_delete=models.CASCADE, null=False, blank=False
+    )
```
tdethier, line 88: The related names here are very confusing, I never understand when I see them in the code 😅

iaso/api/stocks/use_cases.py
```diff
@@ -0,0 +1,92 @@
...
+        StockItemRule.objects.select_related("sku")
+        .prefetch_related("sku__sku_children_parent")
+        .filter(
+            form=instance.form,
+            sku__account=instance.project.account,
+            sku__org_unit_types__in=instance.org_unit.org_unit_type,
+            version__status=StockRulesVersionsStatus.PUBLISHED,
+        )
+        .order("order")
+        .all()
```
tdethier:
```suggestion
```

iaso/api/stocks/use_cases.py
```diff
@@ -0,0 +1,92 @@
...
+        value = int(instance.json[rule.question])
+    except ValueError as error:
+        logger.exception(error)
+        return
+
+    stock_item, _ = StockItem.objects.get_or_create(org_unit=instance.org_unit, sku=rule.sku)
+
+    if stock_item.created_by is None:
+        stock_item.created_by = instance.created_by
+    stock_item.updated_by = instance.created_by
```
tdethier: I don't see `created_by` nor `updated_by` on the model `StockItem`

iaso/api/stocks/use_cases.py
```diff
@@ -0,0 +1,92 @@
...
+    start_date = last_reset.created_at if last_reset else datetime.fromtimestamp(0)
+
+    stock_balance = queryset.filter(created_at__gte=start_date).aggregate(
+        balance=Sum(
+            Case(
+                When(impact=StockImpacts.SUBTRACT, then=-F("value")),
+                default=F("value"),
+            )
+        )
+    )
```
tdethier, line 33: I guess a `StockLedgerItem` with `impact=StockImpacts.RESET` should always have a value of `0`? Because if it's positive, it will be added into the total here

iaso/tests/api/stocks/test_use_cases.py
```diff
@@ -0,0 +1,249 @@
...
+        cls.rule = m.StockItemRule.objects.create(
+            sku=sku,
+            form=form,
+            version=version,
+            question="question_name",
+            impact=m.StockImpacts.ADD,
+        )
+        cls.stock_item = m.StockItem.objects.create(sku=sku, org_unit=org_unit)
+
+    def testStockValue_returns_0_when_no_last_reset_and_no_items(self):
```
tdethier:
```suggestion
    def test_stock_value_returns_0_when_no_last_reset_and_no_items(self):
```

iaso/tests/api/stocks/test_use_cases.py
```diff
@@ -0,0 +1,249 @@
...
+        balance = compute_new_stock_item_value(self.stock_item)
+        self.assertEqual(0, balance)
+
+    def testStockValue_returns_last_reset_value_when_there_are_no_subsequent_items(self):
```
tdethier:
```suggestion
    def test_stock_value_returns_last_reset_value_when_there_are_no_subsequent_items(self):
```

iaso/tests/api/stocks/test_use_cases.py
```diff
@@ -0,0 +1,249 @@
...
+            created_at=datetime.datetime.fromtimestamp(lastResetTime),
+            created_by=self.user,
+            submission=self.instance,
+            question="q1",
+        )
+
+        balance = compute_new_stock_item_value(self.stock_item)
+        self.assertEqual(lastResetValue, balance)
+
+    def testStockValue_correctly_calculates_stock_with_ADD_impacts(self):
```
tdethier:
```suggestion
    def test_stock_value_correctly_calculates_stock_with_ADD_impacts(self):
```

iaso/tests/api/stocks/test_use_cases.py
```diff
@@ -0,0 +1,249 @@
...
+            impact=m.StockImpacts.ADD,
+            created_at=datetime.datetime.fromtimestamp(lastResetTime + 2000),
+            created_by=self.user,
+            submission=self.instance,
+            question="q3",
+        )
+        balance = compute_new_stock_item_value(self.stock_item)
+        self.assertEqual(80, balance)  # 50 (reset) + 10 (add) + 20 (add)
+
+    def testStockValue_correctly_calculates_stock_with_SUBTRACT_impacts(self):
```
tdethier:
```suggestion
    def test_stock_value_correctly_calculates_stock_with_SUBTRACT_impacts(self):
```

iaso/tests/api/stocks/test_use_cases.py
```diff
@@ -0,0 +1,249 @@
...
+            created_at=datetime.datetime.fromtimestamp(lastResetTime + 2000),
+            created_by=self.user,
+            submission=self.instance,
+            question="q3",
+        )
+
+        balance = compute_new_stock_item_value(self.stock_item)
+        self.assertEqual(70, balance)  # 100 (reset) - 10 (subtract) - 20 (subtract)
+
+    def testStockValue_correctly_calculates_stock_with_mixed_impacts_and_no_reset(self):
```
tdethier:
```suggestion
    def test_stock_value_correctly_calculates_stock_with_mixed_impacts_and_no_reset(self):
```

iaso/tests/api/stocks/test_use_cases.py
```diff
@@ -0,0 +1,249 @@
...
+            impact=m.StockImpacts.ADD,
+            created_at=datetime.datetime.fromtimestamp(3000),
+            created_by=self.user,
+            submission=self.instance,
+            question="q3",
+        )
+        balance = compute_new_stock_item_value(self.stock_item)
+        self.assertEqual(40, balance)  # 0 + 30 - 5 + 15
+
+    def testStockValue_correctly_calculates_stock_with_mixed_impacts_and_a_reset(self):
```
tdethier:
```suggestion
    def test_stock_value_correctly_calculates_stock_with_mixed_impacts_and_a_reset(self):
```

iaso/tests/api/stocks/test_use_cases.py
```diff
@@ -0,0 +1,249 @@
...
+            created_at=datetime.datetime.fromtimestamp(lastResetTime + 2000),
+            created_by=self.user,
+            submission=self.instance,
+            question="q3",
+        )
+
+        balance = compute_new_stock_item_value(self.stock_item)
+        self.assertEqual(40, balance)  # 50 (reset) + 10 (add) - 20 (subtract)
+
+    def testStockValue_when_reset_is_last(self):
```
tdethier:
```suggestion
    def test_stock_value_when_reset_is_last(self):
```

iaso/api/stocks/use_cases.py
tdethier: When I read tests I realize that I didn't quite understand the business logic here 😅

iaso/tests/api/stocks/test_use_cases.py
```diff
@@ -0,0 +1,249 @@
+import datetime
+
+from iaso import models as m
+from iaso.api.stocks.use_cases import compute_new_stock_item_value
+from iaso.test import TestCase
+
+
+class StockUseCaseUtilsTestCase(TestCase):
```
tdethier, line 8: I would probably add tests for `apply_stock_management_to_instance` as well, to make sure that the happy path works fine and that all the right `StockSomething` models are created + to check all edge cases with `instance` and issues with `instance.json`

iaso/api/stocks/filters.py
```diff
@@ -0,0 +1,88 @@
...
+
+    @staticmethod
+    def filter_projects(queryset, name, value):
+        project_ids = parse_comma_separated_numeric_values(value, name)
+        return queryset.filter(projects__id__in=project_ids)
+
+    @staticmethod
+    def filter_org_unit_types(queryset, name, value):
+        org_unit_type_ids = parse_comma_separated_numeric_values(value, name)
+        return queryset.filter(org_unit_types__id__in=org_unit_type_ids)
```
tdethier, line 45: Nitpick: this works fine, but you can also use another type of filter that does the parsing for you - example in [OUCR filters](https://github.com/BLSQ/iaso/blob/259e40bbc78e2eed1d56524d92eff69cfcd7fc7e/iaso/api/org_unit_change_requests/filters.py#L32)

Same comment for other filters below in this file

iaso/api/stocks/views.py
```diff
@@ -0,0 +1,374 @@
...
+        ).exists():
+            raise ValidationError("A `StockItem` already exists for this `sku` and this `org_unit`.")
+
+        StockLedgerItem.objects.create(
+            sku=serializer.validated_data["sku"],
+            org_unit=serializer.validated_data["org_unit"],
+            value=serializer.validated_data["value"],
+            impact=StockImpacts.RESET,
+            created_by=self.request.user,
+            created_at=datetime.now(),
```
tdethier, line 163: Django will do it for you 
```suggestion
```

iaso/api/stocks/views.py
```diff
@@ -0,0 +1,374 @@
...
+        return [permission() for permission in permission_classes]
+
+    def perform_create(self, serializer):
+        """
+        POST to create a `StockItem`.
+        """
+        if StockItem.objects.filter(
+            sku=serializer.validated_data["sku"], org_unit=serializer.validated_data["org_unit"]
+        ).exists():
+            raise ValidationError("A `StockItem` already exists for this `sku` and this `org_unit`.")
```
tdethier, line 155: You could do that in your serializer, but it works too

iaso/api/stocks/views.py
```diff
@@ -0,0 +1,374 @@
...
+        response = super().partial_update(request, *args, **kwargs)
+        stock_item = self.get_object()
+
+        StockLedgerItem.objects.create(
+            sku=stock_item.sku,
+            org_unit=stock_item.org_unit,
+            value=stock_item.value,
+            impact=StockImpacts.RESET,
+            created_by=self.request.user,
+            created_at=datetime.now(),
```
tdethier, line 180:
```suggestion
```

iaso/api/stocks/views.py
```diff
@@ -0,0 +1,374 @@
...
+        if StockItem.objects.filter(
+            sku=serializer.validated_data["sku"], org_unit=serializer.validated_data["org_unit"]
+        ).exists():
+            raise ValidationError("A `StockItem` already exists for this `sku` and this `org_unit`.")
+
+        StockLedgerItem.objects.create(
+            sku=serializer.validated_data["sku"],
+            org_unit=serializer.validated_data["org_unit"],
+            value=serializer.validated_data["value"],
+            impact=StockImpacts.RESET,
```
tdethier, line 161: Is it always a reset?

iaso/api/stocks/views.py
```diff
@@ -0,0 +1,374 @@
...
+        PATCH to edit a `StockItem`.
+        """
+        response = super().partial_update(request, *args, **kwargs)
+        stock_item = self.get_object()
+
+        StockLedgerItem.objects.create(
+            sku=stock_item.sku,
+            org_unit=stock_item.org_unit,
+            value=stock_item.value,
+            impact=StockImpacts.RESET,
```
tdethier, line 178: Same question about reset

iaso/api/stocks/views.py
```diff
@@ -0,0 +1,374 @@
...
+            permission_classes = [HasStockManagementReadPermission]
+        else:
+            permission_classes = [HasStockManagementFullPermission]
+        return [permission() for permission in permission_classes]
+
+    def perform_create(self, serializer):
+        """
+        POST to create a `StockLedgerItem`.
+        """
+        serializer.validated_data["created_at"] = datetime.now()
```
tdethier, line 233: In this case, it's important to add it because it's not added automatically at the model level - was it done on purpose?

iaso/api/stocks/views.py
```diff
@@ -0,0 +1,374 @@
...
+        else:
+            permission_classes = [HasStockManagementFullPermission]
+        return [permission() for permission in permission_classes]
+
+    def perform_create(self, serializer):
+        """
+        POST to create a `StockItemRule`.
+        """
+        if serializer.validated_data.get("version").status != StockRulesVersionsStatus.DRAFT:
+            raise PermissionDenied("You can only create items for draft versions")
```
tdethier: I would probably put that in the serializer

iaso/api/stocks/serializers.py
```diff
@@ -0,0 +1,397 @@
...
+            "org_unit_types",
+            "forms",
+            "display_units",
+            "display_precision",
+        ]
+
+    def validate(self, validated_data):
+        request = self.context.get("request")
+        if "projects" in validated_data and not all(
+            p.account == request.user.iaso_profile.account for p in validated_data["projects"]
```
tdethier: This will do the exact same thing, but might avoid fetching the relation 
```suggestion
            p.account_id == request.user.iaso_profile.account_id for p in validated_data["projects"]
```

iaso/api/stocks/serializers.py
```diff
@@ -0,0 +1,397 @@
...
+        fields = [
+            "sku",
+            "org_unit",
+            "value",
+        ]
+
+    def validate(self, validated_data):
+        request = self.context.get("request")
+        sku = validated_data.get("sku")
+        if sku.account != request.user.iaso_profile.account:
```
tdethier: This will do the exact same thing, but might avoid fetching the relation
```suggestion
        if sku.account_id != request.user.iaso_profile.account_id:
```

iaso/api/stocks/serializers.py
```diff
@@ -0,0 +1,397 @@
...
+            "form",
+            "version",
+            "question",
+            "impact",
+            "order",
+        ]
+
+    def validate(self, validated_data):
+        request = self.context.get("request")
+        if validated_data.get("sku").account != request.user.iaso_profile.account:
```
tdethier:
```suggestion
        if validated_data.get("sku").account_id != request.user.iaso_profile.account_id:
```

iaso/api/stocks/serializers.py
```diff
@@ -0,0 +1,397 @@
...
+            raise serializers.ValidationError("User doesn't have access to this SKU")
+        if validated_data.get("version").account != request.user.iaso_profile.account:
```
tdethier:
```suggestion
        if validated_data.get("version").account_id != request.user.iaso_profile.account_id:
```

iaso/api/stocks/serializers.py
```diff
@@ -0,0 +1,397 @@
...
+            raise serializers.ValidationError("User doesn't have access to this version")
+        if validated_data.get("sku").account != validated_data.get("version").account:
```
tdethier:
```suggestion
        if validated_data.get("sku").account_id != validated_data.get("version").account_id:
```

iaso/tests/api/stocks/test_views.py
```diff
@@ -0,0 +1,886 @@
...
+            response = self.client.get(STOCK_ITEM_URL, {"skus": self.sku_1.id})
+            self.assertJSONResponse(response, 200)
+        self.assertEqual(1, response.data["count"])
+
+        with self.assertNumQueries(2):
+            # 1. SELECT COUNT(*)
+            # 2. SELECT FROM StockItem
+            response = self.client.get(STOCK_ITEM_URL, {"skus": f"{self.sku_1.id},{self.sku_2.id}"})
+            self.assertJSONResponse(response, 200)
+        self.assertEqual(1, response.data["count"])
```
tdethier: Nitpick: I would add at least one more `StockItem` in order to better see the filtering

iaso/tests/api/stocks/test_views.py
```diff
@@ -0,0 +1,886 @@
...
+        m.StockItem.objects.create(
+            org_unit=org_unit_2,
+            sku=sku,
+            value=5,
+        )
+
+    def test_not_authenticated_without_rights_list(self):
+        with self.assertNumQueries(0):
+            response = self.client.get(SKU_URL)
+            self.assertJSONResponse(response, 401)
```
tdethier: Nitpick: in general for all API tests, I think it's better if we use constant values from `rest_framework.status` instead of numbers

iaso/tests/api/stocks/test_views.py
```diff
@@ -0,0 +1,886 @@
...
+
+        response = self.client.get(f"{STOCK_ITEM_URL}{self.stock_item_1.pk}/")
+        self.assertEqual(response.json()["value"], 20)
+
+    def test_delete_without_authorization(self):
+        self.client.force_authenticate(self.user_without_rights)
+        response = self.client.delete(f"{STOCK_ITEM_URL}{self.stock_item_1.pk}/")
+        self.assertJSONResponse(response, 403)
+
+    def test_delete(self):
```
tdethier, line 362: Maybe an assert on `StockItem.object.count()` before and after here would be good

iaso/tests/api/stocks/test_views.py
```diff
@@ -0,0 +1,886 @@
...
+            response = self.client.get(LEDGER_ITEM_URL, {"skus": self.sku_1.id})
+            self.assertJSONResponse(response, 200)
+        self.assertEqual(1, response.data["count"])
+
+        with self.assertNumQueries(2):
+            # 1. SELECT COUNT(*)
+            # 2. SELECT StockLedgerItem
+            response = self.client.get(LEDGER_ITEM_URL, {"skus": f"{self.sku_1.id},{self.sku_2.id}"})
+            self.assertJSONResponse(response, 200)
+        self.assertEqual(1, response.data["count"])
```
tdethier, line 463: Same comment on number of objects

iaso/tests/api/stocks/test_views.py
```diff
@@ -0,0 +1,886 @@
...
+            self.assertJSONResponse(response, 200)
+        self.assertEqual(0, response.data["count"])
+
+    def test_create_without_authorization(self):
+        self.client.force_authenticate(self.user_without_rights)
+
+        response = self.client.post(
+            LEDGER_ITEM_URL,
+            data={
+                "id": uuid.uuid4().hex,
```
tdethier, line 495: TIL

iaso/tests/api/stocks/test_views.py
```diff
@@ -0,0 +1,886 @@
...
+
+    def test_patch_rule(self):
+        self.client.force_authenticate(self.user_with_rights)
+        response = self.client.patch(
+            f"{RULE_ITEM_URL}{self.stock_item_rule_1.pk}/", data={"impact": m.StockImpacts.SUBTRACT}
+        )
+        self.assertJSONResponse(response, 200)
+
+        response = self.client.get(f"{RULE_ITEM_URL}{self.stock_item_rule_1.pk}/")
+        self.assertEqual(response.json()["impact"], m.StockImpacts.SUBTRACT)
```
tdethier, line 728: I would probably refresh the python object and assert on it too - in all update test calls

iaso/tests/api/stocks/test_views.py
```diff
@@ -0,0 +1,886 @@
...
+
+    def test_delete_without_authorization(self):
+        self.client.force_authenticate(self.user_without_rights)
+        response = self.client.delete(f"{RULE_ITEM_URL}{self.stock_item_rule_1.pk}/")
+        self.assertJSONResponse(response, 403)
+
+    def test_delete(self):
```
tdethier, line 739: Same comment about counts

iaso/tests/api/stocks/test_views.py
```diff
@@ -0,0 +1,886 @@
...
+    def test_create_without_authorization(self):
+        self.client.force_authenticate(self.user_without_rights)
+
+        response = self.client.post(
+            STOCK_ITEM_URL,
+            data={"sku": self.sku_2.id, "org_unit": self.org_unit_1.id, "value": 10},
+        )
+        self.assertJSONResponse(response, 403)
+
+    def test_create_with_authorization(self):
```
tdethier, line 314: I would add a test that triggers the error when there's already a `StockItem` for a given sku & org unit

iaso/tests/api/stocks/test_views.py
```diff
@@ -0,0 +1,886 @@
...
+
+    def test_delete_for_finalized_version(self):
+        self.client.force_authenticate(self.user_with_rights)
+        response = self.client.delete(f"{RULE_ITEM_URL}{self.stock_item_rule_2.pk}/")
+        self.assertJSONResponse(response, 403)
+
+
+class StockRulesVersionAPITestCase(APITestCase):
+    @classmethod
+    def setUpTestData(cls):
```
tdethier, line 785: Maybe a common base for all view test classes would be good too

iaso/tests/api/stocks/test_views.py
```diff
@@ -0,0 +1,886 @@
...
+
+        response = self.client.patch(
+            f"{RULES_VERSION_URL}{self.version_1.pk}/", data={"status": m.StockRulesVersionsStatus.PUBLISHED}
+        )
+        self.assertJSONResponse(response, 200)
+
+        self.assertEqual(1, m.StockRulesVersion.objects.filter(status=m.StockRulesVersionsStatus.PUBLISHED).count())
+        response = self.client.get(f"{RULES_VERSION_URL}{self.version_1.pk}/")
+        self.assertEqual(response.json()["status"], m.StockRulesVersionsStatus.PUBLISHED)
+        self.assertEqual(m.StockRulesVersion.objects.get(pk=version.pk).status, m.StockRulesVersionsStatus.UNPUBLISHED)
```
tdethier, line 935: I would also make some asserts on python objects here

iaso/tests/api/stocks/test_views.py
```diff
@@ -0,0 +1,886 @@
...
+        response = self.client.get(f"{RULES_VERSION_URL}{self.version_1.pk}/")
+        self.assertJSONResponse(response, 200)
+        self.assertIsNotNone(response.json()["deleted_at"])
+
+    def test_copy_without_rights(self):
+        self.client.force_authenticate(self.user_without_rights)
+        response = self.client.post(f"{RULES_VERSION_URL}{self.version_1.pk}/copy/")
+        self.assertJSONResponse(response, 403)
+
+    def test_copy_with_rights(self):
```
tdethier, line 961: I would add some asserts on counts here

iaso/tests/api/stocks/test_views_mobile.py
```diff
@@ -0,0 +1,414 @@
...
+from iaso import models as m
+from iaso.test import APITestCase
+
+
+SKU_URL = "/api/mobile/stockkeepingunits/"
+LEDGER_ITEM_URL = "/api/mobile/stockledgeritems/"
+RULES_VERSION_URL = "/api/mobile/stockrulesversions/"
+
+
+class StockRulesVersionMobileAPITestCase(APITestCase):
```
tdethier:
```suggestion
class StockKeepingUnitMobileAPITestCase(APITestCase):
```

iaso/tests/api/stocks/test_views_mobile.py
```diff
@@ -0,0 +1,414 @@
...
+        cls.ledger_item = m.StockLedgerItem.objects.create(
+            sku=sku_2,
+            org_unit=org_unit_2,
+            submission=instance,
+            question="question_name_4",
+            value=10,
+            impact=m.StockImpacts.ADD,
+            created_by=cls.user_without_rights,
+            created_at=datetime.now(),
+        )
```
tdethier: Shouldn't you suffix them instead of overwriting `cls.ledger_item`?

tdethier (commented):
Damn this was a huge PR.

Kudos for this feature, it's a lot of work and it was nicely done 👍 

I didn't test this (only looked at the code) because I don't know how to setup the mobile app - some more feedback will probably be made once the web interface is done.
Note: I don't know why, but without any data, both `api/stockledgeritems/` and `api/stockitems/` endpoints never load - I don't even see logs in the backend. I guess it won't be an issue once there is some data.

I made (a lot of) suggestions and comments. Some of them were probably because I didn't fully grasp the business side behind all of this, some of them are simply questions. Please feel free to reach out to me to discuss this

@quang-le I don't know if you have some time to review this, but it would be interesting, at least for the business side

iaso/models/stocks.py
bmonjoie: I have thought about that but I don't think we ever return translated values from the backend. This is all going to be in the FE.

iaso/models/stocks.py
bmonjoie: Projects can be empty. Then, how do you know which account this SKU belongs to?

iaso/models/stocks.py
bmonjoie: > I don't understand the referencing StockItemRule part because there's a FK to that model below

Yeah, that's why I hate comments, I made that comment before I changed the logic...

> If this has to be immutable, I would probably enforce it at the model level to be sure: maybe by overriding .save(), checking if there's a primary key/ID and throwing an error if that's the case

I didn't know we could do that :upside_down_face:

iaso/models/stocks.py
bmonjoie: `created_at` is the value of when it was created on the mobile, `received_at`  is the value the server received it.

iaso/models/stocks.py
bmonjoie: Same answer :upside_down_face:

iaso/admin/base.py
bmonjoie: In the admin, I think we should still have the possibility to correct an error.

iaso/api/stocks/serializers.py
bmonjoie: No because that could have changed. The OrgUnitTypes is an indication to the mobile not to apply rules for all OU but when the mobile sends its items, the backend should not decline them.

iaso/api/stocks/serializers.py
bmonjoie: I guess we could but is that going to add any safety or added value?

iaso/tests/api/stocks/test_serializers.py
bmonjoie: They look similar but they are slightly different to test in isolation. I guess we could but then I'm afraid that looking at the `setupTestData` method is going to be full of irrelevant data for some tests and make it harder to grasp.

iaso/api/stocks/utils.py
bmonjoie: It cannot because it is used because it is used by the zip file importer which doesn't set the request and other information to the serializer.

iaso/api/stocks/utils.py
bmonjoie: I think I removed that and forgot to remove it here

iaso/tests/api/stocks/test_views.py
bmonjoie: Same answer

iaso/models/stocks.py
bmonjoie: I copied that from @mathvdh 's code for entities. All credits go to him :slightly_smiling_face:

iaso/api/stocks/serializers.py
bmonjoie: > - the user has access to these org unit types
> -  the user has access to these forms

I guess we could

> - the forms are linked to the org unit types? I'm not sure about this one

No, because forms are linked to Org Unit Types only for the mobile to display them in a specific place. Restricting it there would be a mistake

iaso/api/stocks/utils.py
bmonjoie: They are tested through the ViewSet tests

iaso/api/stocks/use_cases.py
bmonjoie: > I guess a StockLedgerItem with impact=StockImpacts.RESET should always have a value of 0? Because if it's positive, it will be added into the total here

No, it can have whatever value and it will indeed be added to the total.

bmonjoie: > When I read tests I realize that I didn't quite understand the business logic here 😅

The goal of this method is to compute the current stock value, I know, shocker, right? :stuck_out_tongue: 

To do so, we need to find the latest reset (if there is one), take its value and add/subtract the following ledger item values.

iaso/tests/api/stocks/test_use_cases.py
bmonjoie: Yeah, I left this there in case we use it with Enketo but I don't use it anywhere yet because I was asked not to. Therefore, I suppose I should remove it and make something clean when we add this to Enketo

iaso/api/stocks/views.py
bmonjoie: No, not for StockLedgerItem.

iaso/api/stocks/views.py
bmonjoie: Same remark as before

iaso/api/stocks/views.py
bmonjoie: If you create or update the StockItem directly, then yes.

iaso/api/stocks/views.py
bmonjoie: Same answer

iaso/api/stocks/views.py
bmonjoie: Yes, see my answer about the difference between `created_at` and `received_at`

iaso/api/stocks/serializers.py
tdethier: I believe it will be more safe, yeah, but maybe we can ask others for their opinion

iaso/models/stocks.py
tdethier: Oh, so it's like `OrgUnit.created_at` and `OrgUnit.source_created_at`? If so, I would maybe use the same name. Otherwise I would add your comment in the code

iaso/models/stocks.py
tdethier: These are used in the admin panel. I'm not sure if the FE uses it (or plans to use it at some point).

iaso/tests/api/stocks/test_serializers.py
```diff
@@ -0,0 +1,736 @@
...
+            "sku": self.sku.pk,
+            "submission_id": self.instance.id,
+            "question": "question_name",
+            "value": 10,
+        }
+        serializer = StockLedgerItemWriteSerializer(data=data, context={"request": request})
+        self.assertFalse(serializer.is_valid())
+        self.assertIn("This field is required.", serializer.errors["impact"][0])
+
+    @unittest.expectedFailure
```
tdethier: That works, but if you use `self.assertRaises(SomeException) as e`, you can also make sure you capture the right type of exception and check its content

iaso/models/stocks.py
```diff
@@ -0,0 +1,313 @@
...
+    def __str__(self):
+        return f"{self.name} ({self.short_name})"
+
+
+class StockKeepingUnitChildren(models.Model):
+    parent = models.ForeignKey(
+        "StockKeepingUnit", related_name="parent", on_delete=models.CASCADE, null=False, blank=False
+    )
+    child = models.ForeignKey(
+        "StockKeepingUnit", related_name="child", on_delete=models.CASCADE, null=False, blank=False
```
tdethier:
```suggestion
        "StockKeepingUnit", related_name="parents", on_delete=models.CASCADE, null=False, blank=False
```

iaso/models/stocks.py
```diff
@@ -0,0 +1,313 @@
...
+
+    objects = models.Manager.from_queryset(StockKeepingUnitQuerySet)()
+
+    def __str__(self):
+        return f"{self.name} ({self.short_name})"
+
+
+class StockKeepingUnitChildren(models.Model):
+    parent = models.ForeignKey(
+        "StockKeepingUnit", related_name="parent", on_delete=models.CASCADE, null=False, blank=False
```
tdethier: Here, the reverse is not really relevant because we already have `StockKeepingUnit.children`, so we could:
- use  `related_name="+"` to tell Django not to create the reverse relation
- use `related_name="children_reverse"` (because we're reversing the parent relation)
- not specify anything (Django will set it as `stockkeepingunitchildren_set`)

iaso/tests/api/stocks/test_views.py
tdethier: I think we really should have a test for that

tdethier (changes requested):
Thanks for all the changes :+1: 

I'm requesting changes for 2 small things: 
- reverse names in `StockKeepingUnitChildren`
- missing test for making sure there's an error when creating a `StockItem` with the same SKU & OrgUnit as another `StockItem`

Otherwise it's good for me

iaso/models/stocks.py
tdethier: I prefer the `related_name="+"` version

iaso/models/stocks.py
bmonjoie: Strange, my other comment disappeared. I said that I got confused and indeed, it's in the admin. I believe we can let the admins do whatever they want, no?

tdethier (approved):
I still don't fully understand the business logic, but the backend code looks good :+1:

iaso/models/stocks.py
madewulf: Traductions are not needed for /admin, I would say

PR #2399, opened 2025-08-26
IA-4144 Translate the payments XLS/CSV file

kemar:
Translate the payments XLS/CSV file.

Related JIRA tickets : IA-4144

## Changes

The frontend is using `ExternalLinkIconButton` which creates a direct browser link, thus bypassing the API client in `Api.ts` that adds the `Accept-Language` header.

So the Django backend receives a request **without** the `Accept-Language` header, and `get_language_from_request()` defaults to the `LANGUAGE_CODE = "en"` from `settings.py`.

The implemented solution passes the language as a query parameter.

## How to test

- you need some payment lots
- then go to `/dashboard/payments/lots/`
- then download the XLS file once in EN and once in FR, it should be translated

## Print screen / video

https://github.com/user-attachments/assets/5a45ad10-dbf0-4b2c-81a3-bd4473babf73


Bewi (commented):
LGTM.
Ran locally and the file is translated according to selected language.

Only doubt I have is if wouldn't be better to update the FE component so the Accept-Language header is added. But that'd be more work for this case.

Bewi (approved):

PR #2401, opened 2025-08-27
IA-4444 Exclude unsupported fields types from the duplicates detection

kemar:
Exclude unsupported fields types from the duplicates detection.

Related JIRA tickets : IA-4444

## Changes

- add field type validation in both:
    - the duplicates analysis serializer
    - the form API
- use a centralized constant for the validation logic
- fix `EntityType.DoesNotExist` exception handling (was using `Entity.DoesNotExist`)

## How to test

The first thing you can do is look at the unit tests.

To test the endpoint used by the frontend:

- look at the details of a form in the same way that the UI is calling the endpoint
- any field that has a type not compatible with the duplicates detection should not be listed in the result

```
http://localhost:8081/api/forms/{ID}/?fields=possible_fields_with_latest_version,name,latest_form_version
```

The backend code double checks that rule.

### DIFF
--- a/iaso/api/deduplication/entity_duplicate_analyzis.py
+++ b/iaso/api/deduplication/entity_duplicate_analyzis.py
@@ -30,10 +30,8 @@ class AnalyzePostBodySerializer(serializers.Serializer):
         possible_fields = e_type.reference_form.possible_fields
 
         for f_name in data["fields"]:
-            try:
-                field = [f for f in possible_fields if f["name"] == f_name]
-                field = field[0]
-            except IndexError:
+            field = next((f for f in possible_fields if f["name"] == f_name), None)
+            if not field:
                 raise serializers.ValidationError(f"Field `{f_name}` does not exist on reference form.")
 
             if field["type"] not in EntityDuplicateAnalyzis.SUPPORTED_FIELD_TYPES:
--- a/iaso/api/deduplication/entity_duplicate_analyzis.py
+++ b/iaso/api/deduplication/entity_duplicate_analyzis.py
@@ -8,20 +8,9 @@ import iaso.models.base as base
 import iaso.permissions as core_permissions
 
 from iaso.api.common import HasPermission, ModelViewSet, Paginator
-from iaso.models import Entity, EntityDuplicateAnalyzis, EntityType, Form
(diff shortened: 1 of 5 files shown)


iaso/api/deduplication/entity_duplicate_analyzis.py
```diff
@@ -34,13 +23,21 @@ def validate(self, data):
...
+        possible_fields = e_type.reference_form.possible_fields
 
         for f_name in data["fields"]:
-            if not field_exists(e_type.reference_form, f_name):
-                raise serializers.ValidationError(f"Field {f_name} does not exist on reference form")
+            try:
+                field = [f for f in possible_fields if f["name"] == f_name]
+                field = field[0]
+            except IndexError:
+                raise serializers.ValidationError(f"Field `{f_name}` does not exist on reference form.")
```
tdethier: Nitpick: I'm not a big fan of this way of checking if a field exists, but it works

tdethier (approved):
LGTM :+1:  

The list of possible fields when preparing a new analysis is now greatly reduced on setuper accounts. I guess this will surprise some users ("_why are my fields gone?_") and we'll probably need to tell them that some of their forms might not be properly configured

iaso/api/deduplication/entity_duplicate_analyzis.py
kemar: What would you suggest as an alternative?

tdethier: I did not really think about a cleaner solution.

My first impression would be to call a function instead, but it very similar to the previous code.
Since this is already in a `for` loop, maybe there's an easy way to get the common elements of both lists (`data["fields"]` and `possible_fields`), and after that check for field type... I don't know if it would be much better though

kemar: Does this better suit your expectations? https://github.com/BLSQ/iaso/pull/2401/commits/d5735976a0b3ff1569b17085ca5eba80d8db1f1b

tdethier: Much better, thanks @kemar :smile:

PR #2403, opened 2025-08-29
IA-3684 create ou children not working

beygorghor:
Following this [PR](https://github.com/BLSQ/iaso/pull/2373), create a children of an org unit is not working on save, parent is not saved. Also add children button is present only if we have children already

Related JIRA tickets : IA-3684

- [ ] Are there enough tests?

## Doc

-

## Changes

- always display add button
- pass correct parent org unit id on save if is new org unit, has the param and parent has not been changed

## How to test

- create a children org unit on children tab
- try once using selected parent
- try once with changing the parent
- try to edit and change the parent

## Print screen / video

https://github.com/user-attachments/assets/80a9acaa-f973-4703-bd9b-294a4e26815b

## Notes

-

```
fix: empty instance pop up

Refs: IA-3665
```


Bewi (approved):
Tested locally with and without children, button is available.
I could create the org unit from that button and parent is set properly.
I could create a new org unit from the list and parent is also set properly
I could create a new org unit with no parent

LGTM

PR #2404, opened 2025-08-29
IA-4194 Activate the most recently used account for multi-account users

kemar:
Activate the most recently used account for multi-account users.

Related JIRA tickets: IA-4194

See also #2411 

## Changes

Add an [authentication backend](https://docs.djangoproject.com/fr/4.2/topics/auth/customizing/#specifying-authentication-backends) for tenant users.

This `MultiTenantAuthBackend` use `last_login` to identify the most recently used tenant user.

This works because when users switch accounts, the system calls `login()` which automatically updates `last_login` via the Django built-in signal system.

## How to test?

- login with a multi-accounts user
- switch to another project
- logout
- login again
- you should be connected to the most recently used account

## Print screen / video

https://github.com/user-attachments/assets/23e995c2-25d7-40ca-abb6-1fa9053e43fe

### DIFF
--- a/hat/settings.py
+++ b/hat/settings.py
@@ -711,7 +711,6 @@ HIDE_BASIC_NAV_ITEMS = os.environ.get("HIDE_BASIC_NAV_ITEMS", "no")
 # Django tries the second one, and so on…
 AUTHENTICATION_BACKENDS = [
     "iaso.auth.backends.MultiTenantAuthBackend",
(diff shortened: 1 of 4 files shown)


tdethier (approved):
LGTM :+1: 

I agree that it would be great to define somewhere a "favorite account" to which you automatically login, but this is already a nice feature

PR #2408, opened 2025-09-01
IA-4445 account setup v2

beygorghor:
Improving account setup endpoint and ui to create advanced account.

- possibility to select a language
- sent an email to the user created or not to set the password
- prefill modules
- create an Main org unit type
- create a validated org unit
- create a dummy form

Related JIRA tickets : IA-4445

- [ ] Are there enough tests?

## Doc

-

## Changes

- rewrite api endpoint
- modify react ui to reflect those changes
- log account creation 
- more tests

## How to test

- create an admin user without iaso profie
- connect to the dashboard with this user
- fill up the setup account form and submit it
- connect to the dashboard with this user
- make sure everything is in place and that you can use directly the mobile app
- go to the admin, select modifications, make sure you can see a log for the creation of the account

## Print screen / video

https://github.com/user-attachments/assets/aede82db-61f3-4cb4-a058-35c51e5f01a3

## Notes

- 

```
fix: empty instance pop up

Refs: IA-3665
```


hat/assets/js/apps/Iaso/domains/setup/index.tsx
```diff
@@ -68,12 +68,15 @@ export const SetupAccount: FunctionComponent = () => {
...
-            user_first_name: '',
-            user_last_name: '',
-            password: '',
-            modules: [],
+            account_name: undefined,
+            user_username: undefined,
+            user_first_name: undefined,
+            user_last_name: undefined,
+            user_email: undefined,
+            password: undefined,
```
kemar, line 76: Is `undefined` really a better solution than the empty string?

I'm asking because I know it's almost never a good idea to use `undefined` in JavaScript since it can be overridden.

iaso/api/setup_account.py
```diff
@@ -29,13 +39,22 @@ class SetupAccountSerializer(serializers.Serializer):
...
     user_first_name = serializers.CharField(max_length=30, required=False)
     user_last_name = serializers.CharField(max_length=150, required=False)
-    password = serializers.CharField(required=True)
+    user_email = serializers.EmailField(required=False)
+    password = serializers.CharField(required=False)
     user_manual_path = serializers.CharField(required=False)
-    modules = serializers.JSONField(required=True, initial=["DEFAULT"])  # type: ignore
+    email_invitation = serializers.BooleanField(required=False, default=False)
+    language = serializers.ChoiceField(
+        choices=[("en", "English"), ("fr", "Français")],
```
kemar:
```suggestion
        choices=settings.LANGUAGES
```

We should probably rely on the configured languages here rather than hardcoding them.

iaso/api/setup_account.py
```diff
@@ -105,15 +162,151 @@ def create(self, validated_data):
...
+            "user_email": request.data.get("user_email", ""),
+            "email_invitation": request.data.get("email_invitation", False),
+            "language": request.data.get("language", "en"),
+            "modules": request.data.get("modules", []),
+            "feature_flags": request.data.get("feature_flags", []),
+            "requesting_user": request.user.username if request.user else None,
+            "requesting_user_id": request.user.id if request.user else None,
+        }
+
+        # Parse JSON fields that might come as strings
```
kemar: The `parse_json_field()` method just feels strange to me.

When are JSON fields coming as strings?

iaso/tests/api/test_setup_account.py
```diff
@@ -108,12 +123,214 @@ def test_setup_account_create_with_first_last_name(self):
...
+        valid_emails = [
+            "test@example.com",
+            "user.name@domain.co.uk",
+            "user+tag@example.org",
+            "123@numbers.com",
+            "user@subdomain.example.com",
+        ]
+
+        for email in valid_emails:
+            with self.subTest(email=email):
```
kemar: You can tell you AI that we don't need `subTest()` here 😉

kemar (changes requested):
Almost good.

There is just `parse_json_field()` that feels strange to me. Can you answer my question about it please?

hat/assets/js/apps/Iaso/domains/setup/index.tsx
beygorghor: Keeping an empty string is sending an empty string to the backend, backend is then saving an empty string and shouldn't

iaso/api/setup_account.py
beygorghor: using /api/setupaccount, while creating an account with default value for modules and feature flags, then it saves special chars in the log

kemar (approved):
LGTM

I tend to think though that the `undefined` and `parse_json_field` should be handled at the serializer level.

PR #2410, opened 2025-09-02
IA-4465: displays setup account for super user / staff  in menu

beygorghor:
Setup account page should be accessible in the menu of a superuser too, or staff user

Related JIRA tickets : IA-4465

- [ ] Are there enough tests?

## Changes

- adding entry in the menu / routes if user is super user or staff member

## How to test

-connect with a super user / staff member, make sure setup account page is available in the menu under settings

## Print screen / video

-

## Notes

```
fix: empty instance pop up

Refs: IA-3665
```


tdethier (approved):
The feature works fine, LGTM :+1: 

I only have one remark: it's not caused by this PR, but I would definitely change the FR label for the new menu entry, because it's very confusing. It looks like you're going to configure your own account instead of creating a new one:
image

`Creation d'un nouveau compte` would make much more sense

PR #2411, opened 2025-09-02
IA-4195 Fix multi-tenant authentication bug

kemar:
Fix multi-tenant authentication bug.

Related JIRA tickets: IA-4195

## Changes

This bug was introduced in #2404 

The bug was caused by redundant authentication logic.

The `MultiTenantAuthBackend` already switches from `main_user` to `account_user` during login, but the `profiles.py` was trying to do this again, creating session conflicts.

## How to test?

Just like in #2404 

- login with a multi-accounts user
- switch to another project
- logout
- login again
- you should be connected to the most recently used account


tdethier (approved):
LGTM :+1: 

The initial feature still works locally, but I never was able to reproduce the staging issue on my computer. Hopefully, this will fix it too.

PR #2413, opened 2025-09-03
WC2-818 Optimise entities list for large datasets

kemar:
Optimise entities list for large datasets.

Related JIRA tickets: WC2-818

## Changes

- replace expensive `ArrayAgg` operations with `boolean` + `count` in `EntityQuerySet.with_duplicates()`
- entity list API fields:
    - add `has_duplicates` and `duplicate_count`
    - remove `duplicates`
- update frontend to use boolean flags instead of array operations

This should reduce memory usage and database load as well as network transfer (boolean + int vs large arrays of integers).

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/entities/config.tsx
+++ b/hat/assets/js/apps/Iaso/domains/entities/config.tsx
@@ -144,21 +144,23 @@ export const useColumns = (
                             icon="remove-red-eye"
                             tooltipMessage={MESSAGES.see}
                         />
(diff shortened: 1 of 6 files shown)


tdethier (approved):
I didn't have the right setup to notice the performance difference (I had 125k entities, but without duplicates, so loading each list of duplicate IDs wasn't slow since each list was empty).

However, based on the code only, the performance should be better :+1: 

Let's test that on staging/QA!

PR #2416, opened 2025-09-03
IA-4457: Fix setuper - form and projects

butofleury:

Related JIRA tickets : [IA-4457](https://bluesquare.atlassian.net/browse/IA-4457)

- [ ] Are there enough tests?

## Doc

## Changes

- Fix form update when adding project to form via setuper
- Fixed entity type creation 

## How to test

From setuper folder, launch setuper via python3 setuper.py --additional_projects to generate test data. Then login with the created account open pages:
- [Forms](http://localhost:8081/dashboard/forms/list/accountId/1/isSearchActive/true)
- Projects
- Entity types

## Print screen / video

## Notes

```
fix: empty instance pop up

Refs: IA-3665
```

[IA-4457]: https://bluesquare.atlassian.net/browse/IA-4457?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iaso/models/forms.py
+++ b/iaso/models/forms.py
@@ -355,6 +355,14 @@ class FormVersion(models.Model):
     def __str__(self):
         return "%s - %s - %s" % (self.form.name, self.version_id, self.created_at)
 
+    def as_dict(self):
+        return {
+            "id": self.id,
+            "version_id": self.version_id,
+            "created_at": self.created_at.timestamp() if self.created_at else None,
+            "updated_at": self.updated_at.timestamp() if self.updated_at else None,
+        }
+
 
 def update_possible_fields(sender, instance, **kwargs):
     if not instance.form_descriptor and instance.xls_file:
--- a/iaso/models/forms.py
+++ b/iaso/models/forms.py
@@ -355,14 +355,6 @@ class FormVersion(models.Model):
     def __str__(self):
         return "%s - %s - %s" % (self.form.name, self.version_id, self.created_at)
 
-    def as_dict(self):
-        return {
-            "id": self.id,
-            "version_id": self.version_id,
-            "created_at": self.created_at.timestamp() if self.created_at else None,
-            "updated_at": self.updated_at.timestamp() if self.updated_at else None,
-        }
-
 
 def update_possible_fields(sender, instance, **kwargs):
     if not instance.form_descriptor and instance.xls_file:
--- a/setuper/additional_projects.py
+++ b/setuper/additional_projects.py
@@ -1,4 +1,5 @@
 import random
+
 from create_llin_campaign_forms_submissions import llin_forms
 
 
--- a/setuper/additional_projects.py
+++ b/setuper/additional_projects.py
@@ -65,7 +65,11 @@ def get_project_ids(created_or_updated_projects, iaso_client):
     existing_projects = iaso_client.get("/api/projects/")["projects"]
     project_ids = []
     for current_project in created_or_updated_projects:
-        project_id = [project for project in existing_projects if project["app_id"] == current_project["id"]]
+        project_id = [
+            project
+            for project in existing_projects
+            if project["app_id"] == current_project["id"]
+        ]
         project_ids.append(project_id[0]["id"])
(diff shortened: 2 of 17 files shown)


iaso/models/forms.py
```diff
@@ -355,6 +355,14 @@ def questions_by_path(self):
...
         return "%s - %s - %s" % (self.form.name, self.version_id, self.created_at)
 
+    def as_dict(self):
+        return {
+            "id": self.id,
+            "version_id": self.version_id,
+            "created_at": self.created_at.timestamp() if self.created_at else None,
+            "updated_at": self.updated_at.timestamp() if self.updated_at else None,
+        }
+
```
tdethier: Do you really need that? I don't understand

```suggestion
```

setuper/additional_projects.py
```diff
@@ -120,8 +120,16 @@ def link_forms_to_new_projects(projects, forms, iaso_client):
...
-            current_form["project_ids"] = project_ids
-            iaso_client.patch(f"/api/forms/{current_form['id']}/", json=current_form)
+            form_to_update = {
+                "project_ids": project_ids,
+                "name": current_form["name"],
+                "org_unit_type_ids": current_form["org_unit_type_ids"],
+                "derived": current_form["derived"],
+                "single_per_period": current_form["single_per_period"],
+                "periods_before_allowed": current_form["periods_before_allowed"],
+                "periods_after_allowed": current_form["periods_after_allowed"],
```
tdethier: Do you really need to add all these fields?

tdethier (changes requested):
There's another bug that we need to fix with the setuper, I'm requesting changes to have this extra fix directly in this PR

iaso/models/forms.py
butofleury: yes, there was a bug when creating entity type, the object is created but it return `'FormVersion' object has no attribute 'as_dict'`

Mybe It should be done in a separated ticket

tdethier: Can you check again without `as_dict` with `main` rebased/merged into your branch? I removed that `as_dict` in another PR and I think we should no longer need it, but maybe I forgot something in the API

butofleury: When I check without `as_dict` , I still get the issue. 

Capture d’écran du 2025-09-04 14-37-20


You might forgot something in the API.

tdethier: This is `Form.as_dict()`:

```
def as_dict(self, additional_fields=None, show_version=True):
        res = {
            "form_id": self.form_id,
            "name": self.name,
            "id": self.id,
            "org_unit_types": [t.as_dict() for t in self.org_unit_types.all()],
            "created_at": self.created_at.timestamp() if self.created_at else None,
            "updated_at": self.updated_at.timestamp() if self.updated_at else None,
            "period_type": self.period_type,
            "single_per_period": self.single_per_period,
            "label_keys": self.label_keys,
        }

        if show_version:
            res["latest_form_version"] = self.latest_version.as_dict() if self.latest_version is not None else None
        if additional_fields:
            for field in additional_fields:
                if hasattr(self, field):
                    res[field] = getattr(self, field)

        return res
```

If we call it with `show_version=False` we don't need `FormVersion.as_dict()` (we should always use serializers instead of `as_dict()` methods).

This is called by `EntityType.as_dict()`:

```
    def as_dict(self):
        return {
            "name": self.name,
            "created_at": self.created_at,
            "updated_at": self.updated_at,
            "reference_form": self.reference_form.as_dict() if self.reference_form else None,
            "account": self.account.as_dict(),
        }
```

Do we need data about the form version there? We could pass `show_version=False` there and it would fix our issue, but I'm not sure if that's going to create an issue somewhere else.

If it doesn't, it's worth a shot. Otherwise, I'd say you can add back this `FormVersion.as_dict()` in order to have a quick fix

tdethier (approved):
Good job, this was not an easy one! 

It's working now
image

PR #2418, opened 2025-09-04
IA-4443: Create a filter "Period" on Registry page

butofleury:

Related JIRA tickets : [IA-4443](https://bluesquare.atlassian.net/browse/IA-4443)

- [ ] Are there enough tests?

## Doc

## Changes

Added a periodicity filter on [Registry page](http://localhost:8081/dashboard/registry/accountId/1/orgUnitId/1/formIds/3) to filter submissions registry data according to the period type linked to the forms.

## How to test

Check if there is a form linked to Georegistry project with periodicity and try to enter data for the form. Then open [Registry page](http://localhost:8081/dashboard/registry/accountId/1/orgUnitId/1/formIds/3), start to filter the data linked to the forms by form and period.

## Print screen / video

[Registry-Page.webm](https://github.com/user-attachments/assets/68883780-7aba-4048-8539-bf4bfcd59043)

## Notes

```
fix: empty instance pop up

Refs: IA-3665
```

[IA-4443]: https://bluesquare.atlassian.net/browse/IA-4443?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Bewi (commented):
I wonder if when an OrgUnit has no "period", it should be part of the result even if we have a filter applied on Period.
From my point of view, no period means matching all periods.
What do you think ?

So if I have something like this: 
image

When I filter like this:
image

Should have results IMO. 

So this is if I understood properly the scope of this filter :D

butofleury:
> I wonder if when an OrgUnit has no "period", it should be part of the result even if we have a filter applied on Period. From my point of view, no period means matching all periods. What do you think ?
> 
> So if I have something like this: image
> 
> When I filter like this: image
> 
> Should have results IMO.
> 
> So this is if I understood properly the scope of this filter :D


No Period means to me there is no periodicity defined for the form 

image

It sounds good, when there is no selected periodicity, to get all submissions linked to the form. But when the period type is selected, we should get all submissions linked to the form with the related defined periodicity in the selected start and end period.

@beygorghor, what do you think?

beygorghor:
> > I wonder if when an OrgUnit has no "period", it should be part of the result even if we have a filter applied on Period. From my point of view, no period means matching all periods. What do you think ?
> > So if I have something like this: image
> > When I filter like this: image
> > Should have results IMO.
> > So this is if I understood properly the scope of this filter :D
> 
> No Period means to me there is no periodicity defined for the form
> 
> image
> It sounds good, when there is no selected periodicity, to get all submissions linked to the form. But when the period type is selected, we should get all submissions linked to the form with the related defined periodicity in the selected start and end period.
> 
> @beygorghor, what do you think?

For me if the field is empty means all submissions.  No period, submissions without a period config, ....

Bewi (approved):
LGTM

PR #2433, opened 2025-09-12
[IA-4364] Permissions & modules 2.0

tdethier:
Improved permissions and modules using python objects

Related JIRA tickets : IA-4364

- [x] Are there enough tests?

## Doc

/

## Changes

- Replaced all `str` permissions by python objects
- Replaced all `str` modules by python objects
- Both now interact with each other instead of having dicts with copy pasted values everywhere

## How to test

- Start the app and go to various pages, nothing should change from the standard behavior

## Print screen / video

/

## Notes

- Thanks @kemar for suggesting using python objects!
- Both core + polio are going to be done in this PR
- A similar PR will need to be done in Trypelim once this has been merged

```
fix: empty instance pop up

Refs: IA-3665
```


iaso/tests/api/org_unit_change_requests/test_org_unit_change_requests_mobile.py
```diff
@@ -68,14 +68,10 @@ def setUpTestData(cls):
         m.OrgUnitReferenceInstance.objects.create(org_unit=cls.org_unit_a, form=cls.form_1, instance=cls.instance_1)
         m.OrgUnitReferenceInstance.objects.create(org_unit=cls.org_unit_a, form=cls.form_2, instance=cls.instance_2)
 
-        cls.user = cls.create_user_with_profile(
-            username="user", account=cls.account, permissions=["iaso_org_unit_change_request"]
-        )
+        cls.user = cls.create_user_with_profile(username="user", account=cls.account, permissions=[])
```
tdethier, line 71: `iaso_org_unit_change_request` is an old permission that no longer exists, so I just set an empty list of permissions in that test

hat/menupermissions/models.py
```diff
@@ -47,60 +40,8 @@ class CustomPermissionSupport(models.Model):
...
-        core_permissions._MAPPINGS,
-        core_permissions._COMPLETENESS,
-        core_permissions._ORG_UNITS,
-        core_permissions._LINKS,
-        core_permissions._USERS_ADMIN,
-        core_permissions._PROJECTS,
-        core_permissions._SOURCES,
-        core_permissions._DATA_TASKS,
-        core_permissions._REPORTS,
-    ]
```
tdethier, line 67: This was dead code, so I just removed it

iaso/permissions/base.py
```diff
@@ -0,0 +1,78 @@
...
+
+from iaso.modules import IasoModule
+from iaso.permissions.utils import load_permissions_from_module
+
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class IasoPermission(ABC):
```
kemar, line 19: Why are you using both `@dataclass` and `ABC`?

My first impression is that it looks a bit like an anti-pattern because:

- `@dataclass` is for data containers
- `ABC` is for interface contracts and behaviour definition
- combining them creates hybrid objects that are neither pure data nor pure interfaces

But I may be missing something? Is it just because of the `full_name` method?

In this case, pure `ABC` may be the cleanest fix.

iaso/permissions/base.py
```diff
@@ -0,0 +1,78 @@
...
+    def __str__(self):
+        return self.name
+
+    @abstractmethod
+    def full_name(self) -> str:
+        """
+        Returns the full name of the permission, which is the name prefixed with its application name.
+        This method should be called whenever you use any function that checks for permissions
+        (e.g. HasPermission() in the API).
+        """
```
kemar, line 50: This function currently returns nothing but it has a type annotation that clearly indicates it should return a string.

So this is a bit strange.

We probably should make it _concrete_ rather than _abstract_ because this function is always required and the iaso project cannot run without the core module.

Maybe add the exact expected format in the docstring too:

```
 "<django_app_label>.<permission_codename>"
```

If you choose to go abstract, you should return something: either `pass` or `NotImplementedError("foo")`.

iaso/permissions/base.py
```diff
@@ -0,0 +1,78 @@
...
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class IasoPermission(ABC):
+    """
+    Represents a permission in IASO.
+    """
+
+    name: str
```
kemar: Just wondering if `codename` wouldn't be a more accurate name?

Because I'm seeing this pattern constantly:

```python
perm = Permission.objects.get(codename=PERMISSION.name)
```

This won't change the face of the world but it may be better:

```python
perm = Permission.objects.get(codename=PERMISSION.codename)
```

iaso/permissions/base.py
```diff
@@ -0,0 +1,78 @@
...
+    label: StrOrPromise  # For translation purposes
+    module: Optional[IasoModule] = None
+    group: Optional[str] = None  # Represents a permission group in the web interface
+    category: Optional[str] = (
+        None  # Represents a category whenever there are read/write/admin/non-admin/... permissions
+    )
+    type_in_category: Optional[str] = None  # Represents a type in the category, e.g. "read", "write", "admin"...
+    order_in_category: Optional[int] = (
+        None  # Represents the order of the permission in its category (1 = lowest permissions)
+    )
```
kemar: Are these fields only used for the frontend?

If so, it could be a good idea to prefix them with `ui_`. Alternatively a simple comment could be helpful.

kemar (changes requested):
What a big PR! Kudos for the work.

The organization is way better with module definitions in `iaso/modules.py` and permission definitions in `iaso/permissions/`.

I have some feedback on the form but they are not blocking. You can implement them if you think they add value.

The only thing that is blocking is mixing both `@dataclass` and `ABC`.

This feels like an anti-pattern and there must be a better way by choosing one or the other.

iaso/permissions/base.py
tdethier: I initally went for `dataclass` but realized that `full_name` has to be abstract, yeah. I'll remove the `@dataclass` annotations everywhere

tdethier: So, after doing some changes and discussing them with @kemar, it looks like the current version is the best for now

kemar (approved):
LGTM 👍

PR #2434, opened 2025-09-15
IA-4482 Fix search field in users list not working with some checkboxes

kemar:
When the `ouChildren` checkbox was checked, the backend code was using `original_queryset` instead of the already-filtered `queryset`.

Related JIRA tickets: IA-4482

## Changes

- fix backend issue where search results were overridden when `ouChildren`/`ouParent` checkboxes were checked
- prevent `location` filtering from interfering with parent/children organizational unit filtering
- simplify filtering logic using `Q` objects
- add unit test

## How to test

See the attached video:

https://github.com/user-attachments/assets/b131d891-402c-440d-834f-ef5b5fd7dbc0

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/users/components/Filters.js
+++ b/hat/assets/js/apps/Iaso/domains/users/components/Filters.js
@@ -85,13 +85,21 @@ const Filters = ({ baseUrl, params, canBypassProjectRestrictions }) => {
     const handleChange = useCallback(
         (key, value) => {
             setFiltersUpdated(true);
-            if (key === 'location') {
-                setInitialOrgUnitId(value);
-            }
-            setFilters({
+            let updatedFilters = {
                 ...filters,
                 [key]: value,
-            });
+            };
+            if (key === 'location') {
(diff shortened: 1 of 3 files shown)


beygorghor (changes requested):
I'm still having a weird behavior in the frontend =>

https://github.com/user-attachments/assets/ffe63a60-0679-4bcc-996d-815083f59392

kemar:
> I'm still having a weird behavior in the frontend =>

@beygorghor can you explain this issue in more detail? I don't understand the problem.

What is the expected behaviour here?

beygorghor:
@kemar The check boxes are disabled by default, we don't know why, if you select an org unit they are enabled. If you remove the org unit i assume they should be disabled again, but they don't.

kemar:
> @kemar The check boxes are disabled by default, we don't know why, if you select an org unit they are enabled. If you remove the org unit i assume they should be disabled again, but they don't.

@beygorghor OK, I'll fix that so that they become disabled again!

beygorghor (approved):

PR #2440, opened 2025-09-17
IA-4395 OpenHEXA integration models

bramj:
This PR adds 2 new models to facilitate connecting OpenHEXA instances to IASO:

- `OpenHEXAInstance`: An OpenHEXA server
- `OpenHEXAWorkspace`: A workspace, belongs to an `openhexa_instance` and a IASO `account`

This way we can add (**for now just in the Django admin**) OpenHEXA servers like the main server, but also local hostings, and then connect IASO accounts to one or more OpenHEXA workspaces.
The workspace connection has a JSONB field `config` that allow you to freely set config for whatever you need.

I also took the opportunity to improve the `MetricType` and `MetricValue` admin to make it more useful as well as extract to its own file (the 1000+ `base.py` file is dreadfully large).

Related JIRA tickets : https://bluesquare.atlassian.net/browse/IA-4395

**Okay, but why?**

For Ginger and SNT we need to set up a more explicit connection to OH workspaces. By adding proper models we make the first step towards a cleaner integration, and provide more flexibility than regular ENV vars.

**Why the extra Python dependency?**

I really wanted to avoid this, but the token really had to be encrypted, and there appears to be no built-in Django way of doing this. The code from `EncryptedTextField` is an almost exact copy from OpenHEXA where it has been tried and true for several years. I added the `cryptography` Python package to be able to use `Fernet` for easy foolproof encryption. Since we don't have these skills, it seems like a safer bet than trying to roll our own using the more low-level `pycryptodome`. Since we need that last one for compatibility with the mobile app (encrypting zip setup files) that means we end up with 2 crypto dependencies, sorry about that...


beygorghor (approved):
Nice thank you ! I'll also ask @tdethier  to have a look on this too

.github/workflows/main.yml
```diff
@@ -153,22 +153,23 @@ jobs:
                   # check we don't have missing migration
                   . ./venv/bin/activate &&  python manage.py makemigrations --check
               env:
-                  IASO_ENVIRONMENT: development
-                  RDS_DB_NAME: github_actions
-                  SECRET_KEY: secret
+                  AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
```
tdethier, line 156: Thanks for ordering this 👍

iaso/admin/__init__.py
```diff
@@ -2,6 +2,8 @@
 from django.contrib.auth.models import User
 
 from .base import IasoJSONEditorWidget  # noqa: F401
+from .metric_admin import *
+from .openhexa_admin import *
```
tdethier, line 6: I would have imported the models directly instead of `*` but this works too

iaso/tests/models/test_openhexa.py
```diff
@@ -0,0 +1,124 @@
...
+            account=self.account,
+            slug="duplicate-slug",
+        )
+
+        with self.assertRaises(IntegrityError):
+            OpenHEXAWorkspace.objects.create(
+                openhexa_instance=self.openhexa_instance,
+                account=self.account,
+                slug="duplicate-slug",
+            )
```
tdethier, line 86: Nitpick: I would also check the error message to make sure it's due to a uniqueness error

iaso/utils/models/encrypted_text_field.py
```diff
@@ -0,0 +1,32 @@
...
+    def from_db_value(
+        self,
+        value: typing.Optional[str],
+        expression: typing.Optional[str],
+        connection: typing.Any,
+    ) -> typing.Optional[str]:
+        if value is None:
+            return value
+
+        return Fernet(settings.ENCRYPTED_TEXT_FIELD_KEY).decrypt(value.encode("utf-8")).decode("utf-8")
```
tdethier, line 20: I would maybe raise an exception if `settings.ENCRYPTED_TEXT_FIELD_KEY` is not defined or not in the right format, but this is probably not the right place to do that

iaso/utils/models/encrypted_text_field.py
```diff
@@ -0,0 +1,32 @@
...
+        self,
+        value: typing.Optional[str],
+        connection: typing.Any,
+        prepared: bool = False,
+    ) -> typing.Optional[str]:
+        db_prep_value = super().get_db_prep_value(value, connection, prepared)
+        if db_prep_value is None:
+            return None
+
+        return Fernet(settings.ENCRYPTED_TEXT_FIELD_KEY).encrypt(value.encode("utf-8")).decode("utf-8")
```
tdethier, line 32: Same comment about exception and format

iaso/models/openhexa.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class OpenHEXAInstance(models.Model):
+    class Meta:
+        verbose_name = "OpenHEXA Instance"
+        verbose_name_plural = "OpenHEXA Instances"
+
+    name = models.CharField(max_length=255, unique=True)
+    url = models.URLField()
+    token = EncryptedTextField()
```
tdethier, line 19: Why does the token need to be encrypted in database, but it's in plain text in the admin panel?

tdethier (approved):
The feature works fine 👍 

I like the idea of having dedicated models for this instead of a json config stored somewhere. I would have gone further and replaced `OpenHEXAWorkspace.config` by other models, depending on what needs to be stored there.

I don't mind the extra dependency for the custom encrypted field. It's good that this has already been battle-tested in OpenHEXA. Otherwise, I would have suggested using [this tool](https://pypi.org/project/django-fernet-encrypted-fields/), which basically does the same thing based on Fernet, except that you can also rotate your secret keys.

I just have a question regarding the token in the admin panel, you'll see below.

Looking forward to seeing this replacing all the config stuff.

iaso/models/openhexa.py
```diff
@@ -0,0 +1,53 @@
...
+        ]
+        unique_together = ("openhexa_instance", "account", "slug")
+        verbose_name = "OpenHEXA Workspace"
+        verbose_name_plural = "OpenHEXA Workspaces"
+
+    openhexa_instance = models.ForeignKey(
+        OpenHEXAInstance, on_delete=models.CASCADE, related_name="openhexa_workspaces"
+    )
+    account = models.ForeignKey("iaso.Account", on_delete=models.CASCADE)
+    slug = models.CharField(max_length=255)
```
quang-le:
```suggestion
    slug = models.SlugField(max_length=255)
```

https://docs.djangoproject.com/en/5.2/ref/models/fields/#slugfield

But CharField works just fine

quang-le (commented):
Nice work

iaso/utils/models/encrypted_text_field.py
tdethier: FYI: I'm reusing a similar setup in Pegasus, I added this check in the Django app config, inside the `ready()` method

Bewi (approved):
LGTM

iaso/tests/models/test_openhexa.py
bramj: Good point, will add that

iaso/models/openhexa.py
bramj: Interesting, didn't know about that one, I'll use it because it comes with some validation :+1:

iaso/utils/models/encrypted_text_field.py
bramj: Hm I see why you'd want this, but not sure where this should be added in IASO :thinking:

iaso/models/openhexa.py
bramj: You're right to question that! I don't think it's a big issue since the encryption is there in case the DB contents get lost/stolen.
But I'm changing it so Django admins can edit the token, but not see its value :+1:

iaso/utils/models/encrypted_text_field.py
tdethier: I'll ping you once the Pegasus PR is ready

tdethier: You can check https://github.com/BLSQ/console/pull/1/ - `apps/utils/apps.py`

tdethier:
I suggest deleting and recreating the migrations (we have 3 but only need 1)

PR #2445, opened 2025-09-19
IA-4374: [SaaS] Turn IASO setuper demo account into a sandbox

butofleury:

Related JIRA tickets : [IA-4374](https://bluesquare.atlassian.net/browse/IA-4374)

- [ ] Are there enough tests?

## Doc

## Changes

## How to test

Change `localhost` to `iaso` in the file `setuper/credentials.py`.  e.g: `SERVER = "http://iaso:8081"`

From command line tool launch: 
- open a tab and launch the task worker tasks_worker : `docker compose run iaso manage tasks_worker`
- open a second and launch again the task worker tasks_worker:  `docker compose run iaso manage tasks_worker`
- open a third tab to launch the setuper sandbox task to create or recreate(if it already exist) a sandbox account named "admin" with: `curl -d name=admin http://localhost:8081/tasks/launch_task/iaso.tasks.setuper_sandbox.setuper_sandbox/cron_task_sandbox_user/ -v`

## Print screen / video

**Starting account creation:**
[Capture vidéo du 21-11-2025 09:56:58.webm](https://github.com/user-attachments/assets/43163713-a278-41e9-a543-436377826298)

**Finished to create or recreate the account and its data:**

[Capture vidéo du 21-11-2025 09:59:48.webm](https://github.com/user-attachments/assets/1262bddb-8924-4d66-ad6a-ae144c1b0f5c)

## Notes

```
fix: empty instance pop up

Refs: IA-3665
```

[IA-4374]: https://bluesquare.atlassian.net/browse/IA-4374?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[Capture vidéo du 21-11-2025 09:59:48.webm](https://github.com/user-attachments/assets/819530f5-86cc-4c30-8eab-24db3aab5177)


iaso/management/commands/setuper_sandbox.py
```diff
@@ -0,0 +1,114 @@
...
+        account = Account.objects.filter(name=name).first()
+        Account.objects.filter(name=name).update(name=new_name)
+
+        profiles = Profile.objects.filter(account=account)
+        logger.info(f"Renaming and deactivating all {len(profiles)} users belong to account {name}")
+        new_profiles = self.disable_users(profiles, new_name)
+        updated_profiles = User.objects.bulk_update(new_profiles, ["username", "is_active"])
+        logger.info(f"Disabled {updated_profiles} users")
+
+        projects = Project.objects.filter(account=account).all()
```
tdethier:
```suggestion
        projects = Project.objects.filter(account=account)
```

iaso/management/commands/setuper_sandbox.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def handle(self, *args, **options):
+        name = options["account"]
+        self.stdout.write(self.style.SUCCESS(f" .... Resetting {name} account and its data! ..."))
+
+        current_datetime = int(datetime.now().timestamp())
+        new_name = f"{name}{current_datetime}"
+        logger.info(f"Renaming current account {name} to {new_name}")
+        account = Account.objects.filter(name=name).first()
+        Account.objects.filter(name=name).update(name=new_name)
```
tdethier:
```suggestion
        account.name = new_name
        account.save()
```

iaso/management/commands/setuper_sandbox.py
```diff
@@ -0,0 +1,114 @@
...
+        logger.info(f"Disabled {updated_profiles} users")
+
+        projects = Project.objects.filter(account=account).all()
+        logger.info(f"Renaming app id for all {len(projects)} projects belong to account {name}")
+        projects_to_updated = self.map_project(projects, new_name)
+        updated_projects = Project.objects.bulk_update(projects_to_updated, ["app_id"])
+        logger.info(f"Renamed app id for all {updated_projects} projects")
+
+        data_sources = DataSource.objects.filter(projects__account=account)
+        logger.info(f"Renaming all {len(projects)} data_sources belong to account {name}")
```
tdethier:
```suggestion
        logger.info(f"Renaming all {len(data_sources)} data_sources belong to account {name}")
```

iaso/management/commands/setuper_sandbox.py
```diff
@@ -0,0 +1,114 @@
...
+        updated_profiles = User.objects.bulk_update(new_profiles, ["username", "is_active"])
+        logger.info(f"Disabled {updated_profiles} users")
+
+        projects = Project.objects.filter(account=account).all()
+        logger.info(f"Renaming app id for all {len(projects)} projects belong to account {name}")
+        projects_to_updated = self.map_project(projects, new_name)
+        updated_projects = Project.objects.bulk_update(projects_to_updated, ["app_id"])
+        logger.info(f"Renamed app id for all {updated_projects} projects")
+
+        data_sources = DataSource.objects.filter(projects__account=account)
```
tdethier: There's a many to many relationship between `DataSource` and `Project`, so you should probably add a `.distinct()` here

iaso/management/commands/setuper_sandbox.py
```diff
@@ -0,0 +1,114 @@
...
+        return list(
+            map(
+                lambda data_source: self.change_data_source_name(data_source, new_name),
+                data_sources,
+            )
+        )
+
+    def change_data_source_name(self, data_source, new_account_name):
+        name = data_source.name.replace("admin", new_account_name)
+        data_source.name = name
```
tdethier: Here I would simply append the current timestamp to the existing name (because when `admin` is not in `DataSource.name` the name is not changed and the script crashes later when the setuper is launched). Also, this will work if there are multiple data sources in the same account 

This script crashed while the setuper was running because of this line - my data source still had the old name

iaso/management/commands/setuper_sandbox.py
```diff
@@ -0,0 +1,114 @@
...
+        current_datetime = int(datetime.now().timestamp())
+        new_name = f"{name}{current_datetime}"
+        logger.info(f"Renaming current account {name} to {new_name}")
+        account = Account.objects.filter(name=name).first()
+        Account.objects.filter(name=name).update(name=new_name)
+
+        profiles = Profile.objects.filter(account=account)
+        logger.info(f"Renaming and deactivating all {len(profiles)} users belong to account {name}")
+        new_profiles = self.disable_users(profiles, new_name)
+        updated_profiles = User.objects.bulk_update(new_profiles, ["username", "is_active"])
```
tdethier: Honestly I'm surprised that this line works, because you're sending the payload of profiles and both fields (`username` and `is_active`) need to be accessed through `user`.

It works, so you can leave it like that, but I personally would have made it more readable and would have gotten rid of the bulk update (a simple "for each user, update the name, set is_active to false and save the user")

iaso/management/commands/setuper_sandbox.py
```diff
@@ -0,0 +1,114 @@
...
+        try:
+            command = ["python3", "setuper/setuper.py", "--additional_projects", "-n", f"{account_name}"]
+            process = subprocess.run(command, check=True)
+            self.stdout.write(f"Stdout:\n{process.stdout}")
+            if process.stderr:
+                self.stderr.write(f"Stderr:\n{process.stderr}")
+
+        except subprocess.CalledProcessError as e:
+            self.stderr.write(self.style.ERROR(f"Error executing script '{e}':"))
+            self.stderr.write(self.style.ERROR(f"Command: {e.cmd}"))
```
tdethier:
```suggestion
```

tdethier: It's already included in the line above

iaso/management/commands/setuper_sandbox.py
```diff
@@ -0,0 +1,114 @@
...
+            self.stderr.write(self.style.ERROR(f"Return Code: {e.returncode}"))
+            self.stderr.write(self.style.ERROR(f"Stdout:\n{e.stdout}"))
+            self.stderr.write(self.style.ERROR(f"Stderr:\n{e.stderr}"))
```
tdethier:
```suggestion
```

tdethier: There are both empty even if there's an error

iaso/management/commands/setuper_sandbox.py
```diff
@@ -0,0 +1,114 @@
...
+    def change_user_name(self, profile, new_account_name):
+        user_name = profile.user.username.split(".")
+        new_user_name = new_account_name
+        if len(user_name) > 1:
+            new_user_name = f"{new_account_name}.{user_name[1]}"
+        profile.user.username = new_user_name
+        profile.user.is_active = False
+        return profile.user
+
+    def disable_users(self, profiles, new_account_name):
```
tdethier: Nitpick: the names are a bit confusing because you do disable users when you call this, but you do more than that and the function called does not mention disabling users

iaso/management/commands/setuper_sandbox.py
```diff
@@ -0,0 +1,114 @@
...
+    help = "Custom command to reset admin sandbox account."
+
+    def add_arguments(self, parser):
+        parser.add_argument("account", type=str, help="Account name")
+
+    def change_user_name(self, profile, new_account_name):
+        user_name = profile.user.username.split(".")
+        new_user_name = new_account_name
+        if len(user_name) > 1:
+            new_user_name = f"{new_account_name}.{user_name[1]}"
```
tdethier: ⚠️ the `User.username` field has a max length of 150 characters

iaso/management/commands/setuper_sandbox.py
tdethier: ⚠️ The `DataSource.name` field has a max length of 255 characters

tdethier (changes requested):
I'm requesting changes because I got an issue with data sources while running the script (see comments below).

Otherwise, I made a few suggestions and the script works 👍

cron.yaml
```diff
@@ -17,4 +17,7 @@ cron:
    schedule: "0 3 * * *"
  - name: "vaccine_stock_history_update_for_rounds"
    url: "/tasks/launch_task/plugins.polio.tasks.archive_vaccine_stock_for_rounds.archive_vaccine_stock_for_rounds/polio_cron_task_user/"
-   schedule: "0 2 * * *"
\ No newline at end of file
+   schedule: "0 2 * * *"
+ - name: 'setuper_sandbox'
+   url: '/tasks/setuper_sandbox'
```
tdethier: I don't think it's going to work because it's not the same format, the path to the task is missing, as well as the user that should trigger the task

tdethier: Now that I think about it, I'm a bit afraid that this is going to run on all environments, which we definitely don't want. We should probably use an environment variable and check its value inside the command - if it's set, then it can go further, otherwise it stops immediately. The cleanest way would probably be to determine if we want to run crons or not server by server, but I'm not sure if we can do that.

What do you think @madewulf?

iaso/management/commands/setuper_sandbox.py
```diff
@@ -0,0 +1,109 @@
...
+
+from django.core.management.base import BaseCommand
+
+from iaso.models import Account, DataSource, Profile, Project
+
+
+logger = getLogger(__name__)
+
+
+class Command(BaseCommand):
```
tdethier: If we use a task, this command should probably just call the task, right?

tdethier (changes requested):
I'm requesting changes because I don't think the cron job will work, but we should probably test that on a server

iaso/tasks/setuper_sandbox.py
```diff
@@ -0,0 +1,116 @@
...
+            lambda data_source: change_data_source_name(data_source, current_timestamp),
+            data_sources,
+        )
+    )
+
+
+def change_data_source_name(data_source, current_timestamp):
+    data_source.name = f"{data_source.name}{current_timestamp}"
+    data_source.save()
+    return data_source
```
tdethier: Why do you do `list` then `map` then call a function just to change the data sources names?

You could do a:
```
for data_source in sources:
    data_source.name = f"{data_source.name}{current_timestamp}"
    data_source.save()
```

And if you just want to print the number of data sources updated, you could return `len(data_sources)` in your single function

iaso/tasks/setuper_sandbox.py
```diff
@@ -0,0 +1,116 @@
...
+    return list(
+        map(
+            lambda project: change_project_app_id(project, new_account_name),
+            projects,
+        )
+    )
+
+
+def change_project_app_id(project, new_account_name):
+    current_project = Project.objects.filter(app_id=project.app_id).first()
```
tdethier: You already have the project (in your arguments), you don't need to fetch it again

iaso/tasks/setuper_sandbox.py
```diff
@@ -0,0 +1,116 @@
...
+    app_id = project.app_id.split(".")
+    current_project.app_id = new_account_name
+    if len(app_id) > 1:
+        current_project.app_id = f"{new_account_name}.{app_id[1]}"
```
tdethier: I don't understand why you split the `app_id` why not simply do:

```
current_project.app_id = f"{new_account_name}.{app_id}"
current_project.save()
```

And you can also do that in a `for` loop directly

iaso/tasks/setuper_sandbox.py
```diff
@@ -0,0 +1,116 @@
...
+
+@task_decorator(task_name="setuper_sandbox")
+def setuper_sandbox(name="admin", task=None):
+    ENABLE_SETUPER_SANDBOX = os.environ.get("ENABLE_SETUPER_SANDBOX", "false").lower() == "true"
+    if ENABLE_SETUPER_SANDBOX:
+        logger.warning(f" .... Resetting {name} account and its data! ...")
+        current_datetime = int(datetime.now().timestamp())
+        account_name = name
+        if isinstance(name, list):
+            account_name = name[0]
```
tdethier, line 56: It shouldn't be a list, why did you add this check?

iaso/tasks/setuper_sandbox.py
```diff
@@ -0,0 +1,116 @@
...
+        new_name = f"{account_name}{current_datetime}"
+        logger.info(f"Renaming current account {account_name} to {new_name}")
+        account = Account.objects.filter(name=account_name).first()
+        if account is not None:
+            account.name = new_name
+            account.save()
```
tdethier, line 62: I'd suggest stopping the task here if the account is not found

iaso/tasks/setuper_sandbox.py
```diff
@@ -0,0 +1,116 @@
...
+    except subprocess.CalledProcessError as e:
+        logger.exception(f"Error executing script '{e}':")
+        logger.exception(f"Return Code: {e.returncode}")
+    except Exception as e:
+        logger.exception(f"An unexpected error occurred: {e}")
+
+
+@task_decorator(task_name="setuper_sandbox")
+def setuper_sandbox(name="admin", task=None):
+    ENABLE_SETUPER_SANDBOX = os.environ.get("ENABLE_SETUPER_SANDBOX", "false").lower() == "true"
```
tdethier: It's better to make this check inside `settings.py`  and call the variable from settings here, rather than directly calling `os.environ`

iaso/tasks/setuper_sandbox.py
```diff
@@ -0,0 +1,116 @@
...
+        projects_to_updated = map_project(projects, new_name)
+        updated_projects = Project.objects.bulk_update(projects_to_updated, ["app_id"])
+        logger.info(f"Renamed app id for all {updated_projects} projects")
+
+        data_sources = DataSource.objects.filter(projects__account=account).distinct()
+        logger.info(f"Renaming all {len(data_sources)} data_sources belong to account {account_name}")
+        rename_data_sources = map_data_source(data_sources, current_datetime)
+        logger.info(f"Renamed all {len(rename_data_sources)} data_sources")
+
+        logger.warning(f"Reset {account_name} account!")
```
tdethier: I think that `warning` will trigger a sentry error

```suggestion
        logger.info(f"Reset {account_name} account!")
```

iaso/tasks/setuper_sandbox.py
```diff
@@ -0,0 +1,116 @@
...
+
+from beanstalk_worker import task_decorator
+from iaso.models import Account, DataSource, Profile, Project
+
+
+logger = logging.getLogger(__name__)
+
+
+def change_user_name_and_disable_user(profile, new_account_name):
+    user_name = profile.user.username.split(".")
```
tdethier: Why do you split the username?

iaso/tasks/setuper_sandbox.py
```diff
@@ -0,0 +1,116 @@
...
+    new_user_name = new_account_name
+    if len(user_name) > 1:
+        new_user_name = f"{new_account_name}.{user_name[1]}"
+    profile.user.username = new_user_name
```
tdethier:
```suggestion
    user = profile.user
    user.username = f"{new_account_name}.{user.user_name}"
```

iaso/tasks/setuper_sandbox.py
```diff
@@ -0,0 +1,116 @@
...
+    profile.user.is_active = False
+    profile.user.save()
+    return profile.user
```
tdethier:
```suggestion
    user = profile.user
    user.username = f"{new_account_name}.{user.user_name}"
    user.is_active = False
    user.save()
    return user
```

tdethier (changes requested):
I would make a few changes to make the code clearer, but most importantly, I would stop the task at the beginning if the `Account` is not found (see comments below)

iaso/tasks/setuper_sandbox.py
butofleury: @tdethier , If we stop the task if the account is not found, it will never create the account and it's data on the environement(e.g: demo or staging) when the account is not there. Having the ENABLE_SETUPER_SANDBOX is enough to confirm the need to have the account on the env when the account doesn't yet exist.

iaso/tasks/setuper_sandbox.py
butofleury: @tdethier  I added this check because when sending an argument in task, the parameter is taken as a list. 
For example when I  run: `curl -d name=admin http://localhost:8081/tasks/launch_task/iaso.tasks.setuper_sandbox.setuper_sandbox/cron_task_sandbox_user/ -v` the argument name is taken as a list!

iaso/tasks/setuper_sandbox.py
butofleury: @tdethier I've splitted the `app_id` because the additional projects created via setuper have the appid in the following format `accountName.appId`. 
e.g: https://github.com/BLSQ/iaso/blob/IA-4374-Turn-IASO-setuper-demo-account-into-a-sandbox/setuper/additional_projects.py#L10

image

I've added an improvement for that part, too loop directly on projects

iaso/tasks/setuper_sandbox.py
tdethier: TIL, thanks!

iaso/tasks/setuper_sandbox.py
tdethier: Yeah, ok, why not

iaso/tasks/setuper_sandbox.py
tdethier: Ok, I understand now. You could do that if you want, yeah

tdethier (approved):
LGTM, thanks again for making all the changes 👍 

Let's test this on staging

PR #2446, opened 2025-09-22
IA-4471: speed up migrations

madewulf:
This PR speeds up the database migration process by squashing hundreds of migrations into a few dozens. 
It speeds up the Python CI by at least a minute and will speed a lot the launch of small batches of test in development. (cutting time by 2  to 25 seconds when running docker-compose run iaso manage test plugins.polio.tests.test_subactivities.SubactivitiesAPITestCase -v3 on my machine)

All tests are passing after these migration but some features have no good tests yet, so, I would release this as a post-release feature. 

Related JIRA tickets : [IA-4471]

- [X] My migrations file are included (LOL)
- [X] Are there enough tests?

## How to test

Launch the test suite and see if it works on your installation. 
Do a docker-compose up to have the migrations run on your test database.

## Notes

I verified all the migrations to try to find items that would not have been well included in the merged migrations and lots of the custom sql or python were not copied to the new migrations because they existed to fix bad states in databases that should not happen anymore because the underlying bugs have been fixed. 

It means that the migration runs faster, but there is a small risk of not having well understood one of the migrations that should have been kept. 

I checked also that all index creations were included. 

```
fix: empty instance pop up

Refs: IA-3665
```

[IA-4471]: https://bluesquare.atlassian.net/browse/IA-4471?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hat/menupermissions/migrations/0001_squashed_0075_alter_custompermissionsupport_options.py
```diff
@@ -0,0 +1,137 @@
...
+                    ("iaso_planning", "Planning"),
+                    ("iaso_teams", "Equipes"),
+                    ("iaso_assignments", "Attributions"),
+                    ("iaso_polio_budget", "Budget Polio"),
+                    ("iaso_entities", "Entities"),
+                    ("iaso_storages", "Storages"),
+                    ("iaso_completeness_stats", "Completeness stats"),
+                ),
+            },
+        ),
```
tdethier: There aren't any permissions in that model now, so I guess this part is not really useful. Was this copy pasted from previous migration files?

iaso/migrations/0111_page_account.py
```diff
@@ -7,7 +7,7 @@
 
 class Migration(migrations.Migration):
     dependencies = [
-        ("iaso", "0110_make_readonlyrole"),
+        ("iaso", "0111_page_account"),
```
tdethier: Can a migration file depend on itself?

iaso/migrations/0111_auto_20220106_1725.py
```diff
@@ -9,7 +9,7 @@
 class Migration(migrations.Migration):
     dependencies = [
         migrations.swappable_dependency(settings.AUTH_USER_MODEL),
-        ("iaso", "0110_make_readonlyrole"),
+        ("iaso", "0002_auto_20191001_1142_squashed_0100_auto_20210702_0835"),
```
tdethier, line 12: Now both this one and `0101_page...`  depend on the same migration, isn't this going to make a conflict?

tdethier (commented):
I tried:
- running `migrate` on an empty fresh DB ➡️ it went fine
- running `migrate` on a DB that already had all migrations (from current `develop` branch) ➡️ it went fine, but the last 3 `iaso` migrations (0344, 0345, 0346) were applied. I don't think it will cause an issue since they use `get_or_create` and `CREATE IF NOT EXISTS`


I left a few questions on some files

kemar:
This is great!

I think [you followed this method](https://jacklinke.com/squashing-django-migrations-the-easy-way) to target linear paths in the migrations graph.

I can see that you've already deleted the old migration files, but according to the blog post, squashing is a two-step process:

1. deploy the squashed migrations, but keep the old ones

    - run the squashed migrations
    - verify everything runs without error
    - check that all replaced migrations are marked as applied in the DB

2. transition fully to squashed migrations

    - delete the migrations that were replaced
    - update dependencies in any migrations that relied on them
        - swap filenames to the squashed version in the `dependencies` attribute
    - remove the `replaces = [...]` line from the squashed migration
        - this converts it into a standard migration

Following this two-step process might be safer?

Another thing to consider is inspecting any `RunSQL` or `RunPython` operations to see if `elidable=True` could be used.

Django treats `elidable=True` operations as safe to remove during squashing.

hat/menupermissions/migrations/0001_squashed_0075_alter_custompermissionsupport_options.py
madewulf: Yes, and I don't think it has any impact, in fact. That said, if you look at the other operations in this migration, they kinda cancel this one. So, I'll modify the migration to only have the final state. 

This one migration was fully created by the django squashmigrations command, it just did not catch that.

iaso/migrations/0111_page_account.py
madewulf: > Can a migration file depend on itself?

Apparently, but I'll make it more logical

madewulf:
> This is great!
> 
> I think [you followed this method](https://jacklinke.com/squashing-django-migrations-the-easy-way) to target linear paths in the migrations graph.
> 
> I can see that you've already deleted the old migration files, but according to the blog post, squashing is a two-step process:
> 
> 1. deploy the squashed migrations, but keep the old ones
>    
>    * run the squashed migrations
>    * verify everything runs without error
>    * check that all replaced migrations are marked as applied in the DB
> 2. transition fully to squashed migrations
>    
>    * delete the migrations that were replaced
>    * update dependencies in any migrations that relied on them
>      
>      * swap filenames to the squashed version in the `dependencies` attribute
>    * remove the `replaces = [...]` line from the squashed migration
>      
>      * this converts it into a standard migration
> 
> Following this two-step process might be safer?
> 
> Another thing to consider is inspecting any `RunSQL` or `RunPython` operations to see if `elidable=True` could be used.
> 
> Django treats `elidable=True` operations as safe to remove during squashing.

Thanks for the comment. That said, as we discussed live, I only deleted migrations that were from a previous "squashing" of migrations, and I removed the "replaces" from the squashes of the past. 
So, I made step 2 for the old migrations (from around 2021) in this PR  but it also include the step 1 for a new squash. In maybe a year or two, when all known instances of IASO (especially the local hosted ones) have been migrated, we will be able to perform step2 of this new squash: delete the migrations of the second squash.

kemar (approved):
LGTM

PR #2451, opened 2025-09-24
IA-4485 LQAS planning pipelines

beygorghor:
First PR to implement LQAS sampling in micro-planning.

Related JIRA tickets : IA-4485

- [ ] Are there enough tests?

## Doc

[Here](https://github.com/BLSQ/iaso/blob/IA-4485-planning-pipelines/docs/pages/dev/how_to/planning-pipeline-integration/planning-pipeline-integration.en.md)

## Changes

- Link Planning to pipelines uuids
- allow user to save a pipeline in planning modale, if `openhexa` is configured
- endpoint to know if `openhexa` is configured
- allow to exlcude org units from search by excluding parents
- Custom LQAS form to launch the pipeline.

## How to test

- Make sure the OpenHexa config is correctly set
- Edit a planning and select "LQAS Assignment Pipeline", save it,
- open planning (click on the eye), you should see a sampling button.
- Play with the inputs in the drawer and try to launch a sampling computation 

## Print screen / video

https://github.com/user-attachments/assets/e90cc801-b3a7-42ba-a614-21305786dd89

## Notes

- For now feed-back on pipeline is happening in the drawer, we swill change this an refresh assignments when task is done, or display an error message
- you need a specific openhexa config with this slug (openhexa-config) to make it work:
```

{
  "openhexa_url": "https://api.openhexa.org/graphql/",
  "openhexa_token": "TOKEN",
  "workspace_slug": "iaso-demo-df5d5f", 
  "lqas_pipeline_code": "iaso-test-pipeline"
}

```
You'll need to adapt Iaso cutom connection to use a ngrok url and use specific credential to update the task on your machine

```
fix: empty instance pop up

Refs: IA-3665
```


hat/assets/js/apps/Iaso/domains/assignments/components/OpenhexaIntegrationDrawer.tsx
```diff
@@ -0,0 +1,411 @@
...
+                                    </Button>
+                                )}
+                            </Grid>
+                        </Grid>
+                    </Paper>
+                    <Box position="relative" sx={styles.container}>
+                        {(isFetchingPipeline || isPipelineRunning) && (
+                            <LoadingSpinner absolute fixed={false} />
+                        )}
+                        {/* Step 1: Pipeline selection and parameters */}
```
bmonjoie, line 204: I think you should put each step into its own `FunctionComponent`. This way, it's going to make it a lot clearer what step uses what hooks. Also, it's going to be easier to maintain.

hat/assets/js/apps/Iaso/domains/openHexa/customForms/ExcludedOrgUnits.tsx
```diff
@@ -0,0 +1,72 @@
...
+    parameterValues,
+    planning,
+}) => {
+    const excludedOrgUnitIds =
+        index > 0
+            ? parameterValues?.org_unit_type_exceptions
+                  ?.slice(0, index)
+                  ?.filter(e => Boolean(e))
+                  .join(',')
+            : undefined;
```
bmonjoie, line 31: You should use `useMemo` here.

hat/assets/js/apps/Iaso/domains/openHexa/customForms/ExcludedOrgUnits.tsx
```diff
@@ -0,0 +1,72 @@
...
+            disabled={isFetchingOrgUnits || !orgUnitTypeId}
+            withMarginTop={false}
+            label={MESSAGES.excludedOrgUnits}
+            keyValue="orgUnitId"
+            value={selectedOrgUnitIds}
+            loading={isFetchingOrgUnits}
+            options={orgUnits?.map(ou => ({
+                label: ou.name,
+                value: ou.id,
+            }))}
```
bmonjoie: Probably a good idea to use `useMemo` here too.

hat/assets/js/apps/Iaso/domains/openHexa/customForms/LQASForm.tsx
```diff
@@ -0,0 +1,251 @@
...
+    handleParameterChange: (parameterName: string, value: any) => void;
+};
+
+export const LQASForm: FunctionComponent<Props> = ({
+    planning,
+    setAllowConfirm,
+    parameterValues,
+    handleParameterChange,
+}) => {
+    // console.log('parameterValues', parameterValues);
```
bmonjoie: Don't give @quang-le more reasons to climb in the commit ladder :joy:

iaso/models/microplanning.py
```diff
@@ -139,13 +139,36 @@ class Meta:
     team = models.ForeignKey(Team, on_delete=models.CASCADE)
     org_unit = models.ForeignKey(OrgUnit, on_delete=models.PROTECT)
     published_at = models.DateTimeField(null=True, blank=True)
+    pipeline_uuids = models.JSONField(
```
tdethier: What about an `ArrayField()` [like here](https://github.com/BLSQ/iaso/blob/7fd05d1f3b53b28cfa264de8fb8110ec1a022b64/iaso/models/org_unit.py#L753)?

It feels like a better option than a json where we could store anything, but this works too

iaso/migrations/0345_merge_20250926_1109.py
```diff
@@ -0,0 +1,12 @@
+# Generated by Django 4.2.20 on 2025-09-26 11:09
+
+from django.db import migrations
```
tdethier: I would recreate the migrations after merging/rebasing `develop` in order to avoid getting more merges

hat/assets/js/apps/Iaso/domains/openHexa/customForms/constants.ts
```diff
@@ -0,0 +1,20 @@
...
+        },
+        {
+            label: formatMessage(MESSAGES.urban),
+            value: 'URBAN',
+        },
+        {
+            label: formatMessage(MESSAGES.rural),
+            value: 'RURAL',
+        },
+    ];
```
bmonjoie: You should use `useMemo` here.

iaso/api/org_unit_types/serializers.py
```diff
@@ -271,8 +271,24 @@ def to_representation(self, instance):
...
+    sub_unit_types = serializers.SerializerMethodField()
+
+    class Meta:
+        model = OrgUnitType
+        fields = ["id", "name", "short_name", "depth", "category", "sub_unit_types"]
+        read_only_fields = ["id", "name", "short_name", "depth", "category", "sub_unit_types"]
+
+    def get_sub_unit_types(self, obj):
+        """Recursively serialize sub_unit_types to build complete hierarchy"""
+        sub_types = obj.sub_unit_types.prefetch_related("sub_unit_types").all()
```
tdethier: I'm not sure that `prefetch_related` will work here. I think we should compare the number of SQL queries with/without in the API with django debug toolbar and go with the best one

iaso/api/openhexa/views.py
```diff
@@ -319,3 +321,37 @@ def partial_update(self, request, pk=None):
...
+        """
+        try:
+            # Try to get the config
+            openhexa_config = Config.objects.get(slug=OPENHEXA_CONFIG_SLUG)
+
+            # Check if required fields exist and are not empty
+            content = openhexa_config.content
+            required_fields = ["openhexa_url", "openhexa_token", "workspace_slug"]
+
+            configured = all(field in content and content[field] for field in required_fields)
```
tdethier: You could probably use a serializer for that :sweat_smile:

hat/assets/js/apps/Iaso/domains/openHexa/hooks/useLaunchTask.ts
```diff
@@ -1,19 +1,29 @@
...
         config: body,
         version,
     });
 };
-export const useLaunchTask = (pipelineId?: string, version?: string) => {
+
+export const useLaunchTask = (
+    pipelineId?: string,
+    version?: string,
+    showSucessSnackBar = true,
```
bmonjoie:
```suggestion
    showSuccessSnackBar = true,
```

hat/assets/js/apps/Iaso/domains/openHexa/hooks/usePipelineParameters.tsx
```diff
@@ -25,50 +32,51 @@ export const usePipelineParameters = (pipeline?: any) => {
...
+                                    parameter.default || 0.0;
+                                break;
+                            case 'dict':
+                                initialValues[parameter.code] =
+                                    parameter.default || {};
+                                break;
+                            default:
+                                initialValues[parameter.code] =
+                                    parameter.default || null;
+                        }
```
bmonjoie: This probably should be in its own method.

iaso/api/microplanning.py
```diff
@@ -281,13 +283,29 @@ class Meta:
...
+
+        for uuid_str in value:
+            if not isinstance(uuid_str, str):
+                raise serializers.ValidationError("Each pipeline UUID must be a string")
+            try:
+                uuid.UUID(uuid_str)
+            except ValueError:
+                raise serializers.ValidationError(f"Invalid UUID format: {uuid_str}")
+
+        return value
```
tdethier: I'm wondering if DRF can do all of this for you if you use a `UUIDField` with `many=True` somewhere, but I'm not sure if it's possible

Otherwise, this works, it's fine

iaso/tests/api/test_orgunits.py
```diff
@@ -2416,3 +2416,105 @@ def test_descending_order_with_as_location(self):
         self.client.force_authenticate(self.yoda)
         response = self.client.get("/api/orgunits/?limit=20&order=-name&page=1&asLocation=true")
         self.assertEqual(response.status_code, 200)
+
+    def test_org_unit_excluded_parent_ids_filter(self):
```
tdethier, line 2420: I guess this works, but it's very complicated. Also, I'm not sure if it won't be flaky because of the ordering

I'd rather manually build a specific "simple" pyramid for this test and expect specific results based on this pyramid

iaso/tests/api/test_orgunits.py
```diff
@@ -2416,3 +2416,105 @@ def test_descending_order_with_as_location(self):
...
+
+        # Verify that the count is reduced by the expected amount
+        self.assertEqual(filtered_count, all_count - expected_excluded_count)
+
+        # Verify that none of the filtered org units are in the excluded hierarchy
+        excluded_ids = {ou["id"] for ou in hierarchy_org_units}
+        for ou in filtered_org_units:
+            self.assertNotIn(ou["id"], excluded_ids)
+
+    def test_org_unit_excluded_multiple_parent_ids_filter(self):
```
tdethier, line 2497: Same comment about complexity

iaso/api/microplanning.py
```diff
@@ -313,10 +331,11 @@ def validate(self, attrs):
...
 
-        forms = validated_data.get("forms", self.instance.forms if self.instance else None)
-        for form in forms:
-            if form not in project.forms.all():
-                validation_errors["forms"] = "planningAndForms"
+        forms = validated_data.get("forms", list(self.instance.forms.all()) if self.instance else None)
+        if forms:
+            for form in forms:
+                if form not in project.forms.all():
+                    validation_errors["forms"] = "planningAndForms"
```
bmonjoie: You are looping even after the error is found. You can replace this with this code:
```suggestion
        project_forms = project.forms.all()
        if forms and not all(f in project_forms for f in forms):
             validation_errors["forms"] = "planningAndForms"
```

iaso/tests/test_microplanning.py
```diff
@@ -1122,3 +1257,132 @@ def test_query_mobile_no_modification(self):
 
         response = self.client.post("/api/mobile/plannings/", data={}, format="json")
         self.assertEqual(response.status_code, 403)
+
+
+class PlanningPipelineIntegrationTestCase(APITestCase):
+    """Test Planning model pipeline UUID integration."""
```
tdethier: This should be part of `iaso.tests.models`, it's not an API test

iaso/tests/api/test_org_unit_types_v2.py
```diff
@@ -582,3 +582,137 @@ def assertValidOrgUnitTypeDropdownData(self, data):
...
+        """Test GET /orgunittypes/{id}/hierarchy/ with non-existent ID returns 404"""
+
+        self.client.force_authenticate(self.jane)
+        response = self.client.get(f"{self.BASE_URL}99999/hierarchy/")
+
+        self.assertJSONResponse(response, status.HTTP_404_NOT_FOUND)
+        response_data = response.json()
+        self.assertIn("error", response_data)
+
+    def test_org_unit_type_hierarchy_without_auth(self):
```
tdethier, line 646: :+1: 

I would also check with a user that has no perms in another test

iaso/migrations/0345_merge_20250926_1109.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("iaso", "0344_alter_corepermissionsupport_options"),
+        ("iaso", "0344_planning_pipeline_uuids"),
+    ]
+
+    operations = []
```
bmonjoie: Why don't you just rename your migration to `0345` and point to the previous `0344` for the dependency?

tdethier (commented):
I only read the backend code: there are a few things I would change, but nothing major. I guess this could work like that.

We should take some time to investigate a couple of suggestions I made - I'm not sure if they are feasible or not

hat/assets/js/apps/Iaso/domains/assignments/components/OpenhexaIntegrationDrawer.tsx
beygorghor: this part of the code is going to disappear, it was for testing purpose. Like i said in the desc of this PR, you need a specific config to see this. I'm waiting for Jerome and Paula to finish the job..

hat/assets/js/apps/Iaso/domains/openHexa/customForms/ExcludedOrgUnits.tsx
beygorghor: nope, it's returning a string, it's useless

hat/assets/js/apps/Iaso/domains/assignments/hooks/requests/usePollTask.ts
```diff
@@ -0,0 +1,27 @@
+import { UseQueryResult } from 'react-query';
+import { Task } from 'Iaso/domains/tasks/types';
+import { getRequest } from 'Iaso/libs/Api';
+import { useSnackQuery } from 'Iaso/libs/apiHooks';
+import { DjangoError } from 'Iaso/types/general';
+
+export const usePollTask = (
```
Bewi, line 7: Does this stop if user navigate away ?

hat/assets/js/apps/Iaso/domains/assignments/index.tsx
```diff
@@ -104,6 +106,7 @@ export const Assignments: FunctionComponent = () => {
     });
     const isLoading = isLoadingPlanning || isSaving;
 
+    const { data: hasPipelineConfig } = useGetPipelineConfig();
```
Bewi, line 109: If we never use the entire object, wouldn't it be better to only do an OPTION request to reduce payload ?
Or to have a custom endpoint to reduce reads in DB ?

iaso/migrations/0345_merge_20250926_1109.py
beygorghor: you are right, i always did it like that

hat/assets/js/apps/Iaso/domains/openHexa/customForms/LQASForm.tsx
```diff
@@ -0,0 +1,251 @@
...
+    useEffect(() => {
+        if (
+            parameterValues &&
+            parameterValues?.org_unit_type_sequence_identifiers?.length &&
```
Bewi, line 69: Do you need Elvis operator here ?
You already checked if parameterValues was null
You could have one check less, and I would remove the first one.

hat/assets/js/apps/Iaso/domains/openHexa/customForms/LQASForm.tsx
```diff
@@ -0,0 +1,251 @@
...
+            parameterValues?.org_unit_type_sequence_identifiers?.length > 0
+        ) {
+            const allLevelsFilled =
+                parameterValues?.org_unit_type_sequence_identifiers?.every(
```
Bewi: Same here, no need to check if it is defined again

hat/assets/js/apps/Iaso/domains/openHexa/customForms/LQASForm.tsx
```diff
@@ -0,0 +1,251 @@
...
+            setAllowConfirm(false);
+        }
+    }, [setAllowConfirm, parameterValues]);
+    const { formatMessage } = useSafeIntl();
+    const [expandedLevels, setExpandedLevels] = useState<boolean[]>([false]);
+    const { data: orgUnitTypes, isFetching: isFetchingOrgUnitTypes } =
+        useGetOrgUnitTypesDropdownOptions({
+            projectId: planning.project,
+        });
+    const updateArrayAtIndex = useCallback(
```
Bewi: Feels like those array utils could be re-used.
Same logic as here: 
hat/assets/js/apps/Iaso/domains/openHexa/customForms/ExcludedOrgUnits.tsx

hat/assets/js/apps/Iaso/domains/openHexa/customForms/LQASForm.tsx
```diff
@@ -0,0 +1,251 @@
...
+        elevation: 2,
+    },
+    button: {
+        mt: 2,
+    },
+    icon: {
+        mr: 1,
+    },
+};
+type Criteria = 'RURAL/URBAN' | 'URBAN' | 'RURAL';
```
Bewi: This is hardcoded in at least three places, I would share this to make it easier to use

hat/assets/js/apps/Iaso/domains/openHexa/hooks/useGetPipelineDetails.ts
```diff
@@ -1,11 +1,13 @@
...
 import { getRequest } from 'Iaso/libs/Api';
 import { useSnackQuery } from 'Iaso/libs/apiHooks';
 import { DjangoError } from 'Iaso/types/general';
+import { Pipeline } from '../types/pipeline';
 
 export const useGetPipelineDetails = (
     pipelineId?: string,
-    setError?: (error: DjangoError | null) => void,
-) => {
+    parameterstoRemove: string[] = ['task_id', 'pipeline_id'],
```
Bewi: parameters**To**Remove :D (Upper case on the T)

hat/assets/js/apps/Iaso/domains/openHexa/hooks/usePipelineParameters.tsx
```diff
@@ -85,10 +93,57 @@ export const usePipelineParameters = (pipeline?: any) => {
...
+            if (parameter.multiple) {
+                return (
+                    <InputComponent
+                        type="select"
+                        keyValue={parameter.code}
+                        labelString={parameter.name}
+                        value={currentValue || []}
+                        required={parameter.required}
+                        options={
+                            parameter?.choices?.map(choice => ({
```
Bewi: As suggested by Benjamin, this also could be in a useMemo ?

hat/assets/js/apps/Iaso/domains/plannings/components/PlanningDialog.tsx
```diff
@@ -357,29 +341,86 @@ export const CreateEditPlanning: FunctionComponent<Props> = ({
...
-                        />
+
+                    <Grid container spacing={2}>
+                        {hasPipelineConfig && (
+                            <Grid item xs={6}>
+                                <InputComponent
+                                    type="select"
+                                    multi
+                                    keyValue="pipelineUuids"
+                                    onChange={(keyValue, value) => {
```
Bewi: You can remove curly braces around function call here: 
`(keyValue, value) => onChange(keyValue, value?.split(',') ?? [])`

iaso/api/org_unit_types/serializers.py
beygorghor: I will remove it for now

hat/assets/js/apps/Iaso/domains/plannings/components/PlanningDialog.tsx
```diff
@@ -357,29 +341,86 @@ export const CreateEditPlanning: FunctionComponent<Props> = ({
...
+                        )}
+                        <Grid item xs={6}>
+                            <InputComponent
+                                type="radio"
+                                keyValue="publishingStatus"
+                                onChange={onChange}
+                                value={values.publishingStatus}
+                                errors={getErrors('publishingStatus')}
+                                label={MESSAGES.publishingStatus}
+                                options={[
```
Bewi: this feels like a constant, would move to a useMemo

hat/assets/js/apps/Iaso/domains/assignments/hooks/requests/usePollTask.ts
beygorghor: Yes

hat/assets/js/apps/Iaso/domains/assignments/index.tsx
beygorghor: i'm also using result of this call somewhere else

hat/assets/js/apps/Iaso/domains/openHexa/customForms/LQASForm.tsx
```diff
@@ -89,32 +93,30 @@ export const LQASForm: FunctionComponent<Props> = ({
...
+            const updatedArray = updateArrayAtIndex(index, value, currentArray);
             handleParameterChange(arrayName, updatedArray);
         },
         [parameterValues, handleParameterChange],
     );
-
-    const addToArray = useCallback(
+    const add = useCallback(
         (arrayName: string, value: any = undefined) => {
             const currentArray = parameterValues?.[arrayName] || [];
```
Bewi, line 107: Not important but:
addToArray already do this check `const currentArray = [...(array || [])];`
You could just have: `const updatedArray = addToArray(value, parameterValue?.[arrayName]);`

Same for other array functions

Bewi (approved):
LGTM

PR #2454, opened 2025-09-25
IA-4498 fix parquet export when too long column name based on submissions json keys

mestachs:
Postgres is limited to 64 chars identifier.

Related JIRA tickets : IA-4498 

- [ ] Are there enough tests?

## Doc

## Changes

The problem comes from postgres and postgres_query that limits the number of chars and just truncate them without really a warning. the only solution is generating shorter unique alias while in postgres space then rename to longer names when in duckdb space.

So instead of 
```sql
 COPY (
   SELECT * FROM postgres_query('pg', $$ 
    select 
json->>'Gestion_des_projets_et_suivi_des_activites_internes_sur_5_derniers_mois_2025_4564654654654654' as Gestion_des_projets_et_suivi_des_activites_internes_sur_5_derniers_mois_2025_4564654654654654
  ...
)

```

it will generate a "safe mapping" with shorter column names and once in duckdb layer change to the longer column name

```sql
 COPY (
   SELECT 
   c2 as 'Gestion_des_projets_et_suivi_des_activites_internes_sur_5_derniers_mois_2025_4564654654654654'

FROM postgres_query('pg', $$ 
    select 
json->>'Gestion_des_projets_et_suivi_des_activites_internes_sur_5_derniers_mois_2025_4564654654654654' as c2
  ...
)
```

## How to test

1. create a form
2. upload a form version xlsx : https://docs.google.com/spreadsheets/d/1g3kI7lSNOKiDJjw_XhX72qY_feF5diqXKZBn6DiisVk/edit?gid=0#gid=0
3. go submission screen, select the form, copy the "csv" url, replace `csv=true` by `parquet=true`
  - it should work (was failing before the fix)
5. look at the generated columns with duckdb cli (adapt the file name)
```
describe select * from 'submissions.parquet'
```

## Print screen / video

image

```
fix: failing export when 2 question names starts with same 63 chars

Refs: IA-4498
```

### DIFF
--- a/iaso/api/instances/instances.py
+++ b/iaso/api/instances/instances.py
@@ -560,11 +560,11 @@ class InstancesViewSet(viewsets.ViewSet):
         # actually return parquet file
         form_ids = filters["form_ids"]
         form = Form.objects.get(pk=form_ids)
-        export_queryset = parquet.build_submissions_queryset(queryset, form.id)
+        export_queryset, mapping = parquet.build_submissions_queryset(queryset, form.id)
 
         tmp = tempfile.NamedTemporaryFile(suffix=".parquet", delete=False)
 
-        parquet.export_django_query_to_parquet_via_duckdb(export_queryset, tmp.name)
+        parquet.export_django_query_to_parquet_via_duckdb(export_queryset, tmp.name, mapping)
 
         response = CleaningFileResponse(tmp.name, as_attachment=True, filename="submissions.parquet")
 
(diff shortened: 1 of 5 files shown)


tdethier (approved):
LGTM 👍 

I like the idea of preparing a mapping for each field. I'm not sure about the logic to generate safe aliases though, I would have simply have called them a1, a2... without trying to truncate stuff.


Thanks for including a test!

PR #2477, opened 2025-10-08
IA-4522: Add OU search to Change Request reviews

bmonjoie:
Allows to filter change requests on OrgUnits

Related JIRA tickets : IA-4522

- [ ] Are there enough tests?

## Changes

I copied the logic from [`org_unit_search.py`](https://github.com/BLSQ/iaso/blob/8d87c74ae2a968cdcb424e3d84c63a6ba62f069b/iaso/api/org_unit_search.py#L55-L83) and applied it into a new filter for the change requests.

## How to test

Go to the Change Requests list and enter a value in the "Org unit" text input.
Possible values are:
- Part of the name (case insensitive)
- `ids:X,Y,Z` 
- `refs:XXXX`

```
fix: empty instance pop up

Refs: IA-3665
```


beygorghor (changes requested):
This is working nice for me but:

- maybe use a inputComponent with infos to inform user he can use ids and refs, something like this
Screenshot 2025-10-10 at 12 03 39

- new search filter should be tested
- this search logic could be one function maybe ?

bmonjoie:
Since it's the same code than the OrgUnit search and this one is already tested, I don't see a lot of added values to add tests that will do the same as others.

beygorghor (approved):

PR #2480, opened 2025-10-09
IA-4525: Allow to filter OUs that are open on a specific date

bmonjoie:
Adding a way to filter OU that are open on a specific date

Related JIRA tickets : IA-4525

- [X] Are there enough tests?

## Changes

I have added a `open_date` parameter to the `OrgUnit` search which filters based on the following logic:

**Opening date** : Either null or lower or equal to the given date
**Closed date**: Either null or greater (but not equal to) the given date 

## How to test

Go to the OrgUnit list and filter on a specific date.

```
fix: empty instance pop up

Refs: IA-3665
```

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/app/translations/en.json
+++ b/hat/assets/js/apps/Iaso/domains/app/translations/en.json
@@ -1010,6 +1010,7 @@
     "iaso.orgUnits.name": "Name",
     "iaso.orgUnits.new": "Create an org unit",
     "iaso.orgUnits.onlyDirectChildren": "Only Direct children",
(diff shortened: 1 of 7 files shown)


mestachs (approved):
@beygorghor @bmonjoie  Except my suggestion about the existing search criteria on closed/open date to be moved/removed. Everything looks fine.

image

bmonjoie:
@mestachs I don't think we should remove them, they are not the same thing. 
The open/close dates filters on Org Units that have exactly those values, whereas the new filter will search for any Org Unit that are open at a given date.
Therefore, there are complementary

PR #2481, opened 2025-10-10
[SLEEP] Entities list page: attempt to solve performance issues

bramj:
This PR attempts to address perf issues in the Entities list page (mega-slow for Trypelim with > 1M entities)

- Refactoring: Format entities API to IASO structure
- Improve performance of the entities list page
  - TEMP: Ignore the duplicates, they're not used now
  - Remove 2 problematic `select_related` clauses: this gives the biggest
  gain I think, since it adds SQL joins on massive tables.
  - Use minimal dict for org units when possible
  - Only fetch fields we need for instances

Helpful article about `select_related` vs `prefetch_related`: https://medium.com/@akhilvp/optimizing-django-queries-select-related-vs-prefetch-related-for-large-databases-724da5c743a4

I think the next step could be to add a `last_synced_at` to the `iaso_entities` table, that we maintain synced with the newest instance. This would mean we could avoid the join on the `iaso_instances` table, I think that would be a huge gain.

### DIFF
--- a/iaso/api/entities/serializers.py
+++ b/iaso/api/entities/serializers.py
@@ -1,0 +1,85 @@
+from rest_framework import serializers
+
+from iaso.models import Entity
+from iaso.models.deduplication import ValidationStatus
+from iaso.models.storage import StorageDevice
+
+
+class EntitySerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Entity
+        fields = [
+            "id",
+            "name",
+            "uuid",
+            "created_at",
+            "updated_at",
+            "attributes",
+            "entity_type",
+            "entity_type_name",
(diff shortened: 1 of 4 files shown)


Phil-V:
Thanks Bram! I've reviewed the changes and it works well, I'll merge it soon. 

> I think the next step could be to add a last_synced_at to the iaso_entities table, that we maintain synced with the newest instance. This would mean we could avoid the join on the iaso_instances table, I think that would be a huge gain.


I was thinking the same, there's just no efficient way that I can see to avoid a full table scan without it. I'll ask you for your opinion about potential approaches tomorrow.

Phil-V (approved):

PR #2482, opened 2025-10-10
IA-4527: pipeline task poll

beygorghor:
Change the way we are getting pipeline status, for now if it crashes at the beginning, , IASO task stays in running status

Related JIRA tickets : IA-4527

- [x] Are there enough tests?

## Doc

https://github.com/BLSQ/iaso/blob/c7a3dde4e6492dee54e14bac46fb8a2a5b3c23d4/docs/pages/dev/how_to/planning-pipeline-integration/planning-pipeline-integration.en.md

## Changes

- move launch pipeline to a real task 
- poll openhexa api to receive status using this task
- pipelines is now only sending progress and message

## How to test

With a openhexa config, visit a planning, launch the pipelines and look everything is going well. 
You can override parameters value or token here iaso/api/openhexa/views.py line 227 
`config["connection_token"] = 'fake_token'`
or 
`config["brol"] = "fsdfd"`

Pipeline should fail and iiaso should display the error

## Print screen / video

https://github.com/user-attachments/assets/40037156-4cc9-4cdc-abd9-5aa9e873cd29

## Notes

```
fix: empty instance pop up

Refs: IA-3665
```


iaso/tasks/launch_openhexa_pipeline.py
```diff
@@ -0,0 +1,225 @@
...
+    logger.info(f"Successfully launched pipeline {pipeline_id} v{version} as task {task.pk}")
+
+    task.report_progress_and_stop_if_killed(
+        progress_message=f"Successfully launched pipeline {pipeline_id} v{version} as task {task.pk}"
+    )
+
+    logger.info(f"Started OpenHexa polling task for task {task.pk}")
+
+    # Set up timeout for polling
+    polling_start_time = datetime.now()
```
tdethier: Nitpick: if you use `django.utils.timezone.now()`, you get a timezone-aware datetime, depending on the general Django settings

iaso/tasks/launch_openhexa_pipeline.py
```diff
@@ -0,0 +1,225 @@
...
+            # Check if task was killed
+            task.refresh_from_db()
+            if task.should_be_killed:
+                logger.info(f"Task {task.pk} was killed, stopping polling")
+                return
+
+            # Check if polling timeout has been reached
+            current_time = datetime.now()
+            elapsed_time = current_time - polling_start_time
+            if elapsed_time > max_polling_duration:
```
tdethier: You're comparing here `datetime` with `int`, I don't think that's going to work. You'll probably need to convert one (or two?) of them, I would probably use seconds in both cases for easy comparison.

iaso/tests/api/openhexa/test_views.py
```diff
@@ -489,6 +535,85 @@ def test_get_pipeline_detail_anonymous_user(self):
...
+            version=version,
+            config=config,
+            _immediate=True,
+            task=task,
+        )
+
+        # Verify task was configured correctly
+        task.refresh_from_db()
+
+        # Verify task was configured correctly
```
tdethier:
```suggestion
```

iaso/tests/api/openhexa/test_views.py
```diff
@@ -489,6 +535,85 @@ def test_get_pipeline_detail_anonymous_user(self):
         self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
 
 
+class BackgroundTaskTestCase(OpenHexaAPITestCase):
```
tdethier, line 538: I would probably add tests for checking the retry and timeout behaviors:
- checking what happens if the OH server can't be reached during polling
- checking what happens to the iaso task when the polling result is in `SUCCESS_STATUSES`, `FAILURE_STATUSES` or `IN_PROGRESS_STATUSES`
- checking what happens to the iaso task when there's a timeout

tdethier (changes requested):
There's one thing to change (`datetime` and `int` comparison), otherwise nice job :+1:

hat/assets/js/apps/Iaso/domains/assignments/sampling/OpenhexaIntegrationDrawer.tsx
```diff
@@ -121,15 +120,29 @@ export const OpenhexaIntegrationDrawer: FunctionComponent<Props> = ({
         isLoading: isLaunchingTask,
     } = useLaunchTask(selectedPipelineId, pipeline?.currentVersion?.id, false);
     const taskId = launchResult?.task?.id;
+    const [taskStatus, setTaskStatus] = useState<string | undefined>(undefined);
+
+    const isPipelineActive =
```
Bewi, line 128: Could have a useMemo :D

Bewi: Also, it would be nice to have  a constant for taskStatuses

iaso/tasks/launch_openhexa_pipeline.py
```diff
@@ -0,0 +1,225 @@
...
+
+class MockConfig:
+    """Mock Config object for pipeline configuration."""
+
+    def __init__(self, content_dict):
+        self.content = content_dict
+
+
+@task_decorator(task_name="launch_openhexa_pipeline")
+def launch_openhexa_pipeline(
```
Bewi, line 201: Just a personal preference but I would break this big boy into small method to encapsulate non important bits.
So when we read it, it is easier to know what is happening without logs / small logic pollution

iaso/tasks/launch_openhexa_pipeline.py
beygorghor: ✅

iaso/tasks/launch_openhexa_pipeline.py
beygorghor: ✅

iaso/tests/api/openhexa/test_views.py
beygorghor: ✅

iaso/tests/api/openhexa/test_views.py
beygorghor: ✅

hat/assets/js/apps/Iaso/domains/assignments/sampling/OpenhexaIntegrationDrawer.tsx
beygorghor: bollean computation doesn't require a `useMemo`

beygorghor: TaskStatus was already there in fact ...

iaso/tests/api/openhexa/test_views.py
```diff
@@ -614,6 +612,528 @@ def test_launch_openhexa_pipeline_with_beanstalk_worker(self):
...
+        )
+
+        # Verify task status was updated to ERRORED
+        task.refresh_from_db()
+        self.assertEqual(task.status, "ERRORED")
+
+    @patch("iaso.tasks.launch_openhexa_pipeline.Client")
+    @patch("iaso.tasks.launch_openhexa_pipeline.ExternalTaskModelViewSet.launch_task")
+    def test_polling_running_status(self, mock_launch, mock_client_class):
+        """Test that task status is updated to RUNNING when OpenHexa returns running status."""
```
tdethier, line 735: There's no check for the task's status between the 2 polling calls, so this test actually never checks for `RUNNING` :sweat_smile: 

But having a test with multiple polling calls is a good idea and is closer to what will really happen

tdethier (approved):
Thanks for making the changes, the backend looks good :+1: 

Your test setups are very similar, if you're bored there's probably a lot of refactor that can be done there, but don't worry about it if you don't have time/feel like it

PR #2484, opened 2025-10-13
[SLEEP-1648] Permissions v2 in Trypelim

tdethier:
Rebased `develop` on `trypelim` for permissions v2

Related JIRA tickets : SLEEP-1648

- [ ] Are there enough tests?

## Doc

/

## Changes

- Rebase `develop` on `trypelim`
- Use `is_trypelim_plugin_active()` function everywhere there is some custom behavior to let the CI work properly on the `trypelim` branch even if the plugin is not active 

## How to test

- Run the app, make sure it works normally

## Print screen / video

/

## Notes

- there is [another PR in the trypelim repo](https://github.com/BLSQ/trypelim/pull/201) that switches all trypelim permissions to the new permission model

```
fix: empty instance pop up

Refs: IA-3665
```


iaso/migrations/0001_squashed_0343_importgpkg_default_valid.py
```diff
@@ -730,7 +730,8 @@ class Migration(migrations.Migration):
     ]
 
     operations = [
-        CreateExtension("citext"),
+        TrigramExtension(),
+        CITextExtension(),
```
Phil-V, line 734: Thank you for catching this! 🔥

iaso/api/entities/views.py
```diff
@@ -83,7 +83,11 @@ def get_queryset(self):
 
         queryset = (
             Entity.objects.filter_for_user(self.request.user)
-            .select_related("entity_type")
+            .select_related(
+                "attributes__org_unit",
+                "attributes__created_by",
+                "entity_type",
+            )
```
Phil-V, line 90: This seems to undo the changes from Bram in 0942e1d18e6bfc9ceccf1aa68eb8ae9369d42e68. Maybe by accident or because it breaks tests? He moved the attributes in select_related to prefetch_related for performance reasons.

Phil-V (commented):
Great work and thanks for taking care of the rebase! 

By the way could you use `trypelim-base-25-10-20` as a target for your PR before merging?

iaso/api/entities/views.py
tdethier: Yes, it did break [a test](https://github.com/BLSQ/iaso/blob/3819f2746f5c4edbf89d2917ed5a1683e3cc386e/iaso/tests/api/entities/test_entities.py#L235) - since you're not fetching 2 tables, there are 2 extra queries (so it should be `9` instead of `7`).

I thought that this was due to a mistake (that I did during the rebase or that someone did when moving stuff from the single entity file). But if it's on purpose, then we should revert this and update that test

PR #2487, opened 2025-10-14
IA-4486: restore org unit doesn't restore org unit type

beygorghor:
Restoring version of OU through the historical change is not complete

Related JIRA tickets : IA-4486

- [ ] Are there enough tests?

## Doc

## Changes

use org_unit_type_id and invalidate data

## How to test

Edit an org unit and change the type, go to history tab. restore previous value, org unit types should ba back to initial value

## Print screen / video

https://github.com/user-attachments/assets/dd5b5bbd-aff4-42eb-91f7-97b4306872f4

## Notes

```
fix: empty instance pop up

Refs: IA-3665
```

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/orgUnits/details.tsx
+++ b/hat/assets/js/apps/Iaso/domains/orgUnits/details.tsx
@@ -276,11 +276,10 @@ const OrgUnitDetail: FunctionComponent = () => {
             mappedRevision.groups = group_ids;
             saveOu(mappedRevision).then(res => {
                 refreshOrgUnitQueryCache(res);
-                queryClient.invalidateQueries('currentOrgUnit');
                 onSuccess();
             });
         },
-        [currentOrgUnit, refreshOrgUnitQueryCache, saveOu, queryClient],
+        [currentOrgUnit, refreshOrgUnitQueryCache, saveOu],
     );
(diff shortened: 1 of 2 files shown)


hat/assets/js/apps/Iaso/domains/orgUnits/details.tsx
```diff
@@ -275,10 +276,11 @@ const OrgUnitDetail: FunctionComponent = () => {
             mappedRevision.groups = group_ids;
             saveOu(mappedRevision).then(res => {
                 refreshOrgUnitQueryCache(res);
+                queryClient.invalidateQueries('currentOrgUnit');
```
quang-le: You could put that change in `useRefreshOrgUnit`: it already calls the queryClient, so `refreshOrgUnitQueryCache` could handle the invalidation directly. The hook isn't used anywhere else.

quang-le (changes requested):

hat/assets/js/apps/Iaso/domains/orgUnits/details.tsx
beygorghor: ha yes i tried and it broke the page in fact, i'll try again

quang-le (approved):

PR #2491, opened 2025-10-16
POLIO-2004: make doses per vial configurable in vaccine stock

quang-le:

Related JIRA tickets : POLIO-2004, POLIO-1998

- [x] Are there enough tests?

## Doc

N/A

## Changes

- Add mandatory `doses_per_vial`field to `OutgoingStockMovement`, `IncidentReport`, and `DestructionReport` models
- Add migrations:
    - 1 migration to add the field (optional)
    - 1 migration to populate the field when empty with default values
    - 1 migration to make the field mandatory

- Update impacted tests
- Update `VaccineStockCalculator`:
    - pass the optional `end_date` at the `__init__`level
    - return `doses_per_vial`  in all payloads
- Update Form A (`OutgoingStockMovement`) API:
    - Calculate available earmarked vials based (filtered) on `doses_per_vial`
    - Assign the same `doses_per_vial` to the automatically created earmark movement as the related Form A 

- Remove unused endpoint

## How to test

There's no impact on the front-end yet, so you need to play with the django API or swagger

## Print screen / video

N/A


plugins/polio/api/lqas_im/lqas_data_manager.py
```diff
@@ -336,7 +336,6 @@ def _create_district_data(self, round_obj: Round, district: OrgUnit, district_da
                 cg_caregivers_informed=caregivers_informed,
                 cg_caregivers_informed_ratio=caregivers_informed_ratio,
             )
-            print("ENTRY", lqas_entry)
```
tdethier, line 339: Thanks!

plugins/polio/migrations/0243_auto_20251014_1340.py
```diff
@@ -0,0 +1,54 @@
...
+    EarmarkedStock = apps.get_model("polio", "EarmarkedStock")
+
+    outgoing_stock_movements = OutgoingStockMovement.objects.filter(doses_per_vial__isnull=True).prefetch_related(
+        "vaccine_stock"
+    )
+    incident_reports = IncidentReport.objects.filter(doses_per_vial__isnull=True).prefetch_related("vaccine_stock")
+    destruction_reports = DestructionReport.objects.filter(doses_per_vial__isnull=True).prefetch_related(
+        "vaccine_stock"
+    )
+    earmarked_stocks = EarmarkedStock.objects.filter(doses_per_vial__isnull=True).prefetch_related("vaccine_stock")
```
tdethier, line 26:
```suggestion
    outgoing_stock_movements = OutgoingStockMovement.objects.filter(doses_per_vial__isnull=True).select_related(
        "vaccine_stock"
    )
    incident_reports = IncidentReport.objects.filter(doses_per_vial__isnull=True).select_related("vaccine_stock")
    destruction_reports = DestructionReport.objects.filter(doses_per_vial__isnull=True).select_related(
        "vaccine_stock"
    )
    earmarked_stocks = EarmarkedStock.objects.filter(doses_per_vial__isnull=True).select_related("vaccine_stock")
```

plugins/polio/migrations/0243_auto_20251014_1340.py
```diff
@@ -0,0 +1,54 @@
...
+        stock.doses_per_vial = doses_per_vial
+        stock.save()
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("polio", "0242_destructionreport_doses_per_vial_and_more"),
+    ]
+
+    operations = [migrations.RunPython(add_doses_per_vial, migrations.RunPython.noop)]
```
tdethier, line 54: We can tell Django to get rid of this during the next squash

```suggestion
    operations = [migrations.RunPython(add_doses_per_vial, migrations.RunPython.noop, elidable=True)]
```

plugins/polio/models/base.py
```diff
@@ -2097,11 +2103,12 @@ class MovementTypeEnum(enum.Enum):
 
 
 class VaccineStockCalculator:
-    def __init__(self, vaccine_stock: VaccineStock):
+    def __init__(self, vaccine_stock: VaccineStock, end_date=None):
```
tdethier, line 2106: Good idea

tdethier (approved):
I don't know the business logic behind OutgoingStockMovement and other models, so I couldn't manually create any object like that through the API (I never get the right parameters 😅)

Otherwise, same comment as the previous PR, but it looks good 👍

PR #2493, opened 2025-10-16
SLEEP-1598: export mobile app setup performance

Phil-V:
Speeds up the retrieval of entities for the `export_mobile_app_setup` task by optimizing the database query and using cursor pagination. 

This is Trypelim-only but I'd like to bring these optimisations to IASO and have a more thorough refactor/review at that point. 

Related JIRA tickets : [SLEEP-1598](https://bluesquare.atlassian.net/browse/SLEEP-1598)

- [x] Are there enough tests?

## Changes

- Optimized the mobile entity viewpoint database query
- Broke off the optimization to a separate endpoint to keep everything else unaffected for now (but don't want to keep it that way, as there is some duplicated functionality).
- Used cursor-based pagination for an additional substantial speed-up
- Adapted the task to preserve the existing json schema for the entities.

## Notes

Another approach that was considered, after talking with Benjamin, is to adapt the task to do straight-up database calls in the task instead of going through the api. This might be a better approach, as well as perhaps calling the viewsets directly instead of using http. 

[SLEEP-1598]: https://bluesquare.atlassian.net/browse/SLEEP-1598?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iaso/tasks/export_mobile_app_setup_for_user.py
+++ b/iaso/tasks/export_mobile_app_setup_for_user.py
@@ -20,6 +20,8 @@ import re
 import uuid
 import zipfile
 
+from dataclasses import dataclass
+from typing import Optional
 from urllib.parse import urlencode, urlparse, urlunparse
 
 import requests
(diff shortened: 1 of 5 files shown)


iaso/tasks/export_mobile_app_setup_for_user.py
```diff
@@ -116,27 +118,108 @@ def _get_project_app_details(iaso_client, tmp_dir, app_id):
...
+    count_query_params = {k: v for k, v in call.get("query_params", {}).items() if k not in ["limit", "page", "cursor"]}
+    count_query_params["app_id"] = app_id
+
+    count_url = urlunparse(("", "", count_path, "", urlencode(count_query_params), ""))
+    count_result = _call_endpoint(iaso_client, count_url, None)
+    total_count = count_result.get("count", 0)
+
+    total_pages = (total_count + page_size - 1) // page_size
+
+    logger.info(f"Total count: {total_count}, Total pages: {total_pages} (Page size: {page_size})")
```
bramj, line 146: (I didn't run the code) Not sure I understand this part, there isn't any `total_pages` for cursor pagination, right?

bramj (approved):
Nice work! 👏 

Damn, Django rest_framework is impressive, I didn't know it supported cursor pagination like this.

This looks nice and it would be a huge performance improvement for IASO to have this generalized.

As for this comment:
> Another approach that was considered, after talking with Benjamin, is to adapt the task to do straight-up database calls in the task instead of going through the api. This might be a better approach, as well as perhaps calling the viewsets directly instead of using http.

The idea was to make absolutely sure that the zip export and the mobile app remain aligned, since we'll inevitably forget to make changes on the former when changing the latter. (This already happened when we added the OU change requests to the mobile app and forgot to add them to the zip)

I would avoid hitting the DB directly because of this, calling viewsets directly could be a fair middle ground. If the endpoint becomes fast however, then I suppose it would no longer be an issue.

iaso/tasks/export_mobile_app_setup_for_user.py
Phil-V: This bit can be somewhat hacky and confusing: I needed to reproduce the API response of the non-cursor pagination. (Especially after having checked with Benjamin who told me that the mobile app did make use of the total count, for instance). I decided to reproduce the api completely so that although we use the cursor pagination internally, we return results using the existing schema. For that, the code makes a single call to the existing pagination api to get the count of results and pages, which is then injected into the cursor-based responses.

{
  "count": 36907,
  "results": [...]
  "has_next": true,
  "has_previous": true,
  "page": 2,
  "pages": 37,
  "limit": 1000
}

PR #2500, opened 2025-10-20
IA-4528 ensure path are re-calculated after geopackage import

mestachs:
Some orgunits don't have a path after geopackage import making a lot feature not functional

Related JIRA tickets : IA-4528

- [ ] Are there enough tests?

## Doc

## Changes

it recalculates the path starting from the "top" parents after all orgunits have been imported.

## How to test

1. create a new account will most modules

2. start the worker : `docker compose run iaso manage tasks_worker`

3. create a datasource version based on the geopackage in the jira

image

create a new version (or update V1)

image

4. afterwards all path should be computed, you can verify in django shell or trust the logs in the task

## Print screen / video

image

## Notes

I've modified the test but the path was already filled for every orgunit
I suspect that the geopackage order processing in the test is too simple ( p1 -> p2-> leaf1) but I haven't been able to locate which pattern trigger this behavior;

```
fix: ensure path are re-calculated after geopackage import

Refs: IA-4528
```

### DIFF
--- a/iaso/gpkg/import_gpkg.py
+++ b/iaso/gpkg/import_gpkg.py
@@ -10,6 +10,7 @@ import fiona  # type: ignore
 from django.contrib.auth.models import User
 from django.contrib.gis.geos import MultiPolygon, Point, Polygon
 from django.db import transaction
+from django.db.models import Q
 
 from hat.audit import models as audit_models
 from iaso.models import DataSource, Group, OrgUnit, OrgUnitType, SourceVersion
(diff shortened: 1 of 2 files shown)


iaso/gpkg/import_gpkg.py
```diff
@@ -404,6 +405,9 @@ def import_gpkg_file2(
         ou.parent = parent_ou
         ou.source_ref = ou.source_ref.replace(OLD_INTERNAL_REF, NEW_INTERNAL_REF)
         ou.save()
+
+    recalculate_missing_paths(version, task)
```
madewulf, line 409: I'm quite surprised this is needed in all cases. This has been working for sure in the past. Did you understand how this bug appeared?

madewulf (approved):

iaso/gpkg/import_gpkg.py
mestachs: the more surprising part

```
from django.db.models import Q
import iaso.models as m
m.OrgUnit.objects.filter(Q(path__isnull=True) | Q(path=[])).count()
```
this return 0 in prod

but in staging we have 68834 of them, 24540 on playground

tdethier: maybe a setuper issue?

PR #2501, opened 2025-10-20
FIX CYPRESS 

beygorghor:
Fix failing cypress tests


tdethier (approved):
Cypress issues are [fixed](https://github.com/BLSQ/iaso/actions/runs/18647766582), thanks 👍

PR #2502, opened 2025-10-20
POLIO-2008: add planned campaign to polio main page

quang-le:
add `is_planned` field to campaign and round, then use it in the main polio page

Related JIRA tickets : POLIO-2008, POLIO-2002

- [x] Are there enough tests?

## Doc

Comments in the code

## Changes

- Backend:
    - Add `is_planned` field to `Campaign` and `Round` models
    - Add constraint on `Round` model, so that planned rounds require `target_population` and `percentage_target_population` fields. The constraint couldn't include the `Campaign`-level field because it's working at the model level.
    - Add API level validation so that `target_population` and `percentage_target_population` are required id either round or campaign is planned
    - Exclude planned rounds (`round.is_planned` or `campaign.is_planned`) from preparedness (iaso-level)

- Frontend:
    - Add checkbox in the round and basic info tabs to toggle the `planned status`
    - Update round validation to check for `target_population` and `percentage_target_population` if either round or campaign is planned
    - Disable Scope and Preparedness tabs if campaign/round is planned
    
## How to test

- Go to polio > campaigns
- Create a campaign
- tick the planned box in the Basic Info tab: 
    - The corresponding tick box  should be deactivated in the rounds tab
    - The preparedness and scope tabs should remain deactivated
    - The round tab should be in error (target pop and percentage target pop fields)
- Save: it should save without issue
- Create another campaign
- Tick the planned box in a round:
   - The subsequent rounds should also be planned
   - Scope and Preparedness tabs should be deactivated 
   - The current and subsequent rounds should be in error (target pop and percentage target pop fields)
- Untick the box: a modal should open asking if you want to update the subsequent rounds as well

Repeat with Editing a campaign

## Print screen / video

https://github.com/user-attachments/assets/dffdd290-bde7-42a4-a5fc-fcbec3ca7af6


plugins/polio/models/base.py
```diff
@@ -388,15 +397,20 @@ class Meta:
     age_max = models.IntegerField(null=True, blank=True)
     age_type = models.TextField(null=True, blank=True, choices=AgeChoices.choices)
 
+    target_population = models.IntegerField(null=True, blank=True)
+    percentage_covered_target_population = models.IntegerField(null=True, blank=True)
```
tdethier, line 401: I know you only moved this line, but it's weird that a percentage is an `IntegerField`

plugins/polio/tests/api/test_round_serializer_validations.py
```diff
@@ -0,0 +1,125 @@
...
+        self.assertFalse(serializer.is_valid())
+
+        # Only the missing field should have an error
+        self.assertNotIn("target_population", serializer.errors)
+        self.assertIn("percentage_covered_target_population", serializer.errors)
+
+        # Check that only one field has an error
+        self.assertEqual(len(serializer.errors), 1)
+
+    def test_both_fields_missing_error(self):
```
tdethier, line 114: The same test setup was already done in `test_planned_round_requires_population_fields_via_round_flag`, you could add the extra asserts there and remove this test

plugins/polio/tests/api/test_round_serializer_validations.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def test_single_field_missing_error(self):
+        # Test that only missing fields are included in errors
+        serializer = self._make_serializer(
+            payload={
+                "is_planned": True,
+                "target_population": 1000,  # Provided
+                "percentage_covered_target_population": None,  # Missing
+            }
+        )
```
tdethier, line 87: That's a great idea, I would test for the same thing with the other field in this test as well

plugins/polio/api/rounds/round.py
```diff
@@ -39,6 +39,31 @@ class Meta:
     def get_vaccine_names_extended(self, obj):
         return obj.vaccine_names_extended
 
+    def validate(self, data):
+        # Check if campaign or round is planned
+        is_campaign_planned = None
+        if self.context.get("request"):
+            is_campaign_planned = self.context["request"].data.get("is_planned", None)
```
tdethier, line 46: I thought it was weird to get that through the context, but I understand that context is used in polio to pass values to children serializers, so it's ok

tdethier (changes requested):
The feature works mostly fine 👍 

I had some issues with mandatory fields: while preparing a new planned campaign, I didn't know which fields were mandatory at the round level and couldn't save because of that

Image

After switching to the campaign tab and switching back to the round tab, validation put the fields in red, so I knew what to do. 

Image

It would be great to make these fields dynamically mandatory depending on the checkboxes before validation checks


-------------------

I had another issue afterwards: I unticked the "planned" checkbox at the campaign level, ticked it at the round level and everything was good. But then I also unticked at the round level and the two fields stayed unvalid - at that point, it's not possible to save the form unless to fill out these fields

Image

It looks like validation checks were not triggered again when I unticked the box at the round level

tdethier (approved):
Frontend validation works fine now, thanks 👍

PR #2505, opened 2025-10-21
WC2-838: Add base info to StockRulesVersion details page

bmonjoie:
The StockRulesVersion details page was missing a way to edit the name and publish from there.

Related JIRA tickets : WC2-838

## Changes

Adding the info at the top of the page
Fixing the default sort order on the list page.

## How to test

Go into the details of one of the version and change the name or publish it.

## Print screen / video

image

```
fix: empty instance pop up

Refs: IA-3665
```

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/stock/versions/components/DetailsForm.tsx
+++ b/hat/assets/js/apps/Iaso/domains/stock/versions/components/DetailsForm.tsx
@@ -1,0 +1,61 @@
+import React, { FunctionComponent, useState, useCallback } from 'react';
+import { Box, Button } from '@mui/material';
+import { makeStyles } from '@mui/styles';
+import { useSafeIntl } from 'bluesquare-components';
+
+import InputComponent from 'Iaso/components/forms/InputComponent';
+import { StockRulesVersion } from 'Iaso/domains/stock/types/stocks';
+
+import { useUpdateStockRulesVersion } from 'Iaso/domains/stock/versions/hooks/requests';
+import MESSAGES from '../../messages';
+
+type Props = {
+    version: StockRulesVersion;
+};
+// @ts-ignore
+const useStyles = makeStyles(theme => ({
+    root: {
+        position: 'relative',
+        '& #input-text-name': {
+            paddingRight: theme.spacing(15),
+        },
+    },
+    button: {
+        position: 'absolute !important',
+        right: theme.spacing(3),
+        top: 26,
+    },
+}));
+export const DetailsForm: FunctionComponent<Props> = ({ version }) => {
+    const classes = useStyles();
+    const [name, setName] = useState<string>(version.name);
+    const { formatMessage } = useSafeIntl();
+    const { mutate: updateVersion } = useUpdateStockRulesVersion();
+    const handleSave = useCallback(() => {
+        updateVersion({ name, id: version.id });
+    }, [name, updateVersion, version.id]);
+    const saveDisabled = name === version.name || name === '';
+    return (
+        <Box p={2} className={classes.root}>
+            <InputComponent
+                withMarginTop={false}
+                keyValue="name"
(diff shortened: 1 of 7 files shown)


hat/assets/js/apps/Iaso/domains/stock/versions/components/DetailsForm.tsx
```diff
@@ -0,0 +1,61 @@
...
+        '& #input-text-name': {
+            paddingRight: theme.spacing(15),
+        },
+    },
+    button: {
+        position: 'absolute !important',
+        right: theme.spacing(3),
+        top: 26,
+    },
+}));
```
beygorghor: We need to avoid `useStyles` please use `const styles: SxStyles = ...` see front-end [doc](https://github.com/BLSQ/iaso/blob/develop/docs/pages/dev/reference/guidelines/front-end/front-end.en.md)

hat/assets/js/apps/Iaso/domains/stock/versions/components/VersionBaseInfo.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { PublishVersionModal } from './PublishVersionModal';
+import { StatusCell } from './StatusCell';
+
+const useStyles = makeStyles(theme => ({
+    leftCell: {
+        // @ts-ignore
+        borderRight: `1px solid ${theme.palette.ligthGray.border}`,
+        fontWeight: 'bold',
+    },
+}));
```
beygorghor: Same here

hat/assets/js/apps/Iaso/domains/stock/versions/components/VersionBaseInfo.tsx
```diff
@@ -0,0 +1,82 @@
...
+            <Table size="small" data-test="workflow-base-info">
+                <TableBody>
+                    <Row
+                        label={formatMessage(MESSAGES.status)}
+                        value={
+                            version ? (
+                                <StatusCell status={version.status} />
+                            ) : (
+                                ''
+                            )
```
beygorghor, line 64: You can handle this directly in the Row component . Also i would use placeholder from bluesquare-component

beygorghor (dismissed):
Just fix the style please, useStyles is the old way to proceed and leads to css override issue with MUI 5 !

hat/assets/js/apps/Iaso/domains/stock/versions/components/VersionBaseInfo.tsx
bmonjoie: I don't understand the "You can handle this directly in the Row component" part

hat/assets/js/apps/Iaso/domains/stock/versions/components/VersionBaseInfo.tsx
```diff
@@ -0,0 +1,84 @@
...
+                <TableBody>
+                    <Row
+                        label={formatMessage(MESSAGES.status)}
+                        value={
+                            version ? (
+                                <StatusCell status={version.status} />
+                            ) : (
+                                textPlaceholder
+                            )
+                        }
```
quang-le, line 65: you can move that logic in your custom `Row` component to make you jsx cleaner

quang-le (changes requested):
Just the one suggestion (same as @beygorghor ).
I'm just blocking so that you add a ticket nr to your PR, it reaaaaally helps the release process

hat/assets/js/apps/Iaso/domains/stock/versions/components/VersionBaseInfo.tsx
bmonjoie: As discussed, I tried but then nothing gets printed :shrug:

quang-le (approved):

PR #2507, opened 2025-10-21
POLIO-2011: data model for integrated campaigns

quang-le:
Need to "link" non polio with polio campaigns fro display in calendar

Related JIRA tickets : POLIO-2011, POLIO-1997

- [ ] Are there enough tests?

## Doc

Doc string in the model

## Changes

- Add `CampaignIntegration` model

## How to test
N/A see ticket 

## Print screen / video

N/A


beygorghor (approved):

tdethier (changes requested):
After discussing this with Son, I think it would better to have a different implementation: a foreign key on `Campaign` to `Campaign` called `parent_campaign` (or `polio_campaign` if we consider that the polio campaign is always the main one).

This solution avoids adding an extra model and still allows a main campaign to have multiple integrated campaigns

quang-le:
replaced by https://github.com/BLSQ/iaso/pull/2517

PR #2511, opened 2025-10-22
POLIO-2017: fix param passed to send_mail

quang-le:
Vaccine authorization emails were not sent

Related JIRA tickets :  POLIO-2017

- [ ] Are there enough tests?

## Doc
NA

## Changes

The django method [sanitise_address](https://github.com/django/django/blob/stable/5.2.x/django/core/mail/message.py#L79) expects either a string or a tuple, but our [missing_vaccine_authorization_for_campaign_email_alert](https://github.com/BLSQ/iaso/blob/develop/plugins/polio/api/campaigns/vaccine_authorization_missing_email.py) was passing a tuple with only one value. That was the source of the error.

## How to test

deploy on staging

## Print screen / video
N/A


tdethier (approved):
LGTM 👍

PR #2516, opened 2025-10-23
POLIO-2018 update vaccine stock endpoint to return doses

quang-le:

Related JIRA tickets : POLIO-2018, POLIO-1998

- [x] Are there enough tests?

## Doc

## How to test

With django API, go to `api/polio/vaccine/vaccine_stock/?limit=10& page=1/`

The response should include all the "doses" fields (cf screenshot)

## Print screen / video

Screenshot 2025-10-23 at 15 33 43

### DIFF
--- a/plugins/polio/models/base.py
+++ b/plugins/polio/models/base.py
@@ -2146,31 +2146,6 @@ class VaccineStockCalculator:
         self.list_of_earmarked = None
         self.list_of_earmarked_expanded = None
 
-        self.total_vials_used = None
-        self.total_doses_used = None
-        self.total_vials_destroyed = None
-        self.total_doses_destroyed = None
-        self.total_usable_vials = None
-        self.total_usable_doses = None
-        self.total_unusable_vials = None
-        self.total_unusable_doses = None
-        self.total_vials_received = None
-        self.total_doses_received = None
-        self.total_earmarked_vials = None
-        self.total_earmarked_doses = None
-        self.list_of_vaccines_received = None
-        self.list_of_vaccines_received_expanded = None
-        self.list_of_usable_vials = None
-        self.list_of_usable_vials_expanded = None
-        self.list_of_used_vials = None
-        self.list_of_used_vials_expanded = None
-        self.list_of_unusable_vials = None
-        self.list_of_unusable_vials_expanded = None
-        self.list_of_vaccines_received = None
-        self.list_of_vaccines_received_expanded = None
-        self.list_of_earmarked = None
-        self.list_of_earmarked_expanded = None
-
     def get_vials_used(self):
         # if totals return totals
         if self.total_vials_used is not None and self.total_doses_used is not None:
--- a/plugins/polio/models/base.py
+++ b/plugins/polio/models/base.py
@@ -1536,13 +1536,13 @@ class VaccineStock(models.Model):
         return f"{self.country} - {self.vaccine}"
 
     def usable_vials(self, end_date=None):
-        return VaccineStockCalculator(self).get_list_of_usable_vials(end_date, expanded=True)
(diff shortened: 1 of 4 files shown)


plugins/polio/models/base.py
```diff
@@ -2121,24 +2121,103 @@ def __init__(self, vaccine_stock: VaccineStock, end_date=None):
...
+        self.list_of_usable_vials = None
+        self.list_of_usable_vials_expanded = None
+        self.list_of_used_vials = None
+        self.list_of_used_vials_expanded = None
+        self.list_of_unusable_vials = None
+        self.list_of_unusable_vials_expanded = None
+        self.list_of_vaccines_received = None
+        self.list_of_vaccines_received_expanded = None
+        self.list_of_earmarked = None
+        self.list_of_earmarked_expanded = None
```
tdethier:
```suggestion
```

tdethier: It's the same list twice, right?

plugins/polio/models/base.py
```diff
@@ -2121,24 +2121,103 @@ def __init__(self, vaccine_stock: VaccineStock, end_date=None):
...
     def get_vials_destroyed(self):
+        if self.total_vials_destroyed is not None:
+            return self.total_vials_destroyed
+        if not self.destruction_reports.exists():
+            self.total_vials_destroyed = 0
+            return 0
+        destruction_reports = self.destruction_reports
+        if self.end_date:
+            destruction_reports = destruction_reports.filter(destruction_report_date__lte=self.end_date)
+        total_vials_destroyed = sum(report.unusable_vials_destroyed or 0 for report in destruction_reports)
```
tdethier: Since this is highly similar to `get_doses_destroyed`, could we process both at the same time and set both values? Instead of iterating all the destruction reports twice

plugins/polio/models/base.py
```diff
@@ -2121,24 +2121,103 @@ def __init__(self, vaccine_stock: VaccineStock, end_date=None):
...
-        return total
+        return total_vials, total_doses
 
     def get_vials_destroyed(self):
+        if self.total_vials_destroyed is not None:
+            return self.total_vials_destroyed
+        if not self.destruction_reports.exists():
+            self.total_vials_destroyed = 0
+            return 0
+        destruction_reports = self.destruction_reports
```
tdethier, line 2167: I know you copied this line from below, but I would just remove that from both methods, it doesn't really bring any value

plugins/polio/models/base.py
```diff
@@ -2121,24 +2121,103 @@ def __init__(self, vaccine_stock: VaccineStock, end_date=None):
...
+        self.list_of_vaccines_received_expanded = None
+        self.list_of_usable_vials = None
+        self.list_of_usable_vials_expanded = None
+        self.list_of_used_vials = None
+        self.list_of_used_vials_expanded = None
+        self.list_of_unusable_vials = None
+        self.list_of_unusable_vials_expanded = None
+        self.list_of_vaccines_received = None
+        self.list_of_vaccines_received_expanded = None
+        self.list_of_earmarked = None
```
tdethier: How come the `list_of_xxxx` stuff is not in the API nor in tests? It's for another PR?

tdethier (approved):
It works fine 👍 

I would refactor a bit the destroyed vials/doses methods though, I think we'll gain a bit of performance there

plugins/polio/models/base.py
quang-le: yep, some rebase side-effect I guess

plugins/polio/models/base.py
quang-le: Because we just use them internally in the calculator AFAICT. I still have them to declare it here though. I should probably prefix them then, like `_list_of`

PR #2517, opened 2025-10-24
POLIO-2011: add integrated_to field to Campaign

quang-le:
Need to link a polio campaign to one or several non polio campaigns

Related JIRA tickets :  POLIO-2011

- [ ] Are there enough tests?

## Changes

- Add `integrated_to` field to `Campaign`. with a ForeignKEy to another campaign
- Add a check in the `save` method to prevent campaigns of type polio to be integrated 
- The check could not be added as a DB constraint since `CampaignType` is itself a table

## How to test

N/A we just have models and tests


plugins/polio/tests/models/test_campaign_models.py
```diff
@@ -390,3 +396,186 @@ def test_subactivities_properties(self):
...
+
+        # Verify country was set from the district's country ancestor
+        campaign.refresh_from_db()
+        self.assertEqual(campaign.country, self.country)
+
+    @patch("plugins.polio.models.base.OrgUnit.DoesNotExist")
+    def test_save_handles_org_unit_without_country_ancestor(self, mock_does_not_exist):
+        """Test that save handles OrgUnit.DoesNotExist exception gracefully"""
+        # Mock the exception to be raised
+        mock_does_not_exist.side_effect = OrgUnit.DoesNotExist()
```
tdethier, line 553: I don't think you need to use a mock since you're not doing anything with it. The goal is to make sure that the exception was caught and the `Campaign` was properly saved, so you can simply make sure there's one more entry in the database after creating the new `Campaign` and assert that it has an `id` attribute

tdethier (approved):
LGTM 👍 

It's easier to read than the previous version

PR #2522, opened 2025-10-28
IA-3772: Trim whitespace on org units, entities, forms

Phil-V:
Trim whitespaces when creating or updating Org Units. 

DRF's Serializers do this by default on CharFields (trim_whitespace=True), but in places where we don't use serializers some extra whitespaces could creep in.

Related JIRA tickets : IA-3772

- [x] Are there enough tests?

## Doc

/

## Changes

- Refactor the `import_data` function responsible for creating new org units to use a Serializer.
- Remove an old duplicated mobile api version of `import_data`.
- Sprinkle .strip() in other places where OrgUnits are created/modified but where a full refactoring seemed too ambitious.
- Remove legacy `validated` OrgUnit field that was marked for removal (maintenance task unrelated to this issue). 
- Remove the `validated` attribute on the OrgUnit model (migration).

## How to test

- Create or edit an Org Unit in the dashboard at `/dashboard/orgunits/list/`. Extra whitespaces added to the name should be trimmed when saved.
- Create an Org Unit on the mobile app, both with zip upload enabled and without (toggle the feature flag "Mobile: Show entities screen"). Check that the `name` and `source_created_at` attributes are set correctly when imported. 

## Print screen / video

/ 

## Notes

- Limited my changes to OrgUnits, although the original ticket mentioned other models:
    - ChangeRequests and Forms already use Serializers everywhere and should already trim whitespaces by default.
    - Entities use an Instance for their attributes, so trimming whitespaces would have to happen inside of the Instance's json/xml content. That seemed tricky and out of scope for this ticket, but let me know if there's a straightforward way to do that.

- ~~It seems odd that `source_created_at` is not set in the `import_data` function that I removed. I chose to preserve this behavior in my refactoring but perhaps it was a bug and that logic could be removed.~~

```
fix: empty instance pop up

Refs: IA-3665
```


iaso/models/org_unit.py
```diff
@@ -278,7 +278,6 @@ class OrgUnit(TreeModel):
     name = models.CharField(max_length=255)
     uuid = models.TextField(null=True, blank=True, db_index=True)
     custom = models.BooleanField(default=False)
-    validated = models.BooleanField(default=True, db_index=True)  # TO DO : remove in a later migration
```
quang-le, line 281: That one took a long time for us to remove 😂

iaso/api/serializers.py
```diff
@@ -247,3 +249,114 @@ def get_org_unit_type_short_name(self, org_unit):
...
+
+    def create(self, validated_data):
+        """Implement get_or_create logic and set model fields not provided by the client."""
+        uuid = validated_data.pop("uuid")
+        org_unit, created = OrgUnit.objects.get_or_create(uuid=uuid)
+
+        if not created:
+            return None  # only handle new org units
+
+        t = validated_data.pop("created_at", None)
```
quang-le: Not a big fan of single letter variables

Phil-V: Me neither! Fixed 😉

iaso/models/org_unit.py
Phil-V: thanks for taking a look!

iaso/api/org_units.py
```diff
@@ -805,7 +806,15 @@ def create_org_unit(self, request):
         profile = request.user.iaso_profile
         if request.user:
             org_unit.creator = request.user
-        name = request.data.get("name", None)
+
+        name = request.data.get("name")
```
tdethier, line 809: Did you remove the `None` fallback on purpose? I don't know if a payload can be sent without the `name` key, but if it does, it's going to crash here

iaso/api/org_units.py
```diff
@@ -1011,55 +1015,35 @@ def retrieve(self, request, pk=None):
...
+    for org_unit_data in org_units:
+        serializer = OrgUnitImportSerializer(data=org_unit_data, context=serializer_context)
+        serializer.is_valid(raise_exception=True)
 
-            if not user.is_anonymous:
-                org_unit_db.creator = user
-            org_unit_db.source = "API"
-            if org_unit_location:
-                org_unit_db.location = org_unit_location
+        org_unit_db = serializer.save(**extra_save_kwargs)
```
tdethier: :+1: I like that

iaso/api/serializers.py
```diff
@@ -247,3 +249,114 @@ def get_org_unit_type_short_name(self, org_unit):
...
+
+
+class OrgUnitImportSerializer(serializers.ModelSerializer):
+    """Helper class for the validation and creation of OrgUnit instances from API data."""
+
+    # Map the api's incoming 'id' field to the model's 'uuid' field.
+    id = serializers.CharField(source="uuid")
+
+    name = serializers.CharField(required=False, allow_null=True)
+    accuracy = serializers.FloatField(required=False, allow_null=True)
```
tdethier: I would probably use `DecimalField` to avoid rounding issues (and because `OrgUnitChangeRequestWriteSerializer` uses it for its accuracy field in change requests)

iaso/api/serializers.py
```diff
@@ -247,3 +249,114 @@ def get_org_unit_type_short_name(self, org_unit):
...
+    location = ThreeDimPointField(required=False, allow_null=True, write_only=True)
+
+    # internal non-model fields
+    created_at = serializers.FloatField(required=False, allow_null=True, write_only=True)
+    parent_lookup = serializers.CharField(required=False, allow_null=True, write_only=True)
+    org_unit_type_id_lookup = serializers.CharField(required=False, allow_null=True, write_only=True)
```
tdethier, line 276: I'm wondering if we should use an `IntegerField`

iaso/api/serializers.py
```diff
@@ -247,3 +249,114 @@ def get_org_unit_type_short_name(self, org_unit):
...
+            "created_at",
+            "parent_lookup",
+            "org_unit_type_id_lookup",
+            "parent_id",
+            "org_unit_type_id",
+        ]
+        extra_kwargs = {
+            "created_at": {"write_only": True},
+            "parent_lookup": {"write_only": True},
+            "org_unit_type_id_lookup": {"write_only": True},
```
tdethier: Do you need these 3 lines if you've already passed the `write_only` parameter when declaring the fields above (outside the `Meta` class)?

iaso/api/serializers.py
```diff
@@ -247,3 +249,114 @@ def get_org_unit_type_short_name(self, org_unit):
...
+            "org_unit_type_id": {"required": False},
+        }
+
+    def to_internal_value(self, data):
+        """Pre-process the incoming dictionary and handle legacy data formats."""
+        processed_data = {}
+
+        latitude = data.get("latitude")
+        longitude = data.get("longitude")
+        if latitude is None or longitude is None or (not latitude and not longitude):
```
tdethier:
```suggestion
        if not latitude or not longitude:
```

iaso/api/serializers.py
```diff
@@ -247,3 +249,114 @@ def get_org_unit_type_short_name(self, org_unit):
...
+
+        # there exist versions of the mobile app in the wild with both orgUnitTypeId and org_unit_type_id
+        type_key = data.get("orgUnitTypeId") or data.get("org_unit_type_id")
+        if type_key:
+            processed_data["org_unit_type_id_lookup"] = type_key
+
+        processed_data["id"] = data.get("id")
+        processed_data["name"] = data.get("name")
+        processed_data["accuracy"] = data.get("accuracy")
+        processed_data["created_at"] = data.get("created_at")
```
tdethier, line 325: You're adding these fields without any modification to "get rid of" `data` and only continue further with `processed_data`?

iaso/api/org_units.py
```diff
@@ -1011,55 +1015,35 @@ def retrieve(self, request, pk=None):
...
 
+        if org_unit_db:
```
tdethier:
```suggestion
        new_org_unit = serializer.save(**extra_save_kwargs)

        if new_org_unit:
```

iaso/api/org_units.py
```diff
@@ -586,7 +587,7 @@ def partial_update(self, request, pk=None):
         original_copy = deepcopy(org_unit)
 
         if "name" in request.data:
-            org_unit.name = request.data["name"]
+            org_unit.name = request.data["name"].strip() if request.data["name"] else request.data["name"]
```
tdethier:
```suggestion
        org_unit_name = request.data.get("name", None)
        if org_unit_name:
                    org_unit_name = org_unit_name.strip()
        org_unit.name = org_unit_name
```

tdethier: It's a nitpick, but I thought it was weird to read `if request.data["name"] else request.data["name"]`

iaso/tasks/process_mobile_bulk_upload.py
```diff
@@ -66,7 +66,7 @@ def process_mobile_bulk_upload(api_import_id, project_id, task=None):
             with zipfile.ZipFile(api_import.file, "r") as zip_ref:
                 if ORG_UNITS_JSON in zip_ref.namelist():
                     org_units_data = read_json_file_from_zip(zip_ref, ORG_UNITS_JSON)
-                    new_org_units = import_org_units(org_units_data, user, project.app_id)
+                    new_org_units = import_org_units(org_units_data, user, project.app_id, set_source_created_at=False)
```
tdethier: Why did you set `set_source_created_at` to `False`?

iaso/api/mobile/org_units.py
```diff
@@ -293,7 +292,10 @@ def list(self, request, *args, **kwargs):
 
     @safe_api_import("orgUnit")
     def create(self, _, request):
-        new_org_units = import_data(request.data, request.user, request.query_params.get(APP_ID))
+        data = sorted(request.data, key=lambda ou: float(ou["created_at"]))
+        new_org_units = import_org_units(
+            data, request.user, request.query_params.get(APP_ID), set_source_created_at=False
```
tdethier: Here again I'm not sure why `set_source_created_at` is `False`, but I'm not 100% familiar with the current flow

tdethier (commented):
I left a few comments here and there, but:
- overall the code is much cleaner now.
- since something changed with `source_created_at` and `created_at`, I think we should have more testing on these two fields in our unit tests
- I got a few issues when I tried to create a new OrgUnit with the mobile app, but I'm not sure if it was a setup issue or a bug. We should discuss that together. 

I'll validate or request changes after the discussion 😅

iaso/api/org_units.py
tdethier: nitpick

iaso/api/serializers.py
tdethier: Unless we consider something like `latitude=""` and `longitude="14.251154"` as valid. In that case, you can disregard this suggestion :sweat_smile:

iaso/tasks/process_mobile_bulk_upload.py
Phil-V: Thanks for the review! When I first started refactoring, I saw 2 very similar implementations of import_data (one in `iaso/api/org_units.py` and one in `iaso/api/mobile/org_units.py`). They seemed to have slight differences. The one in the mobile API is used for the create endpoint and the bulk import tasks, and to my understanding _doesn't_ set the `source_created_at` field. Unsure if this was intended, I made sure to replicate this behavior in my refactoring for now. But if we suspect it's a bug, we could simplify the implementation. We could talk further about this.

iaso/api/serializers.py
Phil-V: That was my initial thought as well, but it proved too restrictive and broke some tests that have a fractional part for the `created_at` timestamp (milliseconds?)

iaso/api/serializers.py
Phil-V: Good catch! 👍

iaso/api/serializers.py
Phil-V: Yes this is to avoid mutating the `data` dict in-place, but it's debatable as an approach

tdethier: It's ok, I don't mind

tdethier (approved):
The feature works fine, thanks 👍 

Kudos for implementing a serializer in OrgUnits, we definitely need to rewrite this API and this is the first step towards that!

PR #2524, opened 2025-10-29
POLIO-2019 doses per vial

quang-le:

Related JIRA tickets :  POLIO-2019, POLIO-1998

- [x] Are there enough tests?

## Doc

[Will edit this section when doc is written](https://github.com/BLSQ/polio-doc/pull/4)

## Changes

- Backend:
    - Add an endpoint returning the possible presentations (doses_per_vial) based on a `Config` object. The slug of the config is hard coded/stored in a constant. For use in Supply chain, where we need to show all possible options
    - Add an endpoint returning the possible presentations and the number of usable doses and unusable doses for each presentation. For use in vaccine stock, where some actions are disabled if there's no stock available
    - Add methods to `VaccineStockCalculator` to compute the totals needed for the new API

- Front-end
    - SupplyChain:
        -  PreAlert: 
            - Replace `NumberInput` with `SingleSelect`
            - Set a `defaultDosesPerVial` only if there's only 1 possible presentation
            - Make doses per vial uneditable if there's only one presentation
        - Arrival Report: same as PreAlert, plus:
            - auto-fill `doses_per_vial` when user picks a PO number from the auto-generated list (from pre-alerts)
     - Vaccine Stock:
         - FormA, Earmarked:
             - disable creation if there's no stock of usable doses
              - Replace `NumberInput` with `SingleSelect`
              - Only show presentations for which stock of usable doses > 0 in the dropdown options
              - Edition: add the saved value for presentation/doses per vial to the dropdown to avoid having an errored UI
          - Destruction reports:
              - Same as FormA and Earmarked, but based on stock of unusable doses
          - Incident reports:
              - Never disable Creation, to allow for encoding of Physical Inventory
              - Disable movements from usable stock to unusable if usable stock <= 0
              - Physical Inventory: disable some options based on usable/unusable stock being available 
              
## How to test

With a DB setup for polio, go to the django admin and add a `Config` with the following:
slug=vaccine_doses_per_vial
content=
```json
{
  "bOPV": [
    10,
    20
  ],
  "mOPV2": [
    20,
    50
  ],
  "nOPV2": [
    50
  ]
}
```

Then go to supply chain:
- Try adding a PreAlert: change the values --> the doses and vial values should update. Saving should work OK
- Try adding a vaccine arrival report: select a PO number from the dropdown --> The doses per vial should auto fill

Go to vaccine stock:
- Try adding/editing each type of report for every vaccine. nOPV2 should have a different behaviour: the dropdowns should be (pre)filled with 50 and not be editable

## Print screen / video

https://github.com/user-attachments/assets/80d23082-999c-4d8f-9164-aa41717f8711

https://github.com/user-attachments/assets/7ab8affd-3905-459d-a9ff-429efb23a1e8

### DIFF
--- a/hat/assets/js/apps/Iaso/components/forms/InputComponent.tsx
+++ b/hat/assets/js/apps/Iaso/components/forms/InputComponent.tsx
@@ -285,7 +285,6 @@ const InputComponent: React.FC<InputComponentProps> = ({
                         dataTestId={dataTestId}
                         placeholder={placeholder}
                         onBlur={onBlur}
-                        // onFocus={onFocus}
                     />
                 );
             case 'arrayInput':
--- a/hat/assets/js/apps/Iaso/components/forms/InputComponent.tsx
+++ b/hat/assets/js/apps/Iaso/components/forms/InputComponent.tsx
@@ -284,6 +284,8 @@ const InputComponent: React.FC<InputComponentProps> = ({
                         returnFullObject={returnFullObject}
                         dataTestId={dataTestId}
                         placeholder={placeholder}
+                        onBlur={onBlur}
+                        // onFocus={onFocus}
                     />
                 );
             case 'arrayInput':
--- a/plugins/polio/api/urls.py
+++ b/plugins/polio/api/urls.py
@@ -41,7 +41,7 @@ from plugins.polio.api.polio_org_units import PolioOrgunitViewSet
 from plugins.polio.api.rounds.reasons_for_delay import ReasonForDelayViewSet
 from plugins.polio.api.rounds.round import RoundViewSet
 from plugins.polio.api.rounds.round_date_history import RoundDateHistoryEntryViewset
-from plugins.polio.api.vaccines.doses_per_vaccine import DosesPerVaccineViewset
+from plugins.polio.api.vaccines.doses_per_vaccine import DosesPerVialViewset
 from plugins.polio.api.vaccines.public_vaccine_stock import PublicVaccineStockViewset
 from plugins.polio.api.vaccines.repository_forms import VaccineRepositoryFormsViewSet
 from plugins.polio.api.vaccines.repository_reports import VaccineRepositoryReportsViewSet
@@ -111,7 +111,7 @@ router.register(r"polio/tasks/refreshim/ohh", RefreshIMOutOfHouseholdDataViewset
 router.register(r"polio/tasks/refreshim/hh_ohh", RefreshIMAllDataViewset, basename="refreshimhhohh")
 router.register(r"polio/vaccine/request_forms", VaccineRequestFormViewSet, basename="vaccine_request_forms")
 router.register(r"polio/vaccine/vaccine_stock", VaccineStockManagementViewSet, basename="vaccine_stocks")
-router.register(r"polio/vaccine/doses_per_vial", DosesPerVaccineViewset, basename="doses_per_vial")
+router.register(r"polio/vaccine/doses_per_vial", DosesPerVialViewset, basename="doses_per_vial")
 router.register(r"polio/vaccine/repository", VaccineRepositoryFormsViewSet, basename="vaccine_repository")
 router.register(
     r"polio/vaccine/repository_reports", VaccineRepositoryReportsViewSet, basename="vaccine_repository_reports"
--- a/plugins/polio/api/urls.py
+++ b/plugins/polio/api/urls.py
@@ -41,6 +41,7 @@ from plugins.polio.api.polio_org_units import PolioOrgunitViewSet
 from plugins.polio.api.rounds.reasons_for_delay import ReasonForDelayViewSet
 from plugins.polio.api.rounds.round import RoundViewSet
 from plugins.polio.api.rounds.round_date_history import RoundDateHistoryEntryViewset
+from plugins.polio.api.vaccines.doses_per_vaccine import DosesPerVaccineViewset
 from plugins.polio.api.vaccines.public_vaccine_stock import PublicVaccineStockViewset
 from plugins.polio.api.vaccines.repository_forms import VaccineRepositoryFormsViewSet
 from plugins.polio.api.vaccines.repository_reports import VaccineRepositoryReportsViewSet
@@ -110,6 +111,7 @@ router.register(r"polio/tasks/refreshim/ohh", RefreshIMOutOfHouseholdDataViewset
 router.register(r"polio/tasks/refreshim/hh_ohh", RefreshIMAllDataViewset, basename="refreshimhhohh")
 router.register(r"polio/vaccine/request_forms", VaccineRequestFormViewSet, basename="vaccine_request_forms")
 router.register(r"polio/vaccine/vaccine_stock", VaccineStockManagementViewSet, basename="vaccine_stocks")
+router.register(r"polio/vaccine/doses_per_vial", DosesPerVaccineViewset, basename="doses_per_vial")
 router.register(r"polio/vaccine/repository", VaccineRepositoryFormsViewSet, basename="vaccine_repository")
 router.register(
     r"polio/vaccine/repository_reports", VaccineRepositoryReportsViewSet, basename="vaccine_repository_reports"
--- a/plugins/polio/api/vaccines/doses_per_vaccine.py
+++ b/plugins/polio/api/vaccines/doses_per_vaccine.py
@@ -1,0 +1,53 @@
+from rest_framework import permissions
+from rest_framework.mixins import ListModelMixin
+from rest_framework.viewsets import GenericViewSet
+
+from iaso.api.common import HasPermission
+from iaso.api.config import ConfigSerializer
+from iaso.models.json_config import Config
+from plugins.polio.permissions import (
+    POLIO_CONFIG_PERMISSION,
+    POLIO_PERMISSION,
+    POLIO_VACCINE_STOCK_EARMARKS_ADMIN_PERMISSION,
+    POLIO_VACCINE_STOCK_EARMARKS_NONADMIN_PERMISSION,
+    POLIO_VACCINE_STOCK_EARMARKS_READ_ONLY_PERMISSION,
+    POLIO_VACCINE_STOCK_MANAGEMENT_READ_ONLY_PERMISSION,
+    POLIO_VACCINE_STOCK_MANAGEMENT_READ_PERMISSION,
+    POLIO_VACCINE_STOCK_MANAGEMENT_WRITE_PERMISSION,
+    POLIO_VACCINE_SUPPLY_CHAIN_READ_ONLY_PERMISSION,
+    POLIO_VACCINE_SUPPLY_CHAIN_READ_PERMISSION,
+    POLIO_VACCINE_SUPPLY_CHAIN_WRITE_PERMISSION,
+)
+
+
+class DosesPerVaccineViewset(ListModelMixin, GenericViewSet):
+    """
+    Super custom endpoint to send the possible vaccine presentation (doses per vial) to the front end
+
+    A config with a slug == self.CONFIG_SLUG  is required (create it via the django admin)
+
+    The config data should have 1 key per available vaccine, and the value should be a list of positive integers, eg: 'bOPV':[10,20]
+
+    """
+
+    permission_classes = [
+        permissions.IsAuthenticated,
+        HasPermission(
+            POLIO_CONFIG_PERMISSION,
+            POLIO_PERMISSION,
+            POLIO_VACCINE_SUPPLY_CHAIN_READ_ONLY_PERMISSION,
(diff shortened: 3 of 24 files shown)


hat/assets/js/apps/Iaso/components/forms/InputComponent.tsx
```diff
@@ -284,6 +284,8 @@ const InputComponent: React.FC<InputComponentProps> = ({
                         returnFullObject={returnFullObject}
                         dataTestId={dataTestId}
                         placeholder={placeholder}
+                        onBlur={onBlur}
+                        // onFocus={onFocus}
```
Bewi: Clean up comment

plugins/polio/api/urls.py
```diff
@@ -110,6 +111,7 @@
 router.register(r"polio/tasks/refreshim/hh_ohh", RefreshIMAllDataViewset, basename="refreshimhhohh")
 router.register(r"polio/vaccine/request_forms", VaccineRequestFormViewSet, basename="vaccine_request_forms")
 router.register(r"polio/vaccine/vaccine_stock", VaccineStockManagementViewSet, basename="vaccine_stocks")
+router.register(r"polio/vaccine/doses_per_vial", DosesPerVaccineViewset, basename="doses_per_vial")
```
Bewi: Maybe rename the views to match the route ?

plugins/polio/js/src/domains/VaccineModule/StockManagement/StockVariation/Modals/CreateEditDestruction.tsx
```diff
@@ -45,12 +50,15 @@ export const CreateEditDestruction: FunctionComponent<Props> = ({
...
 }) => {
     const { formatMessage } = useSafeIntl();
     const { mutateAsync: save } = useSaveDestruction();
     const validationSchema = useDestructionValidation();
-    const { data: dosesOptions, isLoading: isLoadingDoses } =
-        useGetDosesOptions(parseInt(vaccineStockId, 10));
+
+    const defaultDosesPerVial =
+        //@ts-ignore
+        (dosesOptions ?? []).length === 1 ? dosesOptions[0].value : undefined;
```
Bewi: It feels a bit hacky, wouldn't it be better to have a default on dosesOptions ?

plugins/polio/js/src/domains/VaccineModule/StockManagement/StockVariation/Modals/CreateEditEarmarked.tsx
```diff
@@ -76,7 +85,24 @@ export const CreateEditEarmarked: FunctionComponent<Props> = ({
         round => round.value === formik.values.round_number,
     );
     const earmarkTypeOptions = useEarmarkOptions();
-
+    const availableDosesPresentations: DropdownOptions<number>[] =
```
Bewi: Feels like this is 99% the same as availableDosesPresentations of CreateEditDestruction, do you think it would be possible to have it in one place with a param for the key to use (doses_available) and earmark/destruction as param ?

plugins/polio/js/src/domains/VaccineModule/StockManagement/StockVariation/VaccineStockVariation.tsx
```diff
@@ -72,6 +73,11 @@ export const VaccineStockVariation: FunctionComponent = () => {
         baseUrl,
     });
 
+    const { data: dosesOptions } = useGetDosesOptions(parseInt(params.id, 10));
+    const hasUsableStock =
```
Bewi, line 77: Would a useMemo make sense here ?

plugins/polio/js/src/domains/VaccineModule/StockManagement/hooks/api.ts
```diff
@@ -253,9 +253,18 @@ export const useGetEarmarkedList = (
     });
 };
 
-export const useGetDosesOptions = (stockId: number) => {
+export const useGetDosesOptions = (
+    stockId: number,
+): UseQueryResult<
+    {
```
Bewi: This is used in multiple places, a type would help to prevent misalignment in the future

plugins/polio/js/src/domains/VaccineModule/StockManagement/StockVariation/Modals/CreateEditDestruction.tsx
Bewi: Or to not allow null and let the parent define [] if dosesOptions is undefined

Bewi (commented):
This is a first review for the FE mainly, I still need to run it locally

plugins/polio/models/base.py
```diff
@@ -2792,3 +2795,61 @@ def get_list_of_earmarked(self, expanded=False):
...
+            if vial["doses_per_vial"] == option:
+                if vial["vials_in"]:
+                    total_unusable_vials += vial["vials_in"]
+                if vial["doses_in"]:
+                    total_unusable_doses += vial["doses_in"]
+                if vial["vials_out"]:
+                    total_unusable_vials -= vial["vials_out"]
+                if vial["doses_out"]:
+                    total_unusable_doses -= vial["doses_out"]
+        return total_unusable_vials, total_unusable_doses
```
tdethier: We can probably refactor both into a single method and pass a parameter to determiner which method should be call to fetch doses and vials data

plugins/polio/models/base.py
```diff
@@ -2792,3 +2795,61 @@ def get_list_of_earmarked(self, expanded=False):
...
+        if not presentation_config:
+            return None
+
+        options = presentation_config.content[self.vaccine_stock.vaccine]
+        if not options:
+            return None
+        results = {}
+        for option in options:
+            results[str(option)] = self._get_unusable_stock_for_presentation(option)[1]
+        return results
```
tdethier, line 2816: If `self._get_(un)usable_stock_for_presentation` is refactored, you can easily extract the code of both into a single method and pass a parameter to select usuable or unsuable because here again the content is the same

plugins/polio/js/src/domains/VaccineModule/StockManagement/StockVariation/Modals/CreateEditDestruction.tsx
quang-le: We only want to preset the value in the UI if there's 1 option, and we want to use the value sent by the API, so this pretty much spells out the business logic. The only reason there's a null check on `dosesOptions` is to prevent errors in the UI while data is being fetched

plugins/polio/js/src/domains/VaccineModule/StockManagement/StockVariation/VaccineStockVariation.tsx
quang-le: No, because it's a primitive (boolean) so there's no reference trap

plugins/polio/api/vaccines/doses_per_vaccine.py
```diff
@@ -0,0 +1,54 @@
...
+            POLIO_VACCINE_STOCK_MANAGEMENT_READ_ONLY_PERMISSION,
+            POLIO_VACCINE_STOCK_MANAGEMENT_READ_PERMISSION,
+            POLIO_VACCINE_STOCK_MANAGEMENT_WRITE_PERMISSION,
+            POLIO_VACCINE_STOCK_EARMARKS_READ_ONLY_PERMISSION,
+            POLIO_VACCINE_STOCK_EARMARKS_NONADMIN_PERMISSION,
+            POLIO_VACCINE_STOCK_EARMARKS_ADMIN_PERMISSION,
+        ),
+    ]
+    serializer_class = ConfigSerializer
+    CONFIG_SLUG = DOSES_PER_VIAL_CONFIG_SLUG
```
tdethier: Nitpick: this works fine, but I would have directly used the constant instead of defining a new constant with the contstant

plugins/polio/js/src/domains/VaccineModule/StockManagement/StockVariation/Modals/CreateEditEarmarked.tsx
quang-le: Good catch, I'll try it, see if I don't get weird state synchro bugs

plugins/polio/tests/api/vaccine_stock_management/test_doses_per_vaccine_endpoint.py
```diff
@@ -0,0 +1,49 @@
...
+        cls.content = {"mOPV2": [50], "nOPV2": [20, 50], "bOPV": [10, 20]}
+        Config.objects.create(slug=DOSES_PER_VIAL_CONFIG_SLUG, content=cls.content)
+
+        # Irrelevant config should be ignored by the queryset filter
+        Config.objects.create(slug="unrelated_config", content={"foo": "bar"})
+
+    def test_anonymous_user_cannot_list(self):
+        self.client.force_authenticate(self.anon)
+        response = self.client.get(BASE_URL)
+        self.assertEqual(response.status_code, 403)
```
tdethier, line 30: It's weird that it's not a 401 - I guess it happens if you don't `force_authenticate` at all

plugins/polio/tests/api/vaccine_stock_management/test_doses_per_vaccine_endpoint.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def test_user_without_perms_cannot_list(self):
+        self.client.force_authenticate(self.user_no_perms)
+        response = self.client.get(BASE_URL)
+        self.assertEqual(response.status_code, 403)
```
tdethier, line 35: Nitpick: I prefer using constants from `rest_framework.status` instead of `int` values

plugins/polio/tests/test_vaccine_stock_management.py
```diff
@@ -1736,225 +1748,146 @@ def test_doses_options_endpoint_invalid_stock_id(self):
...
+        # Check that each result includes unusable_doses field
+        for item in results:
+            self.assertIn("unusable_doses", item)
+            self.assertIsInstance(item["unusable_doses"], int)
+
+        # Verify specific values based on test data
+        # We expect some unusable doses for 20-dose vials based on our test data
+        item_20 = next((item for item in results if item["value"] == 20), None)
+        self.assertIsNotNone(item_20)
+        self.assertGreaterEqual(item_20["unusable_doses"], 0)
```
tdethier: Why `assertGreaterEqual` and not `assertEqual` with the right value?

plugins/polio/api/vaccines/doses_per_vaccine.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def get_queryset(self):
+        return Config.objects.filter(slug=self.CONFIG_SLUG)
```
tdethier: I'm still not a big fan of `Config`, but I guess this works fine 😅

tdethier (approved):
The backend looks good, but I left a few comments.

I tested the feature, it works fine 👍 

I have a couple of suggestions regarding the user experience:
1. when I create a new form A for a nOPV2 vaccine, the `50` value is hardcoded, but it generates a validation error. It stills works, but it's weird to have the error straight away.
Image
 (by the way, the calls made to the backend to load dropdown values are crazy 😵‍💫 

```
http://localhost:8081/api/polio/campaigns/?campaign_category=all&show_test=true&on_hold=true&is_embedded=false&enabled=true
| RESP |  default  |   3314   |    0     |   3314   |    2864    |
Total queries: 3314 in 45.1416s 
```

2. when I prepare a new pre alert, I would expect the order of fields to be the opposite:
Image
Since the doses per vial change the other fields, it makes sense to fill it out first and then do the others in that order

The vaccine arrival reports form has a similar issue.

plugins/polio/api/vaccines/doses_per_vaccine.py
quang-le: Oh yeah, that's a leftover from dev

plugins/polio/api/vaccines/doses_per_vaccine.py
quang-le: Yep, arbitration between time available and need for configurability

plugins/polio/js/src/domains/VaccineModule/StockManagement/StockVariation/VaccineStockVariation.tsx
```diff
@@ -112,7 +112,7 @@ export const VaccineStockVariation: FunctionComponent = () => {
         summary?.country_name,
         summary?.vaccine_type,
     );
-
+    console.log('HAS USABLE', hasUsableStock, dosesOptions, tab);
```
Bewi: Clean up :D

tdethier (approved):
Backend looks good, thanks for the changes!

PR #2525, opened 2025-10-30
WC2-813: DHIS2 Integration

butofleury:

Related JIRA tickets : [WC2-813](https://bluesquare.atlassian.net/browse/WC2-813)

- [ ] Are there enough tests?

## Doc

## Changes

The goal is to aggregate ETL(steps and visits within journey) data per org unit and period(month and year) and store it in the 
`MonthlyStatistics` analytic table and then push to dhis2 and store the push result in the new table `Dhis2SyncResults`

## How to test

As it's about pushing nutrition data to dhis2, it will require to add in env file some additional variable like:
`DHIS2_URL, DHIS2_USER and DHIS2_PASSWORD`

Then, run all migrations with `docker compose run iaso manage migrate wfp`

After, run the command to generate ETL data with push also to dhis2: `docker compose run iaso manage etl_ssd`

## Print screen / video

## Notes

```
fix: empty instance pop up

Refs: IA-3665
```

[WC2-813]: https://bluesquare.atlassian.net/browse/WC2-813?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


plugins/wfp/locale/fr/LC_MESSAGES/django.po
```diff
@@ -46,10 +46,18 @@ msgstr "Retrait volontaire"
...
+msgid "Transfer in from other OTP"
+msgstr "Transfer in from other OTP"
+
 #: plugins/wfp/models.py
 msgid "Transfer To TSFP"
 msgstr "Transfert vers TSFP"
 
+#: plugins/wfp/models.py
+msgid "Transfer in from other TSFP"
+msgstr "Transfer in from other TSFP"
```
tdethier:
```suggestion
msgstr "Transfert depuis autre OTP"

#: plugins/wfp/models.py
msgid "Transfer To TSFP"
msgstr "Transfert vers TSFP"

#: plugins/wfp/models.py
msgid "Transfer in from other TSFP"
msgstr "Transfert depuis autre TSFP"
```

plugins/wfp/migrations/0039_remove_dhis2syncresults_org_unit_and_more.py
```diff
@@ -0,0 +1,21 @@
+# Generated by Django 4.2.25 on 2025-10-30 10:26
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
tdethier: There are so many migrations in this PR, I would simply delete all of them and regenerate a single one with everything, if that has not been deployed anywhere yet.

plugins/wfp/common.py
```diff
@@ -593,6 +597,15 @@ def save_visit(self, visits, journey):
...
             visit.date = current_visit.get("date", None)
             visit.number = visit_number
+            visit.muac_size = current_visit.get("muac", current_visit.get("muac_size"))
+            whz_color = ""
+            if current_visit.get("whz_color", None) == "Y":
+                whz_color = "Yellow"
+            elif current_visit.get("whz_color", None) == "R":
+                whz_color = "Red"
+            elif current_visit.get("whz_color", None) == "G":
+                whz_color = "Green"
```
tdethier, line 608: Nitpick: you could prepare a `dict` for a mapping like that, but it works fine

plugins/wfp/common.py
```diff
@@ -781,23 +819,108 @@ def journey_with_visit_and_steps_per_visit(self, account, programme):
                 ),
             )
             .annotate(org_unit=F("visit__org_unit_id"))
+            .annotate(
+                muac_under_11_5=Sum(
+                    Case(
+                        When(visit__muac_size__lt=11.5, then=Value(1)),
+                        default=Value(0),
+                        output_field=FloatField(),
```
tdethier: Why is the result `FloatField` if you're adding `1` and `0`? `IntegerField`?

plugins/wfp/common.py
```diff
@@ -812,3 +935,105 @@ def journey_with_visit_and_steps_per_visit(self, account, programme):
...
+        row = {}
+        dataElements = None
+        # Reading the dhis2 datalement mapper json file
+        with open("plugins/wfp/dhis2_mapper.json") as mapper:
+            data = json.load(mapper)
+
+        for org_unit, journeys in journey_by_org_units:
+            journeys = list(journeys)
+            journey_by_org_units_period = groupby(journeys, key=itemgetter("period"))
+            row["dataSet"] = "m2GaBFDJDeV"
```
tdethier: Is `"m2GaBFDJDeV"` a fixed ID in DHIS2 in general, or a specific ID for a specific DHIS2 account ? 

I would probably define a constant if it's a global DHIS2 ID, or store a specific DHIS2 ID somewhere (not sure what is the most appropriate way here, because I don't really know what it is: in `settings.py` based on en environment variable, in a dedicated model, in a `Config` object...)

plugins/wfp/common.py
```diff
@@ -812,3 +935,105 @@ def journey_with_visit_and_steps_per_visit(self, account, programme):
...
+        monthlyStatistics = (
+            MonthlyStatistics.objects.prefetch_related("account", "org_unit")
+            .values()
+            .filter(account=account)
+            .filter(org_unit_id__in=[758, 622, 43])
+            # .exclude(period="202510")
+        )
+        journey_by_org_units = groupby(list(monthlyStatistics), key=itemgetter("org_unit_id"))
+        dhis2_aggregated_data = []
+        row = {}
```
tdethier: I don't know if it's done on purpose, but since you define `row` outside your loop on line `960`, you are using the same dict and writing stuff over and over in it instead of using a new dict for each journey. It works, but it can have side effects if you forget to overwrite something

plugins/wfp/common.py
```diff
@@ -812,3 +935,105 @@ def journey_with_visit_and_steps_per_visit(self, account, programme):
...
+                journeys_by_program_type = groupby(list(journey_period), key=itemgetter("programme_type"))
+                dataValues = self.map_dhis2_data(journeys_by_program_type, data)
+
+                dataSet = {
+                    "dataSet": row["dataSet"],
+                    "period": row["period"],
+                    "orgUnit": row["orgUnit"],
+                    "orgUnitId": row["org_unit_id"],
+                    "dataValues": dataValues,
+                }
```
tdethier: And if you define a new `row` on each iteration of your loop, you no longer need to create another dict here

plugins/wfp/common.py
```diff
@@ -812,3 +935,105 @@ def journey_with_visit_and_steps_per_visit(self, account, programme):
...
+
+                for category in categories:
+                    dataElement_by_category = dataElement.get(category)
+
+                    for program, journey_program in journey_by_gender_and_nutrition_program:
+                        dataElement_by_sub_category = dataElement_by_category
+                        if program == "TSFP":
+                            dataElement_by_sub_category = dataElement.get("tsfp_reporting")
+                        elif program == "OTP":
+                            dataElement_by_sub_category = dataElement.get("otp_reporting")
```
tdethier: Nitpick: same comment about dict mapping

plugins/wfp/common.py
```diff
@@ -812,3 +935,105 @@ def journey_with_visit_and_steps_per_visit(self, account, programme):
...
+                        if dataElement_by_sub_category is not None:
+                            if dataElement_by_category is not None:
+                                dataElement_by_main_category = dataElement_by_category[main_category]
+                                for sub_category in sub_categories:
+                                    dataValue = dataElement_by_main_category.get(sub_category)
+                                    if sub_category == exit_type:
+                                        rows[sub_category] = rows["total_with_exit_type"]
+                                    if sub_category == admission_type:
+                                        rows[sub_category] = rows["total_beneficiary"]
```
tdethier: Nitpick: same comment about dict mapping

plugins/wfp/models.py
```diff
@@ -128,3 +138,30 @@ class MonthlyStatistics(models.Model):
...
     given_ration_cbt = models.CharField(max_length=255, choices=RATION_SIZE, null=True, blank=True)
+    oedema = models.FloatField(null=True)
+    muac_under_11_5 = models.FloatField(null=True)  # MUAC < 11.5cm
+    muac_11_5_12_4 = models.FloatField(null=True)  # MUAC between 11.5 and 12.4 cm
+    muac_above_12_5 = models.FloatField(null=True)  # MUAC > 12.5 cm
+    muac_under_23 = models.FloatField(null=True)  # MUAC < 23 cm for PBWG
+    muac_above_23 = models.FloatField(null=True)  # MUAC > 23 cm for PBWG
+    whz_score_2 = models.CharField(max_length=10, null=True)  # WHZ greater than -2 (green)
+    whz_score_3 = models.CharField(max_length=10, null=True)  # WHZ less than -3 (red)
+    whz_score_3_2 = models.CharField(max_length=10, null=True)  # WHZ between -2 and -3 (yellow)
```
tdethier, line 149: Do you need to store all of this because it's a lot of computations and you need to access it quickly?

plugins/wfp/models.py
```diff
@@ -128,3 +138,30 @@ class MonthlyStatistics(models.Model):
...
+    account = models.ForeignKey(Account, on_delete=models.CASCADE, null=True, blank=True)
+    org_unit_id = models.IntegerField(null=True, blank=True, db_index=True)
+    org_unit_dhis2_id = models.TextField(null=True)
+    data_set_id = models.TextField(null=True)
+    programme_type = models.CharField(max_length=255, choices=PROGRAMME_TYPE, null=True, blank=True)
+    status = models.TextField(null=True)
+    period = models.CharField(max_length=8, null=True, blank=True)
+    month = models.CharField(max_length=8, null=True, blank=True)
+    year = models.CharField(max_length=6, null=True, blank=True)
+    response = models.JSONField()
```
tdethier, line 165: Do you need to store the response?

plugins/wfp/management/commands/south_sudan/Pbwg.py
```diff
@@ -155,6 +155,9 @@ def group_visit_by_entity(self, entities):
 
                     current_record["end_date"] = visit_date.strftime("%Y-%m-%d")
                     current_record["duration"] = duration
+                    # print("VISIT ...:", visit)
+                    current_record["muac_size"] = visit.get("muac", visit.get("previous_muac"))
+                    # current_record["muac_size"] = visit.get("muac")
```
tdethier:
```suggestion
                    current_record["muac_size"] = visit.get("muac", visit.get("previous_muac"))
```

plugins/wfp/management/commands/south_sudan/Dhis2.py
```diff
@@ -0,0 +1,60 @@
...
+
+class Dhis2:
+    def sync_data(self, type):
+        entity_type = ETL([type])
+        account = entity_type.account_related_to_entity_type()
+        monthly_data = entity_type.aggregating_data_to_push_to_dhis2(account)
+
+        DHIS2_URL = os.environ.get("DHIS2_URL")
+        DHIS2_USER = os.environ.get("DHIS2_USER")
+        DHIS2_PASSWORD = os.environ.get("DHIS2_PASSWORD")
```
tdethier: We already have a class `ExternalCredentials` in IASO, is there a reason why you used environment variables instead?

plugins/wfp/aggregate_journeys.py
```diff
@@ -3,53 +3,108 @@
...
+
+                row["whz_score_2"] = row["whz_score_2"] + sum(
+                    visit.get("whz_score_2", 0) for visit in all_visits_by_criteria
+                )
+                row["whz_score_3"] = row["whz_score_3"] + sum(
+                    visit.get("whz_score_3", 0) for visit in all_visits_by_criteria
+                )
+                row["whz_score_3_2"] = row["whz_score_3_2"] + sum(
+                    visit.get("whz_score_3_2", 0) for visit in all_visits_by_criteria
+                )
```
tdethier: Otherwise you could go through `all_visits_by_criteria` once and sum each type while iterating

plugins/wfp/aggregate_journeys.py
```diff
@@ -3,53 +3,108 @@
...
+                row["oedema"] = row["oedema"] + sum(visit.get("oedema", 0) for visit in all_visits_by_criteria)
```
tdethier: This could be done in the single loop as well

plugins/wfp/aggregate_journeys.py
```diff
@@ -3,53 +3,108 @@
...
+                    )
 
-                        if step.get("visit__journey__exit_type") is not None:
-                            row["exit_type"] = step.get("visit__journey__exit_type")
+                    row["muac_under_23"] = row["muac_under_23"] + sum(
+                        visit.get("muac_under_23", 0) for visit in aggregated_journey_by_exit_type
+                    )
+                    row["muac_above_23"] = row["muac_above_23"] + sum(
+                        visit.get("muac_above_23", 0) for visit in aggregated_journey_by_exit_type
+                    )
```
tdethier: Same comment about looping only once though `aggregated_journey_by_exit_type`

plugins/wfp/aggregate_journeys.py
```diff
@@ -3,53 +3,108 @@
...
+                    row["given_sachet_rusf"] = row["given_sachet_rusf"] + sum(
+                        visit.get("quantity_given", 0)
+                        for visit in aggregated_journey_by_exit_type
+                        if visit.get("assistance_type") == "rusf"
+                    )
+                    row["given_quantity_csb"] = row["given_quantity_csb"] + sum(
+                        visit.get("csb_quantity", 0)
+                        for visit in aggregated_journey_by_exit_type
+                        if visit.get("assistance_type") in ["csb", "csb1", "csb2"]
+                    )
```
tdethier: This could be done in the single loop as well. Otherwise you're looping `aggregated_journey_by_exist_type` 8 times

tdethier (commented):
I didn't run the code because I don't have a WFP setup - I don't have any data that would be exported.

I made a few suggestions to improve the code here and there. Honestly I lack a lot of business knowledge to understand all of this, but I guess this will work.

@butofleury after making a few changes (especially regarding migrations), would this be testable on staging before merging?

plugins/wfp/models.py
butofleury: Yes, I need to store the response witch return the result of dhis2 post action. So, in case the action fails, this will help to know the reason.

plugins/wfp/models.py
butofleury: Yes, I need to store all those information because it should be pushed to dhis2.

Here is an example of dhis2 dataSet to receive the calculated data from all those fields:

image

tdethier: But do you need to keep track of what has been pushed? Can't you just compute the values and push them without storing them?

I don't mind it, I'm just trying to understand because it's a bit weird to me 😅

tdethier (approved):
Thanks for making the changes, it looks better 👍

Is there a way to test this code on staging before merging this branch? I'd like to see it running with some values to make sure that the expected values are received in DHIS2

plugins/wfp/common.py
```diff
@@ -19,6 +20,8 @@
 
 logger = logging.getLogger(__name__)
 
+DATASET_ID = os.environ.get("DATASET_ID")
```
tdethier: It would probably better to fetch the environment variable in `settings.py` and call settings here

PR #2526, opened 2025-10-30
POLIO-1958: add exclusion list for prep worksheets

quang-le:
Client adds worksheets to the preparedness google sheet, and it breaks everyting

Related JIRA tickets :  POLIO-1958

- [ ] Are there enough tests?

## Doc

[Tell us where the doc can be found (docs folder, wiki, in the code...).](https://github.com/BLSQ/polio-doc/pull/3/files)

## Changes

- Hard coded an exclusion on common sheet names
- Added a config for additional exclusions
- Fixed small bug with round number in the preparedness parser

## How to test
- Generate a preparedness sheet (needs a round with a scope and a future start date)
- Add empty worksheets to the generated sheets
- rename at least 1.
- Add the custom worksheet name to the `preparedness_sheetnames_blacklist`config (add it if necessary: it just a list of strings). In the new sheet, add (in any cell) the string: `"Summary of Regional Level Preparedness"`(this will circumvent the exclusion mechanism already in place)
- From the preparedness tab, launch a refresh: the additional sheets should not trigger errors

## Note

This should go in the next release one way or another since it would solve refresh problems with an actual upcoming campaign

### DIFF
--- a/plugins/polio/preparedness/parser.py
+++ b/plugins/polio/preparedness/parser.py
@@ -4,6 +4,7 @@ from typing import Any, Dict, Optional
 
 from gspread.utils import absolute_range_name, rowcol_to_a1
 
+from iaso.models.json_config import Config
 from plugins.polio.preparedness.calculator import get_preparedness_score
 from plugins.polio.preparedness.client import get_client
 from plugins.polio.preparedness.exceptions import InvalidFormatError
@@ -13,6 +14,8 @@ from plugins.polio.preparedness.spread_cache import CachedSheet, CachedSpread
 logger = getLogger(__name__)
 
 TEMPLATE_VERSION = "v3.5"
+WORKSHEET_EXCLUSION_LIST = ["synth_scores"]
+WORKSHEET_EXCLUSION_SLUG = "preparedness_sheetnames_blacklist"
 
 
 def parse_value(value: str):
@@ -133,9 +136,14 @@ class RoundNumber(str, Enum):
     unknown = "Unknown"
 
 
-def get_national_level_preparedness(spread: CachedSpread):
+def get_national_level_preparedness(spread: CachedSpread, exclusions):
     for worksheet in spread.worksheets():
-        if worksheet.is_hidden:
+        if (
+            worksheet.is_hidden
+            or worksheet.title.lower() in WORKSHEET_EXCLUSION_LIST
+            or worksheet.title.lower().startswith("sheet")
+            or (exclusions and worksheet.title in exclusions)
+        ):
             continue
         cell = worksheet.find_one_of(
             "Summary of National Level Preparedness",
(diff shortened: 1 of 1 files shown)


plugins/polio/preparedness/parser.py
```diff
@@ -351,9 +369,14 @@ def get_regional_level_preparedness_v2(spread: CachedSpread):
...
-def get_national_level_preparedness_v2(spread: CachedSpread):
+def get_national_level_preparedness_v2(spread: CachedSpread, exclusions):
     for worksheet in spread.worksheets():
-        if worksheet.is_hidden:
+        if (
+            worksheet.is_hidden
+            or worksheet.title.lower() in WORKSHEET_EXCLUSION_LIST
+            or worksheet.title.lower().startswith("sheet")
+            or (exclusions and worksheet.title in exclusions)
+        ):
```
tdethier: Since this block is repeated a lot, maybe we should have a function `is_worksheet_excluded`?

plugins/polio/preparedness/parser.py
```diff
@@ -375,31 +398,40 @@ def get_national_level_preparedness_v2(spread: CachedSpread):
...
+        **get_regional_level_preparedness_v2(spread, exclusions),
         "format": TEMPLATE_VERSION,
     }
     preparedness_data["totals"] = get_preparedness_score(preparedness_data)
     return preparedness_data
 
 
-def get_preparedness(spread: CachedSpread):
+def get_preparedness(spread: CachedSpread, exclusions=None):
+    exclusions = Config.objects.filter(slug=WORKSHEET_EXCLUSION_SLUG).first()
```
tdethier: Shouldn't it be this?
```suggestion
    exclusions = Config.objects.filter(slug=WORKSHEET_EXCLUSION_SLUG).first().content
```

plugins/polio/preparedness/parser.py
```diff
@@ -269,7 +282,12 @@ def get_regional_level_preparedness_v2(spread: CachedSpread):
 
     sheet: CachedSheet
     for sheet in spread.worksheets():
-        if sheet.is_hidden:
+        if (
+            sheet.is_hidden
+            or sheet.title.lower() in WORKSHEET_EXCLUSION_LIST
+            or sheet.title.lower().startswith("sheet")
+            or (exclusions and sheet.title in exclusions)
```
tdethier: There might be issues with not using `lower()` here as well, but it depends on what users input in the `Config`.

Maybe you can first `lower()` all exclusions when you fetch it from the `Config` and do it as well on this line

plugins/polio/preparedness/parser.py
tdethier: Image

tdethier (approved):
It works fine after applying the suggested change 👍

PR #2532, opened 2025-11-05
IA-3772: Skip existing org units when importing data (hotfix)

Phil-V:
refs: IA-3772

**Hotfix** for exceptions raised by PR #2522

When attempting to create an OrgUnit that already exists, an AssertionError is thrown:
`AssertionError: `create()` did not return an object instance.`

This PR fixes the error and replicates the original behavior of ignoring the existing OrgUnits.

- [x] Are there enough tests?

## Doc

- 

## Changes

- Added tests that replicate the issue.
- Fixed the issue by checking for existing OrgUnits and ignoring them. 

## How to test

1. Testing the regular upload:
- Create an org unit on mobile (note: you need to configure "sub org unit types to create" on the parent org unit type in the dashboard to allow the creation of org units on mobile).
- Upload the new org unit
- Find the payload in the api imports in the django admin.
- Query the api manually with the same payload, it should be something like this:
```
curl --request POST \
  --url 'http://localhost:8081/api/mobile/orgunits/?app_id=hwbewoz&app_version=2600' \
  --header 'Authorization: Bearer <token>' \
  --header 'content-type: application/json' \
  --data '[
  {
    "id": "2740df90-5307-4bd9-aa5c-be9e6b144ffc",
    "name": "Test Areea Duplicated",
    "validation_status": "",
    "org_unit_type_id": "5",
    "parent_id": "4",
    "time": 0,
    "created_at": 1.762433961328E9,
    "updated_at": 1.762433961328E9
  }
]
```
- The backend should return an empty response as the org unit is already saved and will be ignored.
- You may try further queries with a mix of old and new UUIDs. 

2. Testing the bulk upload.
- You need 2 project feature flags:
   -  Entities > Mobile: Show entities screen
   -  Data Validation > Mobile: Change requests
- Create and upload an org unit with the mobile app
- In the django admin, find the task associated with the upload and check that it's succesfull
- Manually set its status to error
- Restart the task in the dashboard
- Check that it completes again without error. 

You may want to test the features of the original PR #2522 as well 

## Print screen / video

/

## Notes

Didn't want to change anything else, however, in removing the `create()` method on the serializer, I realized that both the code introduced by my PR and the original code set attributes on the OrgUnits that don't exist on the model and were silently ignored. I had to remove them when getting rid of the custom create logic as `source` and `accuracy` were passed to the OrgUnit's constructor and trigger TypeError. 

```
fix: empty instance pop up

Refs: IA-3665
```

### DIFF
--- a/iaso/api/org_units.py
+++ b/iaso/api/org_units.py
@@ -1018,20 +1018,21 @@ def import_org_units(org_units: List[Dict], user, app_id):
     """Import a list of org units."""
     new_org_units = []
     project = Project.objects.get_for_user_and_app_id(user, app_id)
-    if project.account.default_version.data_source.read_only:
+    version = project.account.default_version
+    if version.data_source.read_only:
         raise ValidationError("Creation of org unit not authorized on default data source")
 
     # common values to all new org units
     extra_save_kwargs = {
         "custom": True,
         "validation_status": OrgUnit.VALIDATION_NEW,
-        "version": project.account.default_version,
+        "version": version,
         "creator": user if (user and not user.is_anonymous) else None,
     }
 
     for org_unit_data in org_units:
         uuid = org_unit_data.get("id")
-        if OrgUnit.objects.filter(uuid=uuid).exists():
+        if OrgUnit.objects.filter(uuid=uuid, version=version).exists():
             continue  # skip existing org units
 
         serializer = OrgUnitImportSerializer(data=org_unit_data)
--- a/iaso/api/org_units.py
+++ b/iaso/api/org_units.py
@@ -1025,18 +1025,18 @@ def import_org_units(org_units: List[Dict], user, app_id):
     extra_save_kwargs = {
         "custom": True,
         "validation_status": OrgUnit.VALIDATION_NEW,
-        "source": "API",
         "version": project.account.default_version,
         "creator": user if (user and not user.is_anonymous) else None,
     }
 
     for org_unit_data in org_units:
+        uuid = org_unit_data.get("id")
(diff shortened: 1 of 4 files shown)


iaso/api/org_units.py
```diff
@@ -1025,18 +1025,18 @@ def import_org_units(org_units: List[Dict], user, app_id):
...
         "custom": True,
         "validation_status": OrgUnit.VALIDATION_NEW,
-        "source": "API",
         "version": project.account.default_version,
         "creator": user if (user and not user.is_anonymous) else None,
     }
 
     for org_unit_data in org_units:
+        uuid = org_unit_data.get("id")
+        if OrgUnit.objects.filter(uuid=uuid).exists():
```
tdethier: I think we should add a check on the version as well, because I'm pretty sure we have had issues with uuids not being unique

Phil-V: I was thinking the same but was hesitant to introduce a behavior change in the hotfix. As you pointed out in our conversation, the original logic was only checking the uuid as well: https://github.com/BLSQ/iaso/pull/2522/files#diff-e73cbcd5983aaae3e9f3475ca349f53a7047179d5d21be161ae377a41ae0448eL364

tdethier (approved):
LGTM, thanks for diving a bit deeper on this 👍

iaso/api/org_units.py
Phil-V: After discussing with Martin, we came to the conclusion that the check on the version is indeed the intended behavior and added this check to the PR.

PR #2535, opened 2025-11-06
IA-4578: IASO Colors  🌈 

beygorghor:
Colors used for chips, map shapes, projects, teams ... are hardcoded a bit everywhere in IASO. 
We should save it in the back-end as constant, serve it using the api and replace all those hardcoded values in the front-end

Related JIRA tickets : IA-4578

- [ ] Are there enough tests?

## Doc

-

## Changes

- colors endpoint end test
- refactor ColorPicker
- change all occurence of hardcoded values
- custom migration to add a random color from the list

## How to test

- migrate and make sure your teams a have a color in the admin
- you can get colors on `api/colors` by default as a gradiant palette, using disperse param to have dispersed colors
- test on api colors
- make sure teams colors are used in planning page
- make a search of org units and play with colors in hte tabs
- play with projects colors too.

## Print screen / video

https://github.com/user-attachments/assets/cbf592cd-5575-4bd8-a6bd-ae426aef49ee

## Notes

⚠️ Team color edition will be part of another PR

```
fix: empty instance pop up

Refs: IA-3665
```


iaso/api/colors.py
```diff
@@ -0,0 +1,23 @@
...
+
+@api_view(["GET"])
+@permission_classes([IsAuthenticated])
+def colors_list(request):
+    """Return the list of available colors for the application
+
+    Query parameters:
+    - dispersed: if 'true', returns colors in a dispersed order where similar colors are not adjacent
+    """
+    dispersed = request.query_params.get("dispersed", "").lower() == "true"
```
tdethier: Nitpick: it could be a serializer

iaso/api/teams/serializers.py
```diff
@@ -55,6 +55,7 @@ class Meta:
             "created_at",
             "deleted_at",
             "type",
+            "color",
```
tdethier, line 58: we should add a `validate_color` method to make sure that the value received is actually a valid hex RGB

iaso/tests/api/test_colors.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+class ColorsApiTestCase(APITestCase):
+    @classmethod
+    def setUpTestData(cls):
+        account = m.Account(name="test account")
+        account.save()
+
+        cls.user = user = m.User.objects.create(username="test_user")
+        p = m.Profile(user=user, account=account)
```
tdethier: Nitpick: there's a `create_user_with_profile` utility method

iaso/tests/api/test_colors.py
```diff
@@ -0,0 +1,125 @@
...
+    def test_colors_dispersed_starts_with_blue(self):
+        """Test that dispersed order starts with Light Blue 400, not red"""
+        self.client.force_authenticate(self.user)
+        response = self.client.get("/api/colors/?dispersed=true", format="json")
+        data = response.json()
+
+        first_color = data[0]
+        # Light Blue 400 is at index 9 in COLOR_CHOICES
+        self.assertEqual(first_color["value"], COLOR_CHOICES[9][0])
+        self.assertIn("Light Blue", first_color["label"])
```
tdethier: Nitpick: I would get rid of these tests, I don't think that they bring anything useful that we don't already get from the other tests

```suggestion
```

tdethier (changes requested):
Overall the feature works fine, thank you 👍 

I'm requesting changes because we need the  `TeamSerializer` to validate the color parameter received (if it's not hex, not a valid color, doesn't start with `#`... it should be rejected).

Other miscellaneous comments:
- we should get a picker on the team creation page to select the color
- same for edit (I guess it's the same modal)
- when I look at API calls on the planning page, I noticed that the new endpoint is called twice, but the `dispersed` version is not used

Image

tdethier:
General comment: if we do end up with colors in multiple places, it would maybe worth preparing a custom `ColorField` at the model and serializer levels so that we can simply reuse it everywhere it's required.

We should probably discuss this during a SUM and if we go for that, prepare a separate ticket for that

hat/assets/js/apps/Iaso/components/forms/ColorPicker.tsx
```diff
@@ -13,55 +19,72 @@ const MESSAGES = defineMessages({
...
     };
-    const open = Boolean(anchorEl);
 
     const handleChangeColor = newColor => {
-        handleClick(null);
+        setOpen(false);
         onChangeColor(newColor.hex);
     };
+
+    const handleClickAway = () => {
```
Bewi: useCallbakc?

hat/assets/js/apps/Iaso/hooks/useGetColors.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export const getColor = (
+    i: number,
+    colors?: string[],
+    usedColors?: string[],
+): string => {
+    if (!colors) return 'transparent';
+    if (colors.length === 0) return 'transparent';
+    return colors.filter(color => !usedColors?.includes(color))[
+        i % colors.length
```
Bewi: What's the expected behavior here ?
From what I understand with this module operator, if **_i_** is smaller than colors.lengths, then all good, it will use **_i_*. But colors are filtered, so let's say we have 10 colors, but we filter out 5 colors. 
`I = 6, colors.length = 10 => (6 % 10) = 6   ==> higher than filtered colors length`.
Or am I missing something

Bewi (commented):
I still to run it locally, looks good except for two small comments

iaso/api/teams/serializers.py
beygorghor: ✅

iaso/tests/api/test_colors.py
beygorghor: ✅

iaso/tests/api/test_colors.py
beygorghor: ✅

hat/assets/js/apps/Iaso/components/forms/ColorPicker.tsx
beygorghor: ✅

hat/assets/js/apps/Iaso/hooks/useGetColors.ts
beygorghor: Well spotted !  ✅

beygorghor:
> Overall the feature works fine, thank you 👍
> 
> I'm requesting changes because we need the `TeamSerializer` to validate the color parameter received (if it's not hex, not a valid color, doesn't start with `#`... it should be rejected).
> 
> Other miscellaneous comments:
> 
> * we should get a picker on the team creation page to select the color
> * same for edit (I guess it's the same modal)
> * when I look at API calls on the planning page, I noticed that the new endpoint is called twice, but the `dispersed` version is not used
> 
> Image
Those three points are fixed in another PR already

tdethier (approved):
LGTM 👍

PR #2547, opened 2025-11-12
WC2-849: Entities optimisation 

beygorghor:
Optimisation suggestions were shared by Matteo for the web app endpoints. These are touching on IASO and not only CODA's code, so better be implemented by BLSQ team.

optimize query execution, the old code was joining through instance__ so each entity was duplicated once per matching instance, so when Django was performing the distinct("id") it was doing across a massive join result paying the price of sorting/deduping hundreds of thousands of rows

Related JIRA tickets : WC2-849 

- [ ] Are there enough tests?

## Doc

-

## Changes

apply Matteo changes on Entities model

## How to test

- make sure entities are working like usual

## Print screen / video

## Notes

-

```
fix: empty instance pop up

Refs: IA-3665
```

### DIFF
--- a/iaso/models/entity.py
+++ b/iaso/models/entity.py
@@ -23,7 +23,7 @@ from django.contrib.auth.models import AnonymousUser, User
 from django.contrib.postgres.fields import ArrayField
 from django.core.exceptions import ValidationError
 from django.db import models
-from django.db.models import Exists, OuterRef, Prefetch, Q
+from django.db.models import Exists, OuterRef, Prefetch, Q, Subquery
 
 from iaso.models import Account, Instance, OrgUnit, Project
 from iaso.models.deduplication import EntityDuplicate, ValidationStatus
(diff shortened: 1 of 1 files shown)


iaso/models/entity.py
```diff
@@ -103,13 +103,24 @@ class ProjectNotFoundError(ValidationError):
 
 
 class EntityQuerySet(models.QuerySet):
-    def filter_for_mobile_entity(self, limit_date=None, json_content=None):
+    def _filter_entities_with_instances(self, *, limit_date=None, org_units_qs=None):
+        instances = Instance.objects.all()
```
tdethier, line 107: I'm a bit worried about tenancy here, if we don't provide any org unit queryset, it means that instances are not filtered out and this is going to return all entities

tdethier: It works fine because the queryset is already filtered out with some other parameters first, but we have to be careful because if we call this without additional filtering, all entities could be returned, even the ones from other accounts

tdethier (approved):
LGTM, it will work fine 👍 

But there's a small scenario in which it could cause some issue (see comments)

iaso/models/entity.py
madewulf: I don't think this is a problem because you have this as the result of the function:

self.filter(id__in=Subquery(instances.values("entity_id").distinct()))

you're not returning any of the instances, you're filtering the entities queryset (that already has been filtered out for the current user)  on the possible set of instances and that entity queryset has already been filtered.

madewulf: I'm just rephrasing what you said in your second comment, but I don't see how we would NOT have the entity queryset prefiltered at that point and I fear that adding additional filters will hit the performances

madewulf (approved):

PR #2550, opened 2025-11-14
IA-4575:  Remove the need to click the "search" button for Forms

beygorghor:
This creates friction and looks like a bug

People anyways click on Search by default as nothing shows and then it loads everything anyways
Related JIRA tickets : IA-4575

- [ ] Are there enough tests?

## Doc

- 

## Changes

- enhance curent column selector with an apply button
- dedicated hook to deep link visible columns

## How to test

- Go to forms
- forms should load
- you should be able to change visible columns and use apply button
- a new call is done to fetch new fields
- visible columns should be ... visible
-  refresh the page, columns should remain the same

## Print screen / video

https://github.com/user-attachments/assets/c73d0f48-13cf-4a44-bee3-b58fa93d3acb

## Notes
-

```
fix: empty instance pop up

Refs: IA-3665
```


hat/assets/js/apps/Iaso/domains/forms/config/index.tsx
```diff
@@ -81,10 +95,10 @@ export const useFormVersionsTableColumns = (formId, periodType) => {
                                         tooltipMessage={MESSAGES.edit}
                                     />
                                 )}
-                                // onConfirmed={() => setForceRefresh(true)}
+                                onConfirmed={() => {}}
```
quang-le: From what I see in the code base, there are 2 uses of this component, and none of them actually pass or use `onConfirmed`. Maybe we should remove it from the props instead

hat/assets/js/apps/Iaso/domains/forms/hooks/useColumnsSelectDrawer.tsx
```diff
@@ -0,0 +1,71 @@
...
+                label: column.Header,
+                active: column.id && visibleColumnsKeys.includes(column.id),
+                disabled: false,
+            })) as Option[];
+    }, [visibleColumnsKeys, columns]);
+    const setOptions = useCallback((options: Option[]) => {
+        setVisibleColumnsKeys(
+            options.filter(option => option.active).map(option => option.key),
+        );
+    }, []);
```
quang-le: no need to wrap in `useCallback` if the deps array is empty

hat/assets/js/apps/Iaso/domains/forms/hooks/useColumnsSelectDrawer.tsx
```diff
@@ -0,0 +1,71 @@
...
+    const fields = useMemo(() => {
+        return [
+            ...(params.fields?.split(',') ?? DEFAULT_VISIBLE_COLUMNS),
+            'actions',
+        ];
+    }, [params.fields]);
+    const applyDisabled =
```
quang-le: nitpick: I like booleans to be named `is<thing>` or `has<thing>`, it makes them easy to spot

hat/assets/js/apps/Iaso/domains/forms/index.tsx
```diff
@@ -1,79 +1,42 @@
...
     const { formatMessage } = useSafeIntl();
 
-    const isSearchActive = params?.isSearchActive === 'true';
     // This doesn't duplicate API calls as long as
     // the querystring generated by useGetForms is the same as the one in the table
     const { data: forms, isLoading: isLoadingForms } = useGetForms(
         params,
         undefined,
-        isSearchActive,
+        true,
```
quang-le, line 27: This makes me think it would be more legible if `useGetForms` took a dictionary as param. 
`useGetForms(params, undefined, true)` is impossible to understand without having to check the hook's code

quang-le (changes requested):
I'd say remove the useless `useCallback` and change `useGetForms` so it takes an object and we're good to go.

Nice clean up job!

hat/assets/js/apps/Iaso/domains/forms/config/index.tsx
beygorghor: ✅

hat/assets/js/apps/Iaso/domains/forms/hooks/useColumnsSelectDrawer.tsx
beygorghor: ✅

hat/assets/js/apps/Iaso/domains/forms/hooks/useColumnsSelectDrawer.tsx
beygorghor: ✅

hat/assets/js/apps/Iaso/domains/forms/index.tsx
beygorghor: On the next PR

quang-le (approved):
LGTM!

PR #2555, opened 2025-11-18
IA-4608: Mobile OrgUnit endpoint returns a `parent_id` for parents not in project

bmonjoie:
The mobile application expects the parent_id to point to an OrgUnit that is returned by the endpoint.
When a parent is in another project, the parent_id is invalid.

Related JIRA tickets : IA-4608

- [X] Are there enough tests?

## Changes

I have improved the number of queries made to the backend + check on the parent's type to see if it is returned or not.

## How to test

Assign a parent to an OrgUnit type that is not in the project you are requesting and see if its children have `parent_id` to `null` in the response.

Endpoint is: `/api/mobile/orgunits/?app_id=...`

### DIFF
--- a/iaso/api/mobile/org_units.py
+++ b/iaso/api/mobile/org_units.py
@@ -78,9 +78,12 @@ class ReferenceInstancesSerializer(serializers.ModelSerializer):
 
 
 class MobileOrgUnitSerializer(serializers.ModelSerializer):
+    app_id = None
+
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         # Do not include the geo_json if not requested
+        self.app_id = kwargs["context"].get("app_id")
         if not kwargs["context"].get("include_geo_json"):
             self.fields.pop("geo_json")
 
@@ -119,11 +122,13 @@ class MobileOrgUnitSerializer(serializers.ModelSerializer):
     def get_org_unit_type_name(org_unit: OrgUnit):
         return org_unit.org_unit_type.name if org_unit.org_unit_type else None
 
-    @staticmethod
-    def get_parent_id(org_unit: OrgUnit):
+    def get_parent_id(self, org_unit: OrgUnit):
+        parent = org_unit.parent
         return (
             org_unit.parent_id
-            if org_unit.parent is None or org_unit.parent.validation_status == OrgUnit.VALIDATION_VALID
+            if parent is None
+            or parent.validation_status == OrgUnit.VALIDATION_VALID
+            and (self.app_id is None or any(p.app_id == self.app_id for p in parent.org_unit_type.projects.all()))
             else None
         )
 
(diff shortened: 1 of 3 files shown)


iaso/tests/api/test_mobile_orgunits.py
```diff
@@ -153,8 +162,112 @@ def test_org_unit_with_shapes_limited(self):
...
+            # 15. INSERT INTO "django_cache_table"
+            # 16. RELEASE SAVEPOINT "s128932915669888_x20"
+            response = self.client.get(BASE_URL, data={APP_ID: BASE_APP_ID})
+        self.assertJSONResponse(response, 200)
+        self.assertEqual([self.raditz.id, self.goku.id], response.json()["roots"])
+        self.assertNotIn("count", response.json())
+        self.assertNotIn("next", response.json())
+        self.assertNotIn("previous", response.json())
+        self.assertEqual(len(response.json()["orgUnits"]), 4)
+        self.assertEqual(response.json()["orgUnits"][0]["name"], "Bardock")
```
quang-le, line 189: I would rather name org units after places and use boring names like `parent_level_1` or something

iaso/api/mobile/org_units.py
```diff
@@ -214,41 +219,46 @@ def pagination_class(self):
 
     def get_queryset(self):
         user = self.request.user
-        app_id = AppIdSerializer(data=self.request.query_params).get_app_id(raise_exception=True)
+        app_id = self.get_app_id()
 
         limit_download_to_roots = False
 
+        project = Project.objects.get_for_user_and_app_id(user, app_id)
```
quang-le: Shouldn't you wrap this in a `try...except`? If not `Project` is found this will `raise`

iaso/api/mobile/org_units.py
```diff
@@ -119,11 +122,13 @@ class Meta:
...
-    @staticmethod
-    def get_parent_id(org_unit: OrgUnit):
+    def get_parent_id(self, org_unit: OrgUnit):
+        parent = org_unit.parent
         return (
             org_unit.parent_id
-            if org_unit.parent is None or org_unit.parent.validation_status == OrgUnit.VALIDATION_VALID
+            if parent is None
+            or parent.validation_status == OrgUnit.VALIDATION_VALID
+            and (self.app_id is None or any(p.app_id == self.app_id for p in parent.org_unit_type.projects.all()))
```
quang-le, line 131: `any(p.app_id == self.app_id for p in parent.org_unit_type.projects.all()))`

=>  `parent.org_unit_type.projects.all().filter(app_id=self.app_id).exists()`

It's usually better to use queryset methods when possible

quang-le (changes requested):
Seems to work fine. 
Please have a look at the try/except

iaso/tests/api/test_mobile_orgunits.py
bmonjoie: I'm not going to change all the tests for that :sweat_smile:

iaso/api/mobile/org_units.py
bmonjoie: The exception was already raised by `filter_for_user_and_app_id`

iaso/api/mobile/org_units.py
bmonjoie: If I change it to that, it makes more query to the database. I'm afraid of having a N+1 with this.

quang-le: nvm then

iaso/api/mobile/org_units.py
quang-le: Which is called where?

iaso/api/mobile/org_units.py
tdethier: Yeah unfortunately, this is called for every OrgUnit that goes through the serializer, so it's one of these rare cases where it's better to not use the queryset methods (I guess the best way would be to make all these checks earlier in an annotation, but we can keep this)

iaso/tests/api/test_mobile_orgunits.py
```diff
@@ -153,8 +162,112 @@ def test_org_unit_with_shapes_limited(self):
...
+        self.assertEqual(response.json()["orgUnits"][0]["groups"][0], self.group_2.id)
+        self.assertEqual(response.json()["orgUnits"][1]["name"], "Raditz")
+        self.assertEqual(response.json()["orgUnits"][1]["parent_id"], self.bardock.id)
+        self.assertEqual(0, len(response.json()["orgUnits"][1]["groups"]))
+        self.assertEqual(response.json()["orgUnits"][2]["name"], "Son Gohan")
+        self.assertEqual(response.json()["orgUnits"][2]["parent_id"], None)
+        self.assertEqual(0, len(response.json()["orgUnits"][2]["groups"]))
+        self.assertEqual(response.json()["orgUnits"][3]["name"], "Son Goten")
+        self.assertEqual(response.json()["orgUnits"][3]["parent_id"], None)
+        self.assertEqual(0, len(response.json()["orgUnits"][3]["groups"]))
```
tdethier, line 201: I would have created extra variables for the json result, for the org units in the json,... but this works too 😅

tdethier (commented):
LGTM, but I'll let @madewulf approve

iaso/api/mobile/org_units.py
tdethier: The initial version called `filter_for_user_and_app_id` and inside of that, you had the same `Project.objects.get_for_user_and_app_id(user, app_id)` call

quang-le: If you mean `OrgUnit.objects.filter_for_user_and_app_id`, it doesn indeed call Project's `get_for_user_and_app_id`, but it wraps it in a try/except, so my initial comment still stands

tdethier: You're right @quang-le, and yet when you manually test the endpoint with an unknown `app_id`, it does not crash. 

It's because there's another issue: this line is in `get_queryset` and the list endpoint does actually not call `get_queryset` 😅 If you call `/api/mobile/orgunits/boundingbox/ ` then you get the exception

quang-le (approved):

madewulf (approved):

PR #2557, opened 2025-11-19
[POLIO-2027] Add Performance Dashboard Backend API

Crebert08:

Related JIRA tickets : [POLIO-2027](https://bluesquare.atlassian.net/browse/POLIO-2027)

- [ ]  Is my code clear enough and well documented?
- [x]  My migrations file are included
- [x]  Are there enough tests?

## Doc

## Changes

Added a performance dashboard model and its API that shows these features: Status, Date, Country and Antigen. The API also filters by country and country block

## How to test

In the terminal, start the IASO app in docker, and type this command: `sudo docker compose run iaso manage test plugins.polio.tests.api.performance_dashboard`

## Notes

```
fix: empty instance pop up

Refs: POLIO-2027
```

[POLIO-2027]: https://bluesquare.atlassian.net/browse/POLIO-2027?


plugins/polio/api/perfomance_dashboard/serializers.py
```diff
@@ -0,0 +1,153 @@
...
+        model = OrgUnit
+        fields = ["id", "name"]
+        ref_name = "OrgUnitNestedSerializerForNationalLogisticsPlan"
+
+
+class PerformanceDashboardListSerializer(serializers.ModelSerializer):
+    created_by = UserNestedSerializer(read_only=True)
+    updated_by = UserNestedSerializer(read_only=True)
+    created_at = TimestampField(read_only=True)
+    updated_at = TimestampField(read_only=True)
```
quang-le: Please use `DateTimeField` instead

plugins/polio/api/perfomance_dashboard/views.py
```diff
@@ -0,0 +1,112 @@
...
+    #     print(f"Is user authenticated? {request.user.is_authenticated}")  # DEBUG
+    #     if hasattr(request.user, 'iaso_profile'):
+    #         print(f"User's account from view: {request.user.iaso_profile.account}")  # DEBUG
+    #     else:
+    #         print("User has no iaso_profile")  # DEBUG
+    #
+    #     # Call the parent create method which handles serializer instantiation
+    #     # and should pass the context (including request) automatically,
+    #     # or now explicitly via get_serializer_context
+    #     return super().create(request, *args, **kwargs)
```
quang-le: 🗑️

plugins/polio/api/perfomance_dashboard/views_web_ui.py
quang-le, line 1: what's the use of this file?

plugins/polio/api/perfomance_dashboard/serializers.py
```diff
@@ -0,0 +1,153 @@
...
+    #     print(f"--- Serializer is_valid called ---")
+    #     print(f"Data before validation: {self.initial_data}")
+    #     print(f"Context in is_valid: {self.context}")
+    #     # Call the parent is_valid
+    #     valid = super().is_valid(raise_exception=raise_exception)
+    #     print(f"Validation result: {valid}")
+    #     if not valid:
+    #         print(f"Validation errors: {self.errors}")
+    #     print(f"--- End Serializer is_valid ---")
+    #     return valid
```
quang-le: 🗑️

plugins/polio/api/perfomance_dashboard/common.py
quang-le, line 1: What's the use of this file?

plugins/polio/api/perfomance_dashboard/pagination.py
quang-le, line 1: We have a custom `ModelViewset` that you should use, and that includes pagination

plugins/polio/models/performance_dashboard.py
```diff
@@ -0,0 +1,81 @@
...
+        FINAL = "final", _("Final")
+
+    date = models.DateField()
+    status = models.CharField(max_length=20, choices=Status.choices, default=Status.DRAFT)
+    country = models.ForeignKey(
+        OrgUnit,
+        on_delete=models.PROTECT,
+        related_name="performance_dashboard",
+    )
+    antigen = models.CharField(max_length=20, choices=VACCINES)
```
quang-le: I would rename this field as `vaccine` to be consistent with other models

plugins/polio/tests/api/performance_dashboard/test_filters.py
```diff
@@ -0,0 +1,118 @@
...
+    #     # There are 2 dashboards between Feb 1 and Mar 31, 2023
+    #     params = {"date_from": "2023-02-01", "date_to": "2023-03-31"}
+    #     response = self.client.get(self.PERFORMANCE_DASHBOARD_API_URL, data=params)
+    #     self.assertJSONResponse(response, status.HTTP_200_OK)
+    #     response_data = response.json()
+    #     self.assertEqual(response_data["count"], 2)
+    #
+    #     result_ids = {item["id"] for item in response_data["results"]}
+    #     self.assertIn(self.dashboard_konoha_final_nopv2.id, result_ids)
+    #     self.assertIn(self.dashboard_suna_commented_bopv.id, result_ids)
```
quang-le: Why not test by antigen?

plugins/polio/api/perfomance_dashboard/serializers.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+class PerformanceDashboardListSerializer(serializers.ModelSerializer):
+    created_by = UserNestedSerializer(read_only=True)
+    updated_by = UserNestedSerializer(read_only=True)
+    created_at = TimestampField(read_only=True)
+    updated_at = TimestampField(read_only=True)
+
+    # For read operations, we want to display the country's name
+    country = OrgUnitNestedSerializer(read_only=True)
```
quang-le:
```suggestion
    country_name = CharField(source="country.name", read_only=True)
```

Since we already have a field for `country_id` it may be easier to skip the serializer

plugins/polio/api/perfomance_dashboard/serializers.py
```diff
@@ -0,0 +1,153 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class UserNestedSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = User
+        fields = ["id", "username", "first_name", "last_name"]
+        ref_name = "UserNestedSerializerForNationalLogisticsPlan"
+
```
quang-le: We do have a lot of these, you might be able to re-use one

plugins/polio/api/perfomance_dashboard/serializers.py
```diff
@@ -0,0 +1,153 @@
...
+    #     valid = super().is_valid(raise_exception=raise_exception)
+    #     print(f"Validation result: {valid}")
+    #     if not valid:
+    #         print(f"Validation errors: {self.errors}")
+    #     print(f"--- End Serializer is_valid ---")
+    #     return valid
+
+    def create(self, validated_data):
+        # Set created_by and account automatically from the request user
+        # print("--- Inside create method ---")  # DEBUG
```
quang-le: 🗑️

plugins/polio/api/perfomance_dashboard/serializers.py
```diff
@@ -0,0 +1,153 @@
...
+        request = self.context.get("request")
+        # print(f"Request object: {request}")  # DEBUG
```
quang-le: 🗑️

quang-le: Please remove all these commented prints

quang-le (changes requested):
Looks good overall. 

Please cleanup commented code/prints

plugins/polio/api/perfomance_dashboard/views_web_ui.py
Crebert08: I forgot to delete it. I used to api guidelines to create the API but Thibaullt later told me that the API would not available on the mobile side but I forgot to delete it. This will be fixed. Same as the common.py file

plugins/polio/tests/api/performance_dashboard/test_filters.py
Crebert08: The jira ticket specified that the API will only filter by country and country block

plugins/polio/api/perfomance_dashboard/serializers.py
```diff
@@ -0,0 +1,98 @@
...
+class PerformanceDashboardListSerializer(serializers.ModelSerializer):
+    created_by = UserNestedSerializer(read_only=True)
+    updated_by = UserNestedSerializer(read_only=True)
+    created_at = serializers.DateTimeField(read_only=True)
+    updated_at = serializers.DateTimeField(read_only=True)
+
+    # For read operations, we want to display the country's name
+    country_name = serializers.CharField(source="country.name", read_only=True)
+    # For write operations (create/update), we expect the country ID
+    country_id = serializers.PrimaryKeyRelatedField(source="country", queryset=OrgUnit.objects.all(), write_only=True)
```
tdethier: Since you have another serializer for write operations, you don't need to split this. You should probably use a nested serializer and have a `country` attribute

plugins/polio/api/perfomance_dashboard/serializers.py
```diff
@@ -0,0 +1,98 @@
...
+
+    class Meta:
+        model = PerformanceDashboard
+        fields = [
+            "id",
+            "date",
+            "status",
+            "country_name",  # For read operations (displaying nested country object)
+            "country_id",  # For write operations (accepting country ID)
```
tdethier: Same comment about `country`

plugins/polio/api/perfomance_dashboard/serializers.py
```diff
@@ -0,0 +1,98 @@
...
+    class Meta:
+        model = PerformanceDashboard
+        fields = [
+            "id",
+            "date",
+            "status",
+            "country_id",  # Only country_id is needed for input
+            "antigen",
+        ]
+        # read_only_fields = ["account"] # No longer needed here
```
tdethier:
```suggestion
```

plugins/polio/api/perfomance_dashboard/serializers.py
```diff
@@ -0,0 +1,98 @@
...
+            except Exception as e:
+                logger.error(f"Unexpected error getting profile/account for {request.user}: {e}")
+                raise serializers.ValidationError("Unexpected error encountered while fetching profile/account.")
+        else:
+            # This should ideally not happen if permissions are checked correctly before the serializer
+            logger.error("Request or authenticated user not available in context during creation.")
+            raise serializers.ValidationError("Request context or authenticated user missing.")
+
+        # Call the parent create method with the updated validated_data
+        return super().create(validated_data)
```
tdethier, line 68: If you define how permissions are handled in your viewset, you no longer need to worry about that in your serializer. In that case, you can consider that:
- you do have a user
- they have authenticated
- they have the required permissions

This means that you could almost drop this `create()` method and let DRF do its thing because you don't have any specific business logic to add here (except for the `created_by` field)

plugins/polio/api/perfomance_dashboard/serializers.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def update(self, instance, validated_data):
+        # Set updated_by automatically from the request user
+        request = self.context.get("request")
+        if request and hasattr(request, "user") and request.user.is_authenticated:
```
tdethier: Here as well, if this block is reached, we know for sure that we do have a user

plugins/polio/api/perfomance_dashboard/serializers.py
```diff
@@ -0,0 +1,98 @@
...
+    country_id = serializers.PrimaryKeyRelatedField(source="country", queryset=OrgUnit.objects.all(), write_only=True)
+
+    class Meta:
+        model = PerformanceDashboard
+        fields = [
+            "id",
+            "date",
+            "status",
+            "country_id",  # Only country_id is needed for input
+            "antigen",
```
tdethier: Don't forget about @quang-le's comment on that field, we should probably rename it. If it must have a different name on the web interface, we'll simply change its translation so that the backend logic stays consistent

plugins/polio/api/perfomance_dashboard/serializers.py
```diff
@@ -0,0 +1,98 @@
...
+
+    class Meta:
+        model = PerformanceDashboard
+        fields = [
+            "id",
+            "date",
+            "status",
+            "country_name",  # For read operations (displaying nested country object)
+            "country_id",  # For write operations (accepting country ID)
+            "antigen",
```
tdethier: Same comment about renaming

plugins/polio/api/perfomance_dashboard/permissions.py
```diff
@@ -0,0 +1,43 @@
...
+class HasPerformanceDashboardWritePermission(permissions.BasePermission):
+    """
+    Allows access for users with non-admin or admin logistics permissions.
+    This is for write actions (create, update).
+    """
+
+    def has_permission(self, request, view):
+        return request.user.has_perm(POLIO_PERFORMANCE_NON_ADMIN_PERMISSION.full_name()) or request.user.has_perm(
+            POLIO_PERFORMANCE_ADMIN_PERMISSION.full_name()
+        )
```
tdethier, line 39: There is no distinction between the admin perm and the temporary write perm here, this file should probably be more similar to `polio.api.vaccines.permissions.py`, this means that you'll also probably need to change how these permissions are called. We'll discuss that later

plugins/polio/api/perfomance_dashboard/views.py
```diff
@@ -0,0 +1,98 @@
...
+    def get_serializer_context(self):
+        """
+        Pass the request context to the serializer.
+        This is crucial for the serializer's create/update methods.
+        """
+        # Call the parent implementation to get the default context
+        context = super().get_serializer_context()
+        # Explicitly add the request object to the context
+        context["request"] = self.request
+        return context
```
tdethier: You don't need to do that, DRF will do it for you

plugins/polio/api/perfomance_dashboard/views.py
```diff
@@ -0,0 +1,98 @@
...
+        Dynamically returns the appropriate serializer class based on the action.
+        """
+        if self.action in ["list", "retrieve"]:
+            # For read-only actions (GET), use the serializer that shows detailed, nested data.
+            return PerformanceDashboardListSerializer
+
+        if self.action in ["create", "update", "partial_update"]:
+            # For write actions (POST, PATCH), use the serializer that accepts simple IDs.
+            return PerformanceDashboardWriteSerializer
+        # As a fallback, you can return the default serializer
```
tdethier: You should remove the non-relevant AI comments 😅 
```suggestion
```

plugins/polio/api/perfomance_dashboard/views.py
```diff
@@ -0,0 +1,98 @@
...
+        based on the action being performed.
+        """
+        if self.action in ["list", "retrieve"]:
+            permission_classes = [HasPerformanceDashboardReadOnlyPermission]
+        elif self.action in ["create", "partial_update"]:
+            permission_classes = [HasPerformanceDashboardWritePermission]
+        elif self.action == "destroy":
+            permission_classes = [HasPerformanceDashboardAdminPermission]
+        else:
+            permission_classes = [permissions.IsAuthenticated]
```
tdethier: All scenarios are covered here, if a new one is added and somebody forgets to add it here, let it crash
```suggestion
```

plugins/polio/models/performance_dashboard.py
```diff
@@ -0,0 +1,77 @@
...
+    )
+
+    # Managers
+    objects = DefaultSoftDeletableManager.from_queryset(PerformanceDashboardQuerySet)()
+    objects_only_deleted = OnlyDeletedSoftDeletableManager.from_queryset(PerformanceDashboardQuerySet)()
+    objects_include_deleted = IncludeDeletedSoftDeletableManager.from_queryset(PerformanceDashboardQuerySet)()
+
+    class Meta:
+        verbose_name = _("Performance Dashboard")
+        # verbose_name_plural = _("National Logistics Plans")
```
tdethier:
```suggestion
```

docker-compose.yml
```diff
@@ -93,6 +91,7 @@ services:
             WFP_AUTH_ACCOUNT:
             WFP_AUTH_CLIENT_ID:
             WFP_EMAIL_RECIPIENTS_NEW_ACCOUNT:
+            USE_CELERY:
```
tdethier: Why did you add this?

plugins/polio/tests/api/performance_dashboard/test_views.py
```diff
@@ -0,0 +1,145 @@
...
+    def test_list_returns_only_own_account_dashboards(self):
+        """
+        Test that a user can only list dashboards from their own account.
+        """
+        self.client.force_authenticate(self.user_Hashirama)  # User from Hokage account
+        response = self.client.get(self.PERFORMANCE_DASHBOARD_API_URL)
+        self.assertJSONResponse(response, status.HTTP_200_OK)
+
+        response_data = response.json()
+        # print(f"Response data is: {response_data}")
```
tdethier:
```suggestion
```

plugins/polio/tests/api/performance_dashboard/test_views.py
```diff
@@ -0,0 +1,145 @@
...
+            # Paginated response
+            results = response_data["results"]
+            count = response_data["count"]
+        elif isinstance(response_data, list):
+            # Non-paginated response - results is the list itself
+            results = response_data
+            count = len(results)
+        else:
+            # Unexpected structure
+            self.fail(f"Unexpected response structure: {type(response_data)} - {response_data}")
```
tdethier: You API will always return data in the same format, so you don't need these checks

plugins/polio/tests/api/performance_dashboard/test_views.py
```diff
@@ -0,0 +1,145 @@
...
+        response = self.client.patch(
+            f"{self.PERFORMANCE_DASHBOARD_API_URL}{self.dashboard_2.id}/", data={}, format="json"
+        )
+        self.assertJSONResponse(response, status.HTTP_403_FORBIDDEN)
+
+        # DELETE should be forbidden
+        response = self.client.delete(f"{self.PERFORMANCE_DASHBOARD_API_URL}{self.dashboard_2.id}/")
+        self.assertJSONResponse(response, status.HTTP_403_FORBIDDEN)
+
+    def test_non_admin_user_can_create_and_update(self):
```
tdethier: We need a new test to make sure that the non admin permission time window is enforced: they can update if inside the time window, otherwise they can't

plugins/polio/tests/api/performance_dashboard/test_serializers.py
```diff
@@ -0,0 +1,115 @@
...
+
+        self.assertIn("created_by", data)
+
+        # Check that the values are correct
+        self.assertEqual(data["id"], dashboard.id)
+        self.assertEqual(data["status"], dashboard.status)
+
+        # Check that nested objects are serialized correctly
+
+        self.assertEqual(data["country_name"], dashboard.country.name)
```
tdethier, line 38: It's not nested now, but your AI is right, it should be 😁

plugins/polio/tests/api/performance_dashboard/test_serializers.py
```diff
@@ -0,0 +1,115 @@
...
+            "status": "draft",
+            "antigen": "bOPV",
+            "country_id": self.konoha.id,
+        }
+        # Ensure the user has an iaso_profile and an account
+        if not hasattr(self.user_Hashirama, "iaso_profile"):
+            self.user_Hashirama.iaso_profile = m.IasoProfile.objects.create(
+                user=self.user_Hashirama, account=self.account_hokage
+            )
+            self.user_Hashirama.save()
```
tdethier: You don't need to check that because you define the user however you want in your setup, so you are always sure that they have a profile, especially if you have used the test helper to create the user

plugins/polio/tests/api/performance_dashboard/test_serializers.py
```diff
@@ -0,0 +1,115 @@
...
+        # Create the serializer with the correctly configured DRF request in the context
+        serializer = PerformanceDashboardWriteSerializer(data=data, context={"request": drf_request})
+
+        # Check if the data is valid
+        self.assertTrue(serializer.is_valid(), serializer.errors)
+
+        # --- TEMPORARY DEBUGGING STEP ---
+        is_valid = serializer.is_valid()
+        if not is_valid:
+            print("Serializer validation errors:", serializer.errors)
```
tdethier: You don't want to print if it's not valid, because it should be

You need to  `self.assertTrue(serializer.is_valid())`

plugins/polio/tests/api/performance_dashboard/test_permissions.py
```diff
@@ -0,0 +1,94 @@
+from rest_framework import status
+
+from plugins.polio.models import performance_dashboard as p
+
+from .common_test_data import PerformanceDashboardAPIBase
+
+
+class PerformanceDashboardPermissionsAPITestCase(PerformanceDashboardAPIBase):
```
tdethier: You don't need this class, you have already tested permissions in `test_views`

plugins/polio/api/perfomance_dashboard/filters.py
```diff
@@ -0,0 +1,16 @@
...
+    """
+    Filter for the NationalLogisticsPlan model.
+    """
+
+    country = django_filters.NumberFilter(field_name="country__id")
+    country_block = django_filters.NumberFilter(field_name="country__parent__id")
+
+    class Meta:
+        model = PerformanceDashboard
+        fields = ["status", "antigen"]
```
tdethier: Did you forget these 2 fields or did you forget to add frontend components to add them?

tdethier (changes requested):
Overall, it's not bad 👍 

We can simplify some parts and make everything cleaner. I would add some more tests as well.

The only thing that's really missing is the time window of the non admin perm.

I suggest we take some time together to go through all comments and make the changes

plugins/polio/api/perfomance_dashboard/serializers.py
tdethier: You should still replace these 2 fields

plugins/polio/api/perfomance_dashboard/serializers.py
tdethier: Same comment

plugins/polio/api/perfomance_dashboard/serializers.py
tdethier: Please remove this

plugins/polio/api/perfomance_dashboard/serializers.py
```diff
@@ -0,0 +1,92 @@
...
+            except Exception as e:
+                logger.error(f"Unexpected error getting profile/account for {request.user}: {e}")
+                raise serializers.ValidationError("Unexpected error encountered while fetching profile/account.")
+        else:
+            # This should ideally not happen if permissions are checked correctly before the serializer
+            logger.error("Request or authenticated user not available in context during creation.")
+            raise serializers.ValidationError("Request context or authenticated user missing.")
+
+        # Call the parent create method with the updated validated_data
+        return super().create(validated_data)
```
tdethier, line 68:
```suggestion
    def create(self, validated_data):
        request = self.context["request"]
        user = request.user
        validated_data["created_by"] = user
        validated_data["account"] = user.iaso_profile.account
        return super().create(validated_data)
```

plugins/polio/api/perfomance_dashboard/serializers.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def update(self, instance, validated_data):
+        # Set updated_by automatically from the request user
+        request = self.context.get("request")
+        if request and hasattr(request, "user") and request.user.is_authenticated:
+            validated_data["updated_by"] = request.user
+        # Note: account is typically not changed during an update
+        return super().update(instance, validated_data)
```
tdethier, line 76:
```suggestion
    def update(self, instance, validated_data):
        request = self.context["request"]
        validated_data["updated_by"] = request.user
        return super().update(instance, validated_data)
```

plugins/polio/api/perfomance_dashboard/views.py
```diff
@@ -0,0 +1,64 @@
...
+
+class PerformanceDashboardViewSet(ModelViewSet):
+    """
+    API endpoint for Performance Dashboard.
+
+    This endpoint supports filtering by:
+    - `country` (ID of the country OrgUnit)
+    - `country_block` (ID of the parent OrgUnit of the country, e.g., a region)
+    - `status` (draft, commented, final)
+    - `vaccine` (bOPV, nOPV2, etc.)
```
tdethier: Don't forget to update documentation when you do lists like that

```suggestion
```

plugins/polio/api/perfomance_dashboard/views.py
```diff
@@ -0,0 +1,64 @@
...
+        """
+        Dynamically returns the appropriate serializer class based on the action.
+        """
+        if self.action in ["list", "retrieve"]:
+            return PerformanceDashboardListSerializer
+
+        if self.action in ["create", "update", "partial_update"]:
+            return PerformanceDashboardWriteSerializer
+
+        return super().get_serializer_class()
```
tdethier: There's no serializer class defined at the viewset level, so it's going to crash anyway if this part is reached

```suggestion
```

plugins/polio/js/src/domains/VaccineModule/PerformanceDashboard/modals/CreateEditModal.tsx
```diff
@@ -0,0 +1,139 @@
...
+                    <Field
+                        label={formatMessage(MESSAGES.country)}
+                        name="country_id"
+                        component={SingleSelect}
+                        required
+                        options={countries}
+                        loading={isFetchingCountries}
+                        clearable={false}
+                        isSearchable={false}
+                    />
```
tdethier, line 98: There's an issue with this field when you try to update a line: the country is gone. I guess it's because of the name `country_id`?

Image

plugins/polio/api/perfomance_dashboard/serializers.py
```diff
@@ -0,0 +1,92 @@
...
+            "account",
+            "created_at",
+            "created_by",
+            "updated_at",
+            "updated_by",
+        ]
+        read_only_fields = ["account"]  # Account will be set automatically by the view or create method
+
+
+class PerformanceDashboardWriteSerializer(serializers.ModelSerializer):
```
tdethier, line 47: We don't store which user deletes these objets, so we should probably override the `delete` method and do something like `update()` to store `updated_by=request.user`

plugins/polio/tests/api/performance_dashboard/test_filters.py
```diff
@@ -0,0 +1,50 @@
...
+        self.client.force_authenticate(self.user_Hashirama)
+
+    def test_filter_by_country(self):
+        """
+        Test that we can filter dashboards by country.
+        """
+        # There are 2 dashboards for Konoha in the test data
+        response = self.client.get(f"{self.PERFORMANCE_DASHBOARD_API_URL}?country={self.konoha.id}")
+        self.assertJSONResponse(response, status.HTTP_200_OK)
+        response_data = response.json()
```
tdethier: Nitpick: 
```suggestion
        response_data = self.assertJSONResponse(response, status.HTTP_200_OK)
```

plugins/polio/tests/api/performance_dashboard/test_filters.py
```diff
@@ -0,0 +1,50 @@
...
+        results = response_data.get("results", [])
+        count = len(results)
+        expected_count = PerformanceDashboard.objects.filter(account=self.account_hokage, country=self.konoha).count()
+        self.assertEqual(count, expected_count)
+
+        result_ids = {item["id"] for item in response_data["results"]}
+        self.assertIn(self.dashboard_2.id, result_ids)
+        self.assertIn(self.dashboard_7.id, result_ids)
```
tdethier, line 28: This works fine, but there's an easy way to compare two lists without paying attention to order

```suggestion
        result_ids = [item["id"] for item in response_data["results"]]
        self.assertCountEqual(result_ids, [self.dashboard_2.id, self.dashboard_7.id])
```

plugins/polio/tests/api/performance_dashboard/test_filters.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def test_filter_by_country_block(self):
+        """
+        Test that we can filter dashboards by the country's parent (block).
+        """
+        # There are 2 dashboards for Konoha (which is in Land of Fire block)
+        response = self.client.get(f"{self.PERFORMANCE_DASHBOARD_API_URL}?country_block={self.land_of_fire.id}")
+        self.assertJSONResponse(response, status.HTTP_200_OK)
+        response_data = response.json()
```
tdethier: Same nitpick

plugins/polio/tests/api/performance_dashboard/test_filters.py
```diff
@@ -0,0 +1,50 @@
...
+        results = response_data.get("results", [])
+        count = len(results)
+        self.assertEqual(count, 2)
+
+        result_ids = {item["id"] for item in response_data["results"]}
+        self.assertIn(self.dashboard_2.id, result_ids)
+        self.assertIn(self.dashboard_7.id, result_ids)
```
tdethier: Nitpick: could you maybe use another country block from your setup? It looks like both results are the same, but I'm not sure if it's because the feature works fine or because these values are always returned 😅

plugins/polio/tests/api/performance_dashboard/test_serializers.py
```diff
@@ -0,0 +1,103 @@
...
+
+        self.assertIn("created_by", data)
+
+        # Check that the values are correct
+        self.assertEqual(data["id"], dashboard.id)
+        self.assertEqual(data["status"], dashboard.status)
+
+        # Check that nested objects are serialized correctly
+
+        self.assertEqual(data["country_name"], dashboard.country.name)
```
tdethier, line 38: This should be updated after changing your serializer

plugins/polio/tests/api/performance_dashboard/test_views.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def test_non_admin_user_can_create(self):
+        """
+        Test that a non-admin user can create and update, but not delete.
+        """
+        self.client.force_authenticate(self.user_Kakashi)
+
+        create_data = {"date": "2023-08-01", "status": "draft", "vaccine": "bOPV", "country_id": self.konoha.id}
+        response = self.client.post(self.PERFORMANCE_DASHBOARD_API_URL, data=create_data, format="json")
+        self.assertJSONResponse(response, status.HTTP_201_CREATED)
```
tdethier, line 63: You forgot to add tests for `patch` and `delete`

plugins/polio/tests/api/performance_dashboard/test_views.py
```diff
@@ -0,0 +1,173 @@
...
+
+    @patch("django.utils.timezone.now")
+    def test_non_admin_can_update_recent_record(self, mock_now):
+        """
+        Test that a non-admin user CAN update a recently created record
+        """
+        self.client.force_authenticate(self.user_Kakashi)
+
+        time_of_creation = timezone.make_aware(datetime.datetime(2023, 10, 5))
+        mock_now.return_value = time_of_creation
```
tdethier, line 73: Dates are always annoying to set in unit tests. You can keep this because it works, but next time you need to do something like that, there's an easier way with `time_machine` - see an [example here](https://github.com/BLSQ/iaso/blob/ba5f6c16dca62215fc5a27daf4e045a1ed33496d/iaso/tests/api/org_unit_change_requests/test_org_unit_change_requests.py#L158)

plugins/polio/tests/api/performance_dashboard/test_views.py
```diff
@@ -0,0 +1,173 @@
...
+        response = self.client.get(self.PERFORMANCE_DASHBOARD_API_URL)
+        self.assertJSONResponse(response, status.HTTP_200_OK)
+
+        response_data = response.json()
+        # Check if the response is paginated or just a list
+        if "results" in response_data:
+            results = response_data["results"]
+            count = len(results)
+        else:
+            self.fail("Response is not paginated as expected")
```
tdethier:
```suggestion
```

plugins/polio/tests/api/performance_dashboard/test_views.py
```diff
@@ -0,0 +1,173 @@
...
+        time_of_update = timezone.make_aware(datetime.datetime(2023, 10, 20))
+        mock_now.return_value = time_of_update
+        # Now, on 2023-10-20 (10 days later), try to update it. This should fail.
+        update_data = {"status": "final"}
+        response = self.client.patch(
+            f"{self.PERFORMANCE_DASHBOARD_API_URL}{old_dashboard.id}/", data=update_data, format="json"
+        )
+        self.assertJSONResponse(response, status.HTTP_403_FORBIDDEN)
+
+    def test_admin_user_can_delete(self):
```
tdethier, line 121: I would add a check in this test:
- fetch the performance dashboard object and refresh it from database
- make sure that the `updated_by` field has the user ID (`user_Hashirama`)

plugins/polio/tests/api/performance_dashboard/test_views.py
```diff
@@ -0,0 +1,173 @@
...
+        Test that a non-admin user can create and update, but not delete.
+        """
+        self.client.force_authenticate(self.user_Kakashi)
+
+        create_data = {"date": "2023-08-01", "status": "draft", "vaccine": "bOPV", "country_id": self.konoha.id}
+        response = self.client.post(self.PERFORMANCE_DASHBOARD_API_URL, data=create_data, format="json")
+        self.assertJSONResponse(response, status.HTTP_201_CREATED)
+
+    @patch("django.utils.timezone.now")
+    def test_non_admin_can_update_recent_record(self, mock_now):
```
tdethier, line 66: I would add a check in this test:
- fetch the performance dashboard object and refresh it from DB
- make sure that the `updated_by` field has the user ID (`user_Kakashi`)

plugins/polio/tests/api/performance_dashboard/common_test_data.py
```diff
@@ -0,0 +1,148 @@
...
+        cls.user_Kakashi = cls.create_user_with_profile(
+            username="Kakashi Hatake",
+            account=cls.account_hokage,
+            permissions=[POLIO_PERFORMANCE_NON_ADMIN_PERMISSION],
+        )
+
+        # User with no permissions
+        cls.user_Naruto_no_perms = cls.create_user_with_profile(
+            username="Naruto UZUMAKI", account=cls.account_hokage, permissions=[]
+        )
```
tdethier: Please use snake_case for your variable names (no capital letters)

tdethier (changes requested):
I forgot to add a comment in my previous review ⬆️ 

The feature works fine and the code is much better 👍 
But there are still some things I'd like you to change, I made some suggestions and some comments.

@quang-le could you please go through the frontend again?

plugins/polio/api/perfomance_dashboard/serializers.py
quang-le: @tdethier I disagree: it's easier for the front-end to have a flat response. Since it's just 2 fields, I think nesting is overkill

plugins/polio/api/perfomance_dashboard/views.py
```diff
@@ -0,0 +1,64 @@
...
+
+    The permissions are structured as follows:
+    - **Read-only**: Can only list and retrieve plans.
+    - **Non-admin**: Can create and update plans.
+    - **Admin**: Can delete plans.
+    """
+
+    # The serializer_class is removed because 'get_serializer_class' is implemented
+    # DRF will call get_serializer_clas() instead of looking for this attribute
+    # serializer_class =
```
quang-le: This comment is not necessary.

plugins/polio/js/src/constants/translations/fr.json
```diff
@@ -870,5 +870,17 @@
     "iaso.polio.validation.onsetAfterNotificationError": "La date de début doit précéder la date de notification virus",
     "iaso.polio.validation.onsetAfterOutbreakDeclarationError": "La date de début doit précéder la date de déclaration de l'épidémie",
     "iaso.polio.validation.virusNotificationAfterOutbreakDeclarationError": "La date de notification virus doit précéder la déclaration de l'épidémie",
-    "iaso.task.label.gotoLqasForCountry": "Voir la page LQAS"
+    "iaso.task.label.gotoLqasForCountry": "Voir la page LQAS",
+    "iaso.polio.performanceDashboard": "Tableau des performances",
+    "iaso.polio.performance.delete": "Effacer les données de performance: {name}",
+    "iaso.polio.performance.deleteWarning": "Voulez-vous vraiment supprimer cette entrée ?",
+    "iaso.polio.performance.add": "Ajouter une donnée de performance",
+    "iaso.polio.performance.status.draft": "Brouillon",
```
quang-le:
```suggestion
    "iaso.polio.performance.status.draft": "Ébauche",
```

This how we tranlsate "draft" elsewhere, so please align

plugins/polio/js/src/domains/VaccineModule/PerformanceDashboard/hooks/api/useGetPerformance.ts
```diff
@@ -0,0 +1,31 @@
...
+    const apiParams = useApiParams(safeParams);
+
+    return useSnackQuery<PerformanceList>(
+        [
+            'performance-dashboard',
+            apiParams,
+            apiParams.page,
+            apiParams.limit,
+            apiParams.order,
+        ],
```
quang-le, line 20:
```suggestion
        [ 'performance-dashboard', apiParams ],
```

react-query is smart enough to detect changes in the object's entries

plugins/polio/js/src/domains/VaccineModule/PerformanceDashboard/hooks/api/useGetPerformance.ts
```diff
@@ -0,0 +1,31 @@
...
+const getPerformanceDashboard = (params: any) => {
+    const queryString = new URLSearchParams(params).toString();
+    return getRequest(`/api/polio/performance_dashboard/?${queryString}`);
+};
+
+export const useGetPerformanceDashboard = (params: any) => {
+    const safeParams = useUrlParams(params);
+    const apiParams = useApiParams(safeParams);
+
+    return useSnackQuery<PerformanceList>(
```
quang-le, line 17: Please pass an object as argument. It's more legible and avoids passing `undefined`. This way of using `useSnackBarQuery` is considered legacy

plugins/polio/js/src/domains/VaccineModule/PerformanceDashboard/hooks/api/useGetPerformance.ts
```diff
@@ -0,0 +1,31 @@
...
+            'performance-dashboard',
+            apiParams,
+            apiParams.page,
+            apiParams.limit,
+            apiParams.order,
+        ],
+        () => getPerformanceDashboard(apiParams),
+        undefined,
+        {
+            staleTime: 1000 * 60 * 5, // 5 minutes
```
quang-le, line 24: please check other hooks to see which options we pass by default. We don't have it set at the app level yet (sorry) but we try to keep it consistent

plugins/polio/js/src/domains/VaccineModule/PerformanceDashboard/hooks/api/useSavePerformance.ts
```diff
@@ -0,0 +1,39 @@
...
+        undefined, // No specific error message config here, use default
+        ['performance-dashboard'], // Query keys to invalidate on success (optional, but good practice)
+        {
+            onSuccess: () => {
+                // Invalidate the 'performance-dashboard' query cache after a successful save.
+                // This forces any components using useGetPerformanceDashboard to refetch fresh data.
+                queryClient.invalidateQueries(['performance-dashboard']);
+            },
+        },
+    );
```
quang-le: same as useSnackQuery

plugins/polio/js/src/domains/VaccineModule/PerformanceDashboard/hooks/api/useSavePerformance.ts
```diff
@@ -0,0 +1,39 @@
...
+import {
+    postRequest,
+    patchRequest,
+} from '../../../../../../../../../hat/assets/js/apps/Iaso/libs/Api';
+import { useSnackMutation } from '../../../../../../../../../hat/assets/js/apps/Iaso/libs/apiHooks';
+import { PerformanceData } from '../../types';
+
+const savePerformanceDashboard = (data: Partial<PerformanceData>) => {
+    if (data.id) {
+        // If data has an ID, it means we are updating an existing entry
```
quang-le: This comment is not necessary

plugins/polio/js/src/domains/VaccineModule/PerformanceDashboard/hooks/options.ts
```diff
@@ -0,0 +1,33 @@
...
+const statuses: string[] = ['draft', 'commented', 'final']; // Updated statuses
+const vaccines: string[] = ['nOPV2', 'bOPV', 'nOPV2 & bOPV'];
+
+export const useStatusOptions = (): DropdownOptions<string>[] => {
+    const { formatMessage } = useSafeIntl();
+    return useMemo(() => {
+        return statuses.map(status => {
+            return {
+                value: status,
+                // Assuming messages are like MESSAGES.draft, MESSAGES.commented, MESSAGES.final
```
quang-le: 🗑️

plugins/polio/js/src/domains/VaccineModule/PerformanceDashboard/hooks/options.ts
```diff
@@ -0,0 +1,33 @@
+import { useMemo } from 'react';
+import { useSafeIntl } from 'bluesquare-components';
+import { DropdownOptions } from '../../../../../../../../hat/assets/js/apps/Iaso/types/utils';
+import MESSAGES from '../messages';
+
+const statuses: string[] = ['draft', 'commented', 'final']; // Updated statuses
+const vaccines: string[] = ['nOPV2', 'bOPV', 'nOPV2 & bOPV'];
```
quang-le: why is mOPV2 not included?

plugins/polio/js/src/domains/VaccineModule/PerformanceDashboard/hooks/options.ts
```diff
@@ -0,0 +1,33 @@
...
+export const useVaccineOptions = (): DropdownOptions<string>[] => {
+    return useMemo(() => {
+        return vaccines.map(vaccine => {
+            return {
+                value: vaccine,
+                label: vaccine,
+            };
+        });
+    }, []);
+};
```
quang-le, line 25: If you leave the dependencies array of `useMemo` empty, that means you don't need `useMemo`. You can import `defaultVaccineOptions` and filter out `mOPV2` if needed

plugins/polio/js/src/domains/VaccineModule/PerformanceDashboard/modals/DeleteModal.tsx
quang-le, line 1: There's already a `DeleteModal`component: https://github.com/BLSQ/iaso/blob/develop/hat/assets/js/apps/Iaso/components/DeleteRestoreModals/DeleteRestoreModal.tsx

Please see if you can use that one instead

plugins/polio/js/src/domains/VaccineModule/PerformanceDashboard/table/PerformanceDashboardTable.tsx
```diff
@@ -0,0 +1,34 @@
...
+            params={params}
+            columns={columns}
+            baseUrl={baseUrls.performanceDashboard}
+            columnSelectorEnabled
+            columnSelectorButtonType="button"
+            marginTop={false}
+            extraProps={{
+                loading: isFetching,
+                params,
+                defaultPageSize: 10,
```
quang-le: nitpick: we usually default to 20

plugins/polio/js/src/domains/VaccineModule/PerformanceDashboard/table/usePerformanceDashboardColumns.tsx
```diff
@@ -0,0 +1,96 @@
...
+                                ]}
+                            >
+                                <DeletePerformanceModal
+                                    performanceData={performanceData}
+                                />
+                            </DisplayIfUserHasPerm>
+                        </>
+                    );
+                },
+            },
```
quang-le: If the user has the read only permission, this will display an empty "Actions" column. You should hide the column if the user has none of the `POLIO_PERFORMANCE_NON_ADMIN_PERMISSION` and ` POLIO_PERFORMANCE_ADMIN_PERMISSION`

plugins/polio/migrations/0247_alter_poliopermissionsupport_options_and_more.py
quang-le, line 1: You will need to add a manual migration:
- users that have `iaso_polio_vaccine_stock_earmarks_admin` or `iaso_polio_vaccine_stock_management_write` should get the `iaso_polio_performance_admin`
- users that have `iaso_polio_vaccine_stock_earmarks_nonadmin` or `iaso_polio_vaccine_stock_management_read` should get the `iaso_polio_performance_non_admin`

plugins/polio/tests/api/performance_dashboard/common_test_data.py
quang-le: Also please avoid using pop culture names. We're trying to get rid of the existing ones as they make the tests less legible.

I'd rather have `cls.user_admin` than `cls.user_hashirama`. It really makes a difference when adding/reviewing/maintaining tests

quang-le (changes requested):
Please remove the LLM-generated comments in the front-end, as they are a bit noisy.
Most comments are related to our standard practices, so please make the changes a s we try to keep the codebase consistent.

Don't hesitate to reach out if you need help with the requested migration.

Also: clarify why `mOPV2` was left out of vaccines or add it.

Good work overall. Lots of comments are expected for a first issue, especially when it's a full feature.

plugins/polio/api/perfomance_dashboard/serializers.py
Crebert08: so do I keep these fields or should I delete them ?

tdethier: Alright, if the frontend doesn't want them nested, you can keep them like this :sweat_smile:

plugins/polio/js/src/domains/VaccineModule/PerformanceDashboard/hooks/options.ts
Crebert08: The jira ticket explicitly stated that the choices for antigen will not contain mOPV2. Should I ask Laura if I have to add that choice ?

plugins/polio/api/perfomance_dashboard/views.py
```diff
@@ -55,10 +52,15 @@ def get_serializer_class(self):
...
 
-        return super().get_serializer_class()
+        return PerformanceDashboardListSerializer
+
+    def perform_destroy(self, instance):
+        """
+        Perform a soft delete and log the user who performed the action.
+        """
+        instance.updated_by = self.request.user
+        instance.save(update_fields=["updated_by"])
```
quang-le: This not really necessary since  `instance.delete()` calls `save()` under the hood

plugins/polio/tests/api/performance_dashboard/common_test_data.py
quang-le, line 1: ❤️ future devs will thank you for this change

quang-le (commented):
Nice, don't forget the front-end changes

quang-le (approved):
Nice work. Waiting for @tdethier 's greenlight and we can merge

plugins/polio/js/src/domains/VaccineModule/PerformanceDashboard/hooks/options.ts
```diff
@@ -0,0 +1,26 @@
+import { useMemo } from 'react';
+import { useSafeIntl } from 'bluesquare-components';
+import { DropdownOptions } from '../../../../../../../../hat/assets/js/apps/Iaso/types/utils';
+import MESSAGES from '../messages';
+import { defaultVaccineOptions } from '../../SupplyChain/constants'; // New import
+
+const statuses: string[] = ['draft', 'commented', 'final'];
+// Removed: const vaccines: string[] = ['nOPV2', 'bOPV', 'nOPV2 & bOPV'];
```
tdethier:
```suggestion
import { defaultVaccineOptions } from '../../SupplyChain/constants';

const statuses: string[] = ['draft', 'commented', 'final'];
```

plugins/polio/api/perfomance_dashboard/serializers.py
```diff
@@ -0,0 +1,76 @@
...
+            "country_id",
+            "vaccine",
+            "account",
+            "created_at",
+            "created_by",
+            "updated_at",
+            "updated_by",
+        ]
+        read_only_fields = ["account"]
+        extra_kwargs = {"country_id": {"read_only": True}}
```
tdethier, line 44: Why is it in `extra_kwargs` instead of the attributes like `country_name`?

plugins/polio/migrations/0248_auto_20251202_0947.py
```diff
@@ -0,0 +1,53 @@
...
+    read_only_users = User.objects.filter(user_permissions__codename__in=read_only_permissions).distinct()
+    for user in admin_users:
+        perm = Permission.objects.get(codename="iaso_polio_performance_admin")
+        user.user_permissions.add(perm)
+    for user in non_admin_users:
+        perm = Permission.objects.get(codename="iaso_polio_performance_non_admin")
+        user.user_permissions.add(perm)
+    for user in read_only_users:
+        perm = Permission.objects.get(codename="iaso_polio_performance_read_only")
+        user.user_permissions.add(perm)
```
tdethier: You should fetch once each performance dashboard permission outside of the loop, otherwise you're fetching the same permission over and over again

plugins/polio/migrations/0248_auto_20251202_0947.py
```diff
@@ -0,0 +1,53 @@
...
+    read_only_users = User.objects.filter(user_permissions__codename="iaso_polio_performance_read_only").distinct()
+    for user in admin_users:
+        perm = Permission.objects.get(codename="iaso_polio_performance_admin")
+        user.user_permissions.remove(perm)
+    for user in non_admin_users:
+        perm = Permission.objects.get(codename="iaso_polio_performance_non_admin")
+        user.user_permissions.remove(perm)
+    for user in read_only_users:
+        perm = Permission.objects.get(codename="iaso_polio_performance_read_only")
+        user.user_permissions.remove(perm)
```
tdethier: Same comment

tdethier (approved):
It's much better, nice work 👍 

I made a few comments and I'm already approving this, but please make the change on the migration file, it will greatly reduce the number of DB calls

PR #2559, opened 2025-11-19
IA-4610: Openhexa raise Sentry error

beygorghor:
While fetching the OpenHexa configuration during planning editing, if the account does not have an associated OpenHexa config, the backend API throws an error and triggers Sentry. For this specific endpoint, the absence of a configuration should fail silently—without affecting error reporting for other usages (trying to fetch pipelines, launch metrics command).

Related JIRA tickets : IA-4610

- [ ] Are there enough tests?

## Doc

- 

## Changes

- `get_openhexa_config` option to fail silently

## How to test

First edit a planning without those changes, make sure you have an error in the terminal:
`
ERROR    2025-11-19 08:27:22,103 iaso.utils.openhexa -- No OpenHEXA workspace configured for account ...
`
Apply those changes, error should disappear

## Print screen / video

-

## Notes

```
fix: empty instance pop up

Refs: IA-3665
```

### DIFF
--- a/iaso/api/openhexa/views.py
+++ b/iaso/api/openhexa/views.py
@@ -313,7 +313,7 @@ class OpenHexaPipelinesViewSet(ViewSet):
                 return Response({"configured": False})
 
             account = request.user.iaso_profile.account
-            *_, workspace = get_openhexa_config(account, silent=True)
+            *_, workspace = get_openhexa_config(account)
 
             # If we reach here, config is valid
             response_data = {"configured": True}
--- a/iaso/api/openhexa/views.py
+++ b/iaso/api/openhexa/views.py
@@ -324,9 +324,5 @@ class OpenHexaPipelinesViewSet(ViewSet):
 
             return Response(response_data)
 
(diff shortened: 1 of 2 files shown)


iaso/api/openhexa/views.py
```diff
@@ -313,7 +313,7 @@ def config(self, request):
                 return Response({"configured": False})
 
             account = request.user.iaso_profile.account
-            *_, workspace = get_openhexa_config(account)
+            *_, workspace = get_openhexa_config(account, silent=True)
```
quang-le: Why do you not want to see the logs here? Isn't it useful debug info?

tdethier (changes requested):
I would probably not add a `silent` parameter, but lower the logging level (so `logger.warning()`) instead. 

I think it will not cause a Sentry error, but if it does, we should lower it again (since `ValidationErrors` are raised, it's already enough information)

quang-le (approved):

tdethier (approved):
LGTM 👍

PR #2562, opened 2025-11-20
WC2-852: Fix imports fails due to storage logs

bmonjoie:
`process_mobile_bulk_upload` was wrapped into one giant transaction. This was problematic because we can have multiple zips each with the **full** storage logs. The storage logs could point on entities that are created in one of the zips. If we have entities created in at least two zips, the storage logs can never succeed and therefore the import of all the zips is blocked.

Also, I extracted the logic of finding an entity to attach the storage logs to the proper entity even when there are duplicates.

Related JIRA tickets : WC2-852

- [X] Are there enough tests?

## Changes

See top of the PR

## How to test

It's quite hard to test because you need to send two zips with some entities created in each zips and have storage logs to point to those new entities.

The best is to test that synchronizing using zip is still working fine.

```
fix: empty instance pop up

Refs: IA-3665
```

### DIFF
--- a/iaso/api/instances/instances.py
+++ b/iaso/api/instances/instances.py
@@ -5,7 +5,7 @@ import tempfile
 
 from copy import copy
 from time import gmtime, strftime
-from typing import Any, Dict, Union
+from typing import Any, Dict, Optional, Union
 
 import pandas as pd
 
@@ -44,6 +44,7 @@ from iaso.api.org_units import HasCreateOrgUnitPermission
 from iaso.api.serializers import OrgUnitSerializer
 from iaso.exports import CleaningFileResponse, parquet
 from iaso.models import (
+    Account,
     Entity,
     Form,
     Instance,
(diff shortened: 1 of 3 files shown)


iaso/api/instances/instances.py
```diff
@@ -932,6 +933,42 @@ def instance_logs(self, request, pk=None, logId=None):
...
+        filters = {
+            "uuid": entity_uuid,
+            "entity_type_id": entity_type_id,
+            "account": account,
+        }
+    else:
+        filters = {
+            "uuid": entity_uuid,
+            "account": account,
+        }
```
tdethier, line 949:
```suggestion
    filters = {
        "uuid": entity_uuid,
        "account": account,
    }
    if entity_type_id is not None:
        filters["entity_type_id"] = entity_type_id

```

tdethier (commented):
I couldn't test that because I don't have the required setup, but LGTM 👍
Nitpick: I would add a unit test to show that `import_storage_logs` doesn't raise an exception anymore when an entity is missing

I'll let the others who are more familiar with entities and storage logs approve.

iaso/api/instances/instances.py
bmonjoie: >  Nitpick: I would add a unit test to show that import_storage_logs doesn't raise an exception anymore when an entity is missing

It still does :sweat_smile: I could not get rid of that behavior unfortunately

beygorghor (approved):

bramj (approved):
Splitting up the DB transaction is something I've been wanting to do as well, I believe it will also improve a lot uploads that fail because of entities/org units that are duplicated across zip files (for large uploads such as Trypelim).

Now that we have uniqueness constraints on the UUIDs there should be no risk in splitting off the org units, instances and entities into separate transactions.

For the others I'm less sure as I'm less familiar with them.

**But we should be sure we have these uniqueness constraints on all our environments.** @beygorghor did they put it in place on WFP?
I see the IASO repo doesn't have it yet which makes it easy to have issues on new deployments. Not sure how, but we should add it to the repo, but skip it's creation for existing ENVs or an `IF EXIST`.

@Phil-V do you have any thoughts on this?

Phil-V:
Hi @bramj, I think it should indeed be safe, especially as entities and instances are still wrapped up in the same transaction. But like you said, the uniqueness constraint is critical.

Phil-V:
I have a small nitpick about the progress updates:

doing something like this:
```py
with transaction.atomic():
     log_progress(the_task, 0, "Processing foo") # <- will not be displayed to the user until the transaction is done
     do_stuff("foo")
     log_progress(the_task, 10, "Done processing foo")
```

Won't show "Processing foo" in the UI until the transaction commits and that step is already done. This is ok now that we have a full history in the tasks UI but something to keep in mind. Perhaps the first `log_progress` could be placed outside the atomic block?

madewulf (approved):

madewulf:
To be well tested on staging before release

PR #2571, opened 2025-11-25
[IA-4481] Store user when deleting instances

tdethier:
Store user when deleting instances

Related JIRA tickets : IA-4481

- [x] Are there enough tests?

## Doc

/

## Changes

- Store user + timestamp in `updated_xx` fields when deleting instances
- Fix bug in instances comparison (the history always displayed the same user in both instances, even when the value was a different user)
- Display who deleted the instance in instance detail view

## How to test

- Log into IASO with user A
- Soft delete an instance
- Log out
- Log into IASO with user B
- Check the soft deleted instance: you should now see that it has been deleted by user A
- Restore the instance
- The 2 fields related to deletion now display information about updating the instance and display user B
- Check the instance history
- Compare the initial version with the latest version: you should see different users

## Print screen / video

Details when the instance is not deleted:
image

Details when the instance is deleted:
image

The comparison now displays the proper users:
image

## Notes

/

```
fix: empty instance pop up

Refs: IA-3665
```

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/instances/compare/components/CompareInstanceLogs.tsx
+++ b/hat/assets/js/apps/Iaso/domains/instances/compare/components/CompareInstanceLogs.tsx
@@ -176,7 +176,7 @@ export const CompareInstanceLogs: FunctionComponent = () => {
                                         logAInitialValue),
                             )}
                             loading={isFetchingInstanceLogs}
-                            user={instanceLogA?.user}
+                            user={instanceLogB?.user}
                             infos={instanceLogContent.logB}
                             dropDownLoading={isInstanceLogBFetching}
                         />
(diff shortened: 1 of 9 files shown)


hat/assets/js/apps/Iaso/domains/instances/compare/components/CompareInstanceLogs.tsx
```diff
@@ -176,7 +176,7 @@ export const CompareInstanceLogs: FunctionComponent = () => {
                                         logAInitialValue),
                             )}
                             loading={isFetchingInstanceLogs}
-                            user={instanceLogA?.user}
+                            user={instanceLogB?.user}
```
beygorghor, line 179: 😢

beygorghor (approved):

PR #2577, opened 2025-11-26
IA-4593 search users by ids or refs

mestachs:
Allow advanced users to search by ids/dhis2_ids users, add the id in the excel/csv export.

Related JIRA tickets : IA-4593

- [ ] Are there enough tests?

## Doc

## Changes

- add the user profile's id in the csv and excel export
- allows to search with ids:34, 35
- add a info icon near the search field with helper text

## How to test

- seed a new environnement (via seed command or setuper)
- create users if needed
- export the xlsx and csv
- use the search feature

## Print screen / video

image

```
feat: search users by ids or refs

Refs: IA-4593
```

### DIFF
--- a/iaso/api/profiles/profiles.py
+++ b/iaso/api/profiles/profiles.py
@@ -1,5 +1,3 @@
-import re
-
 from typing import Any, List, Optional, Union
 
 from django.conf import settings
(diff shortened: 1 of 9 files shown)


iaso/api/profiles/profiles.py
```diff
@@ -36,6 +38,16 @@
 PK_ME = "me"
 
 
+def _parse_ids(prefix, search):
+    s = search.replace(prefix, "")
+    try:
+        parsed_ids = re.findall("[A-Za-z0-9_-]+", s)
+    except:
+        print("Failed parsing ids in search", search)
```
tdethier: I would use a logger instead

iaso/api/profiles/profiles.py
```diff
@@ -36,6 +38,16 @@
 PK_ME = "me"
 
 
+def _parse_ids(prefix, search):
```
tdethier: It's a good idea to make it a function 👍 

Nitpick: I would have placed it somewhere in utils and replaced the similar code in other endpoints to call this function

iaso/tests/api/test_profiles.py
```diff
@@ -341,6 +341,7 @@ def test_profile_list_ok(self):
         self.assertValidProfileListData(response.json(), 7)
 
     def test_profile_list_export_as_csv(self):
```
tdethier, line 343: I would add another test with `search=ids:1,2,3` to make sure that it returns the expected profiles

tdethier: There's already a `test_search_by_ids` but it uses the other filter you spotted (using `User` IDs instead of `Profile` IDs). I'd suggest renaming it and adding the new test on profile IDs

tdethier (approved):
The feature works fine, good job 👍 

I would add a new test to cover this new type of search

PR #2580, opened 2025-11-26
[IA-4558] Add a button for the Django admin panel

tdethier:
Add a button for the Django admin panel

Related JIRA tickets : IA-4558

- [ ] Are there enough tests?

## Doc

/

## Changes

- Added a new button to the top bar

## How to test

- Log into IASO as a staff member and superuser :arrow_right: you should see the new button on the top bar
- Click on the button :arrow_right: a new tab opens with the Django admin panel
- Log out from IASO
- Log into IASO as a standard user :arrow_right: you should not see the new button on the top bar

## Print screen / video

image

## Notes

- I initially wanted to add this to the side bar menu, but it's impossible to add a new entry that does not go to `/dashboard/something...`, so I added it there

```
fix: empty instance pop up

Refs: IA-3665
```

### DIFF
--- a/hat/assets/js/apps/Iaso/components/nav/TopBarComponent.js
+++ b/hat/assets/js/apps/Iaso/components/nav/TopBarComponent.js
@@ -15,6 +15,7 @@ import { useCurrentUser } from '../../utils/usersUtils.ts';
 import { CurrentUserInfos } from './CurrentUser/index.tsx';
 import { HomePageButton } from './HomePageButton.tsx';
 import { LogoutButton } from './LogoutButton.tsx';
+import { DjangoAdminPanelButton } from 'Iaso/components/nav/DjangoAdminPanelButton';
 
 const styles = theme => ({
     menuButton: {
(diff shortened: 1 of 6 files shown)


hat/assets/js/apps/Iaso/components/nav/TopBarComponent.js
```diff
@@ -150,6 +151,12 @@ function TopBar(props) {
                                 <HomePageButton />
                             </Box>
 
+                            {(currentUser.is_staff === true || currentUser.is_superuser === true) && (
```
quang-le: The admin is restricted to staff users, so super users will be redirected to the login page

quang-le:
```suggestion
                            { currentUser.is_staff === true && (
```

quang-le (changes requested):
See suggestion

hat/assets/js/apps/Iaso/components/nav/TopBarComponent.js
tdethier: I don't agree with that: 
- we don't use the base Django permissions, so if you're staff and not superuser, you won't be able to do anything on that page
image

- if you use a superuser account, you should have access to everything - by default a superuser made with `manage.py createsuperuser` is not a staff user

If anything, I would restrict that page to superusers only, given how we handle permissions

tdethier: I'll make the change, but since we don't use Django's permission system, anyone who is staff and not superuser won't be able to do anything in the admin panel

image

Maybe restricting to both staff and superuser would be a better option

quang-le: Yes, let's do that then

quang-le (approved):

PR #2585, opened 2025-11-28
POLIO-2038: make polio LQAS country view embeddable

quang-le:
The page needs to be publicly accessible

Related JIRA tickets : POLIO-2038

## Self proofreading checklist

- [x] Did I use eslint and ruff formatters?
- [x] Is my code clear enough and well documented?
- [x] Are my typescript files well typed?
- [ ] New translations have been added or updated if new strings have been introduced in the frontend
- [ ] My migrations file are included
- [x] Are there enough tests?
- [ ] Documentation has been included (for new feature)

## Doc

Tell us where the doc can be found (docs folder, wiki, in the code...).

## Changes

- Country map
    - Add route (front and backend)
    - Adapt endpoints and API calls: mainly pass app_id to endpoints filtering by account
    - Update tests

## How to test

Pre-requisite: a polio table with lqas data, i.e `JsonDataStore` with the relevant data. It can be copied from staging

- when not logged go to `/dashboard/polio/embeddedLqasCountry/` and select date + campaign
- compare to the behaviour of `dashboard/polio/lqas/lqas/` which is the same page but requires auth

Same procedure with `/dashboard/polio/embeddableLqasMap`

## Print screen / video

Country map

Screenshot 2025-12-01 at 13 50 28

PR #2588, opened 2025-12-01
POLIO-2041 embed lqas polio continent map

quang-le:

Related JIRA tickets :  POLIO-2041

- [x] Are there enough tests?

## Doc

## Changes

- Continent map
    - Add route in the front and back end
    - Add check on user or app_id when fetching shapes
    - Pass `app_id` param from front-end when calling API from embedded page
    - add test for anon access

## How to test

Pre-requisite: a polio table with lqas data, i.e `JsonDataStore` with the relevant data. It can be copied from staging

- when not logged go to  `/dashboard/polio/embeddedLqasMap` and select time period (change to 48months directly in url if no data available on local server)
- compare to the behaviour of `dashboard/polio/lqas/lqas-map` which is the same page but requires auth

## Print screen / video

Continent map

Screenshot 2025-12-01 at 14 17 18

## Notes

Merges into https://github.com/BLSQ/iaso/pull/2585


plugins/polio/js/src/domains/LQAS-IM/LQAS/CountryOverview/LqasCountryMapView.tsx
```diff
@@ -69,6 +69,7 @@ export const LqasCountryMapView: FunctionComponent<Props> = ({
             roundNumber,
         });
     }, [campaign, roundNumber]);
+    console.log('SCOPE IDS', scopeIds, campaign, roundNumber);
```
tdethier: 👀 
```suggestion
```

plugins/polio/tests/test_lqas_map.py
```diff
@@ -349,7 +352,21 @@ def test_authorized_user(self):
...
+        response = self.assertJSONResponse(response, 200)
+        data = response["results"]
+        # Passing no app_id when anonymous should return empty org units
+        self.assertEqual(len(data), 0)
+        response = c.get(
+            f"/api/polio/lqasmap/global/?category=lqas&app_id={self.project.app_id}", accept="application/json"
+        )
+        response = self.assertJSONResponse(response, 200)
+        data = response["results"]
+        self.assertGreater(len(data), 0)
```
tdethier, line 369: Shouldn't you get results now that you passed an `app_id`?

plugins/polio/tests/test_lqas_map.py
```diff
@@ -598,6 +615,23 @@ def test_lqas_global_period_filter(self):
...
+        data = response["results"]
+        # Passing no app_id when anonymous should return empty org units
+        self.assertEqual(len(data), 0)
+        response = c.get(
+            f"/api/polio/lqasmap/zoomin/?category=lqas&bounds={self.url_bounds}&app_id={self.project.app_id}",
+            accept="application/json",
+        )
+        response = self.assertJSONResponse(response, 200)
+        data = response["results"]
+        self.assertGreater(len(data), 0)
```
tdethier, line 633: Same question about results & `app_id`

tdethier (commented):
The feature works fine, I can access that page when I'm not logged in and see the same results on both pages👍 

Fun fact - there's a small difference in styling between the pages, I don't really understand why: 
- `Legend` in bold or not
- the size of each field name is different (`Round`)

When logged in:
Image

When anonymous:

Image


I'm not approving yet, because I have some questions on 2 tests, but otherwise LGTM

plugins/polio/tests/test_lqas_map.py
tdethier: Oh now I realize that it's `assertGreater` - shouldn't we check for actual results?

plugins/polio/tests/test_lqas_map.py
tdethier: Oh now I realize that it's `assertGreater` - shouldn't we check for actual results?

plugins/polio/tests/test_lqas_map.py
quang-le: The detailed response is checked in `test_lqas_global`.  This is basically the same test as `test_authorized_user`, which doesn't check the details either. I just check it here because the filtering by app_id doesn't `raise` but returns an empty queryset when no account is found



The `app_id` filtering relies on `OrgUnit.objects.filter_for_user_and_app_id`, so I could double check if this one is tested in iaso and add that test

plugins/polio/tests/test_lqas_map.py
quang-le: same answer

tdethier (approved):
LGTM 👍

PR #2597, opened 2025-12-04
POLIO-2034 performance dashboard threshold setting

quang-le:

Related JIRA tickets : POLIO-2034

- [x] Are there enough tests?

## Doc

To be added in company wiki

## Changes

- Add `PerformanceTresholds`model:
    - success, warning and fail thresholds are stored as JSONFields. serializer validation ensures they have the proper, json-logic compatible shape
- Add API: read operations are public for now, in anticipation of page embedding
- Log changes with `AuditMixin`
- Refactor Performance dashboard API to also be publicly available for read operations (same rationale)
- Refactor `PerformanceDashboard` to use audit logs i.o `updated_by` and `created_by` fields
- Add convenience method for testing with jsonschema  

## How to test

Try to perform CRUD with Django's web API, using different permission profiles


hat/audit/models.py
```diff
@@ -142,6 +142,42 @@ def field_diffs(self):
...
+                    },
+                    "required": ["id", "user_name", "user_id"],
+                },
+                "created_at": {"type": "string"},
+                "org_unit_change_request_id": {"type": ["string", "null"]},
+                "past_value": {"type": "array", "items": past_value_schema},
+                "new_value": {"type": "array", "items": new_value_schema},
+            },
+            "required": ["id", "content_type", "object_id", "source", "user", "created_at", "past_value", "new_value"],
+        }
```
quang-le, line 179: This will be useful when we test the creation of `Modification` (audit logs) with `jsonschema`.  @beygorghor  @tdethier , maybe we could use that a bit more in the tests, it would allow to test the shape of an API response without having to test the values of all the fields, and focus on explicitly testing relevant values

plugins/polio/api/country_plan/permissions.py
```diff
@@ -28,12 +28,9 @@ def has_permission(self, request, view):
...
         - Write access for unsafe methods if user has write or admin permissions.
         """
+
         if request.method in permissions.SAFE_METHODS:
-            return (
-                request.user.has_perm(POLIO_PERFORMANCE_READ_ONLY_PERMISSION.full_name())
-                or request.user.has_perm(POLIO_PERFORMANCE_NON_ADMIN_PERMISSION.full_name())
-                or request.user.has_perm(POLIO_PERFORMANCE_ADMIN_PERMISSION.full_name())
-            )
+            return True
```
quang-le, line 33: @Crebert08  What you did was good. I just changed it to make the endpoints publicly accessible for read-only operations. That will make things easier if/when we need to embed the related pages

plugins/polio/api/country_plan/serializers.py
```diff
@@ -2,24 +2,34 @@
...
 
-from iaso.api.common import UserSerializer
 from iaso.models import OrgUnit
 from plugins.polio.models.performance_dashboard import PerformanceDashboard
 
 
 logger = logging.getLogger(__name__)
 
 
-class OrgUnitNestedSerializer(serializers.ModelSerializer):
```
quang-le, line 13: @Crebert08 This wasn't used

plugins/polio/api/perfomance_dashboard/serializers.py
```diff
@@ -2,24 +2,34 @@
...
+            "country_id",
+            "vaccine",
+            "created_at",
+            "updated_at",
+        ]
 
 
 class PerformanceDashboardListSerializer(serializers.ModelSerializer):
-    created_by = UserSerializer(read_only=True)
-    updated_by = UserSerializer(read_only=True)
```
quang-le, line 22: @Crebert08 it was a good reflex to track the changes made to the data. We already have the `Modification` model and the `AuditMixin` to track those changes and save them in a dedicated table in the DB, so I just refactored your code to make use of the existing tools.

plugins/polio/api/perfomance_dashboard/serializers.py
```diff
@@ -62,15 +70,5 @@ def create(self, validated_data):
         request = self.context.get("request")
         if not request:
             raise serializers.ValidationError("Request context is missing")
-        user = request.user
-        validated_data["created_by"] = user
```
quang-le, line 66: @Crebert08  Not necessary anymore with the `AuditMixin`

plugins/polio/api/perfomance_dashboard/serializers.py
```diff
@@ -62,15 +70,5 @@ def create(self, validated_data):
...
+        validated_data["account"] = request.user.iaso_profile.account
         return super().create(validated_data)
-
-    def update(self, instance, validated_data):
-        request = self.context["request"]
-        if not request:
-            raise serializers.ValidationError("Request context is missing")
-
-        validated_data["updated_by"] = request.user
-        return super().update(instance, validated_data)
```
quang-le, line 76: @Crebert08 same

plugins/polio/api/perfomance_dashboard/views.py
```diff
@@ -56,10 +61,3 @@ def get_serializer_class(self):
             return PerformanceDashboardWriteSerializer
 
         return PerformanceDashboardListSerializer
-
-    def perform_destroy(self, instance):
-        """
-        Perform a soft delete and log the user who performed the action.
-        """
-        instance.updated_by = self.request.user
-        instance.delete()
```
quang-le, line 65: @Crebert08 same

plugins/polio/api/performance_thresholds/serializers.py
```diff
@@ -0,0 +1,61 @@
...
+        is_valid = PerformanceThresholds.is_json_logic_expression(request_data)
+        if not is_valid:
+            raise serializers.ValidationError("Invalid JSON logic for warning threshold")
+        return request_data
+
+    def validate_success_threshold(self, request_data):
+        is_valid = PerformanceThresholds.is_json_logic_expression(request_data)
+        if not is_valid:
+            raise serializers.ValidationError("Invalid JSON logic for success threshold")
+        return request_data
```
quang-le, line 80: @Crebert08 I changed the thresholds on the model to be `JSONField` so we can encode a bit more complex conditions with [JsonLogic](https://jsonlogic.com/).

Since I still want to enforce that those fields are valid JsonLogic, I added the validation here.

@beygorghor @tdethier  The idea behind using JsonLogic is that:
- We already have a front-end component that can handle it
- JsonLogic will be usable as in Python pipelines

plugins/polio/api/performance_thresholds/views.py
```diff
@@ -0,0 +1,59 @@
...
+    """
+    API endpoint for Performance dashboard tresholds.
+
+    This endpoint supports filtering by:
+    - indicator
+
+    The permissions are structured as follows:
+    - **Read-only**: Can only list and retrieve tresholds.
+    - **Non-admin**: Can create and update tresholds.
+    - **Admin**: Can delete tresholds.
```
quang-le: This permission structure was a request in the ticket, but it doesn't make much sense for this feature, at least for now

plugins/polio/js/src/domains/VaccineModule/PerformanceDashboard/modals/DeleteModal.tsx
quang-le, line 1: IDK why this shows up, it's part a the parent (merged) branch

plugins/polio/js/src/domains/VaccineModule/NationalLogisticsPlan/hooks/options.ts
```diff
@@ -20,6 +20,5 @@ export const useStatusOptions = (): DropdownOptions<string>[] => {
     }, [formatMessage]);
 };
 
-export const useVaccineOptions = (): DropdownOptions<string>[] => {
-    return defaultVaccineOptions.filter(option => option.value !== 'mOPV2');
-};
+export const vaccineOptions: DropdownOptions<string>[] =
+    defaultVaccineOptions.filter(option => option.value !== 'mOPV2');
```
quang-le, line 24: @Crebert08 Rule of thumb for hooks:
- if it's not somehow stateful (broadly speaking: doesn't use `useState`, `useEffect`, `useMemo` or another hook that uses one of those 3), it's just a normal function --> don't use the `use` prefix

Here's you can even just export the constant, that will avoid the `filter` operation to be performed each time your component is rendered

plugins/polio/models/country_plan.py
```diff
@@ -18,11 +17,21 @@
...
+
+        if user.is_authenticated:
+            return self.filter(account=user.iaso_profile.account)
+
+        if app_id is not None:  # leaving the possibility to pass app_id if page needs to be embedded
+            try:
+                project = Project.objects.get_for_user_and_app_id(user, app_id)
+                return self.filter(account=project.account)
+            except Project.DoesNotExist:
+                return self.none()
```
quang-le, line 34: @Crebert08 I'm adding a filter on `app_id`, because since the API will be anonymously accessible for read-only, we need a way to filter by account that doesn't rely on the user.
We do this by passing an `app_id` and retrieving the `Project`  associated to it. We then use the account of that project to do the filtering.

plugins/polio/models/country_plan.py
```diff
@@ -74,3 +68,20 @@ class Meta:
...
+                "vaccine": {"type": "string"},
+                "created_at": {"type": "string"},
+                "updated_at": {"type": "string"},
+                "country_name": {"type": "string"},
+                "country_id": {"type": "string"},
+                "status": {"type": "string"},
+                "date": {"type": "string"},
+            },
+            "required": ["id", "vaccine", "country_name", "country_id", "status", "date"],
+        }
```
quang-le, line 87: for testing audit logs

plugins/polio/models/performance_thresholds.py
```diff
@@ -0,0 +1,133 @@
...
+            # this model does not support nested operators
+            if "and" in entry_keys or "or" in entry_keys:
+                found_invalid_data = True
+            else:
+                found_invalid_data = (
+                    len([rule for rule in entry if not PerformanceThresholds.is_json_logic_rule(rule)]) > 0
+                )
+            if found_invalid_data:
+                break
+        return not found_invalid_data
```
quang-le: I put these in the model because they are relevant to the type of data that we will store

plugins/polio/migrations/0249_remove_performancedashboard_created_by_and_more.py
```diff
@@ -0,0 +1,20 @@
...
+
+    operations = [
+        migrations.RemoveField(
+            model_name="performancedashboard",
+            name="created_by",
+        ),
+        migrations.RemoveField(
+            model_name="performancedashboard",
+            name="updated_by",
+        ),
```
tdethier: Maybe we can add this to `0248` and remove this migration file?

plugins/polio/models/country_plan.py
```diff
@@ -74,3 +68,20 @@ class Meta:
     # No unique constraint specified in the ticket, so omitting for now.
     def __str__(self):
         return f"{self.country.name} - {self.date} - {self.vaccine}"
+
+    @staticmethod
+    def json_schema():
```
tdethier, line 73: I'm usually not a fan of static methods in classes (I will most likely never go for that option), but in this case it makes sense 👍

plugins/polio/models/country_plan.py
tdethier: I think we should probably have discussion with Martin & the team on this logging thing, at least for agreeing on when we should use it and when we shouldn't bother. It's not clear for me.

hat/audit/models.py
```diff
@@ -142,6 +142,42 @@ def field_diffs(self):
         new_fields = new_value.get("fields") or {}
         return dict_compare(past_fields, new_fields)
 
+    @staticmethod
+    def make_json_schema(past_value_schema, new_value_schema):
```
tdethier, line 146: Do you expect some models to have different schemas for before & after?

plugins/polio/tests/api/country_plan/test_views.py
```diff
@@ -206,23 +188,46 @@ def test_list_returns_only_own_account_dashboards(self):
         self.assertNotIn(
             self.dashboard_3.id,
             result_ids,
-            "Dashboard from another account (Akatsuki) should not be listed for Hokage user",
         )
 
-    def test_create_sets_audit_fields_correctly(self):
+    def test_audit_log_on_save(self):
```
tdethier, line 189: Could you please add checks in this test to make sure that there's a log as well for:
- creation
- deletion

plugins/polio/models/performance_thresholds.py
```diff
@@ -0,0 +1,133 @@
...
+            try:
+                project = Project.objects.get_for_user_and_app_id(user, app_id)
+                return self.filter(account=project.account)
+            except Project.DoesNotExist:
+                return self.none()
+
+
+class PerformanceThresholds(SoftDeletableModel):
+    account = models.ForeignKey(Account, on_delete=models.CASCADE, related_name="performance_thresholds", null=False)
+    indicator = models.CharField(max_length=200, blank=False)
```
tdethier, line 42: Is there a specific list of indicators? I'm wondering if it should be an enum, but I don't know the business logic

plugins/polio/models/performance_thresholds.py
```diff
@@ -0,0 +1,133 @@
...
+        return f"{self.account.name}-{self.indicator}"
+
+    # make method static for easier testing
+    @staticmethod
+    def is_json_logic_rule(data):
+        if not data:
+            return False
+        keys = data.keys()
+        values = list(data.values())
+        operators = ["==", ">", "<", ">=", "<="]
```
tdethier: I would make that list a constant value at the model level or higher

plugins/polio/models/performance_thresholds.py
```diff
@@ -0,0 +1,133 @@
...
+            correct_reference_format = len([value for value in values if (value[0] == {"var": "value"})]) == len(keys)
+
+            correct_argument_format = len(
+                [value for value in values if (value[1] == {"var": "average"} or isinstance(value[1], int))]
+            ) == len(keys)
+
+        if all_keys_valid and correct_array_format and correct_reference_format and correct_argument_format:
+            return True
+
+        return False
```
tdethier:
```suggestion
        for key, value in data.items():
            if key not in operators:  # change operators depending on where they're defined
                return False
            if not isinstance(value, list) or len(value) != 2:
                return False
            if value[0] != {"var": "value"}:
                return False
            if value[1] != {"var": "average"} and not isinstance(value[1], int):
                return False
        return True
```

or if you prefer one single `if`:

```suggestion
        for key, value in data.items():
            if (
                    key not in operators
                    or not isinstance(value, list)
                    or len(value) != 2
                    or value[0] != {"var": "value"}
                    or (value[1] != {"var": "average"}
                    and not isinstance(value[1], int))
            ):
                return False
        return True
```

plugins/polio/models/performance_thresholds.py
tdethier: Since you don't need to return which part is invalid, I'd suggest simplifying this part:

```suggestion
        if not data:
            return False
        for key, value in data.items():
            if key not in ["and", "or"]:
                # Single rule
                return PerformanceThresholds.is_json_logic_rule(data)
            if not isinstance(value, list):
                return False
            for entry in value:
                entry_keys = entry.keys()
                if "and" in entry_keys or "or" in entry_keys:
                    # this model does not support nested operators
                    return False
                if not PerformanceThresholds.is_json_logic_rule(entry):
                    return False
        return True
```

plugins/polio/tests/models/test_performance_tresholds.py
```diff
@@ -0,0 +1,44 @@
+from iaso import models as m
+from iaso.test import TestCase
+from plugins.polio.models.performance_thresholds import PerformanceThresholds
+
+
+class PerformanceTresholdTestCase(TestCase):
```
tdethier: There's a typo in the class + file name 😅

plugins/polio/tests/models/test_performance_thresholds.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def test_is_json_logic_expression(self):
+        correct_expression_and = {"and": [{">": [self.var, self.avg]}, {"<": [self.var, 95]}]}
+        correct_expression_or = {"or": [{">": [self.var, self.avg]}, {"<": [self.var, 95]}]}
+        incorrect_rule = {
+            "or": [{">": [self.rule_bad_operator]}, {"<": [self.var, 95]}]
+        }  # testing only one incorrect child rule since all rules are tested by test_json_logic_rule_validation
+        incorrect_operator = {"xor": [{">": [self.var, self.avg]}, {"<": [self.var, 95]}]}
+        nested_expression = {"and": [{"and": [{">": [self.var, self.avg]}, {"<": [self.var, 95]}]}]}
+
```
tdethier, line 42: I would add some more asserts here:
- passing a single rule (no `and` or `or`)
- passing a value after the operator that is not a list

plugins/polio/api/performance_thresholds/views.py
tdethier: 😇 

```suggestion
    API endpoint for Performance dashboard thresholds.

    This endpoint supports filtering by:
    - indicator

    The permissions are structured as follows:
    - **Read-only**: Can only list and retrieve thresholds.
    - **Non-admin**: Can create and update thresholds.
    - **Admin**: Can delete thresholds.
```

plugins/polio/tests/api/performance_thresholds/common_data.py
```diff
@@ -0,0 +1,110 @@
...
+        cls.user_read_only = cls.create_user_with_profile(
+            username="read only",
+            account=cls.account,
+            permissions=[POLIO_PERFORMANCE_READ_ONLY_PERMISSION],
+        )
+
+        # No Permissions (Read-only)
+        cls.user_no_perms = cls.create_user_with_profile(username="No perms", account=cls.account, permissions=[])
+        # Anonymous user (unauthenticated)
+        cls.anon = AnonymousUser()
```
tdethier, line 60: Did you forget about your `create_base_users()` method? 😁

plugins/polio/tests/api/performance_thresholds/test_views.py
```diff
@@ -0,0 +1,186 @@
...
+
+class PerformanceThresholdsViewsAPITestCase(PerformanceThresholdsAPIBase):
+    """
+    Test cases for the main actions of the Performance Thresholds API endpoint (ViewSet).
+    """
+
+    def test_read_access_is_public(self):
+        """
+        Unauthenticated users should not be able to access the endpoint.
+        """
```
tdethier: I guess it's an old comment
```suggestion
```

plugins/polio/tests/api/performance_thresholds/test_views.py
```diff
@@ -0,0 +1,186 @@
...
+        self.client.force_authenticate(self.user_non_admin)
+
+        threshold = self.threshold_stock_12m
+
+        update_data = {"success_threshold": self.json_logic_expression_1}
+
+        response = self.client.patch(
+            f"{self.PERFORMANCE_THRESHOLDS_API_URL}{threshold.id}/", data=update_data, format="json"
+        )
+        self.assertJSONResponse(response, status.HTTP_200_OK)
```
tdethier, line 104: We should add an assert inside this and play with time machine to show if there's a 7-days limit on that permission here. Since you're using the same permission class as the dashboard API, my guess is that you won't be able to update a threshold 1 month later

plugins/polio/tests/api/performance_thresholds/test_views.py
```diff
@@ -0,0 +1,186 @@
...
+
+        result_ids = [item["id"] for item in results]
+
+        self.assertNotIn(
+            self.threshold_stock_12m_other_account.id,
+            result_ids,
+        )
+        self.assertEqual(result_ids, expected_ids)
+
+    def test_audit_log_on_save(self):
```
tdethier, line 149: I would also check here that you get logs on creation/deletion

tdethier (changes requested):
Overall, nice work 👍 

The changes to `PerformanceDashboard` are fine, I don't mind the switch to `AuditMixin`.

I'm requesting changes because I think there are a couple of things missing in tests, and to simplify the code that validates the JSON logic fields

plugins/polio/models/performance_thresholds.py
quang-le: None was provided. I'm betting on the `unique_together` constraint to ensure there's not too much pollution. Also typos/mistakes will be easy to spot on the DS side and will have any side effects in the app

plugins/polio/models/performance_thresholds.py
quang-le: I have thought about such a solution, but I wanted to return `False` by default

plugins/polio/models/performance_thresholds.py
quang-le: Same, I prefer to return `False` by default when validating, especially with complex logic

plugins/polio/tests/models/test_performance_tresholds.py
quang-le: worst name since campaigns

plugins/polio/tests/api/performance_thresholds/common_data.py
quang-le: I expanded on the existing tests

plugins/polio/tests/api/performance_thresholds/test_views.py
quang-le: Honestly, I'll go back to the client and challenge the need for that permission structure 😅

hat/audit/models.py
quang-le: creation and hard delete maybe

plugins/polio/migrations/0248_countryplan_performancethresholds_and_more.py
```diff
@@ -0,0 +1,140 @@
...
+                    ("iaso_polio_vaccine_supply_chain_write", "Polio Vaccine Supply Chain Write"),
+                    ("iaso_polio_country_plan_read_only", "Polio Country Plan Read Only"),
+                    ("iaso_polio_country_plan_non_admin", "Polio Country Plan Non Admin"),
+                    ("iaso_polio_country_plan_admin", "Polio Country Plan Admin"),
+                ],
+            },
+        ),
+        migrations.DeleteModel(
+            name="PerformanceDashboard",
+        ),
```
tdethier, line 119: This is going to fail on servers that don't know this model yet 😅

tdethier: Oh no, this model was included in `0247`, my bad

tdethier (approved):
LGTM, nice job both of you @Crebert08 @quang-le 👍 

I managed to CRUD `PerformanceThresholds` through the DRF web view.

FYI, it's not directly related to this PR, but the ordering on the national logistics plans does not work for country name:

Image

plugins/polio/migrations/0248_countryplan_performancethresholds_and_more.py
quang-le: you gave me a sweat here

PR #2598, opened 2025-12-05
IA-4646 fix delete accounts command for local hosting

mestachs:

Related JIRA tickets : IA-4646

- [ ] Are there enough tests?

## Doc

## Changes

## How to test

Explain how to test your PR.

## Print screen / video

## Notes

```
fix: empty instance pop up

Refs: IA-3665
```

### DIFF
--- a/iaso/management/commands/delete_accounts.py
+++ b/iaso/management/commands/delete_accounts.py
@@ -1,4 +1,6 @@
 import datetime
+import random
+import time
 import traceback
 
 from collections import defaultdict
@@ -10,7 +12,7 @@ from django.contrib.contenttypes.models import ContentType
 from django.contrib.sessions.models import Session
 from django.core.management.base import BaseCommand
 from django.db import connection
-from django.db.models import Count, Q, TextField
+from django.db.models import Count, Q, Subquery, TextField
 from django.db.models.functions import Cast
 from django_sql_dashboard.models import Dashboard
 
@@ -20,10 +22,12 @@ from iaso.models import (
     BulkCreateUserCsvFile,
     CommentIaso,
     ExportLog,
+    Form,
     OrgUnitType,
     StorageDevice,
     StorageLogEntry,
     StoragePassword,
+    Task,
     TenantUser,
 )
 from iaso.models.base import (
@@ -33,7 +37,6 @@ from iaso.models.base import (
     ExternalCredentials,
     Mapping,
     Profile,
-    Task,
 )
 from iaso.models.device import Device
 from iaso.models.entity import Entity, EntityType
@@ -54,6 +57,51 @@ def fullname(m):
     if m is None:
         return ""
     return m.__module__ + "." + m.__name__
+
+
+def chunked_delete_records(queryset, chunk_size=1000):
+    """
+    Efficiently deletes a large queryset in smaller chunks.
+    """
+    model = queryset.model
+    total_deleted = 0
+    start_time = time.time()
+
+    # Use .iterator() to avoid loading all 1.2M objects into memory
+    pk_list = list(queryset.values_list("pk", flat=True).iterator())
+    total_to_delete = len(pk_list)
+
+    print(f"Starting deletion of {total_to_delete} records...", model)
+
+    # Iterate over primary keys in chunks
+    for i in range(0, total_to_delete, chunk_size):
+        chunk_pks = pk_list[i : i + chunk_size]
+
+        # Use a transaction block for atomicity within the chunk
+        # with transaction.atomic():
+        try:
+            # Use QuerySet.delete() for efficiency within the database
+            # We filter by 'pk__in' the current chunk
+            chunk_queryset = model.objects.filter(pk__in=chunk_pks)
+            deleted_count = chunk_queryset._raw_delete(using=queryset.db)
+            total_deleted += deleted_count
+
+            elapsed_time = time.time() - start_time
+            print(f"Deleted {total_deleted} / {total_to_delete} records so far in {elapsed_time:.2f}s...")
+        except Exception as err:
+            print("failed to _raw_delete, trying with delete", err, repr(err), err.args)
+            try:
+                chunk_queryset = model.objects.filter(pk__in=chunk_pks)
+                deleted_count = chunk_queryset.delete()
+            except Exception as err:
+                print(
+                    f"Failed to delete chunk :  {total_deleted} / {total_to_delete} | {elapsed_time:.2f}s...",
+                    err,
+                    repr(err),
+                )
+
+    end_time = time.time()
+    print(f"Finished deleting all {total_deleted} records in total elapsed time: {end_time - start_time:.2f} seconds.")
 
 
 def dump_counts():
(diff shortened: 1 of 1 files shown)


iaso/management/commands/delete_accounts.py
```diff
@@ -20,10 +22,12 @@
     BulkCreateUserCsvFile,
     CommentIaso,
     ExportLog,
+    Form,
```
tdethier, line 25: It's already imported below

```suggestion
```

iaso/management/commands/delete_accounts.py
```diff
@@ -56,6 +59,51 @@ def fullname(m):
...
+    total_to_delete = len(pk_list)
+
+    print(f"Starting deletion of {total_to_delete} records...", model)
+
+    # Iterate over primary keys in chunks
+    for i in range(0, total_to_delete, chunk_size):
+        chunk_pks = pk_list[i : i + chunk_size]
+
+        # Use a transaction block for atomicity within the chunk
+        # with transaction.atomic():
```
tdethier:
```suggestion
```

iaso/management/commands/delete_accounts.py
```diff
@@ -119,12 +167,12 @@ def delete_account(self, account):
...
-                Instance.objects.filter(form=form)
-                .values("project__name")
-                .annotate(Count("project__name", distinct=True)),
-            )
+            # print(
+            #    form.name,
+            #    Instance.objects.filter(form=form)
+            #    .values("project__name")
+            #    .annotate(Count("project__name", distinct=True)),
+            # )
```
tdethier:
```suggestion
```

iaso/management/commands/delete_accounts.py
```diff
@@ -164,6 +217,7 @@ def delete_account(self, account):
                                 instance__in=entity_type.reference_form.instances.all()
                             ).delete(),
                         )
+                        chunked_delete_records(entity_type.reference_form.instances.all())
                         print(
                             "reference form : delete instances witht reference form",
                             entity_type.reference_form.instances.all().delete(),
```
tdethier: If you use `chunked_delete_records`, then the print below is useless because all instances have already been deleted, right?

iaso/management/commands/delete_accounts.py
```diff
@@ -180,7 +234,11 @@ def delete_account(self, account):
                     "InstanceFile delete",
                     InstanceFile.objects.filter(instance__in=Instance.objects.filter(project=project)).delete(),
                 )
-                print("Instance delete", Instance.objects.filter(project=project).delete())
+                try:
+                    chunked_delete_records(Instance.objects.filter(project=project))
+                except Exception as e:
+                    print("Instance delete", Instance.objects.filter(project=project).delete())
+                    raise e
```
tdethier: Why do you do another `.delete()` if it's already called in `chunked_delete_records` inside the `try` `catch`?

iaso/management/commands/delete_accounts.py
```diff
@@ -219,11 +277,12 @@ def delete_account(self, account):
                         .all()
                         .delete()
                     )
-
                     print(
                         InstanceFile.objects.filter(instance__org_unit__version__in=data_source.versions.all()).delete()
                     )
-                    print(Instance.objects.filter(org_unit__version__in=data_source.versions.all()).delete())
+                    # print(Instance.objects.filter(org_unit__version__in=data_source.versions.all())._raw_delete(InstanceFile.objects.db))
```
tdethier:
```suggestion
```

iaso/management/commands/delete_accounts.py
```diff
@@ -382,9 +441,11 @@ def handle(self, *args, **options):
         counts_before = dump_counts()
 
         print("keeping ", account_to_keep.name)
-        accounts = Account.objects.filter(~Q(id=account_id_to_keep)).order_by("id")
+        accounts = Account.objects.filter(~Q(id=account_id_to_keep)).order_by("-id")
         print("all the others")
 
+        random.shuffle(accounts)
```
tdethier, line 445: Not sure I understand why you need to shuffle here

iaso/management/commands/delete_accounts.py
```diff
@@ -447,7 +508,7 @@ def handle(self, *args, **options):
                 print(Instance.objects.filter(org_unit__version__in=data_source.versions.all()).delete())
             except Exception as err:
                 traceback.print_exception(type(err), err, err.__traceback__)
-                print("can't delete account", account, err, traceback.format_exc())
+                print("can't delete account", err, traceback.format_exc())
```
tdethier, line 509:
```suggestion
                # traceback.print_exception(type(err), err, err.__traceback__)
                print("can't delete account", err, traceback.format_exc())
```

tdethier (approved):
I made a couple of suggestions and left some questions. For obvious reasons, I didn't test this command on my setup.

Deleting in chunks is a good idea, but I would have personally used Django's `Paginator` to avoid fetching IDs, then objects, then deleting them.
Example:
```
                paginator = Paginator(queryset, chunk_size)
                for page_num in paginator.page_range:
                    page = paginator.get_page(page_num)
                    page.object_list.delete()
```

By the way, this command is very hard to understand and confusing, we should definitely refactor that at some point

mestachs:
> By the way, this command is very hard to understand and confusing, we should definitely refactor that at some point

I understand but it would take a week to get something you might consider "ok" and re testing that thing on a full db. Time we will probably never take.

A lot of the "last delete" are just weird cases hidden in the history of the iaso versions/data history.

> About paginator

the problem with
```
       paginator = Paginator(queryset, chunk_size)
                for page_num in paginator.page_range:
                    page = paginator.get_page(page_num)
                    page.object_list.delete()
```

Deleting per page but from time to time a record won't due to foreign key constraints.

That's also why I do "_rawdelete" (which won't try to get everything in memory) but fallback to .delete() (that manage some cascades but is much more slower), but sometimes it won't delete at all.

the "page ranges" will decrease over time, and will probably end up with weird situations... 
- trying to re-delete already tried records (ending always with the same records in a single page that we "don't manage to delete" )
- skipping a lot of records because we "got to the next page" but that is now at the first page since the first records have been deleted

Another approach is creating a temp table with the ids to delete and iterate with your paginator. But my code managed to delete the 1.400.000+ submissions... so I won't change it.

Since models references in several ways (orgunits, entities and reference instances, change request just made it a mess) it's just a nightmare to know in which order to proceed.
Look at https://iaso-staging.bluesquare.org/models/ and enable all models.

Since new models are always added, it always end up broken because no one ever think about this command, and we fix this in emergency mode the day before a migration.

If I had to redo it I would try based on model relations and build a graph but not even sure it would work (tree alike structure like teams, orgunits will be problematic I guess with self referencing tables). 

Martin suggested to add account_id on every model, not even sure it would suffice, but would help probably with the "scope" of the queries (less need to find a relation of relation of relation that reach an account) but it won't help with the "order of deletion".

tdethier:
I understand, since we do not really use `models.CASCADE`, deleting stuff is a real mess! That's why I already approved the PR

PR #2599, opened 2025-12-08
POLIO-2045: Rename performance dashboard model

quang-le:
In the backend:

The model PerformanceDashboard should be renamed CountryPlan as it is more generic and better describes the model

In the frontend:

Performance dashboard should be renamed National Logistics Plan as it better fits the business reality

Related JIRA tickets : POLIO-2045

- [ ] Are there enough tests?

## Doc

NA
## Changes
- Backend:
    - rename `PerformanceDashboard` model to `CountryPlan`
    - rename the API endpoint similarly
    - rename the permissions
    - rename view and serializer
    - rename permissions

- Frontend
   - Rename `PerformanceDashboard` componenst to `NationalLogisticsPlan` (table, filters etc)
   - rename API hooks, query keys, update API endpoint URL
   - update translations and translation keys
   - update route paths and menu entry
   - rename permissions
   - replace obsolte delete modal

## How to test

Go to Vaccine Module > National Logistics Plan

- check that all CRUD + filtering still works
- check that no mention of performance dashboard remains

## Print screen / video

Screenshot 2025-12-08 at 12 52 55
Screenshot 2025-12-08 at 12 53 08
Screenshot 2025-12-08 at 12 53 13

## Note

merges on POLIO-2034


plugins/polio/api/performance_thresholds/views.py
```diff
@@ -15,7 +15,7 @@
 
 class PerformanceThresholdsViewSet(AuditMixin, ModelViewSet):
     """
-    API endpoint for Performance dashboard thresholds.
+    API endpoint for Performance tresholds.
```
tdethier: 😁
```suggestion
    API endpoint for Performance thresholds.
```

plugins/polio/js/src/domains/VaccineModule/NationalLogisticsPlan/filters/NationalLogisticsPlanFilters.tsx
```diff
@@ -9,18 +9,18 @@ import { baseUrls } from '../../../../constants/urls';
...
 type Props = { params: any };
 
-export const PerformanceDashboardFilters: FunctionComponent<Props> = ({
+export const NationalLogisticsPlanFilters: FunctionComponent<Props> = ({
     params,
 }) => {
     const { formatMessage } = useSafeIntl();
     const { filters, handleSearch, handleChange, filtersUpdated } =
-        usePerformanceDashboardFilters(params);
+        usenationalLogisticsPlanFilters(params);
```
tdethier: The diff is not great, but I simply switched the `n` to `N` in `usenationalLogisticsPlanFilters`

```suggestion
export const useNationalLogisticsPlanFilters = (params: Partial<UrlParams>) => {
    return useFilterState({ baseUrl: baseUrls.nationalLogisticsPlan, params });
};

type Props = { params: any };

export const NationalLogisticsPlanFilters: FunctionComponent<Props> = ({
params,
}) => {
const { formatMessage } = useSafeIntl();
const { filters, handleSearch, handleChange, filtersUpdated } =
        useNationalLogisticsPlanFilters(params);
```

plugins/polio/models/country_plan.py
```diff
@@ -45,17 +45,17 @@ class Status(models.TextChoices):
     country = models.ForeignKey(
         OrgUnit,
         on_delete=models.PROTECT,
-        related_name="performance_dashboard",
+        related_name="country_plan",
```
tdethier:
```suggestion
        related_name="country_plans",
```

plugins/polio/models/country_plan.py
```diff
@@ -45,17 +45,17 @@ class Status(models.TextChoices):
...
     )
     vaccine = models.CharField(max_length=20, choices=VACCINES)
-    account = models.ForeignKey(Account, on_delete=models.CASCADE, related_name="performance_dashboard", null=False)
+    account = models.ForeignKey(Account, on_delete=models.CASCADE, related_name="country_plan", null=False)
```
tdethier:
```suggestion
    account = models.ForeignKey(Account, on_delete=models.CASCADE, related_name="country_plans", null=False)
```

plugins/polio/permissions.py
```diff
@@ -168,30 +168,30 @@ def full_name(self) -> str:
...
-POLIO_PERFORMANCE_ADMIN_PERMISSION = PolioPermission(
-    codename="iaso_polio_performance_admin",
-    label=_("Polio performance Admin"),
+POLIO_COUNTRY_PLAN_ADMIN_PERMISSION = PolioPermission(
+    codename="iaso_polio_country_plan_admin",
+    label=_("Polio Country Plan Admin"),
     module=MODULE_POLIO_PROJECT,
     ui_group=PERMISSION_GROUP_POLIO,
-    ui_category="iaso_polio_performance_permissions",
+    ui_category="iaso_polio_country_plan_permissions",
```
tdethier, line 194: Have the initial 3 permissions already been deployed somewhere? If so, do we need to make a migration to switch the old permissions to the new ones to users who currently have them?

tdethier (approved):
I forgot to include a comment here: LGTM 👍 

I would definitely rename the reverse FKs like I suggested, because otherwise it looks like there's only one object

plugins/polio/permissions.py
quang-le: The initial feature hasn't been released yet, so we're safe

PR #2603, opened 2025-12-09
WC2-862 force authentification to most endpoints

mestachs:
Iaso has always had a mix of authenticated and non authenticated endpoints, in some environnements we want stricter requirement and clearer http error (not 200 with empty results).

Related JIRA tickets : WC2-862

- [ ] Are there enough tests?

## Doc

## Changes

 - made a test that list all endpoints and test several methods
    - from that list found other endpoints that were not covered by the jira
    - identified some "500" errors for some endpoints that clearly don't support anonymous usage
         - `/api/instances/unlock_lock/	POST`
         - `/api/instances/1/add_lock/	POST`
         - `/api/microplanning/plannings/ | GET`
         - `/api/microplanning/plannings/1/ | GET`
         - `/api/orgunits/treesearch/`
         - `/api/superset/token/`
    - identified some endpoint returning data, even if not authenticated    
    - for the others I've added `AuthenticationEnforcedPermission` that will enforce first authentication before proceeding with other aspects (does the user has access to that specific record)
 - fixed the seed command to allow testing of mobile
    - `-` are not allowed in the app_id 
    - everything was uploaded to "unknown-account" so I added "created_by"
 - some tests 
    - were flacky ex mimetype of parquet files are sometime `application/octet-stream` or `application/vnd.apache.parquet` depending on the platform the tests were running. 
    - were missing so I added some of them

## How to test

- launch seed command or setuper
- add in your `.env` : `AUTHENTICATION_ENFORCED=true`
- use the mobile or test manually

## Print screen / video

## Notes

```
fix: empty instance pop up

Refs: IA-3665
```


beanstalk_worker/views.py
```diff
@@ -27,6 +27,9 @@ def task(request):
 
 @csrf_exempt
 def cron(request):
+    if request.headers.get("x-aws-sqsd-taskname") is None:
+        return HttpResponseBadRequest("missing x-aws-sqsd-taskname header")
+
```
madewulf, line 32: Why was that required? Won't it break things on local hosting?

iaso/api/enketo.py
```diff
@@ -309,6 +309,9 @@ def enketo_form_list(request):
     Implement https://docs.getodk.org/openrosa-form-list/#the-manifest-document
 
     Require a param `formID` which is actually an Instance UUID"""
+    if not request.GET.get("formID"):
+        return HttpResponse("formID is required", status=status.HTTP_400_BAD_REQUEST)
```
madewulf: Why not use HttpResponseBadRequest here?

iaso/api/enketo.py
```diff
@@ -355,6 +358,9 @@ def enketo_form_download(request):
     We insert the instance Id In the form definition so the "Form" is unique per instance.
     """
     uuid = request.GET.get("uuid")
+    if not uuid:
+        return HttpResponse("missing uuid", status=status.HTTP_400_BAD_REQUEST)
+
```
madewulf, line 370: Why not use HttpResponseBadRequest?

madewulf (approved):

hat/settings.py
```diff
@@ -726,6 +726,9 @@ def sentry_error_sampler(_, hint):
     "allauth.account.auth_backends.AuthenticationBackend",
 ]
 
+# stricter mode where no public except white listed one
+AUTHENTICATION_ENFORCED = os.environ.get("AUTHENTICATION_ENFORCED", "false") == "true"
```
tdethier, line 730:
```suggestion
AUTHENTICATION_ENFORCED = os.environ.get("AUTHENTICATION_ENFORCED", "false").lower() == "true"
```

iaso/tests/test_auth_enforcement.py
```diff
@@ -0,0 +1,246 @@
...
+    return resp.status_code != 404
+
+
+# polio endpoints are out of scope
+# and wfp auth urls
+def should_skip(path):
+    return path.startswith(("/api/polio", "/wfp_auth", "/dashboard/polio/")) or path == "/dashboard/home/"
+
+
+class TestAuthEnforcement(TestCase):
```
tdethier, line 171: This file is difficult to read, especially the regex and `yield` parts, but it's a very good idea to have these tests 👍

beanstalk_worker/views.py
mestachs: without the header the enpoint just crash, and this endpoint is only "sqsd" driven
not sure we have crons on local hosting by the way (crons where only polio up to now)

> module, method = request.headers["x-aws-sqsd-taskname"].rsplit(".", 1)

mestachs: Note I suspect that we don't use that thing anymore : https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/using-features-managing-env-tiers.html
we post on `/tasks/launch_task/<module.task_name>/<user>`

tdethier (approved):
Good job Stéphan, it was not a easy one to test. I managed to setup the mobile app, make a new submission, upload it and see on the web app after activating the new feature flag 👍 

I made one suggestion regarding the settings (I lost some time because of this, it might happen again when we setup the value on the right servers)

PR #2606, opened 2025-12-09
IA-4648: Show logs when GPKG packages are imported

bmonjoie:
When importing GPKG, no logs were shown or kept

Related JIRA tickets : IA-4648

## Changes

Adding a new database connection to be able to execute queries outside of the current transaction.

## How to test

Import a GPKG package.

## Print screen / video

image

## Notes

I'm not entirely sure of the impact of having another connection to the database, but this is the cleanest way I could find to execute queries outside of the current transaction.
Also, since we only modify the task logs object in that connection, there should not be any risks of deadlocks.

```
fix: empty instance pop up

Refs: IA-3665
```

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/tasks/components/TaskBaseInfo.tsx
+++ b/hat/assets/js/apps/Iaso/domains/tasks/components/TaskBaseInfo.tsx
@@ -84,6 +84,12 @@ export const TaskBaseInfo: FunctionComponent<Props> = ({ task, size }) => {
                         label={formatMessage(MESSAGES.status)}
                         value={task ? <StatusCell task={task} /> : ''}
                     />
+                    {task.status !== 'RUNNING' && task.status !== 'QUEUED' && (
+                        <Row
+                            label={formatMessage(MESSAGES.result_message)}
+                            value={task.progress_message}
+                        />
+                    )}
                 </TableBody>
             </Table>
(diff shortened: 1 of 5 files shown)


hat/assets/js/apps/Iaso/domains/tasks/components/TaskBaseInfo.tsx
```diff
@@ -84,6 +84,12 @@ export const TaskBaseInfo: FunctionComponent<Props> = ({ task, size }) => {
                         label={formatMessage(MESSAGES.status)}
                         value={task ? <StatusCell task={task} /> : ''}
                     />
+                    {task.status !== 'RUNNING' && task.status !== 'QUEUED' && (
```
tdethier, line 87: No constant values, I'm surprised 😁

tdethier (approved):
LGTM 👍 

The transaction stuff was tricky, nice job!

beygorghor (approved):

hat/assets/js/apps/Iaso/domains/tasks/components/TaskBaseInfo.tsx
bmonjoie: It's already defined in an `enum` in TypeScript and it seems to be the idiomatic way of doing it :shrug:

PR #2615, opened 2025-12-11
POLIO-2047: Added Performance threshold in the django admin panel

Crebert08:

Related JIRA tickets : https://bluesquare.atlassian.net/browse/POLIO-2047

- [ ] Are there enough tests?

## Changes

Added the admin panel to perform CRUD operations for the Performance thresholds module

## How to test

Go to `http://localhost:8081/admin/polio/performancethresholds/`
You can add, modify and Delete

## Print screen / video

```
fix: empty instance pop up

Refs: POLIO-2047
```

### DIFF
--- a/plugins/polio/admin.py
+++ b/plugins/polio/admin.py
@@ -446,7 +446,7 @@ class ChronogramTemplateTaskAdmin(TranslatedFieldAdmin, admin.ModelAdmin):
 
 @admin.register(PerformanceThresholds)
 class PerformanceThresholdsAdmin(admin.ModelAdmin):
-    list_display = ("indicator", "account", "updated_at")
+    list_display = ("indicator", "account", "updated_at", "deleted_at")
     list_filter = ("account",)
     search_fields = ("indicator",)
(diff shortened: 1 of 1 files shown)


quang-le (approved):

plugins/polio/admin.py
```diff
@@ -443,6 +444,18 @@ def save_model(self, request, obj, form, change):
         super().save_model(request, obj, form, change)
 
 
+@admin.register(PerformanceThresholds)
+class PerformanceThresholdsAdmin(admin.ModelAdmin):
```
tdethier, line 448: Right now only non-deleted `PerformanceThresholds` appear on the list because it's using the default manager. 

If we want to revert deleted ones, we should probably use the manager that does include soft deleted items

tdethier: What do you think @quang-le? Is this use case going to happen?

quang-le: Not sure, but it would indeed be handy

PR #2622, opened 2025-12-17
IA-4652: Fix looking up instances using JSONLogic with suffixed questions

bmonjoie:
Due to the mobile application's need to assign a type to some calculated questions, the JSONField `Instance.json` might contain keys with a suffix like this: `__int__`. This collides with DRF's way to declare accessing sub-objects properties or even casting.

Related JIRA tickets : IA-4652

- [X] Are there enough tests?

## Changes

The cleanest solution is to annotate the queryset to give those keys a name not containing `__`. 
After the query has been annotated, we can refer to that name and avoid any collisions.

I have added a method to annotate a queryset and I have changed how we access fields with `__` in their names.

## How to test

Search for an instance (or an entity) through a form which contains a suffixed question.

```
fix: empty instance pop up

Refs: IA-3665
```

### DIFF
--- a/iaso/models/entity.py
+++ b/iaso/models/entity.py
@@ -27,7 +27,7 @@ from django.db.models import Exists, OuterRef, Prefetch, Q, Subquery
 
 from iaso.models import Account, Instance, OrgUnit, Project
 from iaso.models.deduplication import EntityDuplicate, ValidationStatus
-from iaso.utils.jsonlogic import jsonlogic_to_q
+from iaso.utils.jsonlogic import annotate_suffixed_json_fields, jsonlogic_to_q
 from iaso.utils.models.soft_deletable import (
     DefaultSoftDeletableManager,
     IncludeDeletedSoftDeletableManager,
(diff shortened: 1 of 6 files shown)


mestachs:
from my small test, it looks ok.
but from my experience of parquet, you might it problem with key length : https://github.com/duckdb/duckdb-postgres/issues/374#issue-3453836808
```
  ("iaso_instance"."json" -> 'test__demo__int') AS "testuudemouuint"

```

apparently there's some black magic at play... it's working but don't know why ;)

image


OK I think I found out

running the produced sql show warning but execute properly
```
NOTICE:  identifier "Gestionuudesuuprojetsuuetuusuiviuudes_activitesuuinternesuusuruu5uuderniers_mois_2025_4564654654654654" will be truncated to "Gestionuudesuuprojetsuuetuusuiviuudes_activitesuuinternesuusuru"
NOTICE:  identifier "Gestionuudesuuprojetsuuetuusuiviuudes_activitesuuinternesuusuruu5uuderniers_mois_2025_1234567891011123" will be truncated to "Gestionuudesuuprojetsuuetuusuiviuudes_activitesuuinternesuusuru"
```
but in fact the UI is just using the json field not that new column
so nothing bad happens

mestachs (approved):

bmonjoie:
@mestachs 

> but from my experience of parquet, you might it problem with key length

That is a very good point, although, the code replace a character by another one. I.e.: `age__int__` (10 chars) becomes `ageuuintuu` (10 chars)
So we would have had issue with the length of those question names anyway, wouldn't have we?

I thought of naming with indexed aliases like `json_alias1`, `json_alias2`, etc. but it required keeping track of an index and it seemed to add unnecessary complexity.

Although, now that I think about it, I wonder if PostGreSQL would complain if we have 2 times the same annotation, like if we said:
```javascript
{ "and": [
     {"<":  [{"var": "age__int__"}, 6] },  
     {">":  [{"var": "age__int__"}, 2] },  
]}
```
We might have to use an indexed alias to avoid this situation.

beygorghor (approved):

PR #2624, opened 2025-12-18
SNT-170: Add is_utility flag to Metrics

Bewi:
Add a utility flag on metrics

Related JIRA tickets : SNT-170

- [ ] Are there enough tests?

## Doc

N/A

## Changes

Add a utility flag on metrics and return only metrics when it is set as false.
Default value is false.

## Print screen / video

## Notes

```
fix: empty instance pop up

Refs: IA-3665
```


iaso/api/metrics/views.py
```diff
@@ -16,7 +16,7 @@ class MetricTypeViewSet(viewsets.ModelViewSet):
     http_method_names = ["get", "options"]
 
     def get_queryset(self):
-        return MetricType.objects.filter(account=self.request.user.iaso_profile.account)
+        return MetricType.objects.filter(account=self.request.user.iaso_profile.account, is_utility=False)
```
tdethier, line 19: Adding a new test to make sure that `MetricTypes` are filtered would be nice, but I don't think that there are any tests on this model/endpoint.

tdethier (approved):
LGTM 👍

iaso/api/metrics/views.py
Bewi: No indeed, we should add them in the future

PR #2628, opened 2025-12-19
IA-4652: Fix looking up instances using JSONLogic with suffixed questions

beygorghor:
Due to the mobile application's need to assign a type to some calculated questions, the JSONField `Instance.json` might contain keys with a suffix like this: `__int__`. This collides with DRF's way to declare accessing sub-objects properties or even casting.

The cleanest solution is to annotate the queryset to give those keys a name not containing `__`. After the query has been annotated, we can refer to that name and avoid any collisions.

IA-4652

Related JIRA tickets : IA-4652

- [ ] Are there enough tests?

## Doc

## Changes

## How to test

Explain how to test your PR.

## Print screen / video

## Notes

```
fix: empty instance pop up

Refs: IA-3665
```

### DIFF
--- a/iaso/utils/jsonlogic.py
+++ b/iaso/utils/jsonlogic.py
@@ -26,8 +26,7 @@ EXTENDED_OPERATOR_LOOKUPS = {
     **COMPARISON_OPERATOR_LOOKUPS,
     "in": "icontains",
 }
-
-DOUBLE_UNDERSCORE_REPLACEMENTS = "uu"
+JSON_ANNOTATED_FIELD_ALIAS = "json_annotated_field"
 
 # This is used to cast a json value from string into a float.
 # Didn't find a simple way to do that.
@@ -66,9 +65,12 @@ def extract_field_name(op: str, jsonlogic: Dict[str, Any], params: Any) -> str:
             f"Unsupported JsonLogic (unknown operator {op}): {jsonlogic}. Supported operators: {list(EXTENDED_OPERATOR_LOOKUPS.keys())}"
         )
 
-    field_position = 1 if op == "in" else 0
-    field = params[field_position]
+    field = params[0]
     if "var" not in field:
+        if op == "in" and "var" in params[1]:
+            return params[1]["var"]
+
+        field_position = 1 if op == "in" else 0
         raise ValueError(
             f"Unsupported JsonLogic. Argument[{field_position}] must contain a variable for given "
             f"operator : {jsonlogic}"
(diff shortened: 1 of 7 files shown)


beygorghor:
Original PR is [here](https://github.com/BLSQ/iaso/pull/2622)

beygorghor:
Staphan comment: 
from my small test, it looks ok.
but from my experience of parquet, you might it problem with key length : https://github.com/duckdb/duckdb-postgres/issues/374#issue-3453836808
```
  ("iaso_instance"."json" -> 'test__demo__int') AS "testuudemouuint"

```

apparently there's some black magic at play... it's working but don't know why ;)

image


OK I think I found out

running the produced sql show warning but execute properly
```
NOTICE:  identifier "Gestionuudesuuprojetsuuetuusuiviuudes_activitesuuinternesuusuruu5uuderniers_mois_2025_4564654654654654" will be truncated to "Gestionuudesuuprojetsuuetuusuiviuudes_activitesuuinternesuusuru"
NOTICE:  identifier "Gestionuudesuuprojetsuuetuusuiviuudes_activitesuuinternesuusuruu5uuderniers_mois_2025_1234567891011123" will be truncated to "Gestionuudesuuprojetsuuetuusuiviuudes_activitesuuinternesuusuru"
```
but in fact the UI is just using the json field not that new column
so nothing bad happens

beygorghor:
@mestachs 

> but from my experience of parquet, you might it problem with key length

That is a very good point, although, the code replace a character by another one. I.e.: `age__int__` (10 chars) becomes `ageuuintuu` (10 chars)
So we would have had issue with the length of those question names anyway, wouldn't have we?

I thought of naming with indexed aliases like `json_alias1`, `json_alias2`, etc. but it required keeping track of an index and it seemed to add unnecessary complexity.

Although, now that I think about it, I wonder if PostGreSQL would complain if we have 2 times the same annotation, like if we said:
```javascript
{ "and": [
     {"<":  [{"var": "age__int__"}, 6] },  
     {">":  [{"var": "age__int__"}, 2] },  
]}
```
We might have to use an indexed alias to avoid this situation.

iaso/utils/jsonlogic.py
bramj, line 611: I didn't try it, but for this method, it didn't work to do the annotation right after this line? https://github.com/BLSQ/iaso/blob/fix/IA-4652_suffixed_questions_not_working_in_json_logic_filters/iaso/api/entity.py#L227

bramj (commented):
Auwch, this is an annoying one, seems like there's no way around it than the annotate everything like you're doing here.
I didn't properly test, but from the looks of it this looks like it's working.
The `iaso/utils/jsonlogic.py` has become very complex, almost feels like it should be extracted as a package.

edil3ra:
TOKEN=$(curl -s -X POST 'http://localhost:8081/api/token/' \
    -H 'Content-Type: application/json' \
    -d '{"username": "iaso", "password": "iaso"}' | jq -r '.access')
    
  FILTER='{"==": [{"var": "age__int__"}, 25]}'
  curl "http://localhost:8081/api/instances/?jsonContent=$(echo -n "$FILTER" | jq -sRr @uri)" \
    -H "Authorization: Bearer $TOKEN" 

  FILTER='{"and": [{"==": [{"var": "gender"}, "female"]}, {">": [{"var": "age__int__"}, 30]}]}'
  curl "http://localhost:8081/api/instances/?jsonContent=$(echo -n "$FILTER" | jq -sRr @uri)" \
    -H "Authorization: Bearer $TOKEN"

I have tested those endpoints and got the correct response

beygorghor:
@madewulf @mestachs can you have a look on this please ?

mestachs (approved):

mestachs (approved):

PR #2636, opened 2025-12-31
POLIO-2039: lqas map link

quang-le:

Related JIRA tickets : POLIO-2039

- [ ] Are there enough tests?

## Doc

NA

## Changes

- Add switch to toggle between 2 types of LQAS filters
- Add url parameters handle both types of filters  
- Remove backend filter on `on_hold` rounds for rounds api as it could create weird UI bugs
- Fix link from afro map to use right params when redirecting

## How to test
With a polio db with Lqas data:

- Go to lqas (afro) map. Select a country and timeframe that match the DB data. Choose a country and click the button to LQAS in the pop up --> a new tab should open with the "classic" filters active for the campaign+ round
- Go to lqas country map: try using both types of filters and toggling between them. Filter state is reset when toggling.
- Repeat from private window with `/dashboard/polio/embeddedLqasCountry/` and `/dashboard/polio/embeddedLqasCountry/`

## Print screen / video

Screenshot 2025-12-31 at 14 51 58
Screenshot 2025-12-31 at 14 52 31

https://github.com/user-attachments/assets/6dfa1232-c717-4ca5-8516-17ec9772abd1


plugins/polio/js/src/domains/LQAS-IM/LQAS/CountryOverview/Filters/LqasCountryViewFilters.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+type Props = {
+    isFetching: boolean;
+    params: LqasUrlParams;
+    side: Side;
+    isEmbedded: boolean;
+    currentUrl: string;
+};
+
+const emptyParams = {
```
Bewi, line 20: This entire right/left thing feels weird.
I would go for {left: { country, campaign, ...}, right: { country, campaign, ...}.
Then when you build your url you can use it the same way you are doing.

plugins/polio/js/src/domains/LQAS-IM/LQAS/CountryOverview/Filters/LqasCountryViewFilters.tsx
```diff
@@ -0,0 +1,97 @@
...
+    const redirectToReplace = useRedirectToReplace();
+
+    const toggleClassicMode = useCallback(() => {
+        setClassicModeOn(!classicModeOn);
+        const newParams = {
+            ...params,
+            ...emptyParams[side],
+            [`${side}FilterType`]: !classicModeOn ? 'country' : 'date',
+        };
+        console.log('NEW PARAMS', newParams);
```
Bewi: Console.log 👯

plugins/polio/js/src/domains/LQAS-IM/LQAS/CountryOverview/Filters/LqasCountryViewFilters.tsx
Bewi: I'd go even further, this shouldn't be aware if it is right or left.
The container should provide params such as : {country, campaign, ...}. 
And this one should only manipulate that.
Then it is up to the parent to decide what to do with it.
I think it could reduce the complexity here and in other filters to prevent checking `if side === 'right'`

Bewi (changes requested):

plugins/polio/js/src/domains/LQAS-IM/LQAS/CountryOverview/Filters/LqasCountryViewFilters.tsx
quang-le: The thing is: we need to be able to deep link this page, so we need to know which params to use from the url.
Also: we've tried something similar to what you propose before, it ended up being harder to manage

Bewi (approved):
LGTM

PR #2644, opened 2026-01-07
IA-4688: selected sampling results

beygorghor:
User should be able to select a sampling result that will be use for assignments step

Related JIRA tickets : IA-4688

- [ ] Are there enough tests?

## Doc

-

## Changes

- added `selected_sampling_results` field to planning model and api + tests
- update frontend to be able to select a sampling n the table
- disable assignments link if planning is not published, or if using pipeline we need a selected sample

## How to test

Make sure you have sampling generated with samplings, you should be able to select one of those to be use with selected planning

## Print screen / video

Screenshot 2026-01-07 at 15 20 14

```
fix: empty instance pop up

Refs: IA-3665
```


hat/assets/js/apps/Iaso/domains/app/translations/en.json
```diff
@@ -898,7 +898,7 @@
     "iaso.label.viewChangeRequestsForUser": "View change requests for user",
     "iaso.label.viewDetails": "View details",
     "iaso.label.viewOrgUnit": "View OrgUnit",
-    "iaso.label.viewPlanning": "View planning",
+    "iaso.label.viewPlanning": "Assignments",
```
quang-le: I think in this case it would be worth changing the key

hat/assets/js/apps/Iaso/domains/plannings/hooks/useCanAssign.tsx
```diff
@@ -0,0 +1,15 @@
+import { Planning } from '../types';
+
+export const useCanAssign = (planning?: Planning): boolean => {
+    if (!planning) return false;
+
+    const hasPipelines =
+        Array.isArray(planning.pipeline_uuids) &&
+        planning.pipeline_uuids.length > 0;
+    const hasSampling = !!planning.selected_sampling_results;
+    const isPublished = !!planning.published_at;
```
quang-le: Is that not the same as 

```suggestion
    const hasSampling = Boolean(planning.selected_sampling_results);
    const isPublished = Boolean(planning.published_at);
```

which is more legible I think

hat/assets/js/apps/Iaso/domains/plannings/utils.ts
```diff
@@ -0,0 +1,15 @@
...
+    if (!planning) return false;
+
+    const hasPipelines =
+        Array.isArray(planning.pipeline_uuids) &&
+        planning.pipeline_uuids.length > 0;
+    const hasSampling = !!planning.selected_sampling_results;
+    const isPublished = !!planning.published_at;
+
+    if (!isPublished) return false;
+    if (!hasPipelines) return true;
```
quang-le, line 13: I would return this line early, since you don't need to compute the rest if `true`

hat/assets/js/apps/Iaso/domains/plannings/utils.ts
quang-le, line 1: This is not really a hook, it's just a normal function that returns a boolean

hat/assets/js/apps/Iaso/domains/plannings/components/PlanningForm.tsx
```diff
@@ -118,8 +123,12 @@ export const PlanningForm: FunctionComponent<Props> = ({
 
         convertError: convertAPIErrorsToState,
     });
-    const { mutateAsync: deletePlanning } = useDeletePlanning(() => {
-        redirectTo(`/${baseUrls.planning}`);
+    const queryClient = useQueryClient();
+    const { mutateAsync: deletePlanning } = useDeletePlanning({
+        onSuccess: () => {
+            queryClient.invalidateQueries(['planningsList']);
```
quang-le: `useDeletePlanning` already invalidates `['planningsList']`, why do we need to repeat the invalidation here?

hat/assets/js/apps/Iaso/domains/plannings/components/SamplingResults.tsx
```diff
@@ -50,7 +52,10 @@ export const SamplingResults: FunctionComponent<Props> = ({ planning }) => {
         () => flattenHierarchy(orgUnitTypeHierarchy?.sub_unit_types || []),
         [orgUnitTypeHierarchy],
     );
+    const { mutateAsync: savePlanning, isLoading: isSavingPlanningResults } =
+        useSavePlanning('edit', undefined, false);
```
quang-le: I'm not a fan of passing `undefined` but 🤷

quang-le: A good way to avoid that  is to replace the args with an object. The refactoring is pretty straightforward if you want to avoid breaking anything.

Not really blocking tho

hat/assets/js/apps/Iaso/domains/plannings/hooks/requests/useDeletePlanning.ts
```diff
@@ -28,7 +30,7 @@ export const useDeletePlanning = ({
     return useSnackMutation({
         mutationFn: deletePlanning,
         options: {
-            onSuccess,
+            onSuccess: onSuccess || defautOnSuccess,
```
quang-le: We should call both I think, since the `defaultOnSuccess` is there to avoid table bugs

hat/assets/js/apps/Iaso/domains/plannings/config.tsx
```diff
@@ -143,8 +145,24 @@ export const usePlanningColumns = (params: any, count: number): Column[] => {
...
 };
 
-export const useSamplingResultsColumns = (planning: Planning): Column[] => {
+export const useSamplingResultsColumns = (
+    planning: Planning,
+    savePlanning: (data: Partial<SavePlanningQuery>) => void,
+): Column[] => {
     const { formatMessage } = useSafeIntl();
+
+    const handleSelectSamplingResult = useCallback(
```
quang-le, line 154: So everytime the user ticks a box we make an API call?

In the next PR we probably should add some state and a save button otherwise, they'll get a warning pop-up at every click

iaso/api/microplanning/serializers.py
```diff
@@ -76,6 +84,10 @@ class Meta:
     project_details = NestedProjectSerializer(source="project", read_only=True)
     target_org_unit_type_details = NestedOrgUnitTypeSerializer(source="target_org_unit_type", read_only=True)
     pipeline_uuids = serializers.ListField(child=serializers.UUIDField(), required=False, allow_empty=True)
+    selected_sampling_results_id = serializers.PrimaryKeyRelatedField(
+        queryset=PlanningSamplingResult.objects.all(), required=False, allow_null=True, write_only=True
+    )
+    selected_sampling_results = NestedPlanningSamplingResultSerializer(read_only=True)
```
quang-le: I don't understand why you need this. Also it seems weird to me to have a field for both the object and the id of the object

quang-le (changes requested):
Minor things front-end side, but it seems to me the serializer needs to be reworked.

hat/assets/js/apps/Iaso/domains/plannings/utils.ts
beygorghor: no we should be able to make assignments without using pipelines but as to be published at least

hat/assets/js/apps/Iaso/domains/plannings/config.tsx
beygorghor: snack bar is not displayed each time, i'll leave like that for now

quang-le (approved):
Just wait for @tdethier opinion on the backend before merging pls

iaso/models/microplanning.py
```diff
@@ -67,6 +67,13 @@ class Meta:
     created_by = models.ForeignKey(User, on_delete=models.SET_NULL, null=True, blank=True)
     created_at = models.DateTimeField(auto_now_add=True)
     updated_at = models.DateTimeField(auto_now=True)
+    selected_sampling_results = models.ForeignKey(
```
tdethier: I would use the singular form, since there's a FK to only one object

tdethier: Additional question: can a `PlanningSamplingResult` be selected by multiple plannings? It looks like it should be a 1to1 relation, but I'm not sure

iaso/api/microplanning/serializers.py
```diff
@@ -125,11 +137,29 @@ def validate(self, attrs):
...
+        data = super().to_representation(instance)
+        sampling = instance.selected_sampling_results
+        data["selected_sampling_results"] = NestedPlanningSamplingResultSerializer(sampling).data if sampling else None
+        return data
+
+
+class PlanningWriteSerializer(PlanningSerializer):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields.pop("selected_sampling_results", None)
```
tdethier: You do this because you want all fields, except `selected_sampling_results`?

If so, I would probably use `except` with `__all__` instead, see [DRF documentation](https://www.django-rest-framework.org/api-guide/serializers/#specifying-which-fields-to-include)

iaso/api/microplanning/serializers.py
tdethier: Same thing for me, I don't understand why you have 2 attributes. Is it because you use the same serializer for read and write operations?

iaso/api/microplanning/serializers.py
```diff
@@ -125,11 +137,29 @@ def validate(self, attrs):
...
         if validation_errors:
             raise serializers.ValidationError(validation_errors)
 
         return validated_data
 
+    def to_representation(self, instance):
+        data = super().to_representation(instance)
+        sampling = instance.selected_sampling_results
+        data["selected_sampling_results"] = NestedPlanningSamplingResultSerializer(sampling).data if sampling else None
+        return data
```
tdethier: I'm not sure why you need this either

iaso/tests/models/test_microplanning.py
```diff
@@ -195,3 +195,19 @@ def test_target_org_unit_type_protect_on_delete(self):
...
+            pipeline_version="v1",
+            pipeline_name="sample-run",
+            parameters={"foo": "bar"},
+        )
+
+        self.planning.selected_sampling_results = sampling
+        self.planning.save()
+        self.planning.refresh_from_db()
+
+        self.assertEqual(self.planning.selected_sampling_results, sampling)
```
tdethier: That's not really necessary since there's no custom logic, this test is just testing Django's ORM. We can keep it but it doesn't really bring any value

tdethier (changes requested):
Kudos to @quang-le, there's indeed something strange here.

I think the relation type should probably be changed, but the serializer can definitely be simplified!

iaso/api/microplanning/serializers.py
tdethier: The weird double attributes are probably due to the fact that this serializer inherits from the other one. I would make this one independant (inherit from `models.ModelSerializer`) and use `except` and `__all__`, it will  probably simplify everything

tdethier (approved):
Much better now!

There are still a few things that should be done to make sure that the validation is correct (see TODOs in the tests). We'll do that in another PR

PR #2650, opened 2026-01-08
IA-4689: diff-file-displays-empty-code-value-as-modified-for-new-ou

edil3ra:
IA-4689: Treat None and empty string as the same for NameFieldType 
Related JIRA tickets : IA-4689

## Changes

Added is_same() method to overrides  NameFieldType and CodeFieldType

## How to test

- Create an org unit in one source with empty code and name fields
- Create a new org unit in another source with missing name fields
- Run diff comparison between the two versions
- Verify that name and code fields show as "same" in the diff output

### DIFF
--- a/iaso/diffing/comparisons.py
+++ b/iaso/diffing/comparisons.py
@@ -26,12 +26,22 @@ class NameFieldType(FieldType):
             return None
         return org_unit.name
 
+    def is_same(self, value, other_value):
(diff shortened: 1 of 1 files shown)


beygorghor (approved):

beygorghor:
Please make sure to fill up PR template

PR #2656, opened 2026-01-12
IA-4683: Fix back button is not working in planning assignments

bmonjoie:
When you open a planning's assignments' screen, using the back arrow button didn't work.

Related JIRA tickets : IA-4683

## Changes

The problem was the back button was using `goBack` which goes to the previous URL. But, when we open the assignments, the URL is changed to include the filters. So pressing back was just going to the URL without the filters which was changed again to include the filters. Therefore, we could not go back and were stuck in a loop.

I have also fixed the code were some removed properties were called on the `Planning` object.

## How to test

1. Go in the plannings
2. Click on the assignments button of one of them. 
3. Wait for the data to load
4. Click back.
5. Enjoy

```
fix: empty instance pop up

Refs: IA-3665
```


beygorghor (approved):
Thanks for this !

PR #2657, opened 2026-01-12
IA-4572: Improve password reset flow

bmonjoie:
Improve design and flow for the login & reset password

Related JIRA tickets : IA-4572

## Changes

Changed the design and the wording on some screens

## How to test

Go through the login and reset password flow.

## Print screen / video

See ticket's comments.

```
fix: empty instance pop up

Refs: IA-3665
```


iaso/locale/fr/LC_MESSAGES/django.po
```diff
@@ -1037,6 +1029,12 @@ msgstr "Virus détecté"
 msgid "Error"
 msgstr "Erreur lors de l'analyse"
 
+#~ msgid "This query parameter is required."
```
Bewi, line 1032: Was this done auto-magically by the translation generation tool ?
I wonder why it keeps it as comment instead of removing it if it is not there anymore.

hat/templates/iaso/reset_password_confirmation.html
```diff
@@ -3,19 +3,55 @@
 {% load static %}
 {% block body %}
   <div class="auth__container">
-      {% if logo_path %}
-        <header class="auth__header">
-          <h1 class="auth__heading"> <img alt="logo" src="{% static logo_path %}" /></h1>
-        </header>
-      {% endif %}
+  {% include "./language_picker.html" %}
```
Bewi: indentation is missing here :D

Bewi (changes requested):
Ok so, it looks good, small remarks but nothing big.

I would reduce the space between the header and the rest.

Only thing I find weird is some alignment / spacing in some screens: 
image

Also, I would make sure the container box never overflows but becomes scrollable in case screen are very small.

I don't ave access to the FIGMA, however, icons look big compared to text size (from MUI size is something like 20x20).
I would add some spacing between the icon and the text too, (try 8px or 16px).
And the font family should be the same as labels.
image


Go back to login button could be smaller, like 16px.



Line height of text on the bottom is too small, which make the text to narrow and feels like a wall of text :D 
image

iaso/locale/fr/LC_MESSAGES/django.po
bmonjoie: Yes, I have not commented out anything

Bewi (approved):
Thanks for the fine tuning 🙏 

Looks good to me now.

PR #2661, opened 2026-01-13
WC2-868: Allowing access to non staff members to ETL debug view

madewulf:
Allowing access to non staff members but limit access following IASO access restrictions on users.

WC2-868


butofleury (approved):
Tested with a user mono account with access to only 2 org units, I can view a debug page for entitiy belongs to an org unit within the region linked to the user.

Otherwise, It show the following page: 
Page-not-found-at-wfp-debug-848-

PR #2666, opened 2026-01-14
[IA-4552] Add the teams columns in the csv and xlsx export

Crebert08:

Related JIRA tickets : https://bluesquare.atlassian.net/browse/IA-4552

- [x] Are there enough tests?

## Doc

## Changes

In the function based view `list_export` present in `/home/crebert/Desktop/iaso/iaso/api/profiles/profiles.py`, I added a string result to fetch the data for teams related to the profile's user. 
In the `BulkCreateUserFromCsvViewSet` model, I updated the list of columns by adding `Teams` so that the data for teams can appear under this new column.

In the test_profiles.py, I updated  `test_profile_list_export_as_xlsx` and `test_profile_list_export_as_csv` by adding the teams column in the headers. I also added a test to see what happens when a user belongs to multiple teams.

in `test_create_users_from_csv`, I changed `test_upload_valid_csv` to stop using a static file which was causing a mismatch between the tests in the local environment and the github CI tests. I also added a few tests on scenarios where a csv is imported with one team, with multiple teams, with an invalid team and with a team from another account.

in the fixtures, I just added the team column to reflect the real needs of the tests. I also added new fixtures corresponding to the new tests.

I also added a new feature regarding the importation of a user by adding the creation of the team from a csv with a team columns. 

I also changed the sample fixture to include the teams columns

## How to test

Go to the User management page and click on the `CSV` and `XLSX` buttons, the app should open a file csv/xlsx file containing the teams related to profile's user.

## Print screen / video

Screenshot from 2026-01-14 16-19-21

## Notes

```
fix: empty instance pop up

Refs: IA-3665
```


iaso/tests/api/test_profiles.py
```diff
@@ -471,14 +473,15 @@ def test_profile_list_export_as_xlsx(self):
...
+                "teams": {0: self.team1.name, 1: None, 2: self.team2.name, 3: None, 4: None, 5: None, 6: None},
                 "phone_number": {0: None, 1: None, 2: None, 3: None, 4: None, 5: None, 6: None},
                 "editable_org_unit_types": {
                     0: None,
                     1: None,
                     2: None,
                     3: None,
                     4: None,
-                    5: self.sub_unit_type.pk,
+                    5: float(self.sub_unit_type.pk),
```
tdethier: Why do you parse an ID in `float`?

iaso/tests/fixtures/test_user_bulk_create_invalid_mail.csv
```diff
@@ -1,4 +1,4 @@
-username,password,email,first_name,last_name,orgunit,orgunit__source_ref,dhis2_id,organization,profile_language,permissions,user_roles,projects,phone_number,editable_org_unit_types
-broly,yOdnjdln8!,biobroly@bluesquarehub.com,broly,bio,9999,,efsd44,,fr,,,,,
-ferdinand,dEF80s:ghd,fdzepplin@bluesquarehub.com,ferdinand,von Zepplin,"Coruscant Jedi Council, 9999, Tatooine",,73JJ37,,fr,,,,,
-rsfg,frdgEF73f9!,efsfg@srgrscom,feesf,dsfsf,"Tatooine, Dagobah",,N363B8D,,FR,,,,,
+username,password,email,first_name,last_name,orgunit,orgunit__source_ref,dhis2_id,organization,profile_language,permissions,user_roles,projects,phone_number,editable_org_unit_types,teams
```
tdethier: `teams` should be between `projects` and `phone_number`, you should move the column but be careful about the other values when you do that

iaso/tests/fixtures/test_user_bulk_create_no_mail.csv
```diff
@@ -1,4 +1,4 @@
-username,password,email,first_name,last_name,orgunit,orgunit__source_ref,dhis2_id,organization,profile_language,permissions,user_roles,projects,phone_number,editable_org_unit_types
-broly,yOdnjdln8!,biobroly@bluesquarehub.com,broly,bio,9999,,,,fr,,,,,
-ferdinand,dEF80s:ghd,fdzepplin@bluesquarehub.com,ferdinand,von Zepplin,"Coruscant Jedi Council, 9999, Tatooine",,,,fr,,,,,
-rsfg,frdgEF73f9!,,feesf,dsfsf,"Tatooine, Dagobah",,,,FR,,,,,
+username,password,email,first_name,last_name,orgunit,orgunit__source_ref,dhis2_id,organization,profile_language,permissions,user_roles,projects,phone_number,editable_org_unit_types,teams
```
tdethier: Same comment about columns order

iaso/tests/fixtures/test_user_bulk_create_semicolon.csv
```diff
@@ -1,3 +1,3 @@
-username;password;email;first_name;last_name;orgunit;orgunit__source_ref;profile_language;dhis2_id;organization;permissions;user_roles;projects;phone_number;editable_org_unit_types
-broly;yodnj!30dln;biobroly@bluesquarehub.com;broly;bio;9999;;fr;dhis2_id_1;;;;;;
-cyrus;yodnj!30dln;cyruswashington@bluesquarehub.com;cyrus;washington;;;fr;dhis2_id_6;;;;;;
+teams,username,password,email,first_name,last_name,orgunit,orgunit__source_ref,profile_language,dhis2_id,organization,permissions,user_roles,projects,phone_number,editable_org_unit_types,teams
```
tdethier: You replaced all the `;` by `,`, please switch them back.

Also:
- you have 2 `teams` columns
- the `teams` column is not in the right place (like other comments)

iaso/tests/fixtures/test_user_bulk_create_valid.csv
```diff
@@ -1,4 +1,4 @@
-username,password,email,first_name,last_name,orgunit,orgunit__source_ref,profile_language,dhis2_id,organization,permissions,user_roles,projects,phone_number,editable_org_unit_types
-broly,yodnj!30dln,biobroly@bluesquarehub.com,broly,bio,,baz,fr,dhis2_id_1,,,,,,
-ferdinand,!dsgd39hfghd,fdzepplin@bluesquarehub.com,ferdinand,von Zepplin,"Coruscant Jedi Council, 9999, Tatooine",baz,fr,dhis2_id_2,,,,,,
-rsfg,frdghb30cd!hd,efsfg@srgrs.com,feesf,dsfsf,"Tatooine, Dagobah",,FR,dhis2_id_3,,,,,,1
+username,password,email,first_name,last_name,orgunit,orgunit__source_ref,profile_language,dhis2_id,organization,permissions,user_roles,projects,phone_number,editable_org_unit_types,teams
```
tdethier: Same comment about columns order

iaso/tests/fixtures/test_user_bulk_create_valid_with_perm.csv
```diff
@@ -1,4 +1,4 @@
-username,password,email,first_name,last_name,orgunit,orgunit__source_ref,permissions,profile_language,dhis2_id,organization,user_roles,projects,phone_number,editable_org_unit_types
-pollux,yodnj!30dln,biobroly@bluesquarehub.com,broly,bio,9999,,"iaso_forms, iaso_submissions",EN,KBFZK83J,,,,,
-ferdinand,!dsgd39hfghd,fdzepplin@bluesquarehub.com,ferdinand,von Zepplin,"Coruscant Jedi Council, 9999, Tatooine",,,fr,9383,,,,,
-castor,frdghb30cd!hd,efsfg@srgrs.com,feesf,dsfsf,"Tatooine, Dagobah",,iaso_projects,EN,SKNF83NF93,,,,,
+username,password,email,first_name,last_name,orgunit,orgunit__source_ref,permissions,profile_language,dhis2_id,organization,user_roles,projects,phone_number,editable_org_unit_types,teams
```
tdethier: Same comment about columns order

iaso/api/profiles/profiles.py
```diff
@@ -561,6 +561,7 @@ def get_row(profile: Profile, **_) -> List[Any]:
                     for item in profile.user_roles.all().order_by("id")
                 ),
                 ",".join(str(item.name) for item in profile.projects.all().order_by("id")),
+                ",".join(str(item) for item in profile.user.teams.all().order_by("id")),
```
tdethier: Two comments here:
- you're fetching the user's teams and joining them by a `, `, which is good. But you're just outputting the `Team` object without selecting any field. So, this is going to call `Team.__str__`, which basically returns only the team name. Nitpick: I would explicitly use `item.name` here instead of doing this implicitly. 
- if a user is part of multiple teams, this line will cause multiple queries to the DB to fetch all data (N+1), so you should add a `prefetch_related("user__teams")` where you prepare the queryset that is going to be used by this function. Since this has an impact on the number of queries that are run, it could be related to your failing test

iaso/api/profiles/bulk_create_users.py
```diff
@@ -38,6 +38,7 @@
     "permissions",
     "user_roles",
     "projects",
+    "teams",
```
tdethier, line 41: Since this list of fields is used when creating users in bulk, I would fully implement this feature and add the newly created users to the requested teams. This will of course require more unit tests to cover this extra feature

iaso/tests/test_create_users_from_csv.py
```diff
@@ -111,9 +112,24 @@ def setUp(self):
...
+        output = io.StringIO()
+        writer = csv.DictWriter(output, fieldnames=reader[0].keys())
+        writer.writeheader()
+        writer.writerows(reader)
+
+        test_file = SimpleUploadedFile("test.csv", output.getvalue().encode("utf-8"), content_type="text/csv")
+
+        with self.assertNumQueries(85):
+            # with open("iaso/tests/fixtures/test_user_bulk_create_valid.csv") as csv_users:
+            response = self.client.post(f"{BASE_URL}", {"file": test_file}, format="multipart")
```
tdethier, line 127: I don't think this is a good way to fix the issue:
- this is basically recreating a new CSV file with the correct org unit ID
- if the ID is wrong, it should be fixed in the initial CSV instead, through templating and passing variables to the template (see [example here](https://github.com/BLSQ/iaso/blob/f5ec4280cfa58cca7e95cbfa534542088c63c064/iaso/tests/api/source_version/test_source_versions.py#L205)). We can discuss this if you want
- however, I don't think that a wrong ID can cause an issue anyway, since the import code ignores invalid/unknown IDs

Please revert these changes, except for `self.asserNumQueries()`  (which might change again once you add a `prefetch_related`

iaso/tests/api/test_profiles.py
```diff
@@ -368,18 +368,19 @@ def test_profile_list_export_as_csv(self):
...
+        expected_csv += f"{self.jane.iaso_profile.id},janedoe,,,,,,,,,,iaso_forms,,,{self.team1.name},,\r\n"
+        expected_csv += f'{self.john.iaso_profile.id},johndoe,,,,,"{self.org_unit_from_sub_type.pk},{self.org_unit_from_parent_type.pk}",{self.org_unit_from_parent_type.source_ref},,,,,,,,,\r\n'
+        expected_csv += f'{self.jim.iaso_profile.id},jim,,,,,,,,,,"{CORE_FORMS_PERMISSION.codename},{CORE_USERS_ADMIN_PERMISSION.codename}",,,{self.team2.name},,\r\n'
+        expected_csv += f"{self.jam.iaso_profile.id},jam,,,,,,,en,,,{CORE_USERS_MANAGED_PERMISSION.codename},,,,,\r\n"
+        expected_csv += f"{self.jom.iaso_profile.id},jom,,,,,,,fr,,,,,,,,\r\n"
+        expected_csv += f"{self.jum.iaso_profile.id},jum,,,,,,,,,,,,{self.project.name},,,{self.sub_unit_type.pk}\r\n"
+        expected_csv += f'{self.user_managed_geo_limit.iaso_profile.id},managedGeoLimit,,,,,{self.org_unit_from_parent_type.id},{self.org_unit_from_parent_type.source_ref},,,,{CORE_USERS_MANAGED_PERMISSION.codename},"{self.user_role_name},{self.user_role_another_account_name}",,,,\r\n'
```
tdethier, line 383: Could you please add another test, similar to this, but with a user that is part of multiple teams? To check if the result behavior is what we expect where there's more than 1 team

tdethier (changes requested):
It's a good start, but some things should be changed:
- column order in tests
- revert the failing test changes
- add a  `prefetch_related` 
- fully implement the feature in bulk create

Please see my comments below and tell me if you have any question 🙂

iaso/tests/test_create_users_from_csv.py
```diff
@@ -111,9 +112,34 @@ def setUp(self):
...
+        # for row in reader:
+        #     if row["username"] == "rsfg":
+        #         row["editable_org_unit_types"] = str(self.org_unit_type_region.id)
+
+        # output = io.StringIO()
+        # writer = csv.DictWriter(output, fieldnames=reader[0].keys())
+        # writer.writeheader()
+        # writer.writerows(reader)
+
+        # test_file = SimpleUploadedFile("test.csv", output.getvalue().encode("utf-8"), content_type="text/csv")
```
tdethier:
```suggestion
```

iaso/tests/test_create_users_from_csv.py
```diff
@@ -111,9 +112,34 @@ def setUp(self):
...
+        csv_content = self.load_fixture_with_jinja_template(
+            path_to_fixtures="iaso/tests/fixtures", fixture_name="test_user_bulk_create_valid.csv", context=context
+        )
+
+        test_file = SimpleUploadedFile(
+            "test_user_bulk_create_valid.csv", csv_content.encode("utf-8"), content_type="text/csv"
+        )
+
+        with self.assertNumQueries(85):
+            # with open("iaso/tests/fixtures/test_user_bulk_create_valid.csv") as csv_users:
```
tdethier:
```suggestion
```

iaso/tests/test_create_users_from_csv.py
```diff
@@ -273,8 +299,22 @@ def test_password_delete_after_import(self):
 
         pswd_deleted = True
 
-        with open("iaso/tests/fixtures/test_user_bulk_create_valid.csv") as csv_users:
-            response = self.client.post(f"{BASE_URL}", {"file": csv_users}, format="multipart")
+        # with open("iaso/tests/fixtures/test_user_bulk_create_valid.csv") as csv_users:
```
tdethier:
```suggestion
```

iaso/tests/test_create_users_from_csv.py
```diff
@@ -308,9 +348,22 @@ def test_created_users_can_login(self):
         self.client.force_authenticate(self.yoda)
         self.source.projects.set([self.project])
 
-        with open("iaso/tests/fixtures/test_user_bulk_create_valid.csv") as csv_users:
-            response = self.client.post(f"{BASE_URL}", {"file": csv_users}, format="multipart")
+        # with open("iaso/tests/fixtures/test_user_bulk_create_valid.csv") as csv_users:
+        #
+
```
tdethier:
```suggestion
```

iaso/tests/test_create_users_from_csv.py
```diff
@@ -339,8 +392,20 @@ def test_users_profiles_have_right_ou(self):
         self.client.force_authenticate(self.yoda)
         self.source.projects.set([self.project])
 
-        with open("iaso/tests/fixtures/test_user_bulk_create_valid.csv") as csv_users:
-            response = self.client.post(f"{BASE_URL}", {"file": csv_users}, format="multipart")
+        # with open("iaso/tests/fixtures/test_user_bulk_create_valid.csv") as csv_users:
```
tdethier:
```suggestion
```

tdethier (changes requested):
Alright, the tests look better now :+1:  You have 2 extra queries for that test now (from 83 to 85), did you check which queries were the 2 new ones? Are you sure these 2 are related to fetching user teams?

I made some suggestions to clean commented code.

Now, you can also implement importing teams from a csv file. Since you are exporting data, users will now want to reimport that file as well, so we need to handle teams during import.

iaso/tests/api/test_profiles.py
```diff
@@ -484,6 +487,25 @@ def test_profile_list_export_as_xlsx(self):
...
+        self.maxDiff = None
+        multi_user = self.create_user_with_profile(username="multiteam", account=self.account)
+
+        multi_user.teams.set([self.team1, self.team2])
+
+        self.client.force_authenticate(self.jane)
+        response = self.client.get("/api/profiles/?csv=true")
+        self.assertEqual(response.status_code, 200)
+
+        response_csv = response.getvalue().decode("utf-8")
```
tdethier: Nitpick: there's a utility `self.assertCsvFileResponse()` method you could use

iaso/tests/fixtures/test_user_bulk_create_valid_with_one_team.csv
```diff
@@ -0,0 +1,2 @@
+username,password,email,first_name,last_name,orgunit,orgunit__source_ref,profile_language,dhis2_id,organization,permissions,user_roles,projects,teams,phone_number,editable_org_unit_types
+rsfg,frdghb30cd!hd,efsfg@srgrs.com,feesf,dsfsf,"Tatooine, Dagobah",,FR,dhis2_id_3,,,,,{{team_names}},,{{org_unit_type_id}}
```
tdethier: This file has exactly the same content as `iaso/tests/fixtures/test_user_bulk_create_valid_with_multiple_teams.csv`, except that it doesn't have any `"` before and after team names. I don't think it's useful to have 2 files for that.

You should probably delete one of them, and the code that calls this template should be responsible for properly adding quotes if needed

iaso/api/fixtures/sample_bulk_user_creation.csv
```diff
@@ -1,2 +1,3 @@
...
+username,password,email,first_name,last_name,orgunit,orgunit__source_ref,permissions,profile_language,dhis2_id,organization,projects,user_roles,teams,phone_number,editable_org_unit_types
+user name should not contain whitespaces,"Min. 8 characters, should include 1 letter and 1 number",,,,Use Org Unit ID to avoid errors,Org Unit external ID,"Possible values: iaso_forms,iaso_mappings, etc.","Possible values: EN, FR",Optional,Optional,projects,user roles,"Use Team names. Use comma separated values inside quotes for multiple teams: ""Team A, Team B""","The phone number as a string (e.g. +32...)","Use comma separated Org Unit Type IDs: 1, 2"
+johndoe,Password123!,john@example.com,John,Doe,Tatooine,,iaso_forms,EN,dhis2_99,MyOrg,ProjectAlpha,Manager,"Alpha Team, Beta Team",+32400000000,1
```
tdethier: It's a good idea to provide an example, but I'm afraid that users will leave it and import `johndoe` every time they import users 🫤
```suggestion
```

iaso/tests/test_create_users_from_csv.py
```diff
@@ -721,3 +758,107 @@ def test_audit_log_on_save(self):
...
+        test_file = SimpleUploadedFile("test.csv", csv_content.encode("utf-8"), content_type="text/csv")
+
+        response = self.client.post(f"{BASE_URL}", {"file": test_file}, format="multipart")
+
+        self.assertEqual(response.status_code, 400)
+
+        expected_error = f"Row 2: Team '{invalid_team_name}' does not exist."
+        self.assertEqual(response.json()["error"], expected_error)
+
+    def test_bulk_create_user_cannot_assign_team_from_another_account(self):
```
tdethier, line 836: Good idea to test that 👍

iaso/api/fixtures/sample_bulk_user_creation.csv
```diff
@@ -1,2 +1,3 @@
...
+username,password,email,first_name,last_name,orgunit,orgunit__source_ref,permissions,profile_language,dhis2_id,organization,projects,user_roles,teams,phone_number,editable_org_unit_types
+user name should not contain whitespaces,"Min. 8 characters, should include 1 letter and 1 number",,,,Use Org Unit ID to avoid errors,Org Unit external ID,"Possible values: iaso_forms,iaso_mappings, etc.","Possible values: EN, FR",Optional,Optional,projects,user roles,"Use Team names. Use comma separated values inside quotes for multiple teams: ""Team A, Team B""","The phone number as a string (e.g. +32...)","Use comma separated Org Unit Type IDs: 1, 2"
```
tdethier: Quotes are not parsed and do not work. You needed quotes in your test setup because you had to define a variable, but users should not manually put quotes in the file

```suggestion
user name should not contain whitespaces,"Min. 8 characters, should include 1 letter and 1 number",,,,Use Org Unit ID to avoid errors,Org Unit external ID,"Possible values: iaso_forms,iaso_mappings, etc.","Possible values: EN, FR",Optional,Optional,projects,user roles,"Use Team names. Use comma separated values for multiple teams: Team A,Team B","The phone number as a string (e.g. +32...)","Use comma separated Org Unit Type IDs: 1, 2"
```

iaso/tests/test_create_users_from_csv.py
tdethier: @Crebert08 the commented line is still there

tdethier (changes requested):
I tested the feature, I could see teams in the downloaded CSV file and was able to create a user with multiple teams in a CSV bulk import 👍 

Thanks for making the changes here and there. 

I'm requesting changes because the CSV template has to be edited and to do some more cleaning up. Otherwise, we're good to go

tdethier (approved):
LGTM 👍 

Sorry you had extra changes to make that were not initially defined in your ticket

Crebert08:
> LGTM 👍
> 
> Sorry you had extra changes to make that were not initially defined in your ticket

No worry :), it actually helped me understand the codebase better, thanks :)) !

PR #2679, opened 2026-01-19
POLIO-2072: refactor campaign filters

quang-le:

The backend filters of the polio campaigns API need to be refactored before they are updated to include integrated campaigns. This PR does the refactoring.

### Related JIRA tickets

POLIO-2072

## Changes

- Add `/api/polio/v2/calendar` endpoint:
    - Uses new `CampaignCalendarViewSet`  and `CalendarCampaignSerializerV2`
    - Does the prefetching in `get_queryset`
    - Abstracts code from existing search filter, so it can be used in the new `CampaignFilterV2` and the legacy `SearchFilterBackend`
    - Adds filtering based on reference date to mitigate the performance cost of making unpaginated calls

## How to test

With a polio DB:
- Go to `Calendar.tsx`:
    - replace the call to `useGetCampaigns` with :
    ```typescript
    const {
        data: campaigns = [],
        isLoading,
        isFetching,
    } = useGetCampaigns({  queryParams,
       queryOptions: { enabled: isTypeSet },
    }
    );
    ``` 
    - Update the import of `useGetCampaigns`to use the calendar specific version
    - Go to the calendar:
        - Page should load
        - Test all filters    

## Print screen / video

https://github.com/user-attachments/assets/622af07a-7e17-43cc-a273-563fa99cd813

## Notes

- Loading times should be faster and API response size smaller
- Adding features specific to integrated campaigns is part of another PR
- Actually updating the front-end as well.

## Doc
Comments in the code


plugins/polio/api/calendar/serializers.py
```diff
@@ -0,0 +1,147 @@
...
+                    fields = ["id"]
+
+            class Meta:
+                model = SubActivityScope
+                fields = ["group", "vaccine"]
+
+            group = NestedGroupSerializer()
+
+        scopes = NestedScopeSerializer(many=True)
+        round_number = serializers.IntegerField(source="round.number")
```
tdethier, line 39: I know it already was like that, but nesting serializers inside serializers is fugly, especially on multiple levels 😅 

Do you think you could extract all nested ones and put them at the same level as the base one, like we do in the other APIs? I don't know if the base one was thoroughly tested

plugins/polio/api/calendar/views.py
```diff
@@ -0,0 +1,108 @@
...
+                    .prefetch_related("scopes")
+                    .prefetch_related("scopes__group")
+                    .prefetch_related(scopes_group_org_units_prefetch)
+                    .prefetch_related("rounds")
+                    .prefetch_related("rounds__datelogs")
+                    .prefetch_related("rounds__datelogs__modified_by")
+                    .prefetch_related("rounds__scopes")
+                    .prefetch_related("rounds__scopes__group")
+                    .prefetch_related(rounds_scopes_group_org_units_prefetch)
+                )
```
tdethier, line 106: You could gather the base queryset with select/prefetch related queries in a single place at the beginning and then filter out based on account ID on the pre fetched queryset

tdethier (approved):
I made the setup explained above and managed to get calendar filters working 👍 

I made a couple of suggestions to improve the code

plugins/polio/api/calendar/serializers.py
quang-le: I could, but then I would have to rewrite a lot of front-end that consumes that API

PR #2680, opened 2026-01-19
IA-4702: adding color to iaso profile

beygorghor:

We need to add the possibility to save a specific color per user profile

### Related JIRA tickets

IA-4702

## Changes

- new field on `iaso_profile` using colors
- be able to save a color using the existing api
- testing the field and the api using color

## How to test

You need to use postman or equivalent to be able to update the color of a specific user using the api.

## Print screen / video

-


edil3ra (commented):
I’ve tested the new profile color fields. it's working as expected.

edil3ra (approved):

iaso/migrations/0360_profile_color.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 4.2.20 on 2026-01-09 14:02
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("iaso", "0359_planning_selected_sampling_result"),
```
tdethier: Another migration was merged in the meantime, you should probably rebase/merge `develop` on your branch and recreate your migration file to avoid issues

iaso/models/base.py
```diff
@@ -578,6 +579,7 @@ class Profile(models.Model):
     user_roles = models.ManyToManyField("UserRole", related_name="iaso_profile", blank=True)
     projects = models.ManyToManyField("Project", related_name="iaso_profile", blank=True)
     phone_number = PhoneNumberField(blank=True)
+    color = models.CharField(max_length=7, null=True, blank=True, default=DEFAULT_COLOR)
```
tdethier: Django recommends [not having null and blank on a charfield](https://docs.djangoproject.com/en/4.2/ref/models/fields/#django.db.models.Field.null). Since there's already going to be a migration, we might as well do this at the same time

```suggestion
    color = models.CharField(max_length=7, blank=True, default=DEFAULT_COLOR)
```

iaso/api/apps/serializers.py
```diff
@@ -82,7 +82,7 @@ def create(self, validated_data):
         name = validated_data.get(self.NAME, None)
         forms = validated_data.get(self.FORMS, None)
         feature_flags = validated_data.get(self.FEATURE_FLAGS, None)
-        color = validated_data.get(self.COLOR, DEFAULT_PROJECT_COLOR)
+        color = validated_data.get(self.COLOR, DEFAULT_COLOR)
```
tdethier, line 92: You should add a `validate_color` method to your serializer, to make sure that it's in the right format. Since this color attribute appears on many different models (and therefore serializers), we should probably have a generic tool for that at some point

iaso/api/profiles/profiles.py
```diff
@@ -376,6 +377,7 @@ def create(self, request):
             user_roles_data = self.validate_user_roles(request)
             projects = self.validate_projects(request, user.profile)
             editable_org_unit_types = self.validate_editable_org_unit_types(request, user.profile)
+            color = request.data.get("color", None)
```
tdethier: Same comment about validating the `color` value

iaso/api/profiles/profiles.py
```diff
@@ -439,6 +442,7 @@ def partial_update(self, request, pk=None):
             user_roles_data = self.validate_user_roles(request)
             projects = self.validate_projects(request, profile)
             editable_org_unit_types = self.validate_editable_org_unit_types(request, profile)
+            color = request.data.get("color", None)
```
tdethier: Same comment about validating `color`

tdethier (commented):
I guess this will work as is, but I would add validation on all endpoints that expect a `color` value (this will also require some more tests). I added some other suggestions as well

PR #2681, opened 2026-01-20
IA-4717: Create a dedicated endpoint for planning to retrieve org units infos

beygorghor:

We need two new endpoints to retrieve org units infos from a specific planning.

One with all the children and another one with the root org unit info:
- `/api/microplanning/orgunits/children/?planning=PLANNING_ID`
- `/api/microplanning/orgunits/root/?planning=PLANNING_ID`

If a a sampling is selected it should return the sampling result list, if not, we should return all org units of target org unit type selected in the planning

We need to have those fields for now:

- id
- name
- GeoJson
- hasGeoJson
- latitude
- longitude

### Related JIRA tickets

IA-4717

## Changes

- New view `PlanningOrgunitsViewSet`
- new serializer `PlanningOrgUnitSerializer`
- testing it 

## How to test

- make sure you have a planning with target org unit type and a root org unit
- visit /api/microplanning/orgunits/children/?planning=XXXX
- it should return a all the org units of this type under the root org unit 
- for the same planning, run a sampling and select it
- visit same endpoint, it should return only org units from the sampling and the root org unit 
- visit /api/microplanning/orgunits/root/?planning=XXXX
- it should return a the root org unit

## Print screen / video
Screenshot 2026-01-20 at 11 27 08

## Notes

-

## Doc

-


iaso/api/microplanning/views.py
```diff
@@ -36,6 +38,50 @@
 )
 
 
+class PlanningOrgunitsViewSet(AuditMixin, ModelViewSet):
```
tdethier: I would not use a `ModelViewSet` because it's designed to facilitate CRUD actions on a specific model, which you don't really need. Here you only want a list view of objects that are linked to another model. I would just use  a `ViewSet` instead. We've done [something similar in SaaS](https://github.com/BLSQ/saas/blob/8774fb17863e61ec130f136d5b245475657989b7/api/account_creation/views.py#L56)

Switching to `ViewSet` will probably make `serializer_class` and `queryset` useless

iaso/api/microplanning/views.py
```diff
@@ -36,6 +38,50 @@
...
+
+        user = request.user
+        try:
+            planning = (
+                self.queryset.filter_for_user(user)
+                .select_related("org_unit", "target_org_unit_type", "selected_sampling_result__group")
+                .get(pk=planning_id)
+            )
+        except Planning.DoesNotExist:
+            return Response({"planning_id": ["Not found."]}, status=status.HTTP_404_NOT_FOUND)
```
tdethier: If you define a serializer with
`planning = serializers.PrimaryKeyRelatedField`

and use this serializer here, all of this is done for you

iaso/api/microplanning/serializers.py
```diff
@@ -408,3 +409,25 @@ def get_assignments(self, planning: Planning):
...
+        model = OrgUnit
+        fields = ["id", "name", "geo_json", "has_geo_json", "latitude", "longitude"]
+        read_only_fields = ["id", "name", "geo_json", "has_geo_json", "latitude", "longitude"]
+
+    def get_geo_json(self, org_unit: OrgUnit):
+        if not org_unit.simplified_geom:
+            return None
+
+        shape_queryset = OrgUnit.objects.filter(id=org_unit.id)
+        return geojson_queryset(shape_queryset, geometry_field="simplified_geom")
```
tdethier: This means that we are going to fetch the DB (N+1) for each single org unit that is passed to the serializer (= each org unit in the results), which is annoying.

I would instead try to annotate the queryset with the missing values, but I'm not sure how difficult it would be

iaso/api/microplanning/views.py
```diff
@@ -36,6 +38,50 @@
...
+        root_org_unit = planning.org_unit
+        org_units = []
+        sampling = planning.selected_sampling_result
+        if sampling and sampling.group_id:
+            org_units = list(org_units_qs.filter(pk__in=sampling.group.org_units.values_list("pk", flat=True)))
+        elif planning.org_unit and planning.target_org_unit_type:
+            org_units = list(
+                org_units_qs.descendants(planning.org_unit).filter(org_unit_type=planning.target_org_unit_type)
+            )
+
```
tdethier, line 67: I would maybe raise an exception in an `else` block?

tdethier (changes requested):
The feature works fine, for plannings with and without sampling 👍 

- I think we can improve the way we fetch the right planning (through a serializer)
- we can have n+1 queries depending on which data is requested - see comments below

Image

tdethier (approved):
LGTM 👍

PR #2683, opened 2026-01-20
POLIO-2063: integrated campaigns API

quang-le:

We need an API endpoint to return integrated campaigns based on a list of campaign ids

### Related JIRA tickets

POLIO-2063

## Changes

- Add `/api/polio/v2/integratedcampaigns`. takes a `obr_names` query param
- Use same serializer as the regular calendar API, since the data is to be used in the same front end
- Add custom filter backend to handle the filtering
- Delete some dead code in `CampaignViewSet`
- Deduplicate `NumberInFilter` and `CharInFilter`(unrelated to the feature)

## How to test

- Create integrated campaigns via the campaigns page
- Using the django web API, pass campaign obr names to the API and check that the results are correct

## Print screen / video

N/A

## Doc

Added some comments in the code


iaso/api/utils/filters.py
```diff
@@ -0,0 +1,9 @@
+import django_filters
+
+
+class NumberInFilter(django_filters.BaseInFilter, django_filters.NumberFilter):
```
tdethier, line 4: Good idea to refactor this 👍

plugins/polio/api/calendar/views.py
```diff
@@ -28,29 +32,26 @@ class CampaignCalendarViewSet(ModelViewSet):
...
-        "country__name",
-    ]
+    # ordering_fields = [
+    #     "obr_name",
+    #     "cvdpv2_notified_at",
+    #     "detection_status",
+    #     "first_round_started_at",
+    #     "last_round_started_at",
+    #     "country__name",
+    # ]
```
tdethier, line 47:
```suggestion
```

plugins/polio/api/campaigns/filters/filters.py
```diff
@@ -41,7 +41,7 @@ class Meta:
     search = django_filters.CharFilter(method="search_filter", label=_("Search"))
     org_unit_groups = django_filters.CharFilter(method="filter_org_unit_groups", label=_("Country groups"))
     campaign_groups = django_filters.CharFilter(method="filter_campaign_groups", label=_("Campaign groups"))
-    campaign_types = django_filters.CharFilter(method="filter_campaign_types", label=_("Campaign types"))
+    # campaign_types = django_filters.CharFilter(method="filter_campaign_types", label=_("Campaign types"))
```
tdethier, line 44:
```suggestion
```

plugins/polio/tests/api/calendar/integrated_campaigns/test_filter.py
```diff
@@ -0,0 +1,141 @@
...
+        queryset = Campaign.objects.all()
+        filtered_queryset = self.filter_backend.filter_queryset(request, queryset, view=None)
+        self.assertEqual(filtered_queryset.count(), 0)
+
+    def test_filter_by_obr_names(self):
+        request = self._request(
+            query_params={"obr_names": f"{self.campaign.obr_name},{self.campaign2.obr_name},{self.campaign3.obr_name}"}
+        )
+        queryset = Campaign.objects.all()
+        filtered_queryset = self.filter_backend.filter_queryset(request, queryset, view=None)
```
tdethier, line 112: Now I understand what you were saying the other day with `APIRequestFactory`.

Honestly, I think it's much harder to read than simply calling the endpoint. If the performance increase is really important we could do that, otherwise I'm not convinced. But you don't need to change these tests, it's good to have a few examples like that to compare.

plugins/polio/tests/api/calendar/integrated_campaigns/test_integrated_campaigns.py
```diff
@@ -0,0 +1,102 @@
...
+        )
+        self.assertJSONResponse(response, 200)
+
+    def test_logged_user_has_access(self):
+        self.client.force_authenticate(self.user)
+        response = self.client.get(
+            URL,
+            format="json",
+        )
+        self.assertJSONResponse(response, 200)
```
tdethier, line 102: I would maybe add asserts on both to show that they see the same results

tdethier (approved):
Overall, nice job 👍 

I made some suggestions and comments below. 

The only weird thing I noticed is that visiting the endpoint without passing any parameters returned an empty list, even though I had an least 1 integrated campaign:

Image

I was expecting this call to return all the integrated campaigns.


But if I start filtering, I get some result:

Image

plugins/polio/api/campaigns/filters/filters.py
quang-le: I actually need to uncomment that. I do it in the next PR

plugins/polio/api/calendar/views.py
quang-le: I left that so I know which filters to reimplement in the front-end

plugins/polio/tests/api/calendar/integrated_campaigns/test_filter.py
quang-le: Yeah, I'm on the fence too, but for reused filters like `DeleteFilterBackend` it can prevent multiple testing

PR #2685, opened 2026-01-21
IA-4718: [LifeNet] Submissions not reaching IASO web after successful submission on mobile app

edil3ra:

fix: round accuracy value during instance import

### Related JIRA tickets

IA-4718

## Changes
iaso/api/instances/serializers.py: Added InstanceImportAccuracySerializer with RoundedDecimalField
iaso/api/instances/instances.py: Added decimal validation with InstanceImportAccuracySerializer

## How to test

because it's hard to simulate instance creation I test it with Django shell.
python manage.py shell

```
import time

import requests

from iaso.models import Form, Instance, OrgUnit

RUN_ID = int(time.time())

BASE_URL = "http://localhost:8081"
USERNAME = "iaso"
PASSWORD = "iaso"
APP_ID = "com.bluesquarehub.iaso.lifenet.datatrac"

# Get actual org_unit and form IDs from mock data
org_unit = OrgUnit.objects.filter(name__startswith='REPLACE').first()
form = Form.objects.filter(name__startswith='REPLACE').first()


auth_response = requests.post(
    f"{BASE_URL}/api/token/",
    data={"username": USERNAME, "password": PASSWORD}
)
token = auth_response.json().get("access")
headers = {"Authorization": f"Bearer {token}"}

# 2. Test import with problematic accuracy values
test_cases = [
    {"accuracy": 699.999, "expected": "700.00"},
    {"accuracy": 2299.999, "expected": "2300.00"},
    {"accuracy": 1458111.7, "expected": None},
]

for i, case in enumerate(test_cases):
    payload = [{
        "id": f"test-{RUN_ID}-{i}",
        "name": f"Test Submission {RUN_ID}-{i}",
        "file": f"test-{RUN_ID}-{i}.xml",
        "formId": form.id,
        "accuracy": case["accuracy"],
        "orgUnitId": org_unit.id,
        "period": f"20251{i:02d}"
    }]

    response = requests.post(
        f"{BASE_URL}/api/instances/",
        params={"app_id": APP_ID},
        headers=headers,
        json=payload
    )


# 3. Verify instances via database
print(f"\n--- Verifying created instances (RUN_ID={RUN_ID}) ---")
instances = Instance.objects.filter(uuid__startswith=f"test-{RUN_ID}")
for inst in instances:
    print(f"  Instance {inst.uuid}: accuracy={inst.accuracy}")

```

# you should get
Instance test-1769011186-0: accuracy=700.00
Instance test-1769011186-1: accuracy=2300.00
Instance test-1769011186-2: accuracy=None

### DIFF
--- a/iaso/api/instances/instances.py
+++ b/iaso/api/instances/instances.py
@@ -41,7 +41,7 @@ from iaso.api.common import (
 )
 from iaso.api.instances.instance_filters import get_form_from_instance_filters, parse_instance_filters
 from iaso.api.instances.json import JsonbPathQueryFirst, JsonPathField, RegexpReplace
-from iaso.api.instances.serializers import FileTypeSerializer
+from iaso.api.instances.serializers import FileTypeSerializer, InstanceImportAccuracySerializer
 from iaso.api.org_units import HasCreateOrgUnitPermission
 from iaso.api.permission_checks import AuthenticationEnforcedPermission
 from iaso.api.serializers import OrgUnitSerializer
@@ -1048,7 +1048,19 @@ def import_data(instances, user, app_id):
         instance.project = project
         instance.name = instance_data.get("name", None)
         instance.period = instance_data.get("period", None)
-        instance.accuracy = instance_data.get("accuracy", None)
+        accuracy_raw = instance_data.get("accuracy", None)
+        if accuracy_raw is not None:
+            accuracy_serializer = InstanceImportAccuracySerializer(data={"accuracy": accuracy_raw})
+            if accuracy_serializer.is_valid():
+                instance.accuracy = accuracy_serializer.validated_data.get("accuracy")
+            else:
+                logger.warning(
+                    "Invalid accuracy value %s for instance %s: %s",
+                    accuracy_raw,
+                    uuid,
+                    accuracy_serializer.errors,
+                )
+                instance.accuracy = None
 
         tentative_org_unit_id = instance_data.get("orgUnitId", None)
         if str(tentative_org_unit_id).isdigit():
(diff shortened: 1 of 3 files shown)


iaso/api/instances/instances.py
```diff
@@ -1048,7 +1048,19 @@ def import_data(instances, user, app_id):
...
+            if accuracy_serializer.is_valid():
+                instance.accuracy = accuracy_serializer.validated_data.get("accuracy")
+            else:
+                logger.warning(
+                    "Invalid accuracy value %s for instance %s: %s",
+                    accuracy_raw,
+                    uuid,
+                    accuracy_serializer.errors,
+                )
+                instance.accuracy = None
```
bmonjoie: I guess it's OK to ignore it if it's invalid but I would prefer to have @madewulf 's input on that.

bmonjoie (changes requested):
There are no tests that test this behavior.

Could you add the following test:

1. Create instance with a correct accuracy  and check that the instance has the proper accuracy attached
2. Create an instance with an accuracy that is more precise than 2 digits and check that the instance has a correctly rounded accuracy attached.
3. Create an instance with an invalid accuracy, check that the accuracy is None.

madewulf (approved):

iaso/api/instances/instances.py
madewulf: Looking at the test, this is to handle a string in the accuracy field. If it's a number with too many commas, it will be round up to the closet 2 commas value. If it's a string, the value will be ignored thanks to this. so this error case might not be necessary. Frankly, this is a detail, but all in all, I would remove the check for completely invalid values. We could still let the import fail in these cases.

PR #2687, opened 2026-01-26
POLIO-2067: integrated campaigns calendar

quang-le:

Display integrated campaigns in polio calendar (regular and embedded)

### Related JIRA tickets

POLIO-2067

## Changes

- Update integrated campaigns API to return `first_round_started_at` and enable front-end ordering
- Abstract data fetching, sorting, ordering and deduplicating to `useGetFormattedCalendarData` and simplify `Calendar` code
- Fix ordering bug
- Changed a few files to TS
- Corrected a few linting issues

## How to test

On a polio DB:
- Integrate 3 campaigns to a polio campaign, using the interface (polio page). Make sure at least one set of round dates overlap
- Go to the calendar and navigate it to reach the "parent" polio campaign
- Play with the filters, in particular "show integrated campaigns":
    - if the "type" filter is not empty and does not include "polio": the checkbox should not be visible
    - if you switch from polio/all campaigns to no polio campaigns and back, the checkbox state should reset
    - if you select the polio campaign and tick the box, the integrated campaigns should be shown and grouped with the parent campaign (with a border to highlight them)
    - When changing the order, the polio+integrated campaigns should be treated as one  

- repeat with embeddedCalendar: visit `/dashboard/polio/embeddedCalendar/` while not logged

## Print screen / video

https://github.com/user-attachments/assets/ee1d3832-bfaf-4618-a69a-3fc9b87621b8

## Doc

Comments in the code


plugins/polio/js/src/domains/Calendar/hooks/useMergedCampaigns/utils.tsx
```diff
@@ -0,0 +1,115 @@
...
+        const first: CalendarCampaign = copy.shift() as CalendarCampaign;
+        const last: CalendarCampaign = copy.pop() as CalendarCampaign;
+        return [
+            { ...first, layout: 'top' },
+            ...copy.map((el: CalendarCampaign) => ({
+                ...el,
+                layout: 'middle' as 'top' | 'middle' | 'bottom', // the TS compiler is being weird so we need to cast the type
+            })),
+            { ...last, layout: 'bottom' },
+        ];
```
beygorghor, line 30:
```suggestion
        return [
            { ...first, layout: 'top' },
            ...copy.map(
                (el: CalendarCampaign): CalendarCampaign => ({
                    ...el,
                    layout: 'middle', 
                }),
            ),
            { ...last, layout: 'bottom' },
        ];
```
This should fix your TS issue

plugins/polio/js/src/domains/Calendar/campaignCalendar/utils/index.tsx
```diff
@@ -223,3 +225,67 @@ export const addRemainingEmptyCells = ({
...
+                return {
+                    borderBottom: borderStyle,
+                    borderRight: borderStyle,
+                };
+            default:
+                return {};
+        }
+    }
+    return {};
+};
```
beygorghor, line 253:
```suggestion
export const getStaticFieldCellStyle = (
    fieldKey: string,
    campaign: MappedCampaign,
): SxProps => {
    const borderStyle = `2px solid ${INTEGRATED_CAMPAIGN_BORDER_COLOR}`;
    const leftFields = ['edit'];
    const rightFields = ['r1StartDate'];
    const topBottomFields = [
        'edit',
        'country',
        'name',
        'campaign_types',
        'r1StartDate',
    ];

    if (!topBottomFields.includes(fieldKey)) return {};

    const style: Record<string, string> = {};
    if (leftFields.includes(fieldKey)) style.borderLeft = borderStyle;
    if (rightFields.includes(fieldKey)) style.borderRight = borderStyle;

    if (campaign.layout === 'top') style.borderTop = borderStyle;
    if (campaign.layout === 'bottom') style.borderBottom = borderStyle;

    return style;
};
```
Maybe something like that ?

plugins/polio/js/src/domains/Calendar/hooks/useGetCampaigns.tsx
beygorghor, line 21: TIL

beygorghor (changes requested):
Very nice job.
Off course we have a edge case where the integrated campaign round(s) are not in the same range as the parent campaign, not sure this will happen for real but it can lead to  inconsistent lay-out. Maybe just remove the border if this is the case...

Image

plugins/polio/js/src/domains/Calendar/campaignCalendar/utils/index.tsx
quang-le: I don't know: it's still a bit verbose, and we lose a bit in legibility I think

quang-le:
> Very nice job. Off course we have a edge case where the integrated campaign round(s) are not in the same range as the parent campaign, not sure this will happen for real but it can lead to inconsistent lay-out. Maybe just remove the border if this is the case...
> Image

This should not be the case, the whole point of integrated campaigns is that they should take place together (there's actually a ticket to add a check on the dates, which is a bit complex).
I think a better solution would be to show the integrated campaign even if the dates are out of display range, but that would be a big change in the calendar codebase I think, not sure the edge case is worth the effort

beygorghor (approved):

PR #2691, opened 2026-01-26
SNT-261 i can create a new metric layer

Bewi:

Add Metric Type (layer) Creation and Deletion

### Related JIRA tickets

SNT-261, SNT-262

## Changes

- Add serializer for metricType write
- Add tests for that serializer
- Add endpoints to allow create/update/delete metric type
- Add tests.

## How to test

Nothing to test in IASO, this is tested through related PR in `snt_malaria`

## Print screen / video

## Notes

snt_malaria PR: https://github.com/BLSQ/snt-malaria/pull/176

### DIFF
--- a/iaso/api/metrics/serializers.py
+++ b/iaso/api/metrics/serializers.py
@@ -32,19 +32,17 @@ class MetricTypeSerializer(serializers.ModelSerializer):
 
 
 class MetricTypeWriteSerializer(serializers.ModelSerializer):
-    code = serializers.CharField(required=True, allow_blank=False)
     name = serializers.CharField(required=True, allow_blank=False)
     category = serializers.CharField(required=True, allow_blank=False)
     description = serializers.CharField(required=False, allow_blank=True)
     units = serializers.CharField(required=False, allow_blank=True)
     unit_symbol = serializers.CharField(required=False, allow_blank=True)
-    origin = serializers.CharField(required=False, allow_blank=True)
-    legend_type = serializers.CharField(required=True, allow_blank=False)
+    origin = serializers.ChoiceField(choices=MetricType.MetricTypeOrigin, required=False, allow_blank=True)
+    legend_type = serializers.ChoiceField(choices=MetricType.LegendType, required=True, allow_blank=False)
 
     class Meta:
         model = MetricType
         fields = [
-            "code",
             "name",
             "category",
             "description",
@@ -54,15 +52,18 @@ class MetricTypeWriteSerializer(serializers.ModelSerializer):
             "origin",
         ]
 
+
+class MetricTypeCreateSerializer(MetricTypeWriteSerializer):
+    code = serializers.CharField(required=True, allow_blank=False)
+
+    class Meta(MetricTypeWriteSerializer.Meta):
+        fields = MetricTypeWriteSerializer.Meta.fields + ["code"]
+
     def validate_code(self, value):
         if any(char.isspace() for char in value):
             raise serializers.ValidationError("Code must not contain whitespace.", code="no_whitespace")
 
         instance = getattr(self, "instance", None)
-
-        if instance is not None:
-            if instance.code != value:
-                raise serializers.ValidationError("codeImmutable", code="code_immutable")
 
         user = self.context["request"].user
         account = user.iaso_profile.account
--- a/iaso/api/metrics/serializers.py
+++ b/iaso/api/metrics/serializers.py
@@ -55,7 +55,11 @@ class MetricTypeWriteSerializer(serializers.ModelSerializer):
         ]
 
     def validate_code(self, value):
+        if any(char.isspace() for char in value):
+            raise serializers.ValidationError("Code must not contain whitespace.", code="no_whitespace")
+
         instance = getattr(self, "instance", None)
+
         if instance is not None:
             if instance.code != value:
                 raise serializers.ValidationError("codeImmutable", code="code_immutable")
--- a/iaso/api/metrics/serializers.py
+++ b/iaso/api/metrics/serializers.py
@@ -1,6 +1,7 @@
 from rest_framework import serializers
 
 from iaso.models import MetricType, MetricValue
+from iaso.utils import legend
 
 
 class MetricTypeSerializer(serializers.ModelSerializer):
@@ -9,6 +10,7 @@ class MetricTypeSerializer(serializers.ModelSerializer):
         fields = [
             "id",
             "account",
+            "code",
             "name",
             "category",
             "description",
@@ -28,6 +30,42 @@ class MetricTypeSerializer(serializers.ModelSerializer):
             "created_at",
             "updated_at",
         ]
+
+
+class MetricTypeWriteSerializer(serializers.ModelSerializer):
+    code = serializers.CharField(required=True, allow_blank=False)
+    name = serializers.CharField(required=True, allow_blank=False)
+    category = serializers.CharField(required=True, allow_blank=False)
+    description = serializers.CharField(required=False, allow_blank=True)
+    units = serializers.CharField(required=False, allow_blank=True)
+    unit_symbol = serializers.CharField(required=False, allow_blank=True)
+    origin = serializers.CharField(required=False, allow_blank=True)
+    legend_type = serializers.CharField(required=True, allow_blank=False)
+
+    class Meta:
+        model = MetricType
+        fields = [
+            "code",
+            "name",
+            "category",
+            "description",
+            "units",
+            "unit_symbol",
+            "legend_type",
+            "origin",
+        ]
+
+    def validate_code(self, value):
+        instance = getattr(self, "instance", None)
+        if instance is not None:
+            if instance.code != value:
+                raise serializers.ValidationError("codeImmutable", code="code_immutable")
+
+        user = self.context["request"].user
+        account = user.iaso_profile.account
+        if MetricType.objects.filter(account=account, code=value).exclude(pk=getattr(instance, "pk", None)).exists():
+            raise serializers.ValidationError("uniqueCode", code="unique_code")
+        return value
 
 
 class MetricValueSerializer(serializers.ModelSerializer):
--- a/iaso/api/metrics/serializers.py
+++ b/iaso/api/metrics/serializers.py
@@ -1,7 +1,6 @@
 from rest_framework import serializers
 
 from iaso.models import MetricType, MetricValue
-from iaso.utils import legend
 
 
 class MetricTypeSerializer(serializers.ModelSerializer):
--- a/iaso/api/metrics/views.py
+++ b/iaso/api/metrics/views.py
@@ -9,7 +9,13 @@ from iaso.api.metrics.filters import ValueAndTypeFilterBackend, ValueFilterBacke
 from iaso.models import MetricType, MetricValue
 from iaso.utils import legend
 
-from .serializers import MetricTypeSerializer, MetricTypeWriteSerializer, MetricValueSerializer, OrgUnitIdSerializer
+from .serializers import (
+    MetricTypeCreateSerializer,
+    MetricTypeSerializer,
+    MetricTypeWriteSerializer,
+    MetricValueSerializer,
+    OrgUnitIdSerializer,
+)
 
 
 # TODO for both viewsets: permission_classes
@@ -18,16 +24,18 @@ from .serializers import MetricTypeSerializer, MetricTypeWriteSerializer, Metric
 class MetricTypeViewSet(viewsets.ModelViewSet):
     serializer_class = MetricTypeSerializer
     ordering_fields = ["id", "name"]
-    http_method_names = ["get", "options", "post", "put", "delete"]
+    http_method_names = ["get", "options", "post", "patch", "delete"]
 
     def get_queryset(self):
         return MetricType.objects.filter(account=self.request.user.iaso_profile.account, is_utility=False)
 
     def get_serializer_class(self):
         return (
-            MetricTypeSerializer
-            if self.action in ["list", "retrieve", "grouped_per_category", "legend_types"]
+            MetricTypeCreateSerializer
+            if self.action == "create"
             else MetricTypeWriteSerializer
+            if self.action in ["update", "partial_update"]
+            else MetricTypeSerializer
         )
 
     def perform_create(self, serializer):
--- a/iaso/api/metrics/views.py
+++ b/iaso/api/metrics/views.py
@@ -3,6 +3,7 @@ from rest_framework import status, viewsets
 from rest_framework.decorators import action
 from rest_framework.response import Response
 
+from iaso.api.apps import serializers
 from iaso.api.common import DropdownOptionsWithRepresentationSerializer
 from iaso.api.metrics.filters import ValueAndTypeFilterBackend, ValueFilterBackend
 from iaso.models import MetricType, MetricValue
@@ -38,11 +39,7 @@ class MetricTypeViewSet(viewsets.ModelViewSet):
 
     def perform_destroy(self, instance):
         if instance.origin == MetricType.MetricTypeOrigin.OPENHEXA.value:
-            return Response(
-                {"detail": "Metric types from OpenHexa cannot be deleted."},
-                status=status.HTTP_403_FORBIDDEN,
-            )
-
+            raise serializers.ValidationError("Cannot delete OpenHexa metric types")
         return super().perform_destroy(instance)
 
     @action(detail=False, methods=["get"])
@@ -76,6 +73,7 @@ class MetricValueViewSet(viewsets.ModelViewSet):
     queryset = MetricValue.objects.all()
     filter_backends = [DjangoFilterBackend, ValueFilterBackend]
     filterset_fields = ["metric_type_id", "org_unit_id"]
+    http_method_names = ["get", "options"]
 
     def get_queryset(self):
         return MetricValue.objects.filter(metric_type__account=self.request.user.iaso_profile.account)
@@ -91,6 +89,7 @@ class MetricOrgUnitsViewSet(viewsets.ModelViewSet):
     queryset = MetricValue.objects.all()
     filter_backends = [DjangoFilterBackend, ValueAndTypeFilterBackend]
     filterset_fields = ["metric_type_id"]
+    http_method_names = ["get", "options"]
 
     def get_queryset(self):
         return MetricValue.objects.filter(metric_type__account=self.request.user.iaso_profile.account)
--- a/iaso/api/metrics/views.py
+++ b/iaso/api/metrics/views.py
@@ -1,12 +1,14 @@
 from django_filters.rest_framework import DjangoFilterBackend
-from rest_framework import viewsets
+from rest_framework import status, viewsets
 from rest_framework.decorators import action
 from rest_framework.response import Response
 
+from iaso.api.common import DropdownOptionsWithRepresentationSerializer
 from iaso.api.metrics.filters import ValueAndTypeFilterBackend, ValueFilterBackend
 from iaso.models import MetricType, MetricValue
+from iaso.utils import legend
 
-from .serializers import MetricTypeSerializer, MetricValueSerializer, OrgUnitIdSerializer
+from .serializers import MetricTypeSerializer, MetricTypeWriteSerializer, MetricValueSerializer, OrgUnitIdSerializer
 
 
 # TODO for both viewsets: permission_classes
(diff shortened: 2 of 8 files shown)


iaso/api/metrics/serializers.py
```diff
@@ -30,6 +31,46 @@ class Meta:
...
+
+    def validate_code(self, value):
+        if any(char.isspace() for char in value):
+            raise serializers.ValidationError("Code must not contain whitespace.", code="no_whitespace")
+
+        instance = getattr(self, "instance", None)
+
+        if instance is not None:
+            if instance.code != value:
+                raise serializers.ValidationError("codeImmutable", code="code_immutable")
```
tdethier: This block is for checking if you are updating an existing `MetricType`, but the block below is for making sure you are not creating another one with the same code?

iaso/api/metrics/serializers.py
```diff
@@ -30,6 +31,46 @@ class Meta:
...
 
 
+class MetricTypeWriteSerializer(serializers.ModelSerializer):
+    code = serializers.CharField(required=True, allow_blank=False)
+    name = serializers.CharField(required=True, allow_blank=False)
+    category = serializers.CharField(required=True, allow_blank=False)
+    description = serializers.CharField(required=False, allow_blank=True)
+    units = serializers.CharField(required=False, allow_blank=True)
+    unit_symbol = serializers.CharField(required=False, allow_blank=True)
+    origin = serializers.CharField(required=False, allow_blank=True)
```
tdethier: You'll probably want to use a `ChoiceField` here since this is an enum, it will handle validation for you. See [example here](https://github.com/BLSQ/iaso/blob/c7aed250e9b5ac9213f31f0dd6a27e66ff9c84e0/iaso/api/data_source_versions_synchronization/serializers.py#L113)

iaso/api/metrics/serializers.py
```diff
@@ -30,6 +31,46 @@ class Meta:
...
+    legend_type = serializers.CharField(required=True, allow_blank=False)
```
tdethier: Same comment about `ChoiceField`

iaso/api/metrics/serializers.py
```diff
@@ -30,6 +31,46 @@ class Meta:
...
+        instance = getattr(self, "instance", None)
+
+        if instance is not None:
+            if instance.code != value:
+                raise serializers.ValidationError("codeImmutable", code="code_immutable")
+
+        user = self.context["request"].user
+        account = user.iaso_profile.account
+        if MetricType.objects.filter(account=account, code=value).exclude(pk=getattr(instance, "pk", None)).exists():
+            raise serializers.ValidationError("uniqueCode", code="unique_code")
```
tdethier, line 71: I think I would set a unique constraint on the model and let the model handle this for you, except if you need to have an error with a specific code for the frontend

iaso/api/metrics/views.py
```diff
@@ -15,11 +18,30 @@
 class MetricTypeViewSet(viewsets.ModelViewSet):
     serializer_class = MetricTypeSerializer
     ordering_fields = ["id", "name"]
-    http_method_names = ["get", "options"]
+    http_method_names = ["get", "options", "post", "put", "delete"]
```
tdethier: I would probably use `patch` instead of `put` but I don't think it matters a lot

iaso/api/metrics/views.py
```diff
@@ -15,11 +18,30 @@
...
 
     def get_queryset(self):
         return MetricType.objects.filter(account=self.request.user.iaso_profile.account, is_utility=False)
 
+    def get_serializer_class(self):
+        return (
+            MetricTypeSerializer
+            if self.action in ["list", "retrieve", "grouped_per_category", "legend_types"]
+            else MetricTypeWriteSerializer
```
tdethier: Nitpick: I would probably default with the read serializer unless you are in specific write operations.

iaso/tests/api/metrics/test_serializers.py
```diff
@@ -32,6 +40,76 @@ def test_fields(self):
...
+            "unit_symbol",
+            "legend_type",
+            "origin",
+        }
+        self.assertEqual(set(serializer.Meta.fields), expected_fields)
+
+    def test_validate_code_whitespace(self):
+        serializer = MetricTypeWriteSerializer()
+        with self.assertRaisesMessage(Exception, "Code must not contain whitespace."):
+            serializer.validate_code("invalid code")
```
tdethier: This test is weird because you're not passing any code 😅 I would expect an error about the code being missing instead.

You should pass a payload with a space in `code` and check if the serializer is valid

iaso/tests/api/metrics/test_serializers.py
```diff
@@ -32,6 +40,76 @@ def test_fields(self):
...
+        metric_type = MetricType(code="same_code")
+        account = Account.objects.create(name="Account")
+        user = self.create_user_with_profile(
+            username="jane_doe",
+            last_name="Doe",
+            first_name="Jane",
+            account=account,
+            permissions=[CORE_ORG_UNITS_PERMISSION],
+        )
+        serializer_context = {"request": type("Request", (), {"user": user})()}
```
tdethier: It's weird, why are you calling `type()` here? 

Here's an [example of setting up a fake request](https://github.com/BLSQ/iaso/blob/c7aed250e9b5ac9213f31f0dd6a27e66ff9c84e0/plugins/polio/tests/api/chronogram/test_chronogram_serializers.py#L303) for the serializer

iaso/tests/api/metrics/test_serializers.py
```diff
@@ -32,6 +40,76 @@ def test_fields(self):
...
+        serializer = MetricTypeWriteSerializer(
+            instance=metric_type, data={"code": "same_code_modified"}, context=serializer_context
+        )
+        is_valid = serializer.is_valid()
+        self.assertFalse(is_valid)
```
tdethier:
```suggestion
        self.assertFalse(serializer.is_valid())
```

iaso/tests/api/metrics/test_serializers.py
```diff
@@ -32,6 +40,76 @@ def test_fields(self):
...
+            "name": "New Metric Type",
+            "category": "New Category",
+            "legend_type": "Threshold",
+            "units": "units",
+            "unit_symbol": "u",
+            "description": "A new metric type",
+            "origin": "CUSTOM",
+        }
+
+        serializer_context = {"request": type("Request", (), {"user": user})()}
```
tdethier: Same comment about setting up a fake request

iaso/tests/api/metrics/test_serializers.py
```diff
@@ -32,6 +40,76 @@ def test_fields(self):
...
+        serializer = MetricTypeWriteSerializer(data=data, context=serializer_context)
+        is_valid = serializer.is_valid()
+        self.assertTrue(is_valid)
```
tdethier:
```suggestion
        self.assertTrue(serializer.is_valid())
```

iaso/tests/api/metrics/test_serializers.py
```diff
@@ -32,6 +40,76 @@ def test_fields(self):
...
+            "units",
+            "unit_symbol",
+            "legend_type",
+            "origin",
+        }
+        self.assertEqual(set(serializer.Meta.fields), expected_fields)
+
+    def test_validate_code_whitespace(self):
+        serializer = MetricTypeWriteSerializer()
+        with self.assertRaisesMessage(Exception, "Code must not contain whitespace."):
```
tdethier:
```suggestion
        with self.assertRaisesMessage(ValidationError, "Code must not contain whitespace."):
```

iaso/tests/api/metrics/test_views.py
```diff
@@ -51,41 +50,104 @@ def test_metric_type_list(self):
...
         )
-        self.assertEqual(response.status_code, status.HTTP_405_METHOD_NOT_ALLOWED)  # Method Not Allowed
+        self.assertJSONResponse(response, status.HTTP_200_OK)
+
+        # Verify the update
+        updated_metric_type = MetricType.objects.get(id=self.metric_type_1.id)
+        self.assertEqual(updated_metric_type.name, "Updated Metric Type 1")
+        self.assertEqual(updated_metric_type.description, "Updated description for Metric Type 1")
+
+    def test_metric_type_update_unauthenticated(self):
```
tdethier: I would add another test where you update a `MetricType` and try to change its code, to make sure that you get the error

iaso/tests/api/metrics/test_views.py
```diff
@@ -51,41 +50,104 @@ def test_metric_type_list(self):
...
                 "description": "Description for Metric Type 3",
-                "origin": MetricType.MetricTypeOrigin.CUSTOM,
+                "category": "Category C",
+                "origin": MetricType.MetricTypeOrigin.CUSTOM.value,
+                "scale": "[1, 2, 3]",
+                "legend_type": MetricType.LegendType.THRESHOLD.value,
             },
         )
-        self.assertEqual(response.status_code, status.HTTP_405_METHOD_NOT_ALLOWED)  # Method Not Allowed
+        self.assertJSONResponse(response, status.HTTP_201_CREATED)
```
tdethier, line 71: I would add asserts to make sure that the view added the other 2 fields. Fetch the `MetricType` from DB and assert on:
- `account`
- `legend_config`

iaso/api/metrics/serializers.py
tdethier: This looks weird to me. I think it's because you are using the same serializer for both CREATE and PATCH/PUT, so you have the same fields. But `code` shouldn't be in your PATCH/PUT serializer.
I would maybe prepare a common base serializer without `code`. Then, 2 serializers that inherit from base for create/update, and the create one has an extra code field and a `validate_code` method to check for spaces

iaso/utils/legend.py
```diff
@@ -0,0 +1,214 @@
...
+
+ORDINAL = {
+    2: [RISK_LOW, RISK_VERY_HIGH],
+    3: [RISK_LOW, RISK_MEDIUM, RISK_VERY_HIGH],
+    4: [RISK_LOW, RISK_MEDIUM, RISK_HIGH, RISK_VERY_HIGH],
+}
+
+
+def get_legend_config(metric_type, scale):
+    print(f"Getting legend config for metric type {metric_type} with scale {scale}")
```
tdethier, line 84:
```suggestion
```

tdethier: Or use a logger if this should be kept 😅

iaso/utils/legend.py
```diff
@@ -0,0 +1,214 @@
...
+    # Temporary: use old way as fallback if legend_type was not defined
+    if metric_type.legend_type is None or metric_type.legend_type == "":
```
tdethier:
```suggestion
    if not metric_type.legend_type:
```

iaso/utils/legend.py
```diff
@@ -0,0 +1,214 @@
...
+        return __get_legend_config(metric_type)
+
+    if metric_type.legend_type == "threshold":
+        scales = get_scales_from_list_or_json_str(scale)
+        print(f"Parsed scales: {scales}")
```
tdethier, line 91:
```suggestion
```

iaso/utils/legend.py
```diff
@@ -0,0 +1,214 @@
...
+        try:
+            numeric_scales = [float(s) for s in scales]
+        except Exception as e:
+            print(f"Error converting scales to numerics: {e}")
```
tdethier, line 95:
```suggestion
```

iaso/utils/legend.py
```diff
@@ -0,0 +1,214 @@
...
+    return __get_legend_config(metric_type)
+
+
+def __get_legend_config(metric_type):
+    if metric_type.category == "Incidence":
+        return {
+            "domain": [5, 50, 100, 200, 300, 500],
+            "range": SEVEN_SHADES,
+        }
+    if metric_type.category == "Prevalence":
```
tdethier, line 118: Shouldn't the category be an enum? If not, I would probably still define constants somewhere for all possible values

iaso/utils/legend.py
```diff
@@ -0,0 +1,214 @@
+import json
```
tdethier, line 1: This whole file looks very customized for a specific SNT use case, are you sure it shouldn't go there?

I'm not sure I understand the purpose of this whole thing

iaso/utils/legend.py
```diff
@@ -0,0 +1,214 @@
...
+def get_scales_from_list_or_json_str(scale):
+    if not scale or scale == "":
+        return []
+
+    if isinstance(scale, list):
+        return scale
+
+    if str.startswith(scale, "[") and str.endswith(scale, "]"):
+        strScale = scale.replace("[", "").replace("]", "")
+        scales = [s.strip() for s in str.split(strScale, ",")]
```
tdethier:
```suggestion
        str_scale = scale.replace("[", "").replace("]", "")
        scales = [s.strip() for s in str.split(str_scale, ",")]
```

iaso/utils/legend.py
```diff
@@ -0,0 +1,214 @@
...
+        try:
+            numeric_scales = [float(s) for s in scales]
+        except Exception as e:
+            print(f"Error converting scales to numerics: {e}")
+            numeric_scales = []
+        return {"domain": numeric_scales, "range": get_range_from_count(len(scales))}
+    if metric_type.legend_type == "ordinal":
+        scales = get_scales_from_list_or_json_str(scale)
+        if 4 > len(scales) < 2:
+            print(f"Metric ordinal has to many or to few scales {len(scales)}")
```
tdethier, line 101:
```suggestion
```

iaso/utils/legend.py
```diff
@@ -0,0 +1,214 @@
...
+        print(f"Parsed scales: {scales}")
+        try:
+            numeric_scales = [float(s) for s in scales]
+        except Exception as e:
+            print(f"Error converting scales to numerics: {e}")
+            numeric_scales = []
+        return {"domain": numeric_scales, "range": get_range_from_count(len(scales))}
+    if metric_type.legend_type == "ordinal":
+        scales = get_scales_from_list_or_json_str(scale)
+        if 4 > len(scales) < 2:
```
tdethier:
```suggestion
        if len(scales) == 3:
```

tdethier (changes requested):
I tried testing the feature, but I don't see the layer tab in the settings page 🫤

I'm requesting changes anyway because I had many comments, mostly about cleaning up some code and improving readability/simplification

iaso/api/metrics/serializers.py
Bewi: I need to validate both cases: 
- Unique code
- And that code didn't change when we update it.

I get the point of a second serializer, I'll do that to make it easier to understand.

iaso/api/metrics/views.py
Bewi: I was using a PUT as I am sending the entire model back.
I might change that based on your comment with the serializer.

iaso/tests/api/metrics/test_serializers.py
Bewi: I struggled to have it work that's why :D

iaso/tests/api/metrics/test_views.py
Bewi: Isn't that covered by the serializer tests ?

iaso/api/metrics/serializers.py
Bewi: I'll leave it like this atm, we want a specific message in the dialog in the FE.
Plus the import tool is also using the MetricType and we don't really check on this yet, so it would require some refactoring.

iaso/utils/legend.py
Bewi: At the moment it is a free thing but ultimately it should become a an enum, still tbd

iaso/utils/legend.py
Bewi: We need to start using logger everywhere SNT

iaso/utils/legend.py
Bewi: Won't change now

iaso/utils/legend.py
Bewi: Won't change now

iaso/utils/legend.py
Bewi: Yep, this is still something we need to review / refactor

iaso/utils/legend.py
Bewi: Won't change now

iaso/utils/legend.py
Bewi: Won't change now

iaso/api/metrics/views.py
```diff
@@ -15,11 +24,32 @@
...
         return MetricType.objects.filter(account=self.request.user.iaso_profile.account, is_utility=False)
 
+    def get_serializer_class(self):
+        return (
+            MetricTypeCreateSerializer
+            if self.action == "create"
+            else MetricTypeWriteSerializer
+            if self.action in ["update", "partial_update"]
+            else MetricTypeSerializer
+        )
```
tdethier, line 39: 👍

iaso/tests/api/metrics/test_serializers.py
```diff
@@ -32,6 +43,79 @@ def test_fields(self):
...
+            "unit_symbol",
+            "legend_type",
+            "origin",
+        }
+        self.assertEqual(set(serializer.Meta.fields), expected_fields)
+
+    def test_validate_code_whitespace(self):
+        serializer = MetricTypeCreateSerializer(data={"code": "invalid code"}, context={"request": self.request})
+        with self.assertRaisesMessage(serializers.ValidationError, "Code must not contain whitespace."):
+            serializer.validate_code("invalid code")
```
tdethier: Nitpick: you shouldn't manually call the validation methods, they should be called by DRF at some point (when you check if the received payload is valid)

```suggestion
            serializer.is_valid()
```

iaso/tests/api/metrics/test_serializers.py
```diff
@@ -32,6 +43,79 @@ def test_fields(self):
...
+
+    def test_validate_code_same(self):
+        metric_type = MetricType(code="same_code")
+        serializer = MetricTypeCreateSerializer(
+            instance=metric_type, data={"code": "same_code_modified"}, context={"request": self.request}
+        )
+        self.assertFalse(serializer.is_valid())
```
tdethier: I think you are testing 2 different things here:
- you are using a create serializer to update an existing metric type (which shouldn't happen, because now you have an update serializer)
- you are trying to update a metric type code

these should be 2 different tests

iaso/tests/api/metrics/test_views.py
```diff
@@ -51,41 +50,109 @@ def test_metric_type_list(self):
...
+
+        created_mt = MetricType.objects.get(code="MT003")
+        self.assertEqual(created_mt.account, self.account)
+        self.assertEqual(created_mt.legend_type, MetricType.LegendType.THRESHOLD.value)
+        self.assertEqual(created_mt.origin, MetricType.MetricTypeOrigin.CUSTOM.value)
```
tdethier, line 76: That's not `legend_config` 😅

iaso/utils/legend.py
```diff
@@ -0,0 +1,214 @@
...
+        print(f"Parsed scales: {scales}")
+        try:
+            numeric_scales = [float(s) for s in scales]
+        except Exception as e:
+            print(f"Error converting scales to numerics: {e}")
+            numeric_scales = []
+        return {"domain": numeric_scales, "range": get_range_from_count(len(scales))}
+    if metric_type.legend_type == "ordinal":
+        scales = get_scales_from_list_or_json_str(scale)
+        if scales == 3:
```
tdethier: I think it was `len(scales)` instead of `scales`

tdethier (approved):
Using 2 different serializers is better 👍 

I'm already approving, but there are a few things that should still be changed. We can do this together if you want. 

Notes while testing:
- I finally was able to test the feature after enabling the dev feature at the account level, the feature works fine 👍 
- When editing a manual layer, the `Create` button should be changed to `Edit`
Image

iaso/tests/api/metrics/test_serializers.py
Bewi: I'll change this, no need to test the change of codes as write serializer doesn't accept it anymore and we already tests fields of that serializer.

Here it should test that we don't create a code that already exists

iaso/tests/api/metrics/test_views.py
Bewi: lengend_this legend_that legend_all_the_things

PR #2694, opened 2026-01-27
IA-4715: Empty Change requests

beygorghor:

While syncing two versions of a data source, if an org unit has no opening date on the origin version and a opening date on the ref version. Preview is saying we have changes but created change requests are empty because only taking modified fields into account.  

### Related JIRA tickets

IA-4715

## Changes

- adapt synchronizer to take into account STATUS_NEW and STATUS_NOT_IN_ORIGIN too
- improve logs to use task logs 

## How to test

- Create a new data source
- make a version with this GPKG 
[SNIS - DHIS2 WHITOUT.gpkg.zip](https://github.com/user-attachments/files/24884501/SNIS.-.DHIS2.WHITOUT.gpkg.zip)
- make another version with this gpkg
[SNIS - DHIS2 WHITH.gpkg.zip](https://github.com/user-attachments/files/24884511/SNIS.-.DHIS2.WHITH.gpkg.zip)
It contains only one org unit, one without opening date and one with
- launch a data source sync, select last version as origin and the first one as target, make sure you select only opening dates
Screenshot 2026-01-27 at 14 26 51

- click on synchronise, give it a name and click on preview, you should have one 1 org unit update
Screenshot 2026-01-27 at 14 27 54

- launch the task
- it should create one change request
Screenshot 2026-01-27 at 14 29 09
- go to validation page
- select the data source you created and the first version
- make sure you search for "data source synchronisation with the name you gave before, you should have one result
Screenshot 2026-01-27 at 14 30 54

- export it to CSV, it should include opening date changes
Screenshot 2026-01-27 at 14 32 13

- open the details of this change request, it should show opening date changes too

Screenshot 2026-01-27 at 14 33 24

## Print screen / video

-

## Notes

-

## Doc

-


iaso/diffing/synchronizer.py
```diff
@@ -309,10 +345,11 @@ def _prepare_new_change_requests(self, diff: dict) -> tuple[Optional[OrgUnitChan
         return org_unit_change_request, group_changes
 
     def _prepare_modified_change_requests(self, diff: dict) -> tuple[OrgUnitChangeRequest, list]:
+        org_unit = diff["orgunit_dhis2"]
         changes = {
             comparison["field"]: comparison["after"]
             for comparison in diff["comparisons"]
-            if comparison["status"] == Differ.STATUS_MODIFIED
+            if comparison["status"] in [Differ.STATUS_MODIFIED, Differ.STATUS_NEW, Differ.STATUS_NOT_IN_ORIGIN]
```
beygorghor, line 357: This si the actual fix, we should also use other status to see if a field pass from None to a specific value or the opposite

iaso/diffing/synchronizer.py
```diff
@@ -323,29 +360,33 @@ def _prepare_modified_change_requests(self, diff: dict) -> tuple[OrgUnitChangeRe
...
+                    .only("pk")
+                    .first()
                 )
-                .only("pk")
-                .first()
-            )
             if parent_org_unit:
                 new_parent_id = parent_org_unit.pk
-            requested_fields.append("new_parent")
+                requested_fields.append("new_parent")
```
bmonjoie: Shouldn't we remove the `if` as we might be removing the parent, or am I missing something?

iaso/diffing/synchronizer.py
```diff
@@ -323,29 +360,33 @@ def _prepare_modified_change_requests(self, diff: dict) -> tuple[OrgUnitChangeRe
...
 
-        if changes.get("name"):
-            new_name = changes["name"]
+        if "name" in changes:
+            new_name = changes["name"] or ""
             requested_fields.append("new_name")
```
bmonjoie: Is it ok to create a change that asks to remove the name of an OrgUnit?

iaso/diffing/synchronizer.py
```diff
@@ -323,29 +365,33 @@ def _prepare_modified_change_requests(self, diff: dict) -> tuple[OrgUnitChangeRe
...
-                OrgUnit.objects.filter(
-                    source_ref=parent_source_ref, version=self.data_source_sync.source_version_to_update
+            parent_org_unit = None
+            if parent_source_ref:
+                parent_org_unit = (
+                    OrgUnit.objects.filter(
+                        source_ref=parent_source_ref, version=self.data_source_sync.source_version_to_update
+                    )
+                    .only("pk")
+                    .first()
```
bmonjoie, line 377: If we don't find it, we'll put the parent to `None`,  I guess it's OK but maybe we should log that as an error. Don't you think?

bmonjoie (approved):
Just a little comment but otherwise, it looks good to me 🙂 👍

PR #2696, opened 2026-01-27
[IA-4524] user management improve csv export permission role projects

Crebert08:

This PR is about changing how the columns: permissions, roles and projects are displayed in the CSV and XLSX export. The former way consisted on making a columns in which all the permissions were displayed, same for roles and projects. The new way consists on showing all the permissions, roles and projects related to the account each one in its column and display `0` and `1` to show if the user has the permission/role/projects.

### Related JIRA tickets

https://bluesquare.atlassian.net/browse/IA-4524

## Changes

- in `iaso/api/profiles/profiles.py`, I fetched all the permissions related to the account, created a logic to see which permission the user has to display 1 or 0 depending on the result, displayed the matrix of the permissions
- in `iaso/tests/api/test_profiles.py`, I changed `test_profile_list_export_as_csv` and `test_profile_list_export_as_csv` to include the permissions matrix and then I added a new test to see if the export include the complete permission matrix while ignoring the permissions inherited from the user roles.

## How to test

Because I only changed the view `list_export`, the only test needed is: `sudo docker compose exec iaso python manage.py test iaso.tests.api.test_profiles`

## Print screen / video

image
image
image
image
image

## Notes

## Doc


Crebert08:
Do you have a preference for the order of the new columns ?

beygorghor:
i'll close this for now, this can break the import process after, and i'm not huge fan of the feature too

PR #2700, opened 2026-01-29
HOTFIX: polio 2079

quang-le:

Performance threshold write actions should be limited in the UI to a specific user

### Related JIRA tickets

POLIO-2079

## Changes

- Add a custom permission for thresholds
- hide create button and table action columns behind new permission
- permission not added to the permission menu in the user page because it's not meant to be attributed

## How to test

- On a polio DB
- Log in as non super user with all polio permissions available in the user page
- Go to Polio > Vaccine Module > performance thresholds
- The create button and the action column should not be visible

- repeat as super user or as user with new permission (given via django admin) --> create button and action column should be visible

## Print screen / video

Screenshot 2026-01-29 at 14 55 34

## Notes

On polio prod, the new permission should be given to the "specific user"
HOTFIX v2026.01.27a


beygorghor (approved):
see original PR

PR #2702, opened 2026-01-30
POLIO-2076 migrate old integrated campaigns

quang-le:

The polio DB still has old integrated campaigns, ie: campaigns with multiple types

### Related JIRA tickets

POLIO-2076
## Changes

- Add migration:
    - if campaign has polio in its types and has more than 1 type:
        - make another campaign per non  polio type
        - set the campaign type to polio only

## How to test

- Before checking out the branch: create 2 campaigns with multiple types (Campaigns UI):
    - 1 with polio type
    - 1 without
- Checkout the branch and run the migration
- Go to campaigns page:
    - The polio campaign should only have the polio type
    - 1 new campaign per (previous) type should have been created, with the same dates, scopes and subactivities
    - The non polio campaign should remain unchanged   

## Print screen / video

Screenshot 2026-01-30 at 11 48 10

## Notes

We only change the type of the polio campaign because we know it's a separate campaign from the non-polio, whereas the non-polio can be "grouped" but are still a different "entity" since they are not directly managed by RRT

## Doc


plugins/polio/migrations/0250_split_old_integrated_campaigns.py
```diff
@@ -0,0 +1,130 @@
...
+            "scopes",
+            "scopes__group__org_units",
+            "account",
+            "rounds__sub_activities",
+            "rounds__sub_activities__scopes",
+            "rounds__sub_activities__scopes__group__org_units",
+        )
+        .annotate(campaign_types_count=Count("campaign_types"))
+        .filter(campaign_types_count__gt=1)
+    )
```
tdethier, line 33:
```suggestion
    campaigns = (
        Campaign.objects.select_related("account")
        .prefetch_related(
            "campaign_types",
            "scopes",
            "scopes__group__org_units",
            "rounds",
            "rounds__scopes",
            "rounds__scopes__group__org_units",
            "rounds__sub_activities",
            "rounds__sub_activities__scopes",
            "rounds__sub_activities__scopes__group__org_units",
        )
        .annotate(campaign_types_count=Count("campaign_types"))
        .filter(campaign_types_count__gt=1)
    )
```

tdethier: Account is a FK, so it can be fetched directly

plugins/polio/migrations/0250_auto_20260129_1429.py
```diff
@@ -0,0 +1,130 @@
...
+                    new_campaign.rounds.set(new_rounds)
+            campaign.campaign_types.set(CampaignType.objects.filter(slug="polio"))
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("polio", "0249_alter_poliopermissionsupport_options"),
+    ]
+
+    operations = [migrations.RunPython(split_multi_type_campaigns, migrations.RunPython.noop)]
```
tdethier:
```suggestion
    operations = [migrations.RunPython(split_multi_type_campaigns, migrations.RunPython.noop, elidable=True)]
```

plugins/polio/migrations/0250_split_old_integrated_campaigns.py
```diff
@@ -0,0 +1,130 @@
+# Generated by Django 4.2.27 on 2026-01-29 14:29
```
tdethier, line 1: Nitpick: I would rename the file to say directly that it splits campaigns into integrated campaigns, something like `0250_split_multiple_campaigns_into_integrated_campaigns.py`

plugins/polio/migrations/0250_split_old_integrated_campaigns.py
```diff
@@ -0,0 +1,130 @@
...
+                        is_preventive=campaign.is_preventive,
+                        is_test=campaign.is_test,
+                        on_hold=campaign.on_hold,
+                        is_planned=campaign.is_planned,
+                        gpei_coordinator=campaign.gpei_coordinator,
+                        gpei_email=campaign.gpei_email,
+                        description=campaign.description,
+                        separate_scopes_per_round=campaign.separate_scopes_per_round,
+                        enable_send_weekly_email=campaign.enable_send_weekly_email,
+                    )
```
tdethier, line 59: Shouldn't you set `integrated_to` as well?

plugins/polio/migrations/0250_split_old_integrated_campaigns.py
```diff
@@ -0,0 +1,130 @@
...
+                            age_min=rnd.age_min,
+                            age_max=rnd.age_max,
+                            age_type=rnd.age_type,
+                            target_population=rnd.target_population,
+                            percentage_covered_target_population=rnd.percentage_covered_target_population,
+                            on_hold=rnd.on_hold,
+                            is_planned=rnd.is_planned,
+                            mop_up_started_at=rnd.mop_up_started_at,
+                            mop_up_ended_at=rnd.mop_up_ended_at,
+                        )
```
tdethier, line 87: We do a lot of "for each object in objects, create a new one with similar values, then add all new objects", I'm wondering if `deepcopy()` could simplify things here. You don't need to change anything, I'm just thinking out loud

plugins/polio/migrations/0250_split_old_integrated_campaigns.py
```diff
@@ -0,0 +1,130 @@
...
+                                for scope in subactivity.scopes.all():
+                                    group = Group.objects.create(name="hidden round scope")
+                                    new_scope = SubactivityScope.objects.create(
+                                        vaccine=scope.vaccine, subactivity=new_subactivity, group=group
+                                    )
+                                    new_scope.group.org_units.set(scope.group.org_units.all())
+                                    new_subactivity_scopes.append(new_scope)
+                                new_subactivity.scopes.set(new_subactivity_scopes)
+                                new_subactivities.append(new_subactivity)
+
```
tdethier, line 124: Shouldn't you set subactivities to the new round here?

tdethier (commented):
I added a few comments, I think that some things are missing but I'm not sure because I lack polio business knowledge

plugins/polio/migrations/0250_split_old_integrated_campaigns.py
quang-le: Nope because I set `round`line 98

tdethier (approved):
LGTM 👍

PR #2704, opened 2026-01-30
IA-4706: remove cypress

quang-le:
We don't want cypress anymore

### Related JIRA tickets

IA-4706

## Changes

- removed cypress tests
- removed commands and deps in package.json
- updated package.lock
- removed GH action

## How to test

- try building webpack

## Print screen / video
N/A


beygorghor (approved):
Nice ! Thanks for this.
D'ont forget to update  / remove the doc regarding cypress, and probably also update iaso-release readme too

PR #2707, opened 2026-01-30
IA-4745: Remove custom translations from user profile api call

beygorghor:

We need a dedicated endpoint to fetch account custom translations  from api/profiles/me call in order to avoid heavy payload.

Also we can remove non iaso permissions from user_group_permissionsresults

### Related JIRA tickets

IA-4745

## Changes

- custom endpoint for custom translations
- tests
- adapt front
- remove non iaso permissions from 'user_group_permissions'

## How to test

Make sure you have some custom translations in your account, you can use those

[custom_translations.json](https://github.com/user-attachments/files/24966288/custom_translations.json)

Launch the dashboard and make sure entities are now beneficiaries

Also check the `/api/profiles/me/`and make sure you don't have non iaso permissions

## Print screen / video

-
## Notes

-

## Doc
-


iaso/models/base.py
```diff
@@ -609,9 +607,11 @@ def get_user_roles_editable_org_unit_type_ids(self):
 
     def as_dict(self, small=False):
         user_roles = self.user_roles.all().order_by("group__name")
-        user_group_permissions = list(
-            map(lambda permission: permission.split(".")[1], list(self.user.get_group_permissions()))
-        )
+        user_group_permissions = [
+            permission.split(".")[1]
+            for permission in self.user.get_group_permissions()
+            if permission.split(".")[1].startswith("iaso_")
```
mestachs, line 613: any risk of breaking other plugins ? (wfp, snt, saas,... do they start with iaso too ?)

iaso/tests/api/test_custom_translations.py
```diff
@@ -0,0 +1,37 @@
...
+        response = self.client.get(f"/api/custom_translations/?account_id={self.ghi.pk}")
+        self.assertJSONResponse(response, 401)
+
+    def test_custom_translations_requires_account_id(self):
+        self.client.force_authenticate(self.jane)
+        response = self.client.get("/api/custom_translations/")
+        data = self.assertJSONResponse(response, 400)
+        self.assertEqual(data, {"account_id": ["Account id is required."]})
+
+    def test_custom_translations_forbidden_for_other_account(self):
```
mestachs, line 32: I would have added a test for non existing account_id (the current behavior looks ok : return 404, but there's no test)

mestachs (commented):
add a test and being sure of not impacting other plugin

mestachs (approved):

iaso/models/base.py
beygorghor: All permissions used in the iaso dashboard are prefixed with `iaso_`  including plugins

PR #2708, opened 2026-02-02
POLIO-1569, POLIO-2024: light polio cleanup

quang-le:

small fixes to polio to cleanup the backlog:
- deduplicates some tests
- make a campaign list column unsortable

### Related JIRA tickets

POLIO-1569, POLIO-2024

## Changes

- Extract `VaccineStockManagementAPITestBase`  from `VaccineStockManagementAPITestCase` and change parent class of former children test classes
- Make status column unsortable on polio campaign page, as it's not working (so is probably not used) and would require significant dev work

## How to test

- Go to polio campaigns, check that status column is not sortable (no arrow on hover)
- Run backend tests


edil3ra (approved):
Looks good to me

PR #2709, opened 2026-02-02
SNT-264 i can import layers data

Bewi:

Import layers data from CSV.

### Related JIRA tickets

SNT-264

## Changes

- Validate scales
- Moved create / update logic to serializer
- Add import metric value from CSV endpoint

## How to test

Go to SNT_Malaria with any account (BFA default setup is enough).

On the left navigation pan, go to settings.
In there you can modify metric layers.
You can try to: 
- Create a layer (Play with legend validator)
- Edit a layer (Play with legend validator)
- Just beside the search bar, click the three dots icon
- Then export the CSV template and fill values as you want based on your layers configuration.
- Once done, make sure it uses "," separator
- Then you can import the csv
- Go to scenarios and open one (create one if you don't have any yet).
- On the main map, verify that your new layers are present dans that selecting one display values and scales properly.
- Verify the legend as well.

## Print screen / video

https://github.com/user-attachments/assets/5bbe487c-344b-40cd-ad32-892afcc577d5

## Notes

Related to SNT_Malaria PR: https://github.com/BLSQ/snt-malaria/pull/181

## Doc

N/A


iaso/api/org_unit_tree/filters.py
```diff
@@ -59,3 +60,11 @@ def filter_data_source_id(self, queryset: QuerySet, _, data_source_id: int) -> Q
             return queryset.filter(version=source.default_version)
 
         return queryset.filter(version__data_source_id=data_source_id)
+
+    def filter_valid_org_units_for_account(queryset: QuerySet, account):
+        return (
+            queryset.filter(version=account.default_version)
+            .filter(validation_status=OrgUnit.VALIDATION_VALID)
+            .filter(Q(location__isnull=False) | Q(simplified_geom__isnull=False))
+            .order_by("name")
```
tdethier: I would not include `order_by` here because it's not the filter component's role to order data. You can add it to the result of calling this method though

iaso/api/org_unit_tree/filters.py
```diff
@@ -59,3 +60,11 @@ def filter_data_source_id(self, queryset: QuerySet, _, data_source_id: int) -> Q
             return queryset.filter(version=source.default_version)
 
         return queryset.filter(version__data_source_id=data_source_id)
+
+    def filter_valid_org_units_for_account(queryset: QuerySet, account):
```
tdethier: Your method signature does not match the other ones. `self` is missing, but it looks like django filters is passing a third parameter every time (that is ignored).

Now that I see that, I'm thinking that this utility method should not be placed here, because it's not related to the `OrgUnitTree` endpoint. I would probably leave it as a utility method outside of this.

iaso/api/metrics/views.py
```diff
@@ -81,11 +81,60 @@ class MetricValueViewSet(viewsets.ModelViewSet):
...
 
+    @action(detail=False, methods=["get"])
+    def csv_template(self, request):
+        account = request.user.iaso_profile.account
+        # Get all custom metric types for the user's account
+        metric_types = MetricType.objects.filter(account=account, origin=MetricType.MetricTypeOrigin.CUSTOM)
+        # Get All org units for the user's account
+        org_units = OrgUnitTreeFilter.filter_valid_org_units_for_account(
+            OrgUnit.objects.all(), account
+        ).prefetch_related("parent")
```
tdethier: `prefetch_related()` will probably not be applied, because you're already filtering and ordering your queryset above. It think it's one more reason for making a dedicated utility function

iaso/models/metric.py
```diff
@@ -57,7 +57,7 @@ class MetricValue(models.Model):
     class Meta:
         unique_together = [["metric_type", "org_unit", "year"]]
 
-    metric_type = models.ForeignKey(MetricType, on_delete=models.PROTECT)
+    metric_type = models.ForeignKey(MetricType, on_delete=models.CASCADE)
```
tdethier, line 60: Is there any specific reason for this?

iaso/api/metrics/views.py
```diff
@@ -81,11 +81,60 @@ class MetricValueViewSet(viewsets.ModelViewSet):
...
+                row.append("")  # Empty value for the metric
+            writer.writerow(row)
+
+        filename = "metric_import_template.csv"
+        response["Content-Disposition"] = f"attachment; filename={filename}"
+        return response
+
+    @action(detail=False, methods=["post"])
+    def import_from_csv(self, request):
+        serializer = ImportMetricValuesSerializer(data=request.data, context={"request": request})
```
tdethier: If you define a specific serializer for your action, you don't need to pass any parameters


```suggestion
    @action(detail=False, methods=["post"], serializer_class=ImportMetricValuesSerializer)
    def import_from_csv(self, request):
        serializer = self.get_serializer()
```

iaso/api/org_unit_tree/filters.py
tdethier: (that's also a sign that this should not be part of a filtering component, but a dedicated utility function)

iaso/api/metrics/serializers.py
```diff
@@ -50,15 +57,50 @@ class Meta:
...
+            "scale",
         ]
 
+    def get_legend_config_from_scale(self, instance):
+        scale = self.initial_data.get("scale")
+        return legend.get_legend_config(instance, scale) if scale else instance.legend_config
+
+    def update(self, instance, validated_data):
+        legend_config = self.get_legend_config_from_scale(instance)
+        return super().update(instance, {**validated_data, "legend_config": legend_config})
```
tdethier, line 70: 👍

iaso/api/metrics/serializers.py
```diff
@@ -50,15 +57,50 @@ class Meta:
             "unit_symbol",
             "legend_type",
             "origin",
+            "scale",
         ]
 
+    def get_legend_config_from_scale(self, instance):
+        scale = self.initial_data.get("scale")
```
tdethier: You should get `scale` from your instance object rather than `self.initial_data` because it's unvalidated data
```suggestion
        scale = instance.scale
```

iaso/api/metrics/serializers.py
```diff
@@ -50,15 +57,50 @@ class Meta:
...
+        return legend.get_legend_config(instance, scale) if scale else instance.legend_config
+
+    def update(self, instance, validated_data):
+        legend_config = self.get_legend_config_from_scale(instance)
+        return super().update(instance, {**validated_data, "legend_config": legend_config})
+
+    def validate(self, data):
+        legend_type = self.initial_data.get("legend_type")
+        scale_count = len(self.initial_data.get("scale", "").split(",")) if self.initial_data.get("scale") else 0
```
tdethier:
```suggestion
        legend_type = data.get("legend_type")
        scale_count = len(data.get("scale", "").split(",")) if data.get("scale") else 0
```

iaso/api/metrics/serializers.py
```diff
@@ -50,15 +57,50 @@ class Meta:
...
 
 class MetricTypeCreateSerializer(MetricTypeWriteSerializer):
     code = serializers.CharField(required=True, allow_blank=False)
 
     class Meta(MetricTypeWriteSerializer.Meta):
         fields = MetricTypeWriteSerializer.Meta.fields + ["code"]
 
+    def create(self, validated_data):
+        account = self.context["request"].user.iaso_profile.account
+        instance = super().create({**validated_data, "account": account})
```
tdethier, line 103: Nitpick: I would probably move the account part to the view because it's not derived from the other attributes of your model (like legend_config). You don't need to change this, but I should maybe make a deep dive on this and add this type of example to the API guidelines

iaso/api/metrics/serializers.py
tdethier: It's dangerous to call `self.initial_data` even in `validate()` because when you get here, a potential `validate_<field_name>()` could already have been called and you would not have its result with `self.initial_data`

iaso/api/metrics/views.py
```diff
@@ -81,11 +81,60 @@ class MetricValueViewSet(viewsets.ModelViewSet):
...
+            metric_type_ids = set(mv.metric_type_id for mv in metric_values)
+            org_unit_ids = set(mv.org_unit_id for mv in metric_values)
+            MetricValue.objects.filter(metric_type_id__in=metric_type_ids, org_unit_id__in=org_unit_ids).delete()
+
+            MetricValue.objects.bulk_create(metric_values)
+
+        return Response(
+            {
+                "total_imported": len(metric_values),
+                "metric_type_import_count": len(set(mv.metric_type_id for mv in metric_values)),
```
tdethier:
```suggestion
                "metric_type_import_count": len(metric_type_ids),
```

iaso/api/metrics/serializers.py
```diff
@@ -73,12 +115,92 @@ def validate_code(self, value):
 
 
 class MetricValueSerializer(serializers.ModelSerializer):
+    id = serializers.IntegerField(read_only=True)
+    year = serializers.IntegerField(required=False, allow_null=True)
+    value = serializers.FloatField(required=False, allow_null=True)
+    string_value = serializers.CharField(required=False, allow_null=True)
+    metric_type = serializers.PrimaryKeyRelatedField(queryset=MetricType.objects.all())
+    org_unit = serializers.PrimaryKeyRelatedField(queryset=OrgUnit.objects.all(), allow_null=False)
+
```
tdethier, line 128: Why did you add this?

iaso/api/metrics/serializers.py
```diff
@@ -73,12 +115,92 @@ def validate_code(self, value):
...
+                _("The following metric types do not exist: ") + ", ".join(missing_metric_types)
+            )
+
+        self.context["existing_metric_types"] = {mt[0]: mt[1] for mt in existing_metric_types}
+
+        if df.shape[0] == 0:
+            raise serializers.ValidationError(_("The CSV must contain at least one value row."))
+
+        org_unit_ids = df["ADM2_ID"].tolist()
+        matching_org_unit_ids = OrgUnit.objects.filter(id__in=org_unit_ids).values_list("id", flat=True)
```
tdethier: Here I would definitely filter OrgUnits based on the account  (+ maybe on profile restrictions?)

iaso/api/metrics/serializers.py
```diff
@@ -73,12 +115,92 @@ def validate_code(self, value):
...
+        existing_metric_types = MetricType.objects.filter(account=account, code__in=metric_type_codes).values_list(
+            "code", "id", "legend_config"
+        )
+        missing_metric_types = metric_type_codes - set(mt[0] for mt in existing_metric_types)
+        if missing_metric_types:
+            raise serializers.ValidationError(
+                _("The following metric types do not exist: ") + ", ".join(missing_metric_types)
+            )
+
+        self.context["existing_metric_types"] = {mt[0]: mt[1] for mt in existing_metric_types}
```
tdethier: Why do you put it in the serializer context? You could just create a new variable because you're using it only below, not in the view

iaso/api/metrics/serializers.py
```diff
@@ -73,12 +115,92 @@ def validate_code(self, value):
...
+
+        # Prepare metric values but don't save yet
+        metric_values = []
+        for index, row in df.iterrows():
+            org_unit_id = row["ADM2_ID"]
+            for code in metric_type_codes:
+                value = row.get(code)
+                if pd.isna(value):
+                    continue
+                metric_type_id = self.context["existing_metric_types"][code]
```
tdethier: This could be fetched from the new variable, not context

iaso/api/metrics/serializers.py
```diff
@@ -73,12 +115,92 @@ def validate_code(self, value):
...
+
+        org_unit_ids = df["ADM2_ID"].tolist()
+        matching_org_unit_ids = OrgUnit.objects.filter(id__in=org_unit_ids).values_list("id", flat=True)
+        missing_org_unit_ids = set(org_unit_ids) - set(matching_org_unit_ids)
+        if missing_org_unit_ids:
+            raise serializers.ValidationError(
+                _("The following org unit IDs do not exist: ") + ", ".join(str(ou_id) for ou_id in missing_org_unit_ids)
+            )
+
+        self.context["dataframe"] = df
```
tdethier: Same, I don't think you need this
```suggestion
```

iaso/api/metrics/serializers.py
```diff
@@ -73,12 +115,92 @@ def validate_code(self, value):
...
+                    # Parse the value as a float
+                    mv.value = float(value)
+                except ValueError:
+                    mv.value = None
+                    mv.string_value = value
+                metric_values.append(mv)
+
+        self.context["metric_values"] = metric_values
+
+        # Do we want to validate all values already here?
```
tdethier, line 212: Good question 😅 

Since you don't do it later in the view, I guess this would be the place to do it

iaso/tests/api/metrics/test_serializers.py
```diff
@@ -129,6 +325,91 @@ def test_fields(self):
...
+        self.assertIn("The CSV must contain 'ADM1_NAME, ADM2_NAME, ADM2_ID' columns.", serializer.errors["file"][0])
+
+    def test_validate_no_row(self):
+        csv_content = "ADM1_NAME,ADM2_NAME,ADM2_ID,MT2"
+        invalid_file = SimpleUploadedFile("test.csv", csv_content.encode(), content_type="text/csv")
+        serializer = ImportMetricValuesSerializer(data={"file": invalid_file}, context={"request": self.request})
+        self.assertFalse(serializer.is_valid())
+        self.assertIn("The CSV must contain at least one value row.", serializer.errors["file"][0])
+
+    def test_validate_missing_org_unit_ids(self):
```
tdethier, line 469: We should have a test for org unit IDs from another account 😇 It should also fail and say that the ID does not exist

iaso/api/metrics/serializers.py
```diff
@@ -50,15 +57,50 @@ class Meta:
             "unit_symbol",
             "legend_type",
             "origin",
+            "scale",
         ]
 
+    def get_legend_config_from_scale(self, instance):
```
quang-le: nitpick: 

```suggestion
    def _get_legend_config_from_scale(self, instance):
```

iaso/api/metrics/serializers.py
```diff
@@ -50,15 +57,50 @@ class Meta:
...
+        scale = self.initial_data.get("scale")
+        return legend.get_legend_config(instance, scale) if scale else instance.legend_config
+
+    def update(self, instance, validated_data):
+        legend_config = self.get_legend_config_from_scale(instance)
```
quang-le: same same 

```suggestion
        legend_config = self._get_legend_config_from_scale(instance)
```

iaso/api/metrics/serializers.py
```diff
@@ -50,15 +57,50 @@ class Meta:
...
 class MetricTypeCreateSerializer(MetricTypeWriteSerializer):
     code = serializers.CharField(required=True, allow_blank=False)
 
     class Meta(MetricTypeWriteSerializer.Meta):
         fields = MetricTypeWriteSerializer.Meta.fields + ["code"]
 
+    def create(self, validated_data):
+        account = self.context["request"].user.iaso_profile.account
+        instance = super().create({**validated_data, "account": account})
+        instance.legend_config = self.get_legend_config_from_scale(instance)
```
quang-le: same same 

```suggestion
        instance.legend_config = self._get_legend_config_from_scale(instance)
```

iaso/tests/api/metrics/test_views.py
```diff
@@ -300,11 +300,53 @@ def test_metric_value_post(self):
...
+                "metric_type": self.metric_type.id,
+                "org_unit": self.org_unit.id,
                 "value": 200.0,
+                "year": 2022,
+                "string_value": "200",
             },
         )
-        self.assertEqual(response.status_code, status.HTTP_405_METHOD_NOT_ALLOWED)  # Method Not Allowed
+        self.assertJSONResponse(response, status.HTTP_401_UNAUTHORIZED)
+
```
tdethier, line 385: We should also have a test here where we try to create a `MetricValue` of a specific `MetricType` that belongs to another account, which should fail

iaso/api/metrics/serializers.py
```diff
@@ -73,12 +115,92 @@ def validate_code(self, value):
 
 
 class MetricValueSerializer(serializers.ModelSerializer):
+    id = serializers.IntegerField(read_only=True)
+    year = serializers.IntegerField(required=False, allow_null=True)
+    value = serializers.FloatField(required=False, allow_null=True)
+    string_value = serializers.CharField(required=False, allow_null=True)
+    metric_type = serializers.PrimaryKeyRelatedField(queryset=MetricType.objects.all())
+    org_unit = serializers.PrimaryKeyRelatedField(queryset=OrgUnit.objects.all(), allow_null=False)
```
quang-le: I always wonder: shouldn't we already filter the queryset by account? but then how to do it at the serializer level?

@tdethier  Any opinions?

iaso/tests/api/metrics/test_views.py
```diff
@@ -300,11 +300,53 @@ def test_metric_value_post(self):
...
+            data={
+                "metric_type": 9999,  # Invalid metric type ID
+                "org_unit": self.org_unit.id,
+                "value": 200.0,
+                "year": 2022,
+                "string_value": "200",
+            },
+        )
+        self.assertJSONResponse(response, status.HTTP_400_BAD_REQUEST)
+
```
tdethier, line 399: And same here, try to create a `MetricValue` linked to an `OrgUnit` that belongs to another account

iaso/api/metrics/serializers.py
quang-le, line 212: In GAPS we're using a worker to validate the values, because the amount of data to validate can be quite big

iaso/tests/api/org_unit_tree/test_filters.py
```diff
@@ -0,0 +1,93 @@
+from django.contrib.gis.geos import MultiPolygon, Point, Polygon
+
+from iaso import models as m
+from iaso.api.org_unit_tree.filters import OrgUnitTreeFilter
+from iaso.models.base import Account
+from iaso.permissions.core_permissions import CORE_ORG_UNITS_PERMISSION
+from iaso.test import APITestCase
+
+
+class OrgUnitTreeFilterTest(APITestCase):
```
tdethier: I think you'll need to get rid of this unfortunately 🙈

iaso/api/org_unit_tree/filters.py
quang-le: The name is misleading, since we only include org units with a location or geometry. I would rename this method, it would make it "safer"/easier to reuse

tdethier (changes requested):
General comments while testing:
- overall, the feature works fine: I was able to create/edit/delete layers 👍 
- I created a layer with `[snt264, snt264]` in scale, which passed frontend validation and was accepted by the backend. But when I tried to edit it, scale was empty: 
Image

I would expect the backend to reject this value


- there are some tenancy issues (being able to create values linked to metric types from other accounts + referencing org units from other accounts), more tests should spot this
- there are some potentially dangerous operations in some serializers (which don't cause any issue in this specific scenario, but we should avoid calling `initial_data`)
- I think the backend should validate `MetricValues` from the imported CSV file
- the changes on `OrgUnitTreeFilter` should be reverted

iaso/api/metrics/serializers.py
tdethier: For the moment, the best solution I've seen in the project is [something like this](https://github.com/BLSQ/iaso/blob/b99285f84305b353f302361d0794349f6448d2c2/iaso/api/microplanning/serializers.py#L262). I'm not 100% convinced by that, I think we could improve it, but it's better than nothing

iaso/api/metrics/serializers.py
Bewi: At the moment we don't really have validation on values nor it is expected.
But in the future, if we want to add some validation in there, we could use the worker indeed.
Problem is that we don't really use the worker yet in SNT neither can the user see the running tasks, so that's quite some bits to think about.

iaso/api/org_unit_tree/filters.py
Bewi: I didn't spot any existing utility for this, so if fine, I'll add a new file for org units utilities.

iaso/models/metric.py
Bewi: Yes, you can delete a metric_type from the interface and there is a confirmation modal for this.
If we don't cascade and that metric_type has values, deletion won't work.
IMO it makes sense to delete associated metric_values as they are strongly linked to a metric_type

iaso/api/metrics/serializers.py
Bewi: That was my first version but that was the only thing done by the view.
Doing it here allows to keep the default view behavior, hence less code.

iaso/api/metrics/serializers.py
Bewi: Nice catch, was used in the view and forgot to refactore this

iaso/api/metrics/serializers.py
Bewi: We won't validate values atm, if needed we will add this later.

iaso/api/metrics/serializers.py
Bewi: I like this

iaso/api/metrics/serializers.py
Bewi: This is because it is now used to write as well and we need PrimaryKeyRelatedFields

iaso/tests/utils/test_org_units.py
```diff
@@ -0,0 +1,92 @@
...
+            path=["1", "2", "8"],
+            version=cls.source_version2,
+            geom=cls.mock_multipolygon,
+        )
+
+    def test_filter_valid_org_units_for_account(self):
+        filtered_qs = get_valid_org_units_with_geography(self.account)
+        self.assertIn(self.angola_district, filtered_qs)
+        self.assertNotIn(self.angola_district_new, filtered_qs)  # New status
+        self.assertNotIn(self.angola_district_rejected, filtered_qs)  # New status
```
tdethier:
```suggestion
        self.assertNotIn(self.angola_district_rejected, filtered_qs)  # Rejected status
```

iaso/api/org_unit_tree/filters.py
```diff
@@ -59,3 +60,11 @@ def filter_data_source_id(self, queryset: QuerySet, _, data_source_id: int) -> Q
...
+        return (
+            queryset.filter(version=account.default_version)
+            .filter(validation_status=OrgUnit.VALIDATION_VALID)
+            .filter(Q(location__isnull=False) | Q(simplified_geom__isnull=False))
+            .order_by("name")
+        )
```
tdethier: We can get rid of this now

```suggestion
```

iaso/models/metric.py
tdethier: It makes sense :+1:

iaso/api/metrics/serializers.py
tdethier: As we discussed, it's probably less code in the view, but way more in tests. You don't need to change this, don't worry

iaso/api/metrics/views.py
```diff
@@ -81,11 +80,58 @@ class MetricValueViewSet(viewsets.ModelViewSet):
...
     def get_queryset(self):
         return MetricValue.objects.filter(metric_type__account=self.request.user.iaso_profile.account)
 
+    @action(detail=False, methods=["get"])
+    def csv_template(self, request):
+        account = request.user.iaso_profile.account
+        # Get all custom metric types for the user's account
+        metric_types = MetricType.objects.filter(account=account, origin=MetricType.MetricTypeOrigin.CUSTOM)
+        # Get All org units for the user's account
+        org_units = get_valid_org_units_with_geography(account).order_by("name").prefetch_related("parent")
```
tdethier: Here again, I don't think that `prefetch_related()` will work because you do it after filtering and ordering 🫤

iaso/tests/api/metrics/test_views.py
```diff
@@ -300,11 +362,126 @@ def test_metric_value_post(self):
             data={
                 "metric_type": self.metric_type.id,
                 "org_unit": self.org_unit.id,
+                "value": 200.0,
                 "year": 2022,
+                "string_value": "200",
+            },
+        )
+        print(response.data)
```
tdethier:
```suggestion
```

tdethier (approved):
I left a few comments to do some more cleaning up, but LGTM 👍

PR #2710, opened 2026-02-02
IA-4751: new endpoint to save color on a profile

beygorghor:

Using existing PATCH api on profiles is a bit risky for now. If we don't provide all the profile info, we will probably erase some of those. We need a dedicated endpoint to be able to just update a profile color.

### Related JIRA tickets

IA-4751

## Changes

- new details update_color action on profile api
- added some tests

## How to test

You can use this Bruno file with https://www.usebruno.com/ application, just don't forget to change the token with the one you will find in  /api/apitoken/

[PROFILE COLOR UPDATE IASO.bru.zip](https://github.com/user-attachments/files/25050696/PROFILE.COLOR.UPDATE.IASO.bru.zip)

## Print screen / video

-

## Notes

-

## Doc

-


iaso/tests/api/test_profiles.py
```diff
@@ -1319,6 +1319,36 @@ def test_update_profile_color(self):
...
+        new_color = "#a1b2c3"
+        data = {"color": new_color}
+
+        response = self.client.patch(f"/api/profiles/{profile.id}/update_color/", data=data, format="json")
+
+        self.assertEqual(response.status_code, 200)
+        profile.refresh_from_db()
+        profile.user.refresh_from_db()
+        self.assertEqual(profile.color, "#A1B2C3")
+        self.assertEqual(response.json()["color"], "#A1B2C3")
```
quang-le:
```suggestion
        self.assertEqual(profile.color, new_color)
        self.assertEqual(response.json()["color"], new_color)
```

iaso/tests/api/test_profiles.py
```diff
@@ -1319,6 +1319,36 @@ def test_update_profile_color(self):
...
+
+    def test_update_profile_color_action_invalid_color(self):
+        self.client.force_authenticate(self.john)
+        profile = Profile.objects.get(user=self.jim)
+        data = {"color": "invalid"}
+
+        response = self.client.patch(f"/api/profiles/{profile.id}/update_color/", data=data, format="json")
+
+        self.assertEqual(response.status_code, 400)
+        self.assertIn("color", response.json())
```
quang-le:
```suggestion
        response = self.client.patch(f"/api/profiles/{profile.id}/update_color/", data=data, format="json")

        result= self.assertJSONResponse(response, status.HTTP_400_BAD_REQUEST) # you need to import status
        self.assertIn("color", result)
```

quang-le (approved):
just a couple of suggestions in the tests

PR #2713, opened 2026-02-03
IA-4752: raise 404 i.o 500 on project not found

quang-le:

The `import_org_units`method would raise a 500 when not finding a project for the given `app_id` which would clutter Sentry

### Related JIRA tickets

IA-4752

## Changes

- wrapped `import_org_units` in a `try/except` in the org unit api and mobile org units api and raise a 404
- Add specific exception for `Project.DoesNotExist` in:
    - mobile bulk upload
    - reimport command
    
- update test to check for error code 

## How to test

The tests should cover it

## Print screen / video

N/A

### DIFF
--- a/iaso/management/commands/reimport_failed_imports.py
+++ b/iaso/management/commands/reimport_failed_imports.py
@@ -3,6 +3,7 @@ from django.core.management.base import BaseCommand
 from hat.api_import.models import APIImport
 from iaso.api.instances.instances import import_data as import_instances
 from iaso.api.org_units import import_org_units
+from iaso.models import Project
 
 
 class Command(BaseCommand):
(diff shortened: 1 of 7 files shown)


iaso/management/commands/reimport_failed_imports.py
```diff
@@ -17,6 +18,8 @@ def handle(self, *args, **options):
                 i.has_problem = False
                 i.save()
                 unit_count = unit_count + 1
+            except Project.DoesNotExist as e:
+                print(f"No Project found for app id {i} {str(e)}")
```
bmonjoie, line 22: Why, `{str(e)}` instead of `, e` like on line 24?

bmonjoie (approved):

iaso/management/commands/reimport_failed_imports.py
quang-le: shhh, I just ctrl+V

PR #2720, opened 2026-02-04
IA-4753: Use Bruno with Iaso

beygorghor:

Have a minimal config to use bruno application inside Iaso. Useful for devs or anyone wanted to poll Iaso api

### Related JIRA tickets

IA-4753

## Changes

- add a new bruno folder 
- create local env
- create first calls

## How to test

You need to edit local environment end enter the **Host** url, **username** and **password**

Play with the calls, add variable values.

Try to update your profile color, it should call api first to fetch your user id and use it to patch the color

## Print screen / video
Screenshot 2026-02-04 at 17 33 42
Screenshot 2026-02-04 at 17 37 53

## Notes

- 

## Doc

[Bruno - The Open Source API Client](https://docs.usebruno.com/)


Bewi (approved):
Love it, variables are a bit confusing but all works fine

PR #2726, opened 2026-02-11
IA-4769: Fix too precise accuracy fails import

bmonjoie:
When the accuracy is too precise, rounding is not enough. We have to take into consideration what the actual size will be before validating.

Some instances cannot be imported because their accuracy is too precise. 
The previous rounding fix was not enough because the total amount of numbers is still too big for the validation.

### Related JIRA tickets

IA-4769

## Changes

I have changed how the validation check for size.

## How to test

1. Call the endpoint `POST /api/instances/` with an accuracy that will be 7 digits or less in total when rounded but with more digits before being rounded.
E.g.:  48.622852 (value from production that failed)

### DIFF
--- a/iaso/tests/api/instances/test_instances.py
+++ b/iaso/tests/api/instances/test_instances.py
@@ -2,6 +2,7 @@ import datetime
 import json
 import typing
 
+from decimal import Decimal
 from unittest import mock
 from unittest.mock import patch
 from uuid import uuid4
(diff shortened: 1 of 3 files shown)


iaso/tests/api/instances/test_instances.py
```diff
@@ -444,6 +444,35 @@ def test_instance_create_with_accuracy_rounded(self):
         last_instance = m.Instance.objects.get(uuid=instance_uuid)
         self.assertEqual(Decimal("12.35"), last_instance.accuracy)
 
+    def test_instance_create_with_accuracy_rounded_with_long_number(self):
+        """POST /api/instances/ with accuracy having more than 2 decimal places should be rounded"""
+        from decimal import Decimal
```
tdethier: You can move it to global imports, there shouldn't be any conflicts

tdethier (approved):
LGTM 👍

PR #2729, opened 2026-02-12
[IA-4617] add bin icon to reject orgunit from list

Crebert08:

This PR is about allowing the user to reject an Org Unit directly from the list.

### Related JIRA tickets

[IA-4617](https://bluesquare.atlassian.net/browse/IA-4617)

## Changes

- In `hat/assets/js/apps/Iaso/domains/orgUnits/components/ActionCell.tsx`, I added a bin icon in the action cell and implemented a custom hook to update the validation_status from 'NEW' or 'Valid' to 'REJECTED and added a custom message to confirm that the validation was modified
- In  `hat/assets/js/apps/Iaso/domains/orgUnits/index.tsx`, I modified the orgUnits components by implementing a variable that filters out the rejected orgUnits by default using the useMemo hook
- The rest is the translations I added and corrected

## How to test

- From the side menu, go the org Units list
- From the the table in the action columns, when you click the bin button, the org units should be rejected and hide by default.
If you want to see the rejected org units, you have to search by validation status: Rejected.

## Print screen / video

https://github.com/user-attachments/assets/e5f4628c-6ac5-4162-aab4-b90301b61bca

## Notes

## Doc

[IA-4617]: https://bluesquare.atlassian.net/browse/IA-4617?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hat/assets/js/apps/Iaso/constants/messages.js
```diff
@@ -102,7 +102,7 @@ const MESSAGES = defineMessages({
         id: 'iaso.label.users',
     },
     userRoles: {
-        defaultMessage: 'User roles',
+        defaultMessage: 'User Roles',
```
beygorghor, line 105: Default language is set to English, it means default message is in English and should correspond to your EN.json translation. Anyway i don't understand why you change this? ti should probably stay ASIS

hat/assets/js/apps/Iaso/domains/orgUnits/components/ActionCell.tsx
```diff
@@ -1,16 +1,31 @@
...
 export const ActionCell: FunctionComponent<Props> = ({ orgUnit }) => {
+    const { mutateAsync: saveOu, isLoading: isSaving } = useSaveOrgUnit(
+        undefined,
+        ['orgunits'],
+        MESSAGES.validationStatusChanged,
+    );
+
     const cell = useMemo(() => {
+        const handleRejectOrgUnit = async () => {
+            await saveOu({
```
beygorghor: async await is not necessary here

hat/assets/js/apps/Iaso/domains/orgUnits/reviewChanges/messages.ts
```diff
@@ -208,7 +208,7 @@ const MESSAGES = defineMessages({
         defaultMessage: 'Data Source Versions Synchronization',
     },
     userRoles: {
-        defaultMessage: 'User roles',
+        defaultMessage: 'User Roles',
```
beygorghor, line 211: Same here

hat/assets/js/apps/Iaso/domains/orgUnits/hooks.ts
```diff
@@ -220,13 +220,14 @@ export type SaveOrgUnitPayload = Omit<Partial<OrgUnit>, 'groups'> & {
 export const useSaveOrgUnit = (
     onSuccess?: () => void,
     invalidateQueryKey?: string[],
+    successMessage?: any,
```
beygorghor: You can use `IntlMessage` from `bluesquare-components`

hat/assets/js/apps/Iaso/domains/orgUnits/index.tsx
```diff
@@ -162,6 +162,25 @@ export const OrgUnits: FunctionComponent = () => {
...
+            return {
+                ...orgUnitsData,
+                orgunits: orgUnitsData.orgunits.filter(
+                    ou => ou.validation_status !== 'REJECTED',
+                ),
+            };
+        }
+
+        return orgUnitsData;
+    }, [orgUnitsData, searches]);
```
beygorghor: No need to do that, this will lead to inconsistencies in your results and while saving, org units results should be refreshed while invalidating the query. I would remove this

hat/assets/js/apps/Iaso/domains/orgUnits/messages.ts
```diff
@@ -625,6 +625,14 @@ const MESSAGES = defineMessages({
         id: 'iaso.orgUnits.addChild',
         defaultMessage: 'Add child',
     },
+    rejectOrgUnit: {
+        id: 'iaso.orgUnits.rejectOrgUnit',
+        defaultMessage: 'Reject Org Unit',
+    },
+    validationStatusChanged: {
+        id: 'iaso.orgUnits.validationStatusChanged',
+        defaultMessage: 'Validation Status changed successfully',
```
beygorghor: Validation status changed successfully ?

hat/assets/js/apps/Iaso/domains/payments/messages.ts
```diff
@@ -34,7 +34,7 @@ const MESSAGES = defineMessages({
         id: 'iaso.forms.title',
     },
     userRoles: {
-        defaultMessage: 'User roles',
+        defaultMessage: 'User Roles',
```
beygorghor, line 37: Same

hat/locale/fr/LC_MESSAGES/django.po
```diff
@@ -604,6 +604,13 @@ msgstr "Réinitialisation du mot de passe sur %(site_name)s"
 #~ msgid "Id"
 #~ msgstr "Id"
 
+#~ msgid "Reject Org Unit"
```
beygorghor, line 607: Same

hat/locale/fr/LC_MESSAGES/django.po
```diff
@@ -604,6 +604,13 @@ msgstr "Réinitialisation du mot de passe sur %(site_name)s"
...
+#~ msgstr "Rejeter l'unité d'organisation"
+
+#~ msgid "Validation Status changed successfully"
```
beygorghor: Same

beygorghor (changes requested):

beygorghor:
Also i would double check what is happening with the multi selection tool, if you org unit is selected manually (not wit the select all), it should be removed from the selection

hat/assets/js/apps/Iaso/domains/orgUnits/components/ActionCell.tsx
quang-le: I would move this function out of `cell` into it's own `useCallback`. It would be more legible

hat/assets/js/apps/Iaso/domains/orgUnits/hooks.ts
quang-le: My rule of thumb when adding parameters, is that once you reach 3, you should refactor to pass an object, especially when you have optional parameters like here. It saves you from having to pass `undefined`. Not blocking, but cleaner

hat/assets/js/apps/Iaso/domains/payments/messages.ts
Crebert08: I changed this because the id of the messages being the same but with different defaultMessage made the CI fail. So I changed the defaultMessage to match

hat/assets/js/apps/Iaso/domains/orgUnits/messages.ts
Crebert08: Same @quang-le

quang-le (changes requested):
There were still requested changes:
- use `IntlMessage` i.o `any` for `successMessage`
- revert the case changes in the translations

please also apply the changes from my comments

hat/assets/js/apps/Iaso/constants/messages.js
Crebert08: I changed this because the id of the messages being the same but with different defaultMessage made the CI fail. So I changed the defaultMessage to match

hat/assets/js/apps/Iaso/domains/orgUnits/components/ActionCell.tsx
```diff
@@ -1,15 +1,30 @@
...
 type Props = {
     orgUnit: OrgUnit;
 };
 
 export const ActionCell: FunctionComponent<Props> = ({ orgUnit }) => {
+    const { mutateAsync: saveOu, isLoading: isSaving } = useSaveOrgUnit(
+        undefined,
+        ['orgunits'],
+        MESSAGES.orgUnitRejected,
+    );
```
quang-le, line 18: refactor this to avoid passing undefined (cf previous review)

quang-le (changes requested):

quang-le (approved):

beygorghor (approved):

PR #2731, opened 2026-02-12
IA-4770: Remove old references to flake8

hugo093:

We still had setup.cfg file that was used only for flake8 (which we removed when switching to ruff). We also had flake8 commands in entrypoint.sh that had to be replaced with ruff commands. 

### Related JIRA tickets

IA-4770

## Changes

- Delete setup.cfg
- Replace flake8 commands by ruff commands

## How to test

Run the docker in test or test_lint

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -41,7 +41,7 @@ case "$1" in
   "test" )
     export TESTING=true
     # Linting tasks first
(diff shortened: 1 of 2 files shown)


tdethier (approved):
LGTM 👍

PR #2732, opened 2026-02-12
IA-4768: Modify handleOrgUnitChange output to return a string and not an array

hugo093:

On the [duplicates page](https://iaso-staging.bluesquare.org/dashboard/entities/duplicates/), picking a location for filtering and hitting search triggers errors when calling the api because the URL would contain `org_unit/["x","y"]/` instead of `org_unit/x,y/`

### Related JIRA tickets

IA-4768

## Changes

* Change the filter state function so it outputs a string instead of an array
* Some type fixing
* Added a todo for what seems to be an unused function

## How to test

Manual testing, select any location in the corresponding filter, it shouldn't return an error anymore upon clicking the search button.

### DIFF
--- a/hat/assets/js/apps/Iaso/hooks/useFilterState.ts
+++ b/hat/assets/js/apps/Iaso/hooks/useFilterState.ts
@@ -201,41 +201,6 @@ export const useMultiTreeviewFilterState = ({
     );
 };
 
-type TreeviewArgs = {
-    paramId: string | undefined;
-    handleChange: (key: string, value: (string | number)[] | undefined) => void;
-};
-
-type TreeviewFilter = {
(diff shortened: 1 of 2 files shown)


hat/assets/js/apps/Iaso/hooks/useFilterState.ts
```diff
@@ -211,6 +211,7 @@ type TreeviewFilter = {
     handleOrgUnitChange: (orgUnit: OrgUnit | undefined) => void;
 };
 
+// todo : remove this, it seems unused ?
```
quang-le: You can remove it 🪓

Phil-V (approved):
Well done! Thanks for the fix 👍

PR #2741, opened 2026-02-13
IA-4719: add missing translations

quang-le:
Some project feature flag translations were missing

### Related JIRA tickets

IA-4719

## Changes

## How to test

Go to projects
Switch language to FR
click create
go to feature flags tab ("Options")
Expand all sections:
- All feature flags should be translated
- all feature flags should have a translated tooltip (hover on left icon) 

## Notes 

Expected translations available in the Jira ticket


Crebert08:
There is a missing tooltip message missing on the: `Mobile: Search through children in OrgUnit tree` feature flags

https://github.com/user-attachments/assets/7e85c94d-5462-4f12-8b8c-68b58c194d78

Otherwise everything else LGTM !

PR #2745, opened 2026-02-16
[IA-4596] don't allow instance creation

Crebert08:

This PR is about solving the issue of a user trying to create an instance from a form without latest version.

### Related JIRA tickets

[IA-4596](https://bluesquare.atlassian.net/browse/IA-4596)

## Changes

Exported the props from `hat/assets/js/apps/Iaso/domains/instances/Components/CreateReassignDialogComponent.tsx` to pass them in the modal in `hat/assets/js/apps/Iaso/domains/instances/Components/CreateSubmissionModal.tsx`. This allowed the definition of the ButtonProps with a disabled attribute which is used to check if a form has a version or not in `hat/assets/js/apps/Iaso/domains/forms/components/FormActions.tsx`

## How to test

- Start the app with `docker compose -f docker-compose.yml -f docker/docker-compose-enketo.yml up`
- Go to the page of the Forms list
- Create a form but do not create a version
The 'add submission' button should be disabled

## Print screen / video

Screenshot from 2026-02-16 14-34-22

## Notes

## Doc

[IA-4596]: https://bluesquare.atlassian.net/browse/IA-4596?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/forms/components/CreateSubmissionModal/CreateSubmissionModalButton.tsx
+++ b/hat/assets/js/apps/Iaso/domains/forms/components/CreateSubmissionModal/CreateSubmissionModalButton.tsx
@@ -1,16 +1,21 @@
 import React, { FunctionComponent } from 'react';
-import { IconButton } from 'bluesquare-components';
 import PublishIcon from '@mui/icons-material/Publish';
+import { IconButton } from 'bluesquare-components';
 import MESSAGES from '../../messages';
 
-type Props = { onClick: () => void };
+type Props = {
+    onClick: () => void;
+    disabled?: boolean;
+};
 
 export const CreateSubmissionModalButton: FunctionComponent<Props> = ({
     onClick,
+    disabled,
 }) => {
     return (
         <IconButton
             onClick={onClick}
+            disabled={disabled}
             overrideIcon={PublishIcon}
             tooltipMessage={MESSAGES.addSubmissionForForm}
         />
--- a/hat/assets/js/apps/Iaso/domains/forms/components/CreateSubmissionModal/CreateSubmissionModalButton.tsx
+++ b/hat/assets/js/apps/Iaso/domains/forms/components/CreateSubmissionModal/CreateSubmissionModalButton.tsx
@@ -5,12 +5,12 @@ import MESSAGES from '../../messages';
 
 type Props = {
     onClick: () => void;
-    disabled?: boolean;
+    disabled: false;
 };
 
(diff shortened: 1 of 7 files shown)


hat/assets/js/apps/Iaso/domains/forms/components/CreateSubmissionModal/CreateSubmissionModalButton.tsx
```diff
@@ -1,16 +1,21 @@
...
 
-type Props = { onClick: () => void };
+type Props = {
+    onClick: () => void;
+    disabled?: boolean;
+};
 
 export const CreateSubmissionModalButton: FunctionComponent<Props> = ({
     onClick,
+    disabled,
```
quang-le: please pass a default value

hat/assets/js/apps/Iaso/domains/instances/components/CreateReAssignDialogComponent.tsx
```diff
@@ -164,4 +164,4 @@ export const ReAssignDialog = makeFullModal(
     UpdateIcon,
 );
 
-export { CreateReAssignDialog };
+export { CreateReAssignDialog, Props };
```
quang-le: `Props` is not name suitable for export, you can use it if you're using your type only in your component, but if you export it you need to be more specific

quang-le (changes requested):

quang-le (changes requested):
(and tell the user that there's not yet xlsform uploaded) ==> This part of the feature is missing. 
Please add a tooltip with the appropriate message when the icon button is disabled

hat/assets/js/apps/Iaso/domains/forms/components/CreateSubmissionModal/CreateSubmissionModalButton.tsx
```diff
@@ -5,12 +5,12 @@ import MESSAGES from '../../messages';
 
 type Props = {
     onClick: () => void;
-    disabled?: boolean;
+    disabled: false;
```
quang-le: The previous typing was correct: you set a default value, but we can still pass `true`

quang-le:
```suggestion
    disabled?:boolean;
```

quang-le (approved):

PR #2747, opened 2026-02-16
WC2-875: Add command to clean-up instance duplicates

bmonjoie:

Some instances have duplicates instances (Instance with the same UUID). This adds a command and a celery task to fix the issue.

### Related JIRA tickets

WC2-875

## Changes

The command looks for duplicate. Then, for each one, it tries to merge them the best it can.

## How to test

With a dump of a database with duplicates. Either execute the command or launch the celery task.

### DIFF
--- a/iaso/management/commands/clean_up_duplicate_submissions.py
+++ b/iaso/management/commands/clean_up_duplicate_submissions.py
@@ -54,7 +54,7 @@ class Command(BaseCommand):
         for index, duplicate in enumerate(duplicates):
             has_content = Instance.objects.filter(
                 uuid=duplicate["uuid"], file_name=duplicate["file_name"], json__isnull=False
-            )
+            ).order_by("pk")
             if has_content.count() == 1:
                 self.stdout.write(
                     f"{index}. Duplicate with uuid '{duplicate['uuid']}' has only one instance with content."
@@ -74,7 +74,11 @@ class Command(BaseCommand):
                 multiple_content += 1
             else:
                 self.stdout.write(f"{index}. Duplicate with uuid '{duplicate['uuid']}' has no instances with content.")
-                first = Instance.objects.filter(uuid=duplicate["uuid"], file_name=duplicate["file_name"]).first()
+                first = (
+                    Instance.objects.filter(uuid=duplicate["uuid"], file_name=duplicate["file_name"])
+                    .order_by("pk")
+                    .first()
+                )
                 # First, let's keep only one
                 self._delete_instances(uuid=duplicate["uuid"], file_name=duplicate["file_name"], first_id=first.id)
                 # We should look if there are Instances with the same `file_name` but no `uuid` and reimport them
--- a/iaso/management/commands/clean_up_duplicate_submissions.py
+++ b/iaso/management/commands/clean_up_duplicate_submissions.py
@@ -1,0 +1,91 @@
+from django.core.management import BaseCommand
+from django.db import transaction
+from django.db.models import Count
+
+from iaso.models import Entity, Instance
+
+
+class Command(BaseCommand):
+    help = "Clean-up duplicate submissions"
+
+    @transaction.atomic
+    def handle(self, *args, **options):
+        self.stdout.write("Computing duplicate submissions...")
+        duplicates = Instance.objects.values("uuid", "file_name").annotate(total=Count("uuid")).filter(total__gt=1)
+        if len(duplicates) <= 0:
+            self.stdout.write(self.style.SUCCESS("No duplicate submissions found!"))
+            return
+
+        self.stdout.write(self.style.ERROR(f"Duplicates: {len(duplicates)}"))
+        single_content, multiple_content, no_content_no_file, no_content_with_file = 0, 0, 0, 0
+        for index, duplicate in enumerate(duplicates):
+            has_content = Instance.objects.filter(
+                uuid=duplicate["uuid"], file_name=duplicate["file_name"], json__isnull=False
+            )
+            if has_content.count() == 1:
+                self.stdout.write(
+                    f"{index}. Duplicate with uuid '{duplicate['uuid']}' has only one instance with content."
+                )
+                self._delete_instances(
+                    uuid=duplicate["uuid"], file_name=duplicate["file_name"], first_id=has_content.first().id
+                )
+                single_content += 1
+            elif has_content.count() > 1:
+                self.stdout.write(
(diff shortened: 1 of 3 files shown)


iaso/management/commands/clean_up_duplicate_submissions.py
```diff
@@ -0,0 +1,91 @@
...
+                    uuid=duplicate["uuid"], file_name=duplicate["file_name"], first_id=has_content.first().id
+                )
+                multiple_content += 1
+            else:
+                self.stdout.write(f"{index}. Duplicate with uuid '{duplicate['uuid']}' has no instances with content.")
+                first = Instance.objects.filter(uuid=duplicate["uuid"], file_name=duplicate["file_name"]).first()
+                # First, let's keep only one
+                self._delete_instances(uuid=duplicate["uuid"], file_name=duplicate["file_name"], first_id=first.id)
+                # We should look if there are Instances with the same `file_name` but no `uuid` and reimport them
+                if duplicate["file_name"] is None:
```
madewulf, line 85: I don't get how you get data in that state, but I'm assuming you've seen an example on one of the databases?

bmonjoie: No, I didn't get the data in that state in the dump I have received, that's why I "log" it as an error afterward. But, by adding this check, if it were to happen, it would not crash my command.

iaso/management/commands/clean_up_duplicate_submissions.py
```diff
@@ -0,0 +1,91 @@
...
+                    self.stdout.write(self.style.ERROR(" - file_name is None, that's odd!"))
+                    continue
+                # We should use the content of the latest file received.
+                file_name = Instance.objects.filter(uuid__isnull=True, file_name=duplicate["file_name"]).order_by(
```
madewulf: The file_name vairable could be renamed file_name_qs to express that it's a queryset

iaso/management/commands/clean_up_duplicate_submissions.py
```diff
@@ -0,0 +1,149 @@
...
+                    uuid=duplicate["uuid"], file_name=duplicate["file_name"], first_id=has_content.first().id
+                )
+                single_content += 1
+            elif has_content.count() > 1:
+                self.stdout.write(
+                    f"{index}. Duplicate with uuid '{duplicate['uuid']}' has {has_content.count()} instances with content."
+                )
+                # Based on the code of `Instance.import_data`, the first one is always the one that has the data.
+                self._delete_instances(
+                    uuid=duplicate["uuid"], file_name=duplicate["file_name"], first_id=has_content.first().id
```
tdethier, line 72: You do `has_content.first()` but your `has_content` queryset doesn't have an `.order_by()`. I'm afraid you can't know for sure which one you're going to get here

iaso/management/commands/clean_up_duplicate_submissions.py
```diff
@@ -0,0 +1,149 @@
...
+            self.stdout.write(self.style.SUCCESS(f" - Multiple instances with content corrected: {multiple_content}"))
+            self.stdout.write(self.style.SUCCESS(f" - No instances with no content: {no_content}"))
+
+    def _delete_instances(self, uuid: str, file_name: Union[str, None], first_id: str):
+        to_delete = Instance.objects.filter(uuid=uuid).exclude(id=first_id)
+        if file_name:
+            to_delete = to_delete.filter(file_name=file_name)
+        self.stdout.write(self.style.WARNING(f" - Deleting {to_delete.count()} instances and linked entities"))
+        Entity.objects.filter(attributes__in=to_delete).delete()
+        to_delete.delete()
```
tdethier, line 153: Could these instances also have `InstanceFile` objects that need to be deleted? I'm not sure if that scenario can happen with the issue we've had

iaso/management/commands/clean_up_duplicate_submissions.py
```diff
@@ -0,0 +1,149 @@
...
+                    f"{index}. Duplicate with uuid '{duplicate['uuid']}' has {has_content.count()} instances with content."
+                )
+                # Based on the code of `Instance.import_data`, the first one is always the one that has the data.
+                self._delete_instances(
+                    uuid=duplicate["uuid"], file_name=duplicate["file_name"], first_id=has_content.first().id
+                )
+                multiple_content += 1
+            else:
+                self.stdout.write(f"{index}. Duplicate with uuid '{duplicate['uuid']}' has no instances with content.")
+                first = Instance.objects.filter(uuid=duplicate["uuid"], file_name=duplicate["file_name"]).first()
```
tdethier: Same comment about `.first()` and `.order_by()`

plugins/wfp/admin.py
```diff
@@ -73,7 +99,7 @@ def queryset(self, request, queryset):
 class BeneficiaryAdmin(admin.ModelAdmin):
     list_filter = ("birth_date", "gender", "account", "guidelines", ProgrammeType)
     list_display = ("id", "birth_date", "gender", "account", "guidelines")
-    actions = [create_uuid_index_action]
+    actions = [create_uuid_index_action, clean_up_duplicates_action, clean_up_duplicates_action_dry_run]
```
tdethier, line 102: TIL

tdethier (approved):
I didn't test that because I don't have a WFP setup, but it looks good. I think there's just an `order_by()` missing

iaso/management/commands/clean_up_duplicate_submissions.py
bmonjoie: I don't think we'll have `InstanceFile`  as there are no attachment in WFP

tdethier: Ok, good

tdethier (approved):
LGTM 👍

iaso/management/commands/clean_up_duplicate_submissions.py
madewulf: This does not seem resolved

bmonjoie: I added `order_by("pk")` so it is resolved. @tdethier were you ok with how I resolved this?

tdethier: Yeah it's ok, unless @madewulf  spotted something I missed

madewulf (approved):

madewulf (approved):

PR #2750, opened 2026-02-17
IA-4776: Implemented data models for validation workflow

hugo093:

Data modelling for the new validation workflow

### Related JIRA tickets

IA-4776

## Changes

* Added a base abstract CreatedAndUpdatedModel
* Added django-autoslug
* Created the main validation workflow models

## Notes 

Still WIP


iaso/models/validation_workflow/run_instances.py
```diff
@@ -0,0 +1,68 @@
...
+
+    workflow = models.ForeignKey(RISPWorkflow, on_delete=models.CASCADE)
+
+    status = models.CharField(max_length=20, choices=WorkflowStatus.choices,
+                              default=WorkflowStatus.ACTIVE)
+
+    # we attach here any object, can be a form submission, an orgunit, whatever.
+    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
+    object_id = models.CharField(max_length=500, blank=True)
+    content_object = GenericForeignKey("content_type", "object_id")
```
bmonjoie, line 37: As discussed, I am not a fan of `GenericForeignKey`, it lowers the performances and, AFAIK, doesn't give proper related fields on the other model. 
I don't think we are going to have many (10+) different kinds of objects we want to validate so I would use nullable ForeignKeys instead.

iaso/models/risp/__init__.py
```diff
@@ -0,0 +1,11 @@
+from .templates import RISPWorkflow, RISPNode, RISPTransition
+from .run_instances import RISPWorkflowInstance, RISPNodeInstance, RISPTransitionInstance
+
+__all__ = [
+    "RISPNode",
+    "RISPNodeInstance",
+    "RISPTransition",
+    "RISPTransitionInstance",
+    "RISPWorkflow",
+    "RISPWorkflowInstance",
```
bmonjoie: RISP is the client's account name. I don't think we should prefix any of the classes with it as this is going to be used by all clients.

iaso/models/risp/run_instances.py
```diff
@@ -0,0 +1,68 @@
+from django.contrib.contenttypes.fields import GenericForeignKey
+from django.contrib.contenttypes.models import ContentType
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+from .templates import RISPWorkflow, RISPNode
+from ..common import CreatedAndUpdatedModel
+
+
+class RISPWorkflowInstance(CreatedAndUpdatedModel):
```
bmonjoie: `Instance` is also the name for `Form`'s submissions. I think it's confusing to use it here.

Maybe `ObjectValidationWorkflow` ?

iaso/models/validation_workflow/run_instances.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+class RISPNodeInstance(CreatedAndUpdatedModel):
+    """
+    Represents a running instance of the workflow
+    """
```
bmonjoie, line 61: This is the same description as above

iaso/models/risp/run_instances.py
```diff
@@ -0,0 +1,68 @@
...
+    status = models.CharField(max_length=20, choices=WorkflowStatus.choices,
+                              default=WorkflowStatus.ACTIVE)
+
+    # we attach here any object, can be a form submission, an orgunit, whatever.
+    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
+    object_id = models.CharField(max_length=500, blank=True)
+    content_object = GenericForeignKey("content_type", "object_id")
+
+
+class RISPNodeInstance(CreatedAndUpdatedModel):
```
bmonjoie: Same remark about `Instance`.

Maybe `ObjectValidationWorkflowNode`

iaso/models/risp/run_instances.py
```diff
@@ -0,0 +1,68 @@
...
+
+    node = models.ForeignKey(RISPNode, on_delete=models.CASCADE)
+
+    status = models.CharField(
+        choices=Status.choices,
+        default=Status.ACTIVE,
+    )
+
+
+class RISPTransitionInstance(CreatedAndUpdatedModel):
```
bmonjoie: Same remark as before.

Maybe `ObjectValidationWorkflowTransition`

iaso/models/risp/run_instances.py
```diff
@@ -0,0 +1,68 @@
...
+        PENDING = "pending", _("Pending")
+        APPROVED = "approved", _("Approved")
+        REJECTED = "rejected", _("Rejected")
+        CANCELLED = "cancelled", _("Cancelled") # in case a parallel branch with higher prio went over
+
+    node_instance = models.ForeignKey(RISPNodeInstance, on_delete=models.CASCADE, related_name='transition_instances')
+
+    status = models.CharField(max_length=20, choices=Status.choices, default=Status.PENDING)
+
+    rejection_comment = models.TextField(blank=True)
```
bmonjoie: You should add a `updated_by` field

iaso/models/risp/templates.py
```diff
@@ -0,0 +1,86 @@
...
+
+from iaso.models import UserRole
+from iaso.models.common import CreatedAndUpdatedModel
+
+
+class RISPWorkflow(CreatedAndUpdatedModel):
+    """
+    The main workflow object that is a template, a static definition of a workflow
+    """
+    uuid = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
```
bmonjoie: In this case, I think it's OK to use an integer as those are created only by the backend.

iaso/models/risp/templates.py
```diff
@@ -0,0 +1,86 @@
...
+
+    name = models.CharField(max_length=256)
+    slug = AutoSlugField(populate_from="name", unique=True)
+    description = models.TextField(blank=True, max_length=1024)
+
+    # is_active = models.BooleanField(default=True) => SoftDeletableModel ? needed ?
```
bmonjoie: I would go with `SoftDeletableModel` as this is what is done everywhere else in the code base.

iaso/models/risp/templates.py
```diff
@@ -0,0 +1,86 @@
...
+
+    owner = models.ForeignKey(get_user_model(), null=True, blank=True, on_delete=models.SET_NULL)
```
bmonjoie: I'm not sure if this is useful. Probably a `created_by` and a `updated_by` fields make more sense to me

iaso/models/risp/templates.py
```diff
@@ -0,0 +1,86 @@
...
+from django.contrib.contenttypes.models import ContentType
+from django.db import models
+from autoslug import AutoSlugField
+from django.utils.translation import gettext_lazy as _
+
+from iaso.models import UserRole
+from iaso.models.common import CreatedAndUpdatedModel
+
+
+class RISPWorkflow(CreatedAndUpdatedModel):
```
bmonjoie: I would name it `ValidationWorkflow`

iaso/models/risp/templates.py
```diff
@@ -0,0 +1,86 @@
...
+        return self.nodes.get(node_type=RISPNode.NodeType.START)
+
+    def get_end_node(self):
+        return self.nodes.get(node_type=RISPNode.NodeType.END)
+
+    def is_entity_allowed(self, entity):
+        ct = ContentType.objects.get_for_model(entity)
+        return self.allowed_content_types.filter(id=ct.id).exists()
+
+class RISPNode(CreatedAndUpdatedModel):
```
bmonjoie: I would name it `ValidationWorkflowNode`

iaso/models/risp/templates.py
```diff
@@ -0,0 +1,86 @@
...
+        choices=MergeStrategy.choices,
+        blank=True,
+        null=True
+    )
+
+    def __str__(self):
+        return f"{self.workflow_id} - {self.name}"
+
+
+class RISPTransition(CreatedAndUpdatedModel):
```
bmonjoie: I would name it `ValidationWorkflowTransition`.

Even though, I'm not sure I understand the role of this class. I think this class (and the corresponding "ObjectValidationWorkflowTransition") should be merged directly into the `ValidationWorkflowNode` class. To me, it adds unnecessary complexity but I might be missing something.

iaso/migrations/0365_validationstateobject_validationstatetemplate_and_more.py
```diff
@@ -0,0 +1,205 @@
...
+
+from django.conf import settings
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+        ("contenttypes", "0002_remove_content_type_name"),
+        ("iaso", "0364_alter_metricvalue_metric_type"),
```
tdethier, line 16: I merged a PR with another migration yesterday, you should rebase `develop` and redo this migration file 😅

iaso/models/validation_workflow/run_instances.py
tdethier: I've never used `GenericForeignKey`, so I don't have any strong opinion (maybe I'll have one after I start using it? 😅)

iaso/models/validation_workflow/templates.py
```diff
@@ -0,0 +1,84 @@
...
+    The main workflow object that is a template, a static definition of a workflow
+    """
+
+    uuid = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+
+    name = models.CharField(max_length=256)
+    slug = AutoSlugField(populate_from="name", unique=True)
+    description = models.TextField(blank=True, max_length=1024)
+
+    # is_active = models.BooleanField(default=True) => SoftDeletableModel ? needed ?
```
tdethier: Yeah I can see workflows being deleted by mistake, I would use `SoftDeletableModel`

iaso/models/validation_workflow/templates.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def is_entity_allowed(self, entity):
+        ct = ContentType.objects.get_for_model(entity)
+        return self.allowed_content_types.filter(id=ct.id).exists()
+
+
+class ValidationNode(CreatedAndUpdatedModel):
+    workflow = models.ForeignKey(ValidationWorkflow, on_delete=models.CASCADE, related_name="nodes")
+    name = models.CharField(max_length=256)
+    slug = AutoSlugField(populate_from="name")
```
tdethier, line 71: Interesting! Does this mean that `slug` will be automatically updated when `name` changes or it's just set once on creation?

iaso/models/validation_workflow/templates.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def __str__(self):
+        return f"{self.workflow_id} - {self.name}"
+
+
+class ValidationTransition(CreatedAndUpdatedModel):
+    from_node = models.ForeignKey(ValidationNode, on_delete=models.CASCADE, related_name="from_node")
+    to_node = models.ForeignKey(ValidationNode, on_delete=models.CASCADE, related_name="to_node")
+
+    roles_required = models.ManyToManyField(UserRole, blank=True)
```
tdethier: So, this means that reusing `UserRole` sounds like the best plan to you?

iaso/models/validation_workflow/run_instances.py
```diff
@@ -0,0 +1,68 @@
+from django.contrib.contenttypes.fields import GenericForeignKey
+from django.contrib.contenttypes.models import ContentType
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+
+from ..common import CreatedAndUpdatedModel
+from .templates import ValidationNode, ValidationWorkflow
+
+
+class ValidationWorkflowInstance(CreatedAndUpdatedModel):
```
tdethier: It's annoying to have `Instance` in the name of these models since it already means something else in IASO and is confusing.
Maybe we should remove `Instance` from these names (or maybe it's time to finally rename `Instance` to `Submission` 🤓)

iaso/models/validation_workflow/run_instances.py
```diff
@@ -0,0 +1,68 @@
...
+
+    node = models.ForeignKey(ValidationNode, on_delete=models.CASCADE)
+
+    status = models.CharField(
+        choices=Status.choices,
+        default=Status.ACTIVE,
+    )
+
+
+class ValidationTransitionInstance(CreatedAndUpdatedModel):
```
tdethier: I agree with Benjamin, I'm not sure about this class. I understand the ones in `templates.py`, but I'm not sure about this and `ValidationNodeInstance`.

Maybe there's something that I'm missing

iaso/models/risp/templates.py
tdethier: Yeah I agree, `created_by` and `updated_by` are the names we usually use in IASO whenever we need to store some kind of author/owner

iaso/models/risp/templates.py
tdethier: In IASO, UUIDs are usually generated by the mobile app and stored in parallel to a standard `int` ID that's used mostly for the backend.

When there's no object creation on mobile, we usually don't add UUIDs

tdethier (commented):
I added some comments here and there. 

Overall, it's good, and there are some nice ideas, but I think I'm missing something on the "instance" stuff. Maybe I misunderstood something.

We can discuss that later if you want

iaso/migrations/0365_validationstateobject_validationstatetemplate_and_more.py
hugo093: Will do at the end , thanks :)

iaso/models/validation_workflow/templates.py
hugo093: It can be yes, depends how you configure it. I've added some tests in models so you have an example. Otherwise the doc is quite simple and explicit

iaso/models/validation_workflow/templates.py
hugo093: Don't have any strong opinion on this tbh :smile:

iaso/models/validation_workflow/run_instances.py
hugo093: Understood, I renamed most of the models to have a more meaningful business logic, also tried to avoid using instance

iaso/tests/test_engine/test_validation_workflow.py
```diff
@@ -0,0 +1,767 @@
...
+            ).exists()
+        )
+
+    def test_complete_step_by_approving(self):
+        WorkflowEngine.start(self.workflow, self.user)
+
+        # get the next step pending => how? => need to find a better way ?
+        step_object = ValidationStepObject.objects.filter(step_template=self.step).first()
+
+        WorkflowEngine.complete_step(step_object, True, self.other_user)
```
tdethier, line 97: I would expect the workflow to know where we're currently at

iaso/tests/test_engine/test_validation_workflow.py
```diff
@@ -0,0 +1,767 @@
...
+        self.assertEqual(step_object.status, ValidationStepObjectStatus.REJECTED)
+        self.assertEqual(step_object.updated_by, self.other_user)
+        self.assertEqual(step_object.comment, "A rejection comment")
+
+        # we check the ending state object has not been created
+        self.assertFalse(
+            ValidationStateObject.objects.filter(
+                workflow_object=workflow_object, state_template=self.end_state
+            ).exists()
+        )
```
tdethier, line 144: I'm not sure why there's no object created here

iaso/models/validation_workflow/run_instances.py
madewulf, line 1: I don't understand the name of the file

iaso/models/validation_workflow/run_instances.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+class ValidationWorkflowObjectStatus(models.TextChoices):
+    ACTIVE = "active", _("Active")
+    APPROVED = "approved", _("Approved")
+    REJECTED = "rejected", _("Rejected")
+    CANCELLED = "cancelled", _("Cancelled")
+
+
+class ValidationWorkflowObject(CreatedAndUpdatedModel):
```
madewulf, line 20: I'm quite sure this should be a SoftDeletableModel

madewulf: As probably all the models that are in this PR

iaso/models/validation_workflow/run_instances.py
madewulf: It's a bit of a pain to use since it adds a level of indirection to foreign keys (you have to figure out what the models pointed at is before joining). It's neat as a concept, but I've found them difficult to optimize and handle.

hugo093: As discussed, it's a good  concept when you really don't know how many content types will be linked and/or there will be many content type. Otherwise adding nullable FKs is more than enough

iaso/models/validation_workflow/run_instances.py
madewulf: Ok, since this is not something that users will be allowed to delete, it does make sense that it's not soft deletable

iaso/models/validation_workflow/templates.py
```diff
@@ -0,0 +1,102 @@
...
+class ValidationStepTemplate(CreatedAndUpdatedModel):
+    name = models.CharField(max_length=100)
+    slug = AutoSlugField(populate_from="name", always_update=True)
+
+    from_state = models.ForeignKey(ValidationStateTemplate, on_delete=models.CASCADE, related_name="outgoing")
+    to_state = models.ForeignKey(ValidationStateTemplate, on_delete=models.CASCADE, related_name="incoming")
+
+    roles_required = models.ManyToManyField(UserRole, blank=True)  # todo : define if we want roles or permissions ?
+
+    priority = models.PositiveIntegerField(default=0)
```
madewulf, line 94: I don't understand why we need this and where this notion comes from

iaso/models/validation_workflow/run_instances.py
```diff
@@ -0,0 +1,106 @@
...
+
+from ..common import CreatedAndUpdatedModel
+from .templates import ValidationStateTemplate, ValidationStepTemplate, ValidationWorkflowTemplate
+
+
+class ValidationWorkflowObjectStatus(models.TextChoices):
+    ACTIVE = "active", _("Active")
+    APPROVED = "approved", _("Approved")
+    REJECTED = "rejected", _("Rejected")
+    CANCELLED = "cancelled", _("Cancelled")
```
madewulf, line 17: What does it mean for a validation state to be cancelled, approved or rejected? 
It feels like my states have states ;-)

iaso/engine/validation_workflow.py
```diff
@@ -0,0 +1,359 @@
...
+        """
+        Moving to next state.
+        """
+
+        if not state_object.workflow_object.is_active:
+            return
+
+        state_template = state_object.state_template
+        # transitions = node_object.outgoing_transition_objects.all()
+        #
```
madewulf, line 173: My impression is that commented code should not be here (but I know it's a draft, so, this will probably be removed)

iaso/engine/validation_workflow.py
```diff
@@ -0,0 +1,359 @@
...
+            state_object.status = ValidationStateObjectStatus.COMPLETED
+            state_object.save(update_fields=["status"])
+
+            workflow_object.status = ValidationWorkflowObjectStatus.APPROVED
+            workflow_object.save(update_fields=["status"])
+            return workflow_object
+
+        # Create ValidationStepObject for all steps outgoing from this state
+        with transaction.atomic():
+            ValidationStepObject.objects.bulk_create(
```
madewulf, line 84: I find it a bit wasteful to create in advance all the possible transitions. Why not just create the transition that really happened and compute the possibilities on the fly when presenting to the users (instead of loading that from te database). It will also lends itsefl better to have workflows that are changed on the fly. (If you remove a step in a workflow in the current implementation, you have to remove all the tentative steps in the workflow objects if you want to allow users to adapt)

iaso/engine/validation_workflow.py
```diff
@@ -0,0 +1,359 @@
...
+                    )
+                    for step_template in state_template.outgoing.all()
+                ),
+                batch_size=1000,
+            )
+
+        # START node auto-completes
+        if state_template.state_type == ValidationStateTemplateType.START:
+            state_object.status = ValidationStateObjectStatus.COMPLETED
+            state_object.save(update_fields=["status"])
```
madewulf, line 99: I don't totally see the sense of having states to be completed. I would say that the state of a workflow object is just the last state it was in, all previous states are in the past  and consequently "completed"

iaso/models/validation_workflow/run_instances.py
hugo093: Have 0 inspiration for another name, but i'll think about it. It just represents the _runtime objects_ models linked to the templates

iaso/models/validation_workflow/templates.py
hugo093: It's for future feature: e.g a parallel branch that has priority over other branches and cancel them immediately once it's approved. Still have to write tests to demonstrate and fine tune this

iaso/models/validation_workflow/run_instances.py
hugo093: Not sure to get the question as this is about the workflow and not the state ?

iaso/engine/validation_workflow.py
hugo093: indeed, draft and will be removed :)

iaso/engine/validation_workflow.py
hugo093: I think it will be useful for parallel branches again, and the fact that it can be cancelled. E.g you have a state A that is completed and splits into two branches , one with a higher prio than the other. Your higher prio branch gets approved, there we would just cancel the step/task (without the user even having created it). Meaning that the prior state was completed but the step/task was cancelled. If we do'nt create the tasks in advance, we could end up with an "empty" branch and would have to make a query to deduct it was cancelled. 

But again, this is in case of complex flow, not a single linear workflow

iaso/engine/validation_workflow.py
hugo093: I agree in the case of a single linear flow, not in the case of a complex merge/split one with priorities and so on.

PR #2751, opened 2026-02-17
[SNT-302] add ColorField + tests

tdethier:

add ColorField + tests

### Related JIRA tickets

SNT-302

## Changes

- add `ColorField` 
- add tests to make sure the field behaves properly

## How to test

- it's just a model field, so there's no manual testing for now
- you can create a model with that field if you want to test it

## Print screen / video

/

## Notes

- if this is approved, we should replace existing `color` attributes by this field, because we're using color more and more in IASO (hence this new field) - this should be part of a dedicated ticket
- this is used in an [SNT PR](https://github.com/BLSQ/snt-malaria/pull/190) 

## Doc

/

### DIFF
--- a/iaso/utils/models/color.py
+++ b/iaso/utils/models/color.py
@@ -1,0 +1,30 @@
+import re
+
+from django.core.exceptions import ValidationError
+from django.db import models
+
+
+HEX_COLOR_REGEX = re.compile(r"#[0-9A-Fa-f]{6}$")
+
+
+def validate_hex_color(value):
+    if not HEX_COLOR_REGEX.match(value):
+        raise ValidationError(f"{value} is not a valid hex color code. It should be in the format #RRGGBB.")
+
+
+class ColorField(models.CharField):
+    description = "A color field that accepts hex color codes (e.g., #RRGGBB)."
+
+    def __init__(self, *args, **kwargs):
+        kwargs.setdefault("max_length", 7)  # Length for hex color code (#RRGGBB)
+        validators = kwargs.get("validators", [])
+        if not validators or validate_hex_color not in validators:
+            validators.append(validate_hex_color)
+        kwargs["validators"] = validators
+        super().__init__(*args, **kwargs)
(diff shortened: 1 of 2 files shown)


iaso/utils/models/color.py
```diff
@@ -0,0 +1,30 @@
...
+
+class ColorField(models.CharField):
+    description = "A color field that accepts hex color codes (e.g., #RRGGBB)."
+
+    def __init__(self, *args, **kwargs):
+        kwargs.setdefault("max_length", 7)  # Length for hex color code (#RRGGBB)
+        validators = kwargs.get("validators", [])
+        if not validators or validate_hex_color not in validators:
+            validators.append(validate_hex_color)
+        kwargs["validators"] = validators
```
Phil-V: Small nitpick here: mutating the validators array passed by the user is not ideal.

I think another approach you could take to simplify is to use `default_validators`:

```py
# from django.core.validators import RegexValidator

    default_validators = [
        RegexValidator(
            regex=HEX_COLOR_REGEX,
            message="%(value)s is not a valid hex color code. It should be in the format #RRGGBB.",
        )
    ]

    def __init__(self, *args, **kwargs):
        kwargs.setdefault("max_length", 7)  # Length for hex color code (#RRGGBB)
        super().__init__(*args, **kwargs)
```

iaso/utils/models/color.py
```diff
@@ -0,0 +1,30 @@
...
+        super().__init__(*args, **kwargs)
+
+    def to_python(self, value):
+        value = super().to_python(value)
+        if value:
+            return value.upper()  # Normalize to uppercase
```
Phil-V, line 26: Good thinking 👍

Phil-V (approved):
Great work!

iaso/utils/models/color.py
tdethier: That looks good, good idea :+1:

PR #2756, opened 2026-02-18
Feat/IA-4748 add assignment deletion to UI

quang-le:

Users need to be able to delete assignments

### Related JIRA tickets

IA-4748

## Changes

- Backend:
    - Refactor `bulk_create_assignments` and `bulk_delete_assignments` to remove N+1 queries
    - refactor  `bulk_delete_assignments` to be able to pass user or team
    - update tests
- Front-end 
    - Add Iconbutton to Assignments TeamTable, to delete all assignments for team/user in selected planning
    - Add button to delete all assignments in assignments page
    - Same in planning details page 

## How to test

- Create a planning with a team of users, another with a team of teams
- For each planning: create assignments for several teams/users
- Try all the delete buttons and icon buttons: the assignments should be deleted. If on assignment page, team table and map should refresh

## Print screen / video

https://github.com/user-attachments/assets/25cf6963-de20-4038-b3f7-ee718203b433


iaso/api/microplanning/views.py
```diff
@@ -225,31 +227,64 @@ def bulk_create_assignments(self, request):
...
-                old_value = [AuditAssignmentSerializer(instance=assignment).data]
-
-            assignment.deleted_at = None
+        org_units = serializer.validated_data["org_units"]
+        assignments_to_update = Assignment.objects.select_related("user", "team", "org_unit").filter(
+            planning=planning, org_unit__in=org_units, deleted_at__isnull=True
+        )
+        assignment_values = list(assignments_to_update.values("id", "org_unit__id"))
+        assignment_ids_to_update = [assignment["id"] for assignment in assignment_values]
+        org_units_to_exclude = {assignment["org_unit__id"] for assignment in assignment_values}
```
tdethier: Nitpick: you don't need to traverse the relation to get the ID

```suggestion
        assignment_values = list(assignments_to_update.values("id", "org_unit_id"))
        assignment_ids_to_update = [assignment["id"] for assignment in assignment_values]
        org_units_to_exclude = {assignment["org_unit_id"] for assignment in assignment_values}
```

iaso/api/microplanning/views.py
```diff
@@ -225,31 +227,64 @@ def bulk_create_assignments(self, request):
...
                 content_object=assignment,
                 source="API " + request.method + request.path,
             )
+            audit_for_update.append(audit)
+            # serialize new value? but updated_at will be wrong
+        with transaction.atomic():
+            Assignment.objects.bulk_update(assignments_to_update, ["user", "team"])
+            now = timezone.now().strftime("%Y-%m-%dT%H:%M:%S.%fZ")
+            for modification in audit_for_update:
+                modification.new_value = [{**modification.new_value[0], "updated_at": now}]
```
tdethier: Why don't you set that field when you get `new_value` in the for loop above?

iaso/api/microplanning/views.py
```diff
@@ -225,31 +227,64 @@ def bulk_create_assignments(self, request):
...
+            new_assignment = Assignment(
+                planning=planning, user=user, created_by=requester, org_unit=org_unit, team=team
+            )
+            created_assignments.append(new_assignment)
+
+        with transaction.atomic():
+            new_assignments = Assignment.objects.bulk_create(created_assignments)
+            for new_assignment in new_assignments:
+                new_value = [AuditAssignmentSerializer(instance=new_assignment).data]
+                audit = Modification(  # Bug: assignment has no id at this stage
```
tdethier: It probably does not have an ID yet because it's in a transaction and the transaction is not over yet when you get to this line. You could maybe bulk create assignments in the transaction, then create audit data after the transaction

iaso/api/microplanning/views.py
```diff
@@ -265,41 +300,60 @@ def bulk_delete_assignments(self, request):
...
         planning = serializer.validated_data["planning"]
+        user = serializer.validated_data.get("user", None)
+        team = serializer.validated_data.get("team", None)
         requester = request.user
 
         # Get all assignments for this planning that are not already deleted
-        assignments = Assignment.objects.filter(planning=planning, deleted_at__isnull=True).filter_for_user(requester)
+        assignments = (
+            Assignment.objects.filter(planning=planning, deleted_at__isnull=True)
+            .select_related("user", "team", "org_unit")
```
tdethier, line 309: 👍

iaso/api/microplanning/views.py
```diff
@@ -265,41 +300,60 @@ def bulk_delete_assignments(self, request):
...
                     "deleted_count": 0,
                     "planning_id": planning.id,
+                    "user": user.id if user else None,
                 }
             )
 
+        old_serialized = {a.id: AuditAssignmentSerializer(a).data for a in assignments}
+
         # Store assignment IDs before update for audit trail
         assignment_ids = list(assignments.values_list("id", flat=True))
```
tdethier, line 331: Since you already have IDs in `old_serialized`, you could do
```suggestion
        assignment_ids = old_serialized.keys()
```

iaso/api/microplanning/views.py
```diff
@@ -265,41 +300,60 @@ def bulk_delete_assignments(self, request):
...
-            Modification.objects.create(
-                user=requester,
-                past_value=old_value,
-                new_value=new_value,
-                content_object=assignment,
-                source="API " + request.method + request.path,
-            )
+        with transaction.atomic():
+            deleted_count = assignments.update(deleted_at=timezone.now())
+            updated_assignments = Assignment.objects.in_bulk(assignment_ids)
```
tdethier, line 334: I find the name confusing because the goal is to delete assignments, not update them (even though I understand that we concretely do an update since it's a soft deletable model). I would call this `deleted_assignments`

iaso/tests/api/microplanning/test_microplanning.py
```diff
@@ -1367,6 +1365,127 @@ def test_bulk_delete_assignments(self):
         for assignment in all_assignments:
             self.assertIsNotNone(assignment.deleted_at)
 
+    def test_bulk_delete_assignments_for_team(self):
```
tdethier, line 1368: Nice coverage 👍

iaso/api/microplanning/serializers.py
```diff
@@ -363,18 +363,28 @@ def validate(self, attrs):
 
 
 class BulkDeleteAssignmentSerializer(serializers.Serializer):
```
tdethier, line 365: Ideally, we should also have unit tests that cover this serializer (without interacting with the view), to make sure that the `user`, `planning` and `team` values are properly filtered tenancy-wise

tdethier (approved):
The feature works fine, I managed to bulk delete assignments on the assigments and planning pages 👍 

I had a few issues with the planning creation (frontend form validation seems off), but I don't think it's related to this ticket.

I left some comments with some nitpicks, but I'm already approving

iaso/api/microplanning/views.py
quang-le: It's an old comment, the tests pass now

PR #2757, opened 2026-02-19
IA-4779: fix DeleteInstanceDialog lifecycle

hugo093:

From the form submissions list, it was not possible to bulk delete form submissions twice in a row, at the second time the confirm button on the modal would always be greyed out. 

### Related JIRA tickets

IA-4779

## Changes

* Fix the react lifecycle of the related component
* Adjust the components to be more TS compliant and MUI compliant

## How to test

In form submission UI : 

* Search for entries
* From the sidebar , select all 
* From the sidebar, delete and confirm
* Without reloading the page, do a new search
* From the sidebar, select all then delete
* The Confirm button in the modal should not be greyed out

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/instances/components/DeleteInstanceDialog.tsx
+++ b/hat/assets/js/apps/Iaso/domains/instances/components/DeleteInstanceDialog.tsx
@@ -1,8 +1,9 @@
-import React, { useState, FunctionComponent, useEffect } from 'react';
+import React, { FunctionComponent } from 'react';
 import DeleteIcon from '@mui/icons-material/Delete';
 import RestoreFromTrashIcon from '@mui/icons-material/RestoreFromTrash';
 import { DialogContentText, IconButton } from '@mui/material';
 import { FormattedMessage } from 'react-intl';
+import { useBoundState } from 'Iaso/hooks/useBoundState';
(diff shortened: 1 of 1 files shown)


hat/assets/js/apps/Iaso/domains/instances/components/DeleteInstanceDialog.tsx
```diff
@@ -33,29 +25,34 @@ const DeleteInstanceDialog: FunctionComponent<Props> = ({
...
-    const classes = useStyles();
-    const [allowConfirm, setAllowConfirm] = useState(true);
+    const [allowConfirm, setAllowConfirm] = useState(false);
+
+    useEffect(() => {
+        if (selection?.selectCount) {
+            setAllowConfirm(true);
+        }
+    }, [selection]);
+
```
quang-le: We have a convenience hook for this (no so nice) pattern:
https://github.com/BLSQ/iaso/blob/develop/hat/assets/js/apps/Iaso/hooks/useBoundState.ts

quang-le:
```suggestion
    const [allowConfirm, setAllowConfirm] = useBoundState(
        false,
        Boolean(selection?.selectCount),
    );

```
The import still needs adding

quang-le (changes requested):
Works fine, just use the convenience hook

quang-le (approved):

PR #2758, opened 2026-02-19
[SNT-302] Add JSON validation fields

tdethier:

Add JSON validation at model level and serializer level 

### Related JIRA tickets

SNT-302

## Changes

- add `JSONSchemaValidator` for validating model-level JSON fields
- add `JSONSchemaField` for adding a JSON field with validation to a serializer

## How to test

- not testable yet, they will be used by other future PRs

## Print screen / video
/

## Notes

/

## Doc

### DIFF
--- a/iaso/utils/serializer/json_schema_field.py
+++ b/iaso/utils/serializer/json_schema_field.py
@@ -5,16 +5,18 @@ from rest_framework import serializers
 
 
 class JSONSchemaField(serializers.JSONField):
-    def __init__(self, *args, **kwargs):
-        self.schema = kwargs.pop("schema", None)
+    def __init__(self, schema, *args, **kwargs):
+        if not schema:
+            raise ValueError("A JSON schema must be provided for validation.")
+        self.schema = schema
+
         super().__init__(*args, **kwargs)
 
     def to_internal_value(self, value):
         data = super().to_internal_value(value)
         try:
             jsonschema.validate(data, self.schema)
-        except jsonschema.exceptions.ValidationError as e:
+        except jsonschema.ValidationError as e:
             raise ValidationError(e.message) from e
-        except jsonschema.exceptions.SchemaError as e:
-            raise ValidationError(e.message) from e
+
         return data
--- a/iaso/utils/serializer/json_schema_field.py
+++ b/iaso/utils/serializer/json_schema_field.py
@@ -1,0 +1,20 @@
+import jsonschema
+
+from django.core.exceptions import ValidationError
+from rest_framework import serializers
+
+
+class JSONSchemaField(serializers.JSONField):
+    def __init__(self, *args, **kwargs):
+        self.schema = kwargs.pop("schema", None)
+        super().__init__(*args, **kwargs)
+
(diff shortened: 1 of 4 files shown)


iaso/utils/validators.py
```diff
@@ -0,0 +1,18 @@
+import jsonschema
+
+from django.core.exceptions import ValidationError
+from django.core.validators import BaseValidator
+
+
+class JSONSchemaValidator(BaseValidator):
```
hugo093: can you add a code and default message and use it in your ValidationError ? (like django does on EmailValidator for example)

iaso/utils/serializer/json_schema_field.py
```diff
@@ -0,0 +1,20 @@
...
+        super().__init__(*args, **kwargs)
+
+    def to_internal_value(self, value):
+        data = super().to_internal_value(value)
+        try:
+            jsonschema.validate(data, self.schema)
+        except jsonschema.exceptions.ValidationError as e:
+            raise ValidationError(e.message) from e
+        except jsonschema.exceptions.SchemaError as e:
+            raise ValidationError(e.message) from e
```
hugo093:
```suggestion
        except (jsonschema.exceptions.ValidationError, jsonschema.exceptions.SchemaError) as e:
            raise ValidationError(e.message) from e
```

iaso/utils/validators.py
```diff
@@ -0,0 +1,18 @@
...
+    def __init__(self, schema, message=None):
+        self.schema = schema
+        super().__init__(limit_value=None, message=message)
+
+    def __call__(self, value):
+        try:
+            jsonschema.validate(value, self.schema)
+        except jsonschema.ValidationError as e:
```
hugo093, line 25: same remark as in json_schema_field.py

iaso/utils/serializer/json_schema_field.py
```diff
@@ -0,0 +1,20 @@
+import jsonschema
+
+from django.core.exceptions import ValidationError
+from rest_framework import serializers
+
+
+class JSONSchemaField(serializers.JSONField):
+    def __init__(self, *args, **kwargs):
+        self.schema = kwargs.pop("schema", None)
```
hugo093: imho we should raise an error if schema is None, as the next step to_internal_value would crash with a TypeError (and it doesn't make sense to use jsonschema without schema)

iaso/utils/validators.py
```diff
@@ -0,0 +1,18 @@
+import jsonschema
+
+from django.core.exceptions import ValidationError
+from django.core.validators import BaseValidator
+
+
+class JSONSchemaValidator(BaseValidator):
+    def __init__(self, schema, message=None):
+        self.schema = schema
```
hugo093, line 15: same comment as above, check that schema is not none

hugo093 (changes requested):
Overall, it's good. I think you could already write basic tests to demonstrate the integration with DRF and models.

hugo093 (approved):
LGTM

PR #2760, opened 2026-02-19
IA-4783: Replace "color charfield" by new ColorField

hugo093:

We have a new ColorField that we can use to replace all the CharField that were made for the use of an hex color.

### Related JIRA tickets

IA-4783

## Changes

Went through the code and replaced those CharFields with ColorField, made migrations. 

## Notes

WFP plugin also has a color field set as a CharField, it's not possible to replace it yet considering it takes values like "green", "red" , "blue" instead of hex. e.g :

```python
            whz_color = ""
            if current_visit.get("whz_color", None) == "Y":
                whz_color = "Yellow"
            elif current_visit.get("whz_color", None) == "R":
                whz_color = "Red"
            elif current_visit.get("whz_color", None) == "G":
                whz_color = "Green"
```

A mapping there is needed before making the migrations


edil3ra (approved):
Migration ran properly

quang-le (commented):
Can you open a new ticket for the WFP mappings please?

PR #2765, opened 2026-02-23
chore: Update snt-malaria-budgeting to 0.7.0

bramj:


quang-le (approved):
again? what a chore indeed 😉

PR #2766, opened 2026-02-23
POLIO-2105: File name not shown when uploading file

beygorghor:

image

### Related JIRA tickets

POLIO-2105

## Changes

- allow backend to return files objects
- fix DocumentUploadWithPreview to avoid nesting array when updating files list 
## How to test

Explain how to test your PR.

## Print screen / video

https://github.com/user-attachments/assets/61afa695-0ed6-4f6f-b907-7324afb47ebc


quang-le (commented):
A lil' backend test 😉

hat/assets/js/apps/Iaso/components/files/pdf/DocumentUploadWithPreview.tsx
quang-le, line 77: You need to update this check, otherwise the condition is never true

quang-le (changes requested):

quang-le (changes requested):
The delete button is broken now

https://github.com/user-attachments/assets/e4505ceb-4afd-485a-8420-50d4bc483def

quang-le (approved):

PR #2774, opened 2026-02-27
IA-4762 be able to change save user profile

hugo093:

It wasn't possible to update the user without erasing the password.

### Related JIRA tickets

IA-4762

## Changes

* Ended up refactoring a bit the profiles API : 
  * Using standard DRF ModelViewSet with serializer
  * Added separate actions for export and update password
  * Remove the update color action
  * Splitting and reorganizing the files
  * Adapting tests
* UI changes
 * Added a UserDetailView
 * Adapted the save/create/delete/export calls
 * Added a Save user password functionality

## How to test

related front end views are dashboard/settings/users/management/ and dashboard/settings/users/management/details/accountId/:accountId/userId/:userId 

## Notes

Many things are still left to do imho : 

* We need to agree on what the list and retrieve should return, right now the list is just an array of what the retrieve would do. It's just made that way so in the list view, opening the edit modal does not make a separate call to get initial data.
* Queries need to be optimized
* React tests need to be written
  * Login, saving password (or updating user without password) and relogging is a critical good E2E test imho
  * Wrote some integration test cases already
  * UserDetailView is fully unit tested
* I don't get the use of useSaveCurrentUser and if it could be removed or not. 
* Most of the dialogs don't use Formik


hat/assets/js/apps/Iaso/domains/assignments/components/teams/TeamTable.tsx
```diff
@@ -61,7 +61,7 @@ export const TeamTable: FunctionComponent<Props> = ({
     const { formatMessage } = useSafeIntl();
 
     const { mutate: updateTeam } = useSaveTeam('edit', false);
-    const { mutate: updateUser } = useSaveProfileColor(false);
+    const { mutate: updateUser } = useSaveProfile(undefined, false);
```
beygorghor: ❤️

hat/assets/js/apps/Iaso/domains/users/components/CreateUserDialog.tsx
```diff
@@ -0,0 +1,341 @@
...
+        position: 'relative',
+    },
+    hiddenOpacity: {
+        position: 'fixed',
+        top: 0,
+        left: -5000,
+        zIndex: -10,
+        opacity: 0,
+    },
+}));
```
beygorghor: please use  `const styles: SxStyles = {` instead it has a higher css priority than make styles and classes
see docs/pages/dev/reference/guidelines/front-end/front-end.en.md

hat/assets/js/apps/Iaso/components/Buttons/DeleteButton.tsx
beygorghor, line 1: This can probably go to a higher level of folder to be reusable no ?

hat/assets/js/apps/Iaso/components/Buttons/EditButton.tsx
beygorghor, line 1: same here

hat/assets/js/apps/Iaso/domains/users/components/PermissionsAttribution.tsx
```diff
@@ -74,9 +74,10 @@ const PermissionsAttribution: React.FunctionComponent<Props> = ({
         // Permission list is not displayed for superuser, no need to fetch it from server
         options: { enabled: !isSuperUser },
     });
+
     const setPermissions = useCallback(
         (codeName: string | string[], isChecked: boolean) => {
-            const newUserPerms = [...currentUser.user_permissions.value];
+            const newUserPerms = [...currentUser.user_permissions.value?.map((v: {id: number, name: string} | number) => typeof v === "object" && "id" in v ? v.id : v)];
```
beygorghor: Those three look similar ... not really readable too

hat/assets/js/apps/Iaso/domains/users/components/UserDetailsView.tsx
```diff
@@ -0,0 +1,358 @@
...
+                                        savePassword={savePassword}
+                                        userId={userId}
+                                    />
+                                    {currentUser?.id?.toString() !== userId &&
+                                        userHasOneOfPermissions(
+                                            [
+                                                Permission.USERS_ADMIN,
+                                                Permission.USERS_MANAGEMENT,
+                                            ],
+                                            currentUser,
```
beygorghor: We have a component for thaht fyi => DisplayIfUserHasPerm

hat/assets/js/apps/Iaso/domains/users/components/UserDetailsView.tsx
```diff
@@ -0,0 +1,358 @@
...
+                                                            component="span"
+                                                            data-testid={"user-color-badge"}
+                                                            sx={{
+                                                                ...styles.badge,
+                                                                backgroundColor:
+                                                                profile?.color,
+                                                            }}
+                                                            tabIndex={0}
+                                                        >
+                                                            {' '}
```
beygorghor: Are you sure this is useful ?

hat/assets/js/apps/Iaso/domains/users/components/UserDetailsView.tsx
```diff
@@ -0,0 +1,358 @@
...
+                            <WidgetPaper title={formatMessage(MESSAGES.generalInfo)} data-testid={"general-info-box"}>
+                                {savingProfile ? (
+                                        <Box sx={{my : 2}}>
+                                            <LoadingSpinner absolute={false} fixed={false}/>
+                                        </Box>
+                                    ) :
+
+                                    <Table size="small">
+                                        <TableBody>
+                                            <Row
```
beygorghor: This row component should return a placeholder if the value is empty, avoiding repetition here

hat/assets/js/apps/Iaso/domains/users/components/UserDetailsView.tsx
beygorghor, line 1: This component should probably splited into more smaller components

hat/assets/js/apps/Iaso/libs/Api.ts
```diff
@@ -55,7 +55,7 @@ export const iasoFetch = async (
     const url =
         typeof resource === 'string' ? resource : (resource.url ?? resource);
     const method = init?.method ?? 'GET';
-    console.log('API', init);
```
beygorghor, line 58: ❤️

beygorghor (changes requested):
I really appreciate the user details page.

I found a small bug while searching a user with AsyncSelect on Instances page. Probably due t the payload that is paginated now:

Image
 
Image

Can you make sure this message is not displayed while editing a user and we didn't touch the phone number ?

Image
This dialog can be smaller, also maybe and some labels on those inputs

Image
Maybe also change the title here to User detail ?

Image

A last one, password change button can be orange, using warning color

hat/assets/js/apps/Iaso/domains/assignments/components/teams/TeamTable.tsx
hugo093: Now that i see this, i should change the function definition to avoid this "undefined"..

hat/assets/js/apps/Iaso/components/Buttons/DeleteButton.tsx
hugo093: I would put it in bluesquare-component tbh, but as you install from a specific commit, i will put it somewhere else :)

iaso/locale/fr/LC_MESSAGES/django.po
```diff
@@ -1127,21 +1126,72 @@ msgstr "Virus détecté"
...
 msgstr "Erreur lors de l'analyse"
 
-#~ msgid "This query parameter is required."
-#~ msgstr "Ce paramètre de requête est requis."
+# iaso/api/profiles/views
+msgid "Set up a password for your new account on {domain}"
+msgstr "Configurer un mot de passe pour votre nouveau compte sur {domain}"
+
+#: iaso/templates/emails/create_password_email.html:2
+msgid ""
```
tdethier, line 1134: Did you generate all these new translations with the command `make_messages`?

iaso/api/setup_account.py
```diff
@@ -251,7 +251,7 @@ def create(self, validated_data):
 
         # Send email invitation if requested
         if email_invitation and user.email:
-            from iaso.api.profiles.profiles import ProfilesViewSet
+            from iaso.api.profiles.views import ProfilesViewSet
```
tdethier: I wonder why that was a local import in the first place

hat/settings.py
```diff
@@ -276,6 +276,7 @@
         "BACKEND": "django.template.backends.django.DjangoTemplates",
         "DIRS": [
             "./hat/templates",
+            os.path.join(BASE_DIR, "iaso", "templates"),
```
tdethier, line 279: 👍

iaso/api/profile_logs/serializers.py
```diff
@@ -63,7 +18,7 @@ class NestedUserForListSerializer(serializers.ModelSerializer):
     user_id = serializers.SerializerMethodField(read_only=True)
 
     class Meta:
-        model = User
+        model = get_user_model()
```
tdethier, line 21: I don't think we really need to call this since we did not implement a custom `User` class, but I don't mind

iaso/api/profile_logs/serializers.py
```diff
@@ -1,56 +1,11 @@
...
 
-import django_filters
-
-from django.conf import settings
-from django.contrib.auth.models import User
-from django.db.models import Q
-from django.db.models.query import QuerySet
-from django.utils.translation import gettext_lazy as _
-from rest_framework import filters, serializers
+from django.contrib.auth import get_user_model
```
tdethier, line 3: Thanks for splitting this file in multiple ones 👍

iaso/api/profiles/serializers/common.py
```diff
@@ -0,0 +1,48 @@
...
+        super().__init__(**kwargs)
+
+    def to_internal_value(self, data):
+        value = super().to_internal_value(data)
+
+        if isinstance(value, PhoneNumber):
+            return value
+
+        serializer = self.parent
+        country_code = serializer.initial_data.get(self.country_code_field)
```
tdethier, line 26: I'm always a bit scared to see `initial_data`  being used, I guess we don't have any choice because `to_internal_value` is called when data is being validated?

iaso/api/profiles/serializers/common.py
```diff
@@ -0,0 +1,48 @@
...
+        if not number.is_valid():
+            self.fail("invalid")
+
+        return number
+
+    def to_representation(self, value):
+        if not value:
+            return None
+
+        return value.as_e164
```
tdethier, line 48: TIL

iaso/api/profiles/serializers/common.py
```diff
@@ -0,0 +1,48 @@
...
+        self.country_code_field = country_code_field
+        super().__init__(**kwargs)
+
+    def to_internal_value(self, data):
+        value = super().to_internal_value(data)
+
+        if isinstance(value, PhoneNumber):
+            return value
+
+        serializer = self.parent
```
tdethier, line 25: It's interesting to interact with other fields this way, I've never done that. If not done properly it can quickly become a mess I guess 😅

iaso/api/profiles/serializers/list.py
```diff
@@ -0,0 +1,100 @@
...
+        return self._get_user_infos(obj).email
+
+    def get_country_code(self, obj):
+        return region_code_for_number(obj.phone_number).lower() if obj.phone_number else None
+
+    def get_editable_org_unit_type_ids(self, obj):
+        try:
+            editable_org_unit_type_ids = obj.annotated_editable_org_unit_types_ids
+        except AttributeError:
+            editable_org_unit_type_ids = [out.pk for out in obj.editable_org_unit_types.all()]
```
tdethier: Nitpick to avoid fetching everything with `all()`: 
```suggestion
            editable_org_unit_type_ids = list(obj.editable_org_unit_types.values_list("id", flat=True))
```

iaso/api/profiles/serializers/retrieve.py
```diff
@@ -0,0 +1,322 @@
...
+    is_superuser = serializers.BooleanField(source="user.is_superuser", read_only=True)
+    user_roles = serializers.PrimaryKeyRelatedField(source="get_ordered_user_roles", many=True, read_only=True)
+    user_roles_permissions = NestedUserRoleSerializer(many=True, read_only=True, source="get_ordered_user_roles")
+
+    country_code = serializers.SerializerMethodField()
+    phone_number = serializers.CharField(source="phone_number.as_e164", read_only=True)
+
+    projects = NestedProjectSerializer(many=True, read_only=True, source="get_ordered_projects")
+
+    other_accounts = NestedAccountSerializer(many=True, read_only=True, source="user.tenant_user.get_other_accounts")
```
tdethier, line 239: I don't think this is necessary, does the frontend need to know all the accounts of a user when editing a profile?

iaso/api/profiles/serializers/retrieve.py
```diff
@@ -0,0 +1,322 @@
...
+
+    editable_org_unit_type_ids = serializers.SerializerMethodField()
+    user_roles_editable_org_unit_type_ids = serializers.ListField(
+        source="get_user_roles_editable_org_unit_type_ids", read_only=True
+    )
+    account = NestedAccountExtendedSerializer(read_only=True)  # todo : Account.as_small_dict
+    org_units = NestedOrgUnitSerializer(many=True, source="get_ordered_org_units")
```
tdethier, line 246: I don't think we need to go that deep for both attributes. Usually, just returning IDs to the front is enough (and the front makes necessary retrieve calls to get more information). If you just need a name/color/other "simple" field like that, then it's fine but I don't think we need all this information

@beygorghor any opinion? does the front need to receive so many fields? check the content of both nested serializers and other sub nested serializers as well

iaso/api/profiles/serializers/create.py
```diff
@@ -0,0 +1,210 @@
...
+from django.utils.translation import gettext_lazy as _
+from rest_framework import serializers
+
+from iaso.api.profiles.serializers.common import CountryAwarePhoneNumberField
+from iaso.models import OrgUnit, OrgUnitType, Profile, Project, TenantUser, UserRole
+from iaso.models.tenant_users import UserCreationData, UsernameAlreadyExistsError
+from iaso.utils.colors import COLOR_FORMAT_ERROR, validate_hex_color
+
+
+class ProfileCreateSerializer(serializers.ModelSerializer):
```
tdethier, line 12: These serializers are so huge, it's a good idea to split them into multiple files 👍

iaso/api/profiles/serializers/create.py
```diff
@@ -0,0 +1,210 @@
...
+        # we get the account
+        account = self._get_account()
+        user_roles = data.get("user_roles")
+
+        if not user_roles:
+            return
+
+        if UserRole.objects.filter(pk__in=[user_role.pk for user_role in user_roles]).exclude(account=account).exists():
+            raise serializers.ValidationError(
+                {"user_roles": _("One or more user roles do not belong to the provided account.")}
```
tdethier: Why not both in a single method?

iaso/api/profiles/serializers/create.py
```diff
@@ -0,0 +1,210 @@
...
+    user_roles = serializers.PrimaryKeyRelatedField(
+        allow_empty=True, allow_null=True, many=True, queryset=UserRole.objects.all(), required=False
+    )
+    projects = serializers.PrimaryKeyRelatedField(
+        allow_empty=True, allow_null=True, queryset=Project.objects.all(), required=False, many=True
+    )
+    org_units = serializers.PrimaryKeyRelatedField(
+        allow_empty=True, allow_null=True, many=True, queryset=OrgUnit.objects.all(), required=False
+    )
+    editable_org_unit_type_ids = serializers.PrimaryKeyRelatedField(
```
tdethier, line 32: For all `PrimaryKeyRelatedField`, we usually restrict the queryset in `__init__` so that you don't need to check for exlusions in `validate_...` - see [example here](https://github.com/BLSQ/iaso/blob/cedb765da66ddd85bfb2387a8b9cad27b38c23b6/iaso/api/metrics/serializers.py#L122)

iaso/api/profiles/serializers/create.py
```diff
@@ -0,0 +1,210 @@
...
+                "account_id": account.id,
+                "user_id": user.id,
+            }
+        )
+
+        return profile
+
+    def to_representation(self, instance):
+        """
+        Legacy compatibility : imho those fields shouldn't be there in the create response.
```
tdethier: Yeah I'm not sure that they are useful for the frontend either. Maybe we could get rid of that

iaso/api/profiles/serializers/update.py
```diff
@@ -0,0 +1,175 @@
...
+    )
+
+    user_roles = serializers.PrimaryKeyRelatedField(
+        allow_empty=True, allow_null=True, many=True, queryset=UserRole.objects.all(), required=False
+    )
+
+    org_units = serializers.PrimaryKeyRelatedField(
+        allow_empty=True, allow_null=True, many=True, queryset=OrgUnit.objects.all(), required=False
+    )
+    editable_org_unit_type_ids = serializers.PrimaryKeyRelatedField(
```
tdethier, line 46: Same comment about `PrimaryKeyRelatedField`

iaso/api/profiles/serializers/update.py
```diff
@@ -0,0 +1,175 @@
...
+            "email",
+            "color",
+            "phone_number",
+            "country_code",
+            "dhis2_id",
+            "user_permissions",
+            "org_units",
+            "editable_org_unit_type_ids",
+            "projects",
+            "user_roles",
```
tdethier, line 73: When I see this, I'm thinking that maybe there could be a base write serializer for everything that's common to both create and update, but I'm not sure if that would make the code easier to understand

iaso/api/profiles/views.py
```diff
@@ -0,0 +1,526 @@
...
+from iaso.api.profiles.serializers.update import ProfileMeUpdateSerializer, ProfileUpdatePasswordSerializer
+from iaso.models import OrgUnit, Profile, TenantUser
+from iaso.permissions.core_permissions import CORE_USERS_ADMIN_PERMISSION, CORE_USERS_MANAGED_PERMISSION
+from iaso.utils import is_mobile_request
+
+
+PK_ME = "me"
+
+
+class HasProfilePermission(permissions.BasePermission):
```
tdethier: Nitpick: you could even move this to a `permissions.py` file next to this `views.py`

iaso/api/profiles/views.py
```diff
@@ -0,0 +1,526 @@
...
+        if request.user.has_perm(CORE_USERS_ADMIN_PERMISSION.full_name()):
+            return True
+
+        if request.user.has_perm(CORE_USERS_MANAGED_PERMISSION.full_name()):
+            return self.has_permission_over_user(request, pk, view.action)
+
+        return view.action in ["retrieve", "list", "export_csv", "export_xlsx"]
+
+    # We could `return False` instead of raising exceptions,
+    # but it's better to be explicit about why the permission was denied.
```
tdethier: 👍

iaso/api/profiles/views.py
```diff
@@ -0,0 +1,526 @@
...
+    def has_permission_over_user(request, pk, view_action):
+        if view_action in ["retrieve", "list", "export_csv", "export_xlsx"]:
+            return True
+
+        if not pk:
+            new_user_org_units = request.data.get("org_units", [])
+            if len(new_user_org_units) == 0:
+                raise PermissionDenied(
+                    f"User with '{CORE_USERS_MANAGED_PERMISSION}' can not create a new user without a location."
+                )
```
tdethier: I would probably have checked that in the related serializers. I know that a `PermissionDenied` in the serializer is not great, but sometimes permissions can't really be checked by the permission class if the business logic is complex

iaso/api/profiles/views.py
```diff
@@ -0,0 +1,526 @@
...
+
+        if pk == request.user.id:
+            raise PermissionDenied(f"User with '{CORE_USERS_MANAGED_PERMISSION}' cannot edit their own permissions.")
+
+        org_units = OrgUnit.objects.hierarchy(request.user.iaso_profile.org_units.all()).values_list("id", flat=True)
```
tdethier: Nitpick: I would probably rename this, to something like `requester_org_units` or `request_user_org_units`

iaso/api/profiles/views.py
```diff
@@ -0,0 +1,526 @@
...
+    """
+
+    http_method_names = ["get", "post", "patch", "put", "delete", "head", "options", "trace"]
+    permission_classes = [permissions.IsAuthenticated, HasProfilePermission]
+    pagination_class = ProfilePagination
+
+    filter_backends = [OrderingFilter, DjangoFilterBackend]
+    filterset_class = ProfileListFilter
+    ordering = ["id"]  # default ordering
+    ordering_fields = ["id", "user__username", "annotated_first_user_role"]
```
tdethier, line 88: All columns on the list view can currently be used to sort profiles

iaso/api/profiles/views.py
```diff
@@ -0,0 +1,526 @@
...
+        if self.action == "list":
+            return ProfileListSerializer
+        if self.action in ["update", "partial_update"]:
+            if self.kwargs.get(self.lookup_url_kwarg or self.lookup_field, "") == PK_ME:
+                return ProfileMeUpdateSerializer
+            return ProfileUpdateSerializer
+        if self.action in ["update_password"]:
+            return ProfileUpdatePasswordSerializer
+
+        raise NotImplementedError(f"Serializer not implemented for action {self.action}")
```
tdethier, line 104: 👍

iaso/api/profiles/views.py
```diff
@@ -0,0 +1,526 @@
...
+        if self.action == "list":
+            if self.request.query_params.get("managedUsersOnly", "").lower() in ["true", "1"]:
+                qs = ManagedUsersPolicy.authorize_list(self.request.user, qs)
+
+            qs = qs.annotate(
+                # Adds a sortable field containing each user's alphabetically first role name,
+                # enabling consistent frontend sorting of users with multiple roles.
+                annotated_first_user_role=Min("user_roles__group__name")
+            ).prefetch_related(
+                "user",
```
tdethier: `select_related` since it's a 1to1?

iaso/api/profiles/views.py
```diff
@@ -0,0 +1,526 @@
...
+            return self.filter_queryset(self.get_queryset()).get(user=self.request.user)
+        return super().get_object()
+
+    def retrieve(self, request, *args, **kwargs):
+        try:
+            return super().retrieve(request, *args, **kwargs)
+        except Profile.DoesNotExist:
+            if kwargs.get(self.lookup_field, "") == PK_ME:
+                serializer = ProfileUserFallbackRetrieveSerializer(instance=self.request.user)
+                return Response(serializer.data, status=status.HTTP_200_OK)
```
tdethier, line 155: I'm not sure why the call to receive the current user profile would trigger a `Profile.DoesNotExist`

iaso/api/profiles/views.py
```diff
@@ -0,0 +1,526 @@
...
+
+    @action(detail=False, methods=["get"], url_path="export-xlsx", url_name="export-xlsx")
+    def export_xlsx(self, request, *args, **kwargs):
+        queryset = self.filter_queryset(self.get_queryset())
+        return self.list_export(queryset, file_format=FileFormatEnum.XLSX)
+
+    @action(detail=True, methods=["patch", "put"], url_path="update-password", url_name="update-password")
+    def update_password(self, request, *args, **kwargs):
+        instance = self.get_object()
+        serializer = self.get_serializer(data=request.data)
```
tdethier, line 171: That's awesome 👏

iaso/api/profiles/views.py
```diff
@@ -0,0 +1,526 @@
...
+            instance=profile, old_data_dump=None, request_user=self.request.user, source=source
+        )
+
+        # send email
+        send_invite = serializer.validated_data.get("send_email_invitation")
+        email = serializer.validated_data.get("email")
+        language = serializer.validated_data.get("language", settings.LANGUAGE_CODE)
+
+        if send_invite and email:
+            transaction.on_commit(lambda: self.send_email_invitation(profile, language))
```
tdethier, line 242: 👍

iaso/api/profiles/views.py
```diff
@@ -0,0 +1,526 @@
...
+                new_org_unit_ids={ou.id for ou in serializer.validated_data["org_units"]},
+            )
+
+        # from super().update to avoid the get_object and serializer again
+        self.perform_update(serializer)
+
+        if getattr(instance, "_prefetched_objects_cache", None):
+            # If 'prefetch_related' has been applied to a queryset, we need to
+            # forcibly invalidate the prefetch cache on the instance.
+            instance._prefetched_objects_cache = {}
```
tdethier, line 358: You need to do this because otherwise the result serializer would still contain old values from before the update?

iaso/api/profiles/filters.py
```diff
@@ -0,0 +1,90 @@
...
+    user_roles = django_filters.BaseInFilter(
+        field_name="user__iaso_profile__user_roles__pk", lookup_expr="in", distinct=True
+    )
+    projects = django_filters.BaseInFilter(
+        field_name="user__iaso_profile__projects__pk", lookup_expr="in", distinct=True
+    )
+    permissions = django_filters.BaseInFilter(
+        field_name="user__user_permissions__codename", lookup_expr="in", distinct=True
+    )
+    orgUnitTypes = django_filters.CharFilter(method="filter_org_unit_type")
```
tdethier, line 24:
```suggestion
    org_unit_types = django_filters.CharFilter(method="filter_org_unit_type")
```

or is this going to break the frontend?

iaso/api/profiles/policies.py
```diff
@@ -0,0 +1,140 @@
...
+            return
+
+        if user.has_perm(CORE_USERS_ADMIN_PERMISSION.full_name()):
+            return
+
+        if user.has_perm(CORE_USERS_MANAGED_PERMISSION.full_name()):
+            managed_org_units = set(OrgUnit.objects.hierarchy(org_units).values_list("id", flat=True).distinct())
+
+            for org_unit_id in [ou.id for ou in org_units]:
+                if org_unit_id not in list(managed_org_units) and not user.is_superuser:
```
tdethier: I think `and not user.is_superuser` is not necessary, because of the early return with the admin permission, but this should be tested

iaso/api/profiles/policies.py
```diff
@@ -0,0 +1,140 @@
...
+
+    @staticmethod
+    def authorize_update(user, profile, new_org_unit_ids):
+        # admins can do anything
+        if user.has_perm(CORE_USERS_ADMIN_PERMISSION.full_name()):
+            return
+
+        # Only managed users are restricted
+        if not user.has_perm(CORE_USERS_MANAGED_PERMISSION.full_name()):
+            return
```
tdethier, line 38: Maybe this works with the rest of the code, but when you're just reading this policy, this part is wrong:
- users that don't have any users permission can't edit profiles at all.
- users with managed user permission can edit profiles of users that are part of their health pyramid
- users with admin user permission can edit an profile in their account

iaso/api/profiles/policies.py
```diff
@@ -0,0 +1,140 @@
...
+                if org_unit_id not in list(managed_org_units) and not user.is_superuser:
+                    raise PermissionDenied(
+                        f"User with {CORE_USERS_MANAGED_PERMISSION} cannot assign an OrgUnit outside of their own health "
+                        f"pyramid. Trying to assign {org_unit_id}."
+                    )
+
+        return
+
+    @staticmethod
+    def authorize_update(user, profile, new_org_unit_ids):
```
tdethier: I would probably rename the `user` parameter in all policy methods, because there's a user linked to all profiles and it's not clear which user this is. `requester` instead?

iaso/api/profiles/policies.py
```diff
@@ -0,0 +1,140 @@
...
+        # admins can do anything
+        if user.has_perm(CORE_USERS_ADMIN_PERMISSION.full_name()):
+            return
+
+        # Only managed users are restricted
+        if not user.has_perm(CORE_USERS_MANAGED_PERMISSION.full_name()):
+            return
+
+        profile_org_units = user.iaso_profile.org_units.all()
```
tdethier: here as well it's confusing, because you pass a `profile` parameter, but `profile_org_units` are not org units that belong to this profile. `requester_org_units` instead?

iaso/api/profiles/policies.py
```diff
@@ -0,0 +1,140 @@
...
+
+        # No restriction configured => allow
+        if not profile_org_units.exists():
+            return
+
+        managed_org_unit_ids = set(OrgUnit.objects.hierarchy(profile_org_units).values_list("id", flat=True))
+
+        existing_org_unit_ids = set(profile.org_units.values_list("id", flat=True))
```
tdethier: I would use `profile_org_unit_ids` here

iaso/api/profiles/views.py
```diff
@@ -0,0 +1,526 @@
...
+        if send_invite and email:
+            transaction.on_commit(lambda: self.send_email_invitation(profile, language))
+
+    @transaction.atomic
+    def create(self, request, *args, **kwargs):
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+
+        # run the advanced checks : authorization/policy validation
+        OrgUnitPolicy.authorize_create(self.request.user, serializer.validated_data.get("org_units", None) or [])
```
tdethier, line 250: I always struggle with `get("something",  None) or "something else"`, why not do `get("something", "something else")` directly?

iaso/tasks/profiles_bulk_update.py
```diff
@@ -3,23 +3,119 @@
...
 from iaso.permissions.core_permissions import (
     CORE_TEAMS_PERMISSION,
     CORE_USERS_ADMIN_PERMISSION,
+    CORE_USERS_MANAGED_PERMISSION,
 )
 from iaso.permissions.utils import raise_error_if_user_lacks_admin_permission
+from iaso.utils import search_by_ids_refs
+
+
+def get_filtered_profiles(
```
tdethier, line 26: The goal is to not break the bulk update task while redoing profiles, but ideally, we'd get rid of this function and have dedicated filters for this task, or maybe reuse some profile filters, right?

iaso/tests/test_multi_tenant.py
```diff
@@ -231,13 +232,13 @@ def test_source_access(self):
...
 
     def test_profile_access(self):
-        response = self.raccoon_client.get("/api/profiles/", accept="application/json")
+        response = self.raccoon_client.get(reverse("profiles-list"), accept="application/json")
         content = json.loads(response.content)
-        self.assertEqual(content["profiles"][0]["user_name"], "raccoon")
+        self.assertEqual(content["results"][0]["user_name"], "raccoon")
 
-        response = self.yoda_client.get("/api/profiles/", accept="application/json")
+        response = self.yoda_client.get(reverse("profiles-list"), accept="application/json")
```
tdethier, line 239: Nitpick: I understand why you use `reverse` but since we don't use it anywhere in the code base, nor in this file, I'd rather keep not using it, to stay consistent. I'd define a const above and call it on each test instead

iaso/tests/api/test_profiles.py
```diff
@@ -5,10 +5,12 @@
 from django.contrib.auth import get_user_model
 from django.contrib.auth.models import Group, Permission
 from django.core import mail
-from django.test import override_settings
+from django.test import override_settings, tag
```
tdethier: Maybe it doesn't have to be done in this PR because it's already so huge, but test-wise, I would definitely:
- reproduce the same organization as the API directory and have dedicated test files for each part ( 1 for views, 1 for each serializer type, 1 for filters...)
- add all missing tests that don't exist yet (serializers) and move parts of this file to their specific file (filters)

iaso/tests/api/test_profiles.py
```diff
@@ -302,52 +369,78 @@ def test_profile_me_no_profile(self):
...
+
+        self.client.force_authenticate(self.jane)
+        response = self.client.get(reverse("profiles-detail", kwargs={"pk": self.jane.iaso_profile.id}))
         self.assertJSONResponse(response, 200)
         response_data = response.json()
         self.assertValidProfileData(response_data)
         self.assertEqual(response_data["user_name"], "janedoe")
-        response = self.client.patch(profile_url)
+
+    @tag("patch")
```
tdethier: So you do use tags? We don't so far, so you don't need to add them if you don't use them 😅

iaso/tests/api/test_profiles.py
```diff
@@ -236,59 +246,116 @@ def setUpTestData(cls):
...
+        else:  # if straightforward list responses without result keys, for custom list methods or bulk create
+            self.assertIsInstance(list_data, list)
+            self.assertEqual(expected_length, len(list_data))
+
+        if paginated:
+            self.assertHasField(list_data, "has_next", bool)
+            self.assertHasField(list_data, "has_previous", bool)
+            self.assertHasField(list_data, "page", int)
+            self.assertHasField(list_data, "pages", int)
+            self.assertHasField(list_data, "limit", int)
```
tdethier: Why did you copy this from `APITestCase`?

iaso/tests/api/test_profiles.py
```diff
@@ -236,59 +246,116 @@ def setUpTestData(cls):
...
+            "user_roles": user_roles,
+            "user_roles_permissions": user_roles_permissions,
+            "org_units": org_units,
+        }
+        return data
 
-        response = self.client.get("/api/profiles/me/")
+    def test_retrieve_profile_me_without_auth(self):
+        """GET /profiles/me/ without auth should result in a 401"""
+        response = self.client.get(reverse("profiles-detail", kwargs={"pk": "me"}))
```
tdethier, line 309: Same comment about `reverse`

iaso/tests/api/test_profiles.py
```diff
@@ -302,52 +369,78 @@ def test_profile_me_no_profile(self):
...
 
-    def test_profile_list_read_only_permissions(self):
+    def test_profile_list_number_queries_without_read_only_permissions(self):
         """GET /profiles/ with auth (user has read only permissions)"""
-
         self.client.force_authenticate(self.jane)
-        with self.assertNumQueries(13):
-            response = self.client.get("/api/profiles/")
+        # todo : fix this later, should be 13
+        with self.assertNumQueries(40):
```
tdethier: That's quite an increase 😅

iaso/tests/api/test_profiles.py
```diff
@@ -594,23 +683,25 @@ def test_create_profile_no_perm(self):
...
+    @tag("create")
     def test_create_user_with_user_roles_and_permissions(self):
         self.client.force_authenticate(self.jim)
         data = {
             "user_name": "unittest_user_name",
             "password": "unittest_password",
             "first_name": "unittest_first_name",
             "last_name": "unittest_last_name",
-            "email": "unittest_last_name",
+            "email": "john.doe@test.com",
```
tdethier, line 655: 👍

iaso/tests/api/test_profiles.py
```diff
@@ -772,17 +869,109 @@ def test_create_profile_with_send_email(self):
             "email": "test@test.com",
         }
 
-        response = self.client.post("/api/profiles/", data=data, format="json")
-        self.assertEqual(response.status_code, 200)
+        with self.captureOnCommitCallbacks(execute=True) as callbacks:
```
tdethier, line 808: TIL

iaso/tests/api/test_profiles.py
```diff
@@ -772,17 +869,109 @@ def test_create_profile_with_send_email(self):
...
+            response = self.client.post(reverse("profiles-list"), data=data, format="json")
 
+        self.assertEqual(response.status_code, 201)
+        self.assertEqual(len(callbacks), 1)
         self.assertEqual(len(mail.outbox), 1)
         email = mail.outbox[0]
+
+        profile = Profile.objects.get(pk=response.json()["id"])
```
tdethier: You can do `result = self.assertJSONResponse(response, 201)` and receive the JSON directly instead of calling `response.json()`

iaso/tests/api/test_profiles.py
```diff
@@ -772,17 +869,109 @@ def test_create_profile_with_send_email(self):
...
+            "last_name": "unittest_last_name",
+            "send_email_invitation": True,
+            "email": "test@test.com",
+            "language": "fr",
+        }
+
+        with self.captureOnCommitCallbacks(execute=True) as callbacks:
+            response = self.client.post(reverse("profiles-list"), data=data, format="json")
+
+        self.assertEqual(response.status_code, 201)
```
tdethier: Same comment about `assertJSONResponse`

iaso/tests/api/test_profiles.py
```diff
@@ -794,26 +983,27 @@ def test_create_profile_with_no_password_and_not_send_email(self):
             "email": "test@test.com",
         }
 
-        response = self.client.post("/api/profiles/", data=data, format="json")
+        response = self.client.post(reverse("profiles-list"), data=data, format="json")
         self.assertEqual(response.status_code, 400)
 
         response_data = response.json()
-        self.assertEqual(response_data["errorKey"], "password")
+        self.assertHasError(response_data, "password", "This field is required.")
```
tdethier: 👍

tdethier (changes requested):
Overall, very nice job. This was an annoying API and rewriting everything with DRF will make our lives much easier when we need to update this endpoint 👍 

I was able to CRUD profiles, change passwords and so on.


While testing, I noticed this:
- when updating a user's password, the legend shows `*` instead of "new password" or something similar 
 Image


- when updating a profile that does not have a set phone number, this warning shows up every time, even if I don't touch the phone number field
Image

- there's no possibility of resetting the phone number field, once a country code is there, you can no longer save your profile until you put a valid phone number

- when setting location restrictions, I managed to trigger an error after selecting an org unit in the tree view (I'm not sure how, because I could not reproduce that afterwards)
Image

- I would probably add a panel for "org unit write types" on the details page, it's the only modal tab that is not displayed there

- there are a lot of duplicated queries now:
```
list view:
http://localhost:8081/api/profiles/?limit=20&order=user__username&page=1&managedUsersOnly=true
| RESP |  default  |    74    |    0     |    74    |     60     |
Total queries: 74 in 2.0643s 
INFO     2026-03-04 16:02:43,362 django.server -- "GET /api/profiles/?limit=20&order=user__username&page=1&managedUsersOnly=true HTTP/1.1" 200 21049


retrieve view:
http://localhost:8081/api/profiles/641/
| RESP |  default  |    23    |    0     |    23    |     12     |
Total queries: 23 in 0.4025s 
INFO     2026-03-04 16:02:36,119 django.server -- "GET /api/profiles/641/ HTTP/1.1" 200 2187


delete view:
http://localhost:8081/api/profiles/641/
| RESP |  default  |    28    |    59    |    87    |     20     |
Total queries: 87 in 0.6851s 
INFO     2026-03-04 16:02:40,386 django.server -- "DELETE /api/profiles/641/ HTTP/1.1" 204 0


post view:
http://localhost:8081/api/profiles/
| RESP |  default  |    18    |    6     |    24    |     8      |
Total queries: 24 in 0.4840s 
INFO     2026-03-04 16:08:05,002 django.server -- "POST /api/profiles/ HTTP/1.1" 201 290


update view:
http://localhost:8081/api/profiles/642/
| RESP |  default  |    23    |    5     |    28    |     16     |
Total queries: 28 in 0.4685s 
INFO     2026-03-04 16:08:36,660 django.server -- "PATCH /api/profiles/642/ HTTP/1.1" 200 211

```


- we should get rid of repeated queries, especially the list view because it's the slowest and it's often called
- I think the retrieve serializer contains lots of fields that we don't actually need
- some parts of this refactor are not done yet (see comments at the top of `views.py`), do we ship this first version first, create another ticket for the v2 and handle it later, or do we fix this already?

iaso/tests/test_multi_tenant.py
hugo093: I just use reverse so if we change the url one day , we don't have to change it everywhere in the tests

iaso/api/profiles/views.py
hugo093: see this example :
 ```shell
x = {'a': None}
x.get('a', "default")
print(x.get('a', "default"))
> None
print(x.get('a', 'default') or 'default')
> default
print(x.get('b', 'default'))
> default```

iaso/api/profiles/filters.py
hugo093: It's going to break the front end i'm afraid , but in the future i'd like to use something like https://pypi.org/project/djangorestframework-camel-case/ to have camelCase on the front-end and snake-case on the backend

iaso/api/setup_account.py
hugo093: maybe a circular import, moved it at the top level

iaso/locale/fr/LC_MESSAGES/django.po
hugo093: I think so yes why ?

iaso/api/profiles/serializers/common.py
hugo093: yes :)

iaso/api/profiles/serializers/retrieve.py
hugo093: It seems like so, but i'm not yet sure that we actually need all the info for this serializer (front end related)

iaso/api/profiles/views.py
hugo093: I agree, maybe i'd continue improving this later :) this is a part I didn't really touched

iaso/api/profiles/views.py
hugo093: Actually I deactivated most of the sorting on the list view UI for the frontend as it wasn't really working

iaso/api/profiles/views.py
hugo093: Yes :) but tbh it was already there before

iaso/api/profiles/policies.py
hugo093: Maybe we should review this in another ticket actually

iaso/tasks/profiles_bulk_update.py
hugo093: Exactly, this is the old "filter" method that i just moved there for legacy reasons as bulk_update is not really part of this ticket

hugo093:
@tdethier  @beygorghor 
Tried to implement most of the remarks made here but as this PR is getting bigger and bigger I'd suggest that, if you agree of course, we merge this and test it. On the other hand I'll create a bunch of tickets: 
* Backend:
  * Fix the mix of camelCase and snake_case
  * Optimize and test queries
  * Use the standard way for PrimaryKeyRelatedField 
  * Decide on the responses payloads for list  and retrieve
  * Review the policies and permission
  * Implement a ColorFieldSerializer
  * Test
  * Decide on sortable fields
  * Refactor the profile_bulk_update task ?
* Front end 
  * Replace all the forms to use Formik
  * Allow the phone field to be cleared
  * Dynamic columns on the search
  * Sortable columns
  * Adding an org unit write type panel in the detail view
  * Tests

Do you agree?

tdethier (approved):
I agree with your general comment on the PR, let's create tickets for all elements on your list and fix all of this later. 
The feature works fine for now, and it will help other implementations too (people started complaining about this password bug on SNT)

PR #2775, opened 2026-03-02
Enable ruff security rules and related fixes

hugo093:

Security rules were not enabled in the ruff config 

### Related JIRA tickets

IA-4809

## Changes

* Enabled "S" rule for ruff in pyproject.toml
* Did the right fixes when I could, otherwise left a todo 

## How to test

* run the command `ruff check .`

## Notes

* We need to decide what to do in case of empty try - except : log warn ?
* Perhaps the most urgent fix to decide is the one located in iaso/isao/models/instances.py with the pseudo check on the url


iaso/utils/__init__.py
```diff
@@ -116,8 +117,12 @@ def geojson_queryset(queryset, geometry_field, pk_field="id", fields=[]):
...
         raise ValueError("invalid geom field name")
 
-    features_queryset = queryset.extra(
-        select={"geojson_queryset_result": f"ST_AsGeoJSON({geometry_field})::json"}
+    features_queryset = queryset.annotate(
+        geojson_queryset_result=Func(
+            geometry_field,
+            function="ST_AsGeoJSON",
+            output_field=JSONField(),
+        )
```
mestachs: are you sure it's equivalent ?

mestachs:
global feeling
 - adding short timeout looks risky in some calls
    - ex writing data to dhis2 might be slow, reading the pyramid too
    - ex iaso task being a self client in the zip creation (especially for entities I would ask Philippe to really test data with A LOT of data, there are a lot of entities there)
    - I would specify at least per product 
      - DHIS2, superset, iaso, s3, enketo, powerbi, ona/odk => constants 
      - for dhis2 I would specify higher (different contant) timeout for "big write" and the pyramid synchronisation : apparently we use dhis2 python package (don't know if there's timeout specified)
      - to mitigate this I would allow to override it via env variables so in case of trouble we don't need a hot fix
      - iaso/dhis2/authentication.py can "low" timeout
      - this ona/odk might be slow too : https://github.com/BLSQ/iaso/blob/2dccb9f942bb6ee9cbc1a23a3405fff5506830de/plugins/polio/api/common.py#L145
    - note I think it's good idea (already had apps deadlocked/all threads starved due to unresponsive external server), I would just be a bit more cautious (moving from a really permissive environment to a 30 sec timeout is perhaps a bit too optimistic)
 - not sure everything is covered by a test (polio admin html thing)
 - I hope the xml parser is not too risky, definitively need to be tested by a real submission via the mobile and enketo

iaso/tests/utils_parquet.py
```diff
@@ -78,10 +78,10 @@ def compare_or_create_snapshot(parquet_path, snapshot_path, stable_columns=None,
     )
 
 
-# used to debug if the test fails
+# used to debug if the test fails => noqa
 def read_parquet(f):
     with duckdb.connect() as con:
-        result = con.execute(f"SELECT * FROM read_parquet('{f.name}')")
+        result = con.execute(f"SELECT * FROM read_parquet('{f.name}')")  # noqa: S608
```
mestachs: I would have fixed it instead of adding an exception to the rule

iaso/exports/duckdb_util.py
```diff
@@ -22,7 +22,8 @@ def export_django_query_to_parquet_via_duckdb(qs: QuerySet, output_file_path: st
     # initially was full_sql = sql % tuple(map(adapt_param, params)) but supporting all types is complicated
     dsn = connection.get_connection_params()
 
-    tmpdir = "/tmp/duckdb_tmp"
+    # todo : could we use a TempFile here instead ?
```
mestachs: I don't know it's used to let duckdb overflow on disk instead of doing everything in memory, probably a random directory is better (it's normally covered by the test, so you can try to do the switch)

PR #2777, opened 2026-03-03
[SNT-293] integrate DATABASES from plugin_settings.py

michiwend:

Pulls DATABASES from plugin_settings. Allows any plugin do define its own (additional) databases. 

### Related JIRA tickets

SNT-293

## Changes

This change integrates the DATABASES configuration block from plugin_settings, so plugins can define their own databases. This is needed in SNT where we integrate with multiple PostgreSQL databases through unmanaged Django ORM models.


tdethier (approved):
LGTM 👍

PR #2781, opened 2026-03-04
IA-4824: LIVE_COMPONENTS not working

beygorghor:

While running webpack using LIVE_COMPONENTS=true it should use local repository of bluesquare-components

### Related JIRA tickets

IA-4824

## Changes

Fix Webpack config to use 'bluesquare-components' locally 

## How to test

Make sure you pull [this](https://github.com/BLSQ/bluesquare-components/pull/209) branch  from bluesquare-components, repository should be at the same level as IASO repository.

Run `LIVE_COMPONENTS=true npm run dev`or `LIVE_COMPONENTS=true docker compose up`
Open 'bluesquare-components' code and make some changes, it should live reload and you should see your changes directly

## Notes

This [PR](https://github.com/BLSQ/bluesquare-components/pull/209) is fixing some typings issue we had while exporting, displaying some warning on the compilation.

### DIFF
--- a/hat/webpack.dev.js
+++ b/hat/webpack.dev.js
@@ -166,11 +166,14 @@ module.exports = {
                     eager: true,
                     requiredVersion: false,
                 },
-                'bluesquare-components': {
-                    singleton: true,
-                    eager: true,
-                    requiredVersion: false,
-                },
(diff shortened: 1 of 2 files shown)


Bewi (approved):
Tested locally, works all good!

PR #2783, opened 2026-03-04
New model and engine for validation workflow

hugo093:

Final data modelling of the validation workflow feature

### Related JIRA tickets

IA-4776


iaso/engine/validation_workflow.py
```diff
@@ -0,0 +1,115 @@
...
+
+from iaso.models import Instance, ValidationStatus
+from iaso.models.validation_workflow.validation_status import Status
+
+
+class ValidationWorkflowEngine:
+    @staticmethod
+    @transaction.atomic
+    def start(workflow_template, user, entity: Instance, parent_entity: Optional[Instance] = None):
+        if entity and not workflow_template.is_entity_allowed(entity):
```
quang-le: I know we're running out of kinda generic words, but `Entity` is a core model in iaso, so it's a bit confusing here

iaso/engine/validation_workflow.py
```diff
@@ -0,0 +1,115 @@
...
+        # if there are any nodes further on that have already been approved/rejected, not allowed, it's too late
+
+    @staticmethod
+    @transaction.atomic
+    def undo_node(validation_status: ValidationStatus, user):
+        # todo
+        ValidationWorkflowEngine._can_undo_node(validation_status)
+
+        if validation_status.status != Status.ACCEPTED:
+            raise ValueError("Cannot undo node that isn't accepted")
```
quang-le: shouldn't that be part of `_can_undo_node`?

iaso/engine/validation_workflow.py
```diff
@@ -0,0 +1,115 @@
...
+
+        validation_status.updated_by = None
+        validation_status.status = Status.UNKNOWN
+        validation_status.comment = ""
```
quang-le, line 196: maybe out of scope for now, but I think it would make sense to keep track of who undid a node, i.e. forced a status change, maybe force a comment, or generate one like "reverted by <user>" or something

iaso/engine/validation_workflow.py
```diff
@@ -0,0 +1,115 @@
...
+        validation_status.save()
+
+        # delete next related validation status
+        validation_status.get_next_validation_statuses().delete()
```
quang-le: If I understand the current code, this will only get and delete the statuses for the node immediately after the current status' node. But what if a user reverts a node down 2 levels?

iaso/models/validation_workflow/templates.py
```diff
@@ -0,0 +1,66 @@
...
+class ValidationNode(CreatedAndUpdatedModel):
+    """
+    Represents a node in the workflow aka a Task to do in order to continue the validation.
+    """
+
+    workflow = models.ForeignKey(ValidationWorkflow, on_delete=models.CASCADE, related_name="nodes")
+    name = models.CharField(max_length=256)
+    slug = AutoSlugField(populate_from="name")
+    color = ColorField(blank=True, null=True)
+    next_nodes = models.ManyToManyField("self", symmetrical=False, related_name="previous_nodes")
```
quang-le: Are these all the nodes that come after, or just the ones that follow directly?

iaso/models/validation_workflow/templates.py
```diff
@@ -0,0 +1,96 @@
...
+        ).first()
+
+
+class ValidationNode(CreatedAndUpdatedModel):
+    """
+    Represents a node in the workflow aka a Task to do in order to continue the validation.
+    """
+
+    workflow = models.ForeignKey(ValidationWorkflow, on_delete=models.CASCADE, related_name="nodes")
+    name = models.CharField(max_length=256)
```
quang-le, line 58: name should probably be unique within the workflow

iaso/models/validation_workflow/templates.py
```diff
@@ -0,0 +1,108 @@
...
+    Static definition of a node in the workflow (aka a Task) to do in order to continue the validation.
+    """
+
+    workflow = models.ForeignKey(ValidationWorkflow, on_delete=models.CASCADE, related_name="node_templates")
+    name = models.CharField(max_length=256)
+    slug = AutoSlugField(populate_from="name", unique=True, always_update=True, unique_with="workflow_id")
+
+    color = ColorField(blank=True, null=True)
+    next_node_templates = models.ManyToManyField("self", symmetrical=False, related_name="previous_node_templates")
+    roles_required = models.ManyToManyField(UserRole, blank=True)
```
madewulf, line 63: I still don't get how you are going to naturally represent situation like:
- health worker makes a submission it's in state (submitted)
-  it needs to be approved by A and B -> it gets into states (waiting_for_a, waiting_for_be)
- a rejects -> it should go to (rejected) not to (rejected, waiting_for_b)

PR #2786, opened 2026-03-05
WC2-875: Fix issue with the duplication cleanup script

bmonjoie:

After we ran the script in production, we encountered an error at the commit.
It seems there are `Instance` pointing to some entities we want to delete.

### Related JIRA tickets

WC2-875

## Changes

When deleting entities, we first move all the instances that are linked to them to the one we keep.

## How to test

Not really possible :shrug:


tdethier (commented):
LGTM 👍

tdethier (approved):

PR #2789, opened 2026-03-06
[IA-4773] Create `MetricType` permission

tdethier:

Create `MetricType` permission

### Related JIRA tickets

IA-4773

## Changes

- created a new `iaso_metric_types` permission
- implemented permission checks on the metrics API
- added translations for the new permission
- added some missing tests in metrics API + refactored some tests

## How to test

- Go to the user permissions screen or user roles screen
- Check that the `Admin` section now has a permission for metric types
- Test the metric type API with and without the metric type permission (there's no frontend for that in IASO, but there's one in SNT) - see examples in unit tests
  - anyone who's logged in can see data (GET/HEAD/OPTIONS)
  - create/update/delete requires the new permission

## Print screen / video

image

## Notes

/

## Doc

/


iaso/api/metrics/views.py
hugo093: you can remove this now ;)

hat/assets/js/apps/Iaso/domains/app/translations/en.json
```diff
@@ -1175,6 +1175,8 @@
     "iaso.permissions.iaso_entity_type_write": "Entity types management",
     "iaso.permissions.iaso_forms_stats": "Form statistics",
     "iaso.permissions.iaso_forms_stats_tooltip": "See statistics of number forms submissions",
+    "iaso.permissions.iaso_metric_types": "Metric types management",
```
hugo093, line 1178: just a side note but i find it weird that we don't fetch this from the backend. Seems like we duplicate info

iaso/tests/api/metrics/test_views.py
```diff
@@ -191,318 +242,391 @@ def test_metric_type_grouped_per_category(self):
...
+        self.assertEqual(csv[0], expected_header)
+        expected_org_unit_values = ["", self.org_unit.name, str(self.org_unit.id)]
+        self.assertEqual(csv[1], expected_org_unit_values)
+
+    def test_metric_value_csv_template_unauthenticated(self):
+        response = self.client.get(f"{self.BASE_URL}csv_template/")
+        self.assertJSONResponse(response, status.HTTP_401_UNAUTHORIZED)
+
+    def test_metric_value_import_from_csv_with_perm(self):
+        # TODO: write this test
```
hugo093, line 622: still todo ?

hugo093 (changes requested):
Just two todos to remove but otherwise LGTM

hat/assets/js/apps/Iaso/domains/app/translations/en.json
tdethier: Yeah, translations are in a weird state. Some are in the front, some in the back, but there hasn't been any discussion on how we should manage them. It would be worth it

iaso/tests/api/metrics/test_views.py
tdethier: Yeah, there was no test on that endpoint at all. I added some and realized that I don't know the business logic behind that endpoint, so it was not worth implementing this test in this PR. But we should definitely do it at some point

hugo093 (approved):

PR #2790, opened 2026-03-06
Implement djangorest framework camelcase for profiles API

hugo093:

THere is a mix of camelCase and snake_case in front-end and back-end for the profile API.

### Related JIRA tickets

IA-4829, IA-4739

## Changes

* Added djangorestframework_camel_case and custom filters and middlewares
* Adapted tests
* Added compatibility for mobile endpoint
* Fixed a front end bug on create profile

## How to test

Run the tests under iaso.tests.api.test_profiles or just call the /api/v2/profiles/ endpoint with camelCase parameters, it will work even though they are defined as snake_case in the backend

For compatibility, /api/profiles/ will behave the same as before


iaso/middleware.py
```diff
@@ -0,0 +1,16 @@
...
+class CustomCamelCaseMiddleWare(CamelCaseMiddleWare):
+    def __call__(self, request):
+        # we only execute this middleware on profiles API at the moment
+        route = ""
+        try:
+            route = resolve(request.path_info).route
+        except Resolver404:
+            pass
+
+        if route.startswith("/api/profiles"):
```
Phil-V: This will only run for v1, is that intended @hugo093 ?

iaso/tests/api/test_profiles.py
```diff
@@ -283,45 +281,68 @@ def get_new_user_data(self):
...
+        user.iaso_profile.save()
+
+        self.client.force_authenticate(user)
+        response = self.client.get(reverse("profiles-detail", kwargs={"pk": "me"}))
+        response_data = self.assertJSONResponse(response, 200)
+        for k in ["id", "first_name", "last_name", "user_name", "email", "phone_number", "organization", "projects"]:
+            self.assertIn(k, response_data)
+
+        for k in ["id", "name", "app_id"]:
+            self.assertIn(k, response_data["projects"][0])
```
Phil-V, line 331: Great! 👍

iaso/middleware.py
Phil-V: My bad sorry! I got my wires crossed during manual testing, it works as intended

Phil-V (approved):
Good job!

iaso/middleware.py
hugo093: you're right actually , it should be /api/v2/profiles :) Otherwise it would apply on legacy and it's not good

PR #2793, opened 2026-03-09
IA-4822: Unauthorized users are able to modify org unit types

beygorghor:

User without org uni type permission can edit , create or delete an org unit type

### Related JIRA tickets

IA-4822

## Changes

- adapt viewset permission,
- adapt and write tests
- used bruno file to test API

## How to test

Open bruno file PATCH Org unit types without permissions. 
go to var tab and copy paste the org unit unit id org the type you want to change
add the apitoken of a user without `CORE_ORG_UNITS_TYPES_PERMISSION`
Try to patch, it should say:
```
{
  "detail": "You do not have permission to perform this action."
}

```


iaso/api/org_unit_types/viewsets.py
```diff
@@ -23,19 +24,37 @@
...
+    """
+
+    def has_permission(self, request, view):
+        if request.method in permissions.SAFE_METHODS:
+            return True
+        if not request.user or not request.user.is_authenticated:
+            return False
+        return request.user.has_perm(CORE_ORG_UNITS_TYPES_PERMISSION.full_name())
+
+
```
hugo093: can you move this to a permissions.py file ? :) (Also I see this pattern a lot, maybe we could make some generic class :P )

hugo093 (changes requested):
Nit-picking for the permission , otherwise LGTM

iaso/api/org_unit_types/viewsets.py
beygorghor: i'll move it to a separate file, for the generic thing, maybe we can do that on a separate ticket ?

hugo093 (approved):
LGTM

PR #2795, opened 2026-03-09
IA-4848: Entities api refactor

Phil-V:

Refactor the Entities API to pave the way for cursor pagination and performance optimisations. 

### Related JIRA tickets

IA-4848

## Changes

The API should remain identical (with one exception, see note below).

- Refactored the list view to use Django Rest Framework's architecture.
- Added tests for ordering

## How to test

- Visit the Entities List page (Menu > Entities > Entities List) and
    - Test all the filters (search, groups, location, etc.)
    - Test the ordering of the results for different columns
    - Test the map view tab
    - Test the csv/xlsx exports
- Go to the Entity Types settings (Menu > Entities > Entity Types), and change the "List Fields" for the entity type. This determines the columns displayed in the entities list. Add some fields or remove all fields if they were already set.
- Go back to the entity list, refresh the page, check that the columns are displayed properly and test that csv / xlsx exports work with/without the extra columns.

## Print screen / video

/

## Notes

While I strove to keep the api identical, there are a couple of potentially breaking changes:

- The [original code](https://github.com/BLSQ/iaso/blob/c95d7df5509d4f36b7d4ad5a9cfc9ebd85851e21/iaso/api/entity.py#L368) calls `get_and_save_json_of_xml` on the Instance model when retrieving the entities attributes. The intention here I suspect was to backfill the `json` attribute if missing. It's strange to do this in the view, so I got rid of it. Removing this is unlikely to actually cause issues, but if it does we need to resolve it in other ways I think.
- Calling the endpoint without a `limit` used to return _everything_ without pagination ([code](https://github.com/BLSQ/iaso/blob/c95d7df5509d4f36b7d4ad5a9cfc9ebd85851e21/iaso/api/entity.py#L481)). This can potentially crash all but the tiniest Iaso production setups. To avoid this, I've enabled pagination by default and set a high default limit `page_size = 10000`. I think this should be a comfortable compromise. The result follows the same structure (just with extra pagination-related attributes) so existing scripts or clients that rely on the `result` and `columns` keys should not break.
- Removed a [likely unintended filter](https://github.com/BLSQ/iaso/blob/39b7b622ba94993a8aa3107e49bad70a8ff3a023/iaso/api/entity.py#L191) mapping entity type to entity name.

Note for the reviewer: EntitySerializer has been moved with no changes (so all the other endpoints on the viewset that rely on it should be unaffected by this PR). 

## Doc

/


iaso/api/entities/serializers.py
```diff
@@ -0,0 +1,272 @@
...
+            "entity_type_name",
+            "created_at",
+            "org_unit_name",
+            "org_unit_id",
+            "last_saved_instance",
+        ]
+
+    id = serializers.IntegerField(label="ID")
+    uuid = serializers.UUIDField(label="UUID")
+    entity_type_name = serializers.SerializerMethodField(label="Entity Type")
```
hugo093: instead of the SerializerMethodField (and the associated method) it could be : `serializers.Charfield(label="Entity Type", source="entity_type.name", default="", read_only=True)`

iaso/api/entities/serializers.py
```diff
@@ -0,0 +1,272 @@
...
+    created_at = serializers.DateTimeField(format=EXPORTS_DATETIME_FORMAT, label="Creation Date")
+    org_unit_name = serializers.SerializerMethodField(label="HC")
+    org_unit_id = serializers.SerializerMethodField(label="HC ID")
```
hugo093: instead of the SerializerMethodField (and the associated method) it could be : serializers.Charfield(label="HC ID", source="attributes.org_unit.id", default="", read_only=True)

iaso/api/entities/serializers.py
```diff
@@ -0,0 +1,272 @@
...
+            "org_unit_name",
+            "org_unit_id",
+            "last_saved_instance",
+        ]
+
+    id = serializers.IntegerField(label="ID")
+    uuid = serializers.UUIDField(label="UUID")
+    entity_type_name = serializers.SerializerMethodField(label="Entity Type")
+    created_at = serializers.DateTimeField(format=EXPORTS_DATETIME_FORMAT, label="Creation Date")
+    org_unit_name = serializers.SerializerMethodField(label="HC")
```
hugo093: same remark :)

iaso/api/entities/serializers.py
```diff
@@ -0,0 +1,272 @@
...
+    source_name = serializers.SerializerMethodField()
+
+    def get_org_unit_type(self, obj):
+        return obj.org_unit_type and obj.org_unit_type.name
+
+    def get_org_unit_type_depth(self, obj):
+        return obj.org_unit_type and obj.org_unit_type.depth
+
+    def get_parent_name(self, obj):
+        return obj.parent and obj.parent.name
```
hugo093: same remark for those, using source on a serializer would do the trick

iaso/api/entity.py
```diff
@@ -1,107 +1,90 @@
...
-    results = []
-    e1qs = entity.duplicates1.filter(validation_status=ValidationStatus.PENDING)
-    e2qs = entity.duplicates2.filter(validation_status=ValidationStatus.PENDING)
-    if e1qs.count() > 0:
-        results = results + list(map(lambda x: x.entity2.id, e1qs.all()))
-    elif e2qs.count() > 0:
-        results = results + list(map(lambda x: x.entity1.id, e2qs.all()))
-    return results
+
+class EntityListPaginator(pagination.PageNumberPagination):
```
hugo093: move this to a pagination.py file, ideally this file should only contain the viewset

iaso/api/entity.py
```diff
@@ -1,107 +1,90 @@
-import datetime
```
hugo093, line 1: can you rename this file to views.py to be compliant with the rest of the code ?

iaso/api/entity.py
```diff
@@ -1,107 +1,90 @@
...
-    def get_duplicates(self, entity: Entity):
-        return _get_duplicates(entity)
 
-    def get_nfc_cards(self, entity: Entity):
-        nfc_count = StorageDevice.objects.filter(entity=entity, type=StorageDevice.NFC).count()
-        return nfc_count
+logger = getLogger(__name__)
+
+
+class EntityLocationPaginator(pagination.PageNumberPagination):
```
hugo093: move this to a pagination.py file

iaso/api/entity.py
```diff
@@ -1,107 +1,90 @@
...
+                "has_previous": self.page.has_previous(),
+                "page": self.page.number,
+                "pages": self.page.paginator.num_pages,
+                "limit": self.page.paginator.per_page,
+                "columns": columns,
+            }
+        )
+
+
+class EntityTypeColumnSerializer(serializers.Serializer):
```
hugo093: move this to serializers.py

iaso/api/entity.py
```diff
@@ -1,107 +1,90 @@
...
+
+    page_size_query_param = "limit"
+    # large default page_size for legacy API compatibility
+    # not specifying a limit used to return everything
+    page_size = 10000
+
+    def paginate_queryset(self, queryset, request, view=None):
+        """Store a reference to the view to access later."""
+
+        self.view = view
```
hugo093: nit-picking here, but could be self._view as it's more private + defining it at top level and setting it to None by default

iaso/api/entities/filters.py
```diff
@@ -0,0 +1,140 @@
...
+from rest_framework import filters
+from rest_framework.exceptions import ValidationError
+
+from iaso.models import Entity, Instance, OrgUnit
+from iaso.utils.date_and_time import date_string_to_end_of_day, date_string_to_start_of_day
+from iaso.utils.jsonlogic import entities_jsonlogic_to_q
+
+
+class CharInFilter(BaseInFilter, CharFilter):
+    pass
```
hugo093: I'd say to move that to iaso.api.common as somebody else might need it

iaso/api/entities/filters.py
```diff
@@ -0,0 +1,140 @@
...
+    search = CharFilter(method="filter_search")
+
+    class Meta:
+        model = Entity
+        fields = []
+
+    def filter_show_deleted(self, queryset, name, value):
+        if value:
+            return queryset.filter(deleted_at__isnull=True)
+        return queryset
```
hugo093: could be a  BooleanFilter with lookup_expr=isnull  ?

iaso/api/entities/serializers.py
```diff
@@ -0,0 +1,272 @@
...
+    attributes = serializers.SerializerMethodField()
+    submitter = serializers.SerializerMethodField()
+    org_unit = serializers.SerializerMethodField()
+    duplicates = serializers.SerializerMethodField()
+    nfc_cards = serializers.SerializerMethodField()
+
+    def get_attributes(self, entity: Entity):
+        if entity.attributes:
+            return entity.attributes.as_dict()
+        return None
```
hugo093, line 200: Should be a nested serializer instead of using as_dict (and seeing the related as_dict method, i'm sorry)

iaso/api/entities/serializers.py
```diff
@@ -0,0 +1,272 @@
...
+
+    def get_org_unit(self, entity: Entity):
+        if entity.attributes and entity.attributes.org_unit:
+            return entity.attributes.org_unit.as_dict_for_entity()
+        return None
```
hugo093, line 205: same remark, should be a nested serializer

iaso/api/entities/serializers.py
```diff
@@ -0,0 +1,272 @@
...
+    attributes = serializers.PrimaryKeyRelatedField(read_only=True)
+    longitude = serializers.SerializerMethodField()
+    latitude = serializers.SerializerMethodField()
+    entity_type = serializers.SerializerMethodField()
+    name = serializers.SerializerMethodField()
+    last_saved_instance = serializers.SerializerMethodField()
+    has_duplicates = serializers.SerializerMethodField()
+
+    def get_org_unit(self, obj):
+        return obj.attributes and EntityListNestedOrgUnitSerializer(obj.attributes.org_unit).data
```
hugo093: EntityListNestedOrgUnitSerializer(source="attributes.org_unit") in the serializer definition instead?

iaso/api/entities/filters.py
```diff
@@ -0,0 +1,140 @@
...
+            field_name = field.lstrip("-")
+            base_field = field_name.split("__")[0]
+
+            if base_field in model_fields:
+                final_ordering.append(field)
+            else:
+                prefix = "-" if field.startswith("-") else ""
+                final_ordering.append(f"{prefix}attributes__json__{field_name}")
+
+        return final_ordering
```
hugo093, line 131: Ngl, i've a hard time understanding the purpose of this.. imho it should be fixed front end wise and we should keep it more simple in the backend ? Even if we need to duplicates ordering fields by adding attributes__json

iaso/api/entities/filters.py
```diff
@@ -0,0 +1,140 @@
...
+
+        return queryset
+
+
+class EntityOrderingFilter(filters.OrderingFilter):
+    """Support sorting on the Entity's attributes."""
+
+    def get_ordering(self, request, queryset, view):
+        # Accept either parameter, for legacy API compatibility
+        ordering = request.query_params.get("order_columns") or request.query_params.get("order")
```
hugo093, line 105: are we going to change this later on the front and keep order only ? Also for the second parameter , use getattr(view, "ordering", None) instead of "order"

iaso/api/entities/renderers.py
```diff
@@ -0,0 +1,142 @@
...
+            yield writer.writeheader().encode()
+            yield writer.writerow(first_row).encode()
+
+            for row in iterator:
+                yield writer.writerow(row).encode()
+
+        return stream_data(data_iterator)
+
+
+class XlsxStreamingRenderer(BaseStreamingExportRenderer):
```
hugo093, line 100: could we have used something like https://github.com/django-commons/drf-excel instead ?

iaso/api/entity.py
hugo093, line 230: Now that i see this, it could be refactored as well : 
* use get_object to return the proper entity
* perform_destroy for the right operation
* do we really need to return some data through a serializer for some soft deleted item?

iaso/api/entities/views.py
```diff
@@ -122,40 +105,55 @@ class EntityViewSet(ModelViewSet):
...
+        EntityDateFilterBackend,
     ]
+    filterset_class = EntityFilterSet
     permission_classes = [permissions.IsAuthenticated, HasPermission(CORE_ENTITIES_PERMISSION)]  # type: ignore
+    content_negotiation_class = LegacyExportContentNegotation
+    renderer_classes = (
+        renderers.JSONRenderer,
+        renderers.BrowsableAPIRenderer,
+        CSVStreamingRenderer,
+        XlsxStreamingRenderer,
```
hugo093, line 68: I'd remove this and put @renderer_classes(...) on the list action itself because that's the only place where we use it if i'm not mistaking

iaso/tests/api/entities/test_entities.py
```diff
@@ -112,7 +115,7 @@ def test_retrieve_entity(self):
         response = self.client.get("/api/entities/", format="json")
 
         self.assertEqual(response.status_code, 200)
-        self.assertEqual(len(response.json()), 2)
+        self.assertEqual(len(response.json().get("result")), 2)
```
hugo093: take a look at assertJSONResponse and maybe assertValidListData ;)

iaso/tests/api/entities/test_entities.py
```diff
@@ -192,8 +195,8 @@ def test_list_entities_search_filter(self):
         the_result = response.json()["result"][0]
         self.assertEqual(the_result["id"], newly_added_entity.id)
 
-        # Case 2: search by entity name - make sure it's case-insensitive
-        response = self.client.get("/api/entities/?search=cLiEnT", format="json")
+        # Case 2: search by entity name - make sure it's case-insensitive and ignores white space
+        response = self.client.get("/api/entities/?search= cLiEnT &foo=bar", format="json")
```
hugo093: use data parameter in .get to construct GET params

hugo093 (changes requested):
Good job overall, just a few things to improve. But seems like a big improvement from what was there before. Maybe in the future we could improve some other things (like moving the export actions to another endpoints ?) but we'll see

iaso/api/entities/views.py
Phil-V: It's something I've been trying to do, but for some reason I cannot get it to work. After importing `renderer_classes` from `rest_framework.decorators` and defining it on the view it seems to be just ignored.

More broadly I'm less sure now that renderers were the right move (although it did help organize the chaos while preserving the current API) and separate actions as you suggested might be more straightforward.

Phil-V:
Just to keep the scope of changes manageable, I'll create a separate clean-up PR when done, with some non-backward compatible changes that you flagged:
- get rid of the custom EntityOrderingFilter entirely
- possibly move the exports to actions

Pushed back for a future PR:
- refactor EntitySerializer (use nested serializer instead of dict method)
- refactor destroy endpoint

iaso/api/entities/serializers.py
Phil-V: Must have been confused during local testing and rewrote a bunch of them to use SerializerMethodFields, it's a good thing you mentioned this and it's much better now ☺️

iaso/api/entities/serializers.py
Phil-V: Will leave this unfixed for now, I think I'll keep it out of scope for this PR (to minimize api changes for now)

iaso/api/entities/renderers.py
Phil-V: Sounds like a good option! Could use in tandem with styles on the serializers fields.

hugo093 (approved):
LGTM, good job

PR #2799, opened 2026-03-10
IA-4850: Do not return reference instances with deleted forms

bmonjoie:

The mobile receives instances that are linked to deleted forms, which prevents it from importing them.

### Related JIRA tickets

IA-4850

## Changes

The endpoints filters out instances that are linked to deleted forms.

## How to test

1. Configure an OrgUnit to have at least one reference form
2. Add a reference instance to the OrgUnit
3. call `/api/mobile/orgunits/{ID}/reference_instances/?app_id={YOUR APP ID}` (with the proper ID and YOUR_APP_ID)
4. Ensure that the reference instance is returned
5. Mark the form as deleted
6. Redo call from 3.
7. Ensure the reference instance is NOT returned.

### DIFF
--- a/iaso/tests/api/test_mobile_orgunits.py
+++ b/iaso/tests/api/test_mobile_orgunits.py
@@ -1,3 +1,5 @@
+import datetime
+
 import time_machine
 
 from django.contrib.gis.geos import MultiPolygon, Point, Polygon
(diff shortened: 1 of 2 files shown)


iaso/tests/api/test_mobile_orgunits.py
```diff
@@ -412,9 +414,16 @@ def test_reference_instances(self):
             form=form2, org_unit=self.raditz, json={"key": "bar"}, form_version=form_version2
         )
         instance_file1 = InstanceFile.objects.create(instance=instance2, file="test1.jpg")
+        # Instance 3. Form is soft-deleted
+        form3 = Form.objects.create(name="Form 3", deleted_at=datetime.datetime.now())
```
tdethier, line 418: Nitpick: it's better to use now() from django.utils.timezone because it's aware of the Django time zone settings

tdethier (approved):
LGTM 👍

PR #2803, opened 2026-03-10
[IA-4851]: Added generic ColorBadge component (+ tests) and use it in teams column

hugo093:

In the teams search UI, a colorPicker was present in the table and is quite confusing for the user

### Related JIRA tickets

IA-4851

## Changes

* Made a generic ColorBadge component
* Adapted the user detail view to use that component
* Adapted the teams list to display a badge instead of a color picker in the table

## How to test

* Go to Users -> Management 
  * search for any user and pick one (eye icon)
  * Color badge should be displayed
* Go to Users -> Teams
  * Search for teams
  * The color displayed in the column should be a badge and not a color picker anymore


beygorghor (approved):
Maybe use ColorBadge  in users table, like we are doing on the teams page

PR #2804, opened 2026-03-11
IA-4849: Entity list cursor pagination

Phil-V:

Implement cursor pagination for the entities list view.

### Related JIRA tickets

IA-4849

## Changes

- Support cursor pagination in the entities API (when the `cursor` GET parameter is passed).
- Separate the count of results into a separate endpoint and UI component, to be retrieved only if necessary. 
- Add cursor pagination controls to the entities list page.

## How to test

Visit the Entities List Page and test that the pagination, ordering of the columns, accuracy of the count and filters all work without issues.

## Print screen / video

/

## Notes

Known issues:

- An issue can arise in the ordering of json attributes with an explicit "null" value, that are treated differently by postgres. This is explained in the code and fixable but not fixed as of yet. 
- I've noticed an existing issue with that causes ordering to fail when the name of a json attribute shadows the name of an attribute on the entity model (the frontend passes "name" to the ordering in both cases for instance). This isn't addressed by this PR.

Note: 
- This is a subtask and a follow-up PR with performance optimisation is on the way (IA-4847)

## Doc

/


iaso/tests/api/entities/test_entities.py
```diff
@@ -1404,3 +1404,380 @@ def test_ordering_by_attributes_span(self):
         self.assertEqual(result[0]["id"], self.entity_1.id)  # Burkina Faso
         self.assertEqual(result[1]["id"], self.entity_3.id)  # karo
         self.assertEqual(result[2]["id"], self.entity_2.id)  # Yaba
+
+
+class WebEntityCursorPaginationAPITestCase(EntityAPITestCase):
```
tdethier: I would maybe refactor this `test_entities.py` file, it's so huge. Or at least move this to a dedicated test file `test_pagination.py`, in an `entities` directory

iaso/tests/api/entities/test_entities.py
```diff
@@ -1404,3 +1404,380 @@ def test_ordering_by_attributes_span(self):
...
+        next_cursor = data_p1["next"]
+
+        # Delete the last item on page 1 (which the 'next' cursor is referencing)
+        last_item_id = data_p1["result"][1]["id"]
+        m.Entity.objects.filter(id=last_item_id).delete()
+
+        res_p2 = self.client.get("/api/entities/", data=params | {"cursor": next_cursor}, format="json")
+        data_p2 = self.assertJSONResponse(res_p2, 200)
+
+        self.assertTrue(len(data_p2["result"]) > 0)
```
tdethier: Shouldn't the result still be `2`? Should we assert on a specific value instead of `> 0`?

Nitpick: otherwise, there's `self.assertGreater()`

iaso/api/entities/pagination.py
```diff
@@ -52,3 +60,227 @@ def get_paginated_response(self, data):
...
+    def _apply_cursor_filter(self, queryset, cursor, direction_reverse):
+        """
+        Filter the queryset based on the cursor's position.
+
+        Build a "cascading" OR condition.
+        For example, if ordering by (field_A, field_B, field_id) with cursor
+        values (a, b, id), we must fetch records where:
+            (field_A > a) OR
+            (field_A == a AND field_B > b) OR
+            (field_A == a AND field_B == b AND field_id > id)
```
tdethier, line 237: I don't think you have tests where you order by multiple values, could you please add some to make sure that this method works as expected?

hat/assets/js/apps/Iaso/domains/entities/components/CursorPagination.tsx
```diff
@@ -0,0 +1,69 @@
...
+        >
+            <Box
+                sx={{
+                    //  Reset `mr={8}` on PageRowSelect
+                    '& > div': {
+                        marginRight: '0 !important',
+                    },
+                }}
+                mr={4}
+            >
```
tdethier, line 41: Not sure about this, is it allowed @beygorghor? 😁

tdethier (approved):
The feature works fine 👍 

I was able to order by various columns and paginate properly, but I don't have enough data locally to notice performance changes. I left some comments regarding some refactor and tests.

I would probably create new tickets to address the two issues mentioned in this PR notes section (if they don't already exist)

hat/assets/js/apps/Iaso/domains/entities/components/CursorPagination.tsx
beygorghor: We can do that, but it's not recommended, we need to adapt PageRowSelect in bluesquare-components in order to receive some styling props. This can be one in another ticket

Phil-V: Thanks for the review 🙏 will create a separate PR for the styling props

PR #2805, opened 2026-03-11

oschvr:

### Related JIRA tickets

IA-4811

## Changes

## How to test

**** TODO

## Print screen / video

Screenshot 2026-03-11 at 11 38 27

## Notes

## Doc

***** TODO


docker-compose.yml
```diff
@@ -4,38 +4,14 @@ services:
...
-            # jupyter notebooks
-            - ./notebooks:/opt/notebooks
-            # Optional Used to load dev bluesquare-components JS. See `Live Bluesquare components` in doc
-            - ../bluesquare-components:/opt/bluesquare-components
-            # Optional Used to load dev snt-malaria-budgeting Python package.
-            - ../snt-malaria-budgeting:/opt/snt-malaria-budgeting
         links: &hat_links
             - db
         environment: &hat_environment
+            DISABLE_SSL_REDIRECT: true
```
tdethier: Why do you need to add this variable? If it's really useful, I would place it alphabetically 😇

docker-compose.yml
```diff
@@ -114,28 +90,11 @@ services:
         logging: *iaso_logging
         ports:
             - '5433:5432'
-        volumes:
-            - ../pgdata-iaso-16:/var/lib/postgresql/data
         environment:
             POSTGRES_DB: ${RDS_DB_NAME}
             POSTGRES_USER: postgres
             POSTGRES_PASSWORD: postgres
```
tdethier: If this docker compose file is intended for production only, I guess we can also get rid of the `db` section, we're not deploying databases in docker

docker-compose.yml
```diff
@@ -114,28 +90,11 @@ services:
...
-            - iaso
-        ports:
-            - '3000:3000'
-        volumes: *hat_volumes
-        environment: *hat_environment
-        logging: *iaso_logging
-        command: start_webpack
 
     # celery worker for wfp plugin
     redis:
```
tdethier: Same remark here, I'm not sure this is useful for production

tdethier: This sounds like we should definitely have 2 different docker compose files

docker/prod/Dockerfile
```diff
@@ -0,0 +1,113 @@
...
+    && rm -rf /var/lib/apt/lists/*
+
+# Create virtual environment
+RUN python -m venv /venv
+ENV PATH="/venv/bin:$PATH"
+
+# Install python dependencies + gunicorn
+COPY requirements.txt ./
+RUN pip install --no-cache-dir -r requirements.txt && \
+    pip install --no-cache-dir gunicorn
```
tdethier, line 48: I know you're going to say "not now", but this reminds me that we should definitely switch to `uv`

docker/prod/Dockerfile
```diff
@@ -0,0 +1,113 @@
...
+    find /venv -type f -name "*.dist-info" -type d -exec rm -rf {} + 2>/dev/null || true && \
+    find /venv -type f -name "*.so" -exec strip --strip-debug {} + 2>/dev/null || true
+
+# ── Stage 3: Production image ─────────────────────────────────────────────────
+# Slim image with runtime-only libraries.
+FROM python:3.9-slim-bullseye AS prod
+
+# Create non-root user early so COPY --chown can use it (avoids chown layer duplication)
+RUN groupadd --gid 1001 appuser && \
+    useradd --uid 1001 --gid appuser --shell /bin/bash --create-home appuser
```
tdethier, line 63: Any specific reason for setting a `gid` and `uid`?

docker/prod/Dockerfile
```diff
@@ -0,0 +1,113 @@
...
+
+# Runtime-only dependencies:
+#  - libgdal28: GDAL shared library (runtime, no headers)
+#  - libpq5: PostgreSQL client library (for psycopg2)
+#  - postgresql-client: psql CLI (used in wait_for_dbs.sh)
```
tdethier, line 68: `wait_for_dbs.sh` is like a manual health check for making sure that we start the backend only when DB is ready, we should probably get rid of it too and use docker health checks

docker/prod/Dockerfile
```diff
@@ -0,0 +1,113 @@
...
+
+# Copy project code (all owned by appuser, no separate chown layer needed)
+COPY --chown=appuser:appuser hat ./hat
+COPY --chown=appuser:appuser iaso ./iaso
+COPY --chown=appuser:appuser setuper ./setuper
+COPY --chown=appuser:appuser beanstalk_worker ./beanstalk_worker
+COPY --chown=appuser:appuser django_sql_dashboard_export ./django_sql_dashboard_export
+COPY --chown=appuser:appuser plugins ./plugins
+COPY --chown=appuser:appuser scripts ./scripts
+COPY --chown=appuser:appuser README.md manage.py entrypoint.sh tsconfig.json babel-register.js mypy.ini pyproject.toml ./
```
tdethier: I don't think `mypy.ini` is still used, but I'm not sure

docker/prod/Dockerfile
```diff
@@ -0,0 +1,113 @@
...
+
+# test data and notebooks
+COPY --chown=appuser:appuser testdata ./testdata
```
tdethier: If it's for prod, I'm not sure that we need `testdata` (I'd like to get rid of this completely, or move these files to `tests/fixtures/` at least

docker/prod/Dockerfile
```diff
@@ -0,0 +1,113 @@
...
+COPY --chown=appuser:appuser notebooks /opt/notebooks
```
tdethier: Here again, I'm not sure what it is and why we have that in the repo, but I think we could get rid of this

requirements.txt
```diff
@@ -146,6 +145,7 @@ requests-toolbelt==0.9.1
 # ------------------------------------------------------------------------------
 ipython==8.18.1  # https://github.com/ipython/ipython/tags
 django-extensions==3.2.3  # https://github.com/django-extensions/django-extensions/tags
+django-querycount==0.8.3  # https://github.com/bradmontgomery/django-querycount/tags
```
tdethier, line 149: It's already in `requirements-dev.txt`, why do we need to add it here?

tdethier (commented):
I managed to build the new image after increasing the docker desktop memory cap 👍
 
`docker image ls` shows `iaso:latest - 2.02GB - 455MB - U`, so that's a nice improvement.

I wanted to run the setuper on that setup, so I started a worker (worked fine), ran the setuper from outside the container and immediately encountered an issue:
```
iaso-1  |   File "/venv/lib/python3.9/site-packages/django/db/models/sql/compiler.py", line 1821, in execute_sql
iaso-1  |     for sql, params in self.as_sql():
iaso-1  |   File "/venv/lib/python3.9/site-packages/django/db/models/sql/compiler.py", line 1745, in as_sql
iaso-1  |     value_rows = [
iaso-1  |   File "/venv/lib/python3.9/site-packages/django/db/models/sql/compiler.py", line 1746, in <listcomp>
iaso-1  |     [
iaso-1  |   File "/venv/lib/python3.9/site-packages/django/db/models/sql/compiler.py", line 1747, in <listcomp>
iaso-1  |     self.prepare_value(field, self.pre_save_val(field, obj))
iaso-1  |   File "/venv/lib/python3.9/site-packages/django/db/models/sql/compiler.py", line 1695, in pre_save_val
iaso-1  |     return field.pre_save(obj, add=True)
iaso-1  |   File "/venv/lib/python3.9/site-packages/django/db/models/fields/files.py", line 317, in pre_save
iaso-1  |     file.save(file.name, file.file, save=False)
iaso-1  |   File "/venv/lib/python3.9/site-packages/django/db/models/fields/files.py", line 93, in save
iaso-1  |     self.name = self.storage.save(name, content, max_length=self.field.max_length)
iaso-1  |   File "/venv/lib/python3.9/site-packages/django/core/files/storage/base.py", line 49, in save
iaso-1  |     name = self._save(name, content)
iaso-1  |   File "/venv/lib/python3.9/site-packages/django/core/files/storage/filesystem.py", line 82, in _save
iaso-1  |     os.makedirs(directory, exist_ok=True)
iaso-1  |   File "/usr/local/lib/python3.9/os.py", line 215, in makedirs
iaso-1  |     makedirs(head, exist_ok=exist_ok)
iaso-1  |   File "/usr/local/lib/python3.9/os.py", line 215, in makedirs
iaso-1  |     makedirs(head, exist_ok=exist_ok)
iaso-1  |   File "/usr/local/lib/python3.9/os.py", line 225, in makedirs
iaso-1  |     mkdir(name, mode)
iaso-1  | PermissionError: [Errno 13] Permission denied: '/opt/app/media'
```

So, it looks like there's some rights issue with the media setup.


I also left a few comments and questions, I think we'll be able to get rid of even more stuff

docker-compose.yml
oschvr: Since I used the `bundle` Dockerfile as starting point, this is present here as default, https://github.com/BLSQ/iaso/blob/develop/docker/bundle/Dockerfile#L69

Normally the SSL offloading and redirection happens at the LoadBalancer/Proxy (i.e. nginx) 

This implies that IASO handles the TLS certificates, is that correct?

docker-compose.yml
oschvr: Correct me if I'm wrong but the idea of having a docker-compose is to be able to spinup the whole setup in one command (`docker compose up`)

This can be targeted for for `dev` and `prod` but alternatively, end users might not want to use `docker-compose` but something like kubernetes, and for that you only need the Dockerfile

docker-compose.yml
oschvr: correct, see above

docker/prod/Dockerfile
oschvr: hehe you're starting to know me, but hey, I think this optimization excercise can accomodate switching to `uv`

After reading this piece https://docs.astral.sh/uv/guides/migration/pip-to-project/ we would need to pin the `requirements.txt` at [project][dependencies]: https://github.com/BLSQ/iaso/blob/develop/pyproject.toml

What do you think? Should we include it ?

docker/prod/Dockerfile
oschvr: To get deterministic group and user ids all build will have them same, maybe useful in the future.

Also 0-999 are usually reserved for sys users

Not necessary really, id even say cosmetic

docker/prod/Dockerfile
oschvr: Is there a chance that we can do this psql ping using the python/django postgres ORM ?

That way we can discard installing `postgresql-client` and run something like 

`docker compose exec iaso wait_for_db.py`

requirements.txt
oschvr: this had to do with enabling/disabling `DEBUG=true` which instructs the webpack to build files or not.

DEBUG=true will build them, but the Dockerfile only includes `requirements.txt` which doesn't have this dependency, thus rendering the files would fail.

This one I am not quite sure how to reproduce, perhaps by clarifying the points above we can discard this change.

tdethier: Alright, since we know that we don't want DEBUG, we can definitely remove this line :)

docker/prod/Dockerfile
tdethier: I think that I would do something similar to [GAPS](https://github.com/BLSQ/gaps/blob/db1986ba34b48872f8b924778386f2ae1a0777fa/docker-compose.base.yml#L4). This means adding a new python dependency, if it's lighter than `postgresql-client` it might be worth it.

However, now that I think about it, this might a non-issue: we'll might want to keep `postgresql-client` if it allows us to start a DB shell on the prod server (which is sometimes useful).

oschvr:
I accidentally merged `main` into this PR, where I should have used `develop`. closing this in favour of https://github.com/BLSQ/iaso/pull/2832.

PR #2810, opened 2026-03-12
IA-4865: Add an endpoint for mobile to retrieve instance's information

bmonjoie:

The mobile needs an endpoint to retrieve instance information in order to download them.

### Related JIRA tickets

IA-4865

## Changes

Allowed `/api/mobile/instances/{id}/` endpoint and return the same data as the reference instances.

## How to test

Call `/api/mobile/instances/{id}/` with the ID of a known instance.

### DIFF
--- a/iaso/api/instances/views_mobile.py
+++ b/iaso/api/instances/views_mobile.py
@@ -1,13 +1,13 @@
 from django.db.models import Prefetch
 from rest_framework.decorators import action
-from rest_framework.exceptions import MethodNotAllowed
 from rest_framework.generics import get_object_or_404
+from rest_framework.mixins import RetrieveModelMixin
 from rest_framework.permissions import BasePermission
+from rest_framework.viewsets import GenericViewSet
 
-from iaso.api.common import ModelViewSet
-from iaso.api.instances.instances import HasInstancePermission, InstanceFileSerializer
-from iaso.api.instances.serializers import FileTypeSerializer
-from iaso.api.mobile.org_units import ReferenceInstancesSerializer
+from iaso.api.common import Paginator
+from iaso.api.instances.instances import HasInstancePermission
+from iaso.api.instances.serializers import FileTypeSerializer, InstanceFileSerializer, MobileInstancesSerializer
 from iaso.models import (
     Instance,
     InstanceFile,
@@ -24,16 +24,19 @@ class DenyAll(BasePermission):
         return False
 
 
-class InstancesMobileViewSet(ModelViewSet):
+class InstancesMobileViewSet(GenericViewSet, RetrieveModelMixin):
     """Mobile Instances API
 
+    GET /api/mobile/instances/<id>/
     GET /api/mobile/instances/<id>/attachments
     """
 
     permission_classes = [HasInstancePermission]
     http_method_names = ["get"]
     lookup_field = "uuid"
-    serializer_class = ReferenceInstancesSerializer
+    lookup_url_kwarg = "uuid"
+    serializer_class = MobileInstancesSerializer
+    pagination_class = Paginator
 
     def get_queryset(self):
         request = self.request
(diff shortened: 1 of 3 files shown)


iaso/api/instances/views_mobile.py
hugo093, line 24: this can be removed

iaso/api/instances/views_mobile.py
```diff
@@ -26,16 +30,23 @@ class InstancesMobileViewSet(ModelViewSet):
     GET /api/mobile/instances/<id>/attachments
```
hugo093, line 31: since we use just one endpoint that isn't destroy or list, i think we can just inherit from rest_framework.viewsets.GenericViewSet instead. 

Like that you could remove the def list()..

iaso/api/instances/views_mobile.py
```diff
@@ -26,16 +30,23 @@ class InstancesMobileViewSet(ModelViewSet):
...
         queryset = queryset.filter_for_user(request.user).filter_on_user_projects(user=request.user)
+        queryset = queryset.prefetch_related(
+            Prefetch("instancefile_set", queryset=InstanceFile.objects_with_file_extensions.all())
+        )
         return queryset
 
+    def list(self, request, *args, **kwargs):
+        raise MethodNotAllowed("GET")
+
     @action(["GET"], detail=True, permission_classes=[HasInstancePermission])
```
hugo093: permission_classes=[HasInstancePermission] can be removed as you defined it at the top level

iaso/api/instances/views_mobile.py
```diff
@@ -26,16 +30,23 @@ class InstancesMobileViewSet(ModelViewSet):
     GET /api/mobile/instances/<id>/attachments
     """
 
-    permission_classes = [DenyAll]
+    permission_classes = [HasInstancePermission]
     http_method_names = ["get"]
     lookup_field = "uuid"
+    serializer_class = ReferenceInstancesSerializer
```
hugo093: seeing the endpoint below aka attachment, it's not used ? so you should change the attachement endpoint def and modify this serializer_class

iaso/api/instances/views_mobile.py
```diff
@@ -26,16 +30,23 @@ class InstancesMobileViewSet(ModelViewSet):
...
         request = self.request
         queryset: InstanceQuerySet = Instance.objects
         queryset = queryset.filter_for_user(request.user).filter_on_user_projects(user=request.user)
+        queryset = queryset.prefetch_related(
+            Prefetch("instancefile_set", queryset=InstanceFile.objects_with_file_extensions.all())
+        )
         return queryset
 
+    def list(self, request, *args, **kwargs):
+        raise MethodNotAllowed("GET")
```
hugo093: see remark above, could be removed (it would just trigger a 404)

iaso/api/instances/views_mobile.py
hugo093: self.get_object(), the whole block try, except could be even moved into the def of get_object()

iaso/api/instances/views_mobile.py
```diff
@@ -26,16 +30,23 @@ class InstancesMobileViewSet(ModelViewSet):
     GET /api/mobile/instances/<id>/attachments
     """
 
-    permission_classes = [DenyAll]
+    permission_classes = [HasInstancePermission]
     http_method_names = ["get"]
     lookup_field = "uuid"
```
hugo093, line 36: add also lookup_url_kwarg = "uuid"

iaso/api/instances/views_mobile.py
hugo093, line 71: Define in a pagination.py file the paginator that correspond to this and use it at top level of your viewset

iaso/api/instances/views_mobile.py
```diff
@@ -26,16 +30,23 @@ class InstancesMobileViewSet(ModelViewSet):
...
+        queryset = queryset.prefetch_related(
+            Prefetch("instancefile_set", queryset=InstanceFile.objects_with_file_extensions.all())
+        )
         return queryset
 
+    def list(self, request, *args, **kwargs):
+        raise MethodNotAllowed("GET")
+
     @action(["GET"], detail=True, permission_classes=[HasInstancePermission])
     def attachments(self, request, uuid):
```
hugo093: this whole endpoint should be cleaned up and redone properly imho

hugo093 (changes requested):
I'm requesting changes because I think that any other endpoints than /attachements should raise a 404 and not a 405. For the refactor and cleaning of the attachements endpoint in itself, it can be done in another ticket

iaso/api/instances/views_mobile.py
bmonjoie: This is not part of the PR :slightly_smiling_face:

iaso/api/instances/views_mobile.py
bmonjoie: Changing it to `GenericViewSet` requires me to redefine `def retrieve(self, request)`, I can also import `RetrieveModelMixin`  but then, it's basically the same as using `ModelViewSet`.
Also, either redefining `def retrieve(self, request)` or implementing `RetrieveModelMixin`, automatically change the return error code from 404 to 405 for delete and patch.

iaso/api/instances/views_mobile.py
bmonjoie: Yes, but I liked that it was explicit, since this is an action and not part of the `ModelViewSet`.

iaso/api/instances/views_mobile.py
bmonjoie: Why would you assume it is not used?
It's an action, it has a different serializer than the `ModelViewSet`

hugo093 (approved):

PR #2812, opened 2026-03-13
[IA-4855]: First draft of API design

hugo093:

Find a way to coordinate and establish a "contract" between backend , frontend and mobile without needing to develop everything in django

### Related JIRA tickets

IA-4855

## Changes

Created an openAPI file to be able to visualize the endpoints , requests and responses

## How to test

If you have a JetBrain IDE:

* Make sure the plugin _OpenAPI Specifications_ is installed
* Open the file, a preview alongside the editor should be available

Otherwise:

You can go to [swagger editor](https://editor.swagger.io/)   and load the yaml file to preview it

### DIFF
--- a/iaso/api/validation_workflows/api_design.yaml
+++ b/iaso/api/validation_workflows/api_design.yaml
@@ -1,0 +1,754 @@
+openapi: 3.1.0
+info:
+  title: Validation Workflow API - OpenAPI 3.1
+  description: |-
+    This is a first draft of what the API would look like.
+  version: 1.0.0
+tags:
+  - name: Validation workflows
+    description: Everything about the static workflow definitions
+  - name: Validation workflow node templates
+    description: Everything about the static workflow node templates definitions
+  - name: Validation nodes
+    description: Everything about the validation nodes aka tasks
+
+
+paths:
+
+  /validation-workflows:
+    get:
+      x-permissions-required:
+        - authenticated
+        - iaso.iaso_validation_workflow.read
+      description: Filter and list all validation workflows linked to the user (accountId)
+      tags:
+        - Validation workflows
+      parameters:
+        - name: forms
+          in: query
+          schema:
+            type: array
+            items:
+              type: integer
+          description: related (multiple) form artifact allowed
+          required: false
+        - name: name
+          in: query
+          schema:
+            type: string
+          required: false
+      responses:
+        "200":
+          description: "OK"
+          content:
+            application/json:
+              schema:
+                properties:
+                  count:
+                    type: integer
+                  hasNext:
+                    type: boolean
+                  hasPrevious:
+                    type: boolean
+                  page:
+                    type: integer
+                  pages:
+                    type: integer
+                  limit:
+                    type: integer
+                  results:
+                    type: array
+                    items:
+                      $ref: '#/components/schemas/ValidationWorkflowList'
+    post:
+      x-permissions-required:
+        - authenticated
+        - iaso.iaso_validation_workflow.write
+      description: Create a validation workflow
+      tags:
+        - Validation workflows
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/CreateValidationWorkflowSchema'
+        required: true
+      responses:
+        "201":
+          description: "Workflow created"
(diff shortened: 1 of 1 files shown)


iaso/api/validation_workflows/api_design.yaml
```diff
@@ -0,0 +1,754 @@
...
+        - authenticated
+        - iaso.iaso_validation_workflow.write
+      description: Create a validation workflow
+      tags:
+        - Validation workflows
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/CreateValidationWorkflowSchema'
```
quang-le, line 74: So we pass the name and we receive a the slug in the response?
I would let users define the slug at least optionally

iaso/api/validation_workflows/api_design.yaml
```diff
@@ -0,0 +1,754 @@
...
+
+    delete:
+      x-permissions-required:
+        - authenticated
+        - iaso.iaso_validation_workflow.write
+      parameters:
+        - $ref: '#/components/parameters/validationWorkflowSlugParam'
+      tags:
+        - Validation workflows
+      summary: Delete a specific validation workflow
```
quang-le, line 151: What happens to the nodes when we soft delete the workflow?
Either they should be (soft-)deleted as well, or filtered out the node's API responses

iaso/api/validation_workflows/api_design.yaml
```diff
@@ -0,0 +1,754 @@
...
+    parameters:
+      - $ref: '#/components/parameters/validationWorkflowSlugParam'
+      - $ref: '#/components/parameters/instanceIdParam'
+    get:
+      tags:
+        - Validation nodes
+      x-permissions-required:
+        - authenticated
+        - iaso.iaso_submissions.read
+      summary: See details of the active validation process for that instance
```
quang-le, line 341: I will need to see the user role(s) needed to validate/reject the current node (name for display, id for actual checking)

iaso/api/validation_workflows/api_design.yaml
```diff
@@ -0,0 +1,754 @@
...
+                    type:
+                      - string
+                      - 'null'
+                  history:
+                    type: array
+                    items:
+                      $ref: '#/components/schemas/InstanceValidationWorkflowRetrieve'
+
+
+  /validation-workflows/{validationWorkflowSlug}/instance/{instanceId}/node-templates/{nodeTemplateSlug}/complete:
```
quang-le: I don't really mind, but do we really need both the WF slug and the instance id?

iaso/api/validation_workflows/api_design.yaml
quang-le: I will also need the node ids, for the undo feature

iaso/api/validation_workflows/api_design.yaml
hugo093: On my side the slug is auto generated from the name, so imho it should stay like that

iaso/api/validation_workflows/api_design.yaml
hugo093: no indeed :)

iaso/api/validation_workflows/api_design.yaml
hugo093: If a workflow is soft deleted, i would filter out all data related to it (nodes, validationnodes, ..)

iaso/api/validation_workflows/api_design.yaml
quang-le: Ok, but then how's the name slugified?  Will stuff like `Vérification des formulaires` and `Verification des formulaires` generate different slugs? If not we might need a unique constraint on the name as well

quang-le (approved):

iaso/api/validation_workflows/api_design.yaml
hugo093: I'm using [django-autoslug](https://django-autoslug.readthedocs.io/en/stable/) for this, there are already tests under iaso.tests.models.test_validation_workflow. I can add a unique constraint on name if needed

PR #2813, opened 2026-03-13
[IA-4732] Allow to search by organization + tests

hugo093:

It should be possible to filter users by organization through the search input on "list user" UI

### Related JIRA tickets

IA-4732

## Changes

Adapt  the search filter to contain organization as well

## How to test

* Python tests were added
* Front-end:
  * Go to Management => Users 
  * In the "search" input, type a value related to an organization (icontains, partial or exact), it should filter the users that icontains this organization

### DIFF
--- a/iaso/api/profiles/filters.py
+++ b/iaso/api/profiles/filters.py
@@ -49,6 +49,7 @@ class ProfileListFilter(django_filters.rest_framework.FilterSet):
             | Q(user__tenant_user__main_user__username__icontains=value)
             | Q(user__first_name__icontains=value)
             | Q(user__last_name__icontains=value)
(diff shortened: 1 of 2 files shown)


tdethier (approved):
The feature works fine 👍 

But I spotted a weird bug due to the v2 I think: since organizations are not returned by the list call and opening the edit modal does not trigger a retrieve call, the `Organization` field is empty, even though there's a value that should be there

Image

This means that we'll probably have the same issue with all fields that are not part of the list serializer: organization, phone number, home page...

hugo093:
Will create a ticket for the edit fields not being filled in in the list view

PR #2814, opened 2026-03-13
[IA-4869] new endpoint for user dropdown

hugo093:

User select box are using the new profiles list endpoint which is paginated and wasn't before

### Related JIRA tickets

IA-4869

## Changes

Backend:

* New dropdown endpoint
  * Not paginated by default
  * Paginated if "limit" query param is passed
  * Tests

Front end:
* Adapted most of the calls to use the new dropdown endpoints 
  * Team UI
  * Links UI
  * Entities UI
  * Instances
* Quickly fixed the polio country configuration UI, but will need to have a proper setup to improve it
    
## How to test

* Go to the mentioned UIs above:
  * Autocomplete user select box should work and be paginated to max 100 results
  * Other user select boxes shouldn't be paginated


hat/assets/js/apps/Iaso/domains/instances/hooks/useGetProfilesDropdown.tsx
```diff
@@ -1,34 +1,29 @@
...
+        queryKey: ['profilesDropdown', ids],
         queryFn: () => {
             if (ids === undefined || ids.length < 1) {
                 return [];
             }
-            return getRequest(makeUrlWithParams('/api/profiles/', { ids }));
+            return getRequest(
+                makeUrlWithParams('/api/profiles/dropdown/', {
+                    ids,
+                    limit: 100,
```
quang-le: I know users shouldn't select more than 100 items in a dropdown, but this will limit the the results displayed by the `AsyncSelect`

hat/assets/js/apps/Iaso/domains/links/hooks/filters.ts
```diff
@@ -67,17 +67,12 @@ export const useGetProfilesOptions = (): UseQueryResult<
     DropdownOptions<number>[]
 > => {
     return useSnackQuery({
-        queryKey: ['profiles', 'dropdown'],
-        queryFn: () => getRequest('/api/profiles/'),
+        queryKey: ['profilesDropdown'],
```
quang-le: Not sure about this change: are you sure the query cache will be invalidated when a write operation is performed on the profiles API?

hat/assets/js/apps/Iaso/domains/teams/hooks/requests/useGetProfilesDropdown.ts
```diff
@@ -1,29 +1,16 @@
...
-import getDisplayName, { Profile } from '../../../../utils/usersUtils';
 
 export const useGetProfilesDropdown = (): UseQueryResult<
     DropdownOptions<number>,
     Error
 > => {
     return useSnackQuery(
-        ['profiles'],
-        () => getRequest('/api/profiles/'),
+        ['profilesDropdown'],
```
quang-le: same same

plugins/polio/js/src/domains/Config/CountryNotification/Table/CountryNotificationsConfigTable.tsx
```diff
@@ -71,7 +71,7 @@ export const CountryNotificationsConfigTable: FunctionComponent<Props> = ({
                         language={settings.row.original.language}
                         users={settings.row.original.users}
                         teams={settings.row.original.teams}
-                        allUsers={allUsers?.profiles}
+                        allUsers={allUsers?.results}
```
quang-le: This will not work. I think this part requires a more thorough refactor:
- Use `UserAsyncSelect` in the modal
- Be able to exclude from API response users without email
- Probably some cleaning up with query keys and response shapes/types

hat/assets/js/apps/Iaso/domains/teams/hooks/requests/useGetProfilesDropdown.ts
quang-le: Also: we should only keep one of these `useGetProfilesDropDown` hooks and make it confiugurable for the 3 use cases we currently have (teams, instances, polio)

quang-le (changes requested):

hat/assets/js/apps/Iaso/domains/instances/hooks/useGetProfilesDropdown.tsx
hugo093: I discussed this with @beygorghor , since it's an autocomplete, do we really need to display more than 100 results? If so I can remove this limit param

quang-le: Since it's the request that will display the actual selection I would remove it. We're already limiting the response size by passing the ids

hat/assets/js/apps/Iaso/components/filters/UserAsyncSelect.tsx
```diff
@@ -1,26 +1,60 @@
...
+    ComponentProps<typeof AsyncSelect>,
+    | 'keyValue'
+    | 'label'
+    | 'multi'
+    | 'value'
+    | 'onChange'
+    | 'fetchOptions'
+    | 'clearable'
+    | 'debounceTime'
+>;
```
quang-le, line 26: ❤️

hat/assets/js/apps/Iaso/domains/users/hooks/useGetProfilesDropdown.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    const shouldTriggerWithEmptyQuery =
+        typeof triggerWithEmptyQuery === 'function'
+            ? triggerWithEmptyQuery()
+            : triggerWithEmptyQuery;
+
+    return useSnackQuery({
+        queryKey: [
+            'profiles',
+            params ?? {},
```
quang-le: this null check is not necessary, since you already spread `{}` by default line 48

hat/assets/js/apps/Iaso/domains/users/hooks/useGetProfilesDropdown.ts
```diff
@@ -0,0 +1,83 @@
...
+            shouldTriggerWithEmptyQuery ? 'triggerWithEmptyQuery' : undefined,
+        ].filter(Boolean),
+        queryFn: () => {
+            if (isEmpty(baseParams) && !shouldTriggerWithEmptyQuery) {
```
quang-le, line 65: what's the benefit of using this approach,i.o just the `enabled` option?

hat/assets/js/apps/Iaso/components/filters/UserAsyncSelect.tsx
```diff
@@ -1,26 +1,60 @@
...
 
-export const UserAsyncSelect: FC<Props> = ({
+export const UserAsyncSelect = ({
     handleChange,
     filterUsers,
+    additionalFilters = {},
```
quang-le: also not really necessary, since the underlying `useGetProfilesDropdown` can handle this prop being undefined

quang-le (commented):
I still need to run it, be here are already a few comments

hat/assets/js/apps/Iaso/domains/users/hooks/useGetProfilesDropdown.ts
hugo093: iirc the use case i had : because using enabled option would mean that the query wouldn't trigger and we would still use the old result. And it caused problems when going from one "state" to another

hat/assets/js/apps/Iaso/domains/instances/hooks/requests/getUsersDropDown.tsx
```diff
@@ -1,16 +1,26 @@
...
-                value: user.user_id,
-                label: getDisplayName(user),
-            };
-        }),
+type Props = {
+    query: string;
+    limit?: number;
+    queryClient: QueryClient;
+};
+export const getUsersDropDown = async ({
```
quang-le, line 12: You need to be able to apply the exclude users without email here as well, otherwise you'll be able to find and select them using text search, but they won't be displayed (as `Chip`) once selected

quang-le (changes requested):
just the last one detail for polio, otherwise LGTM

quang-le (approved):

PR #2815, opened 2026-03-13
[IA-4867] use a retrieve call to display data in edit user modal

hugo093:

In the list users view, the edit modal didn't have all the fields filled in initially

### Related JIRA tickets

IA-4867

## Changes

* Make a get profile call when opening the edit modal to fill in the values from there

## How to test

* Go to management => users
* Click on edit icon of any user listed 
* All the fields should be filled in according to the user initial values
* Go on the detail view of any user (eye icon)
* Click update user
* All the fields should be filled in according to the user initial values

## Notes

* Redoing the EditUserDialog form in a more standard way (Formik) will be done in a separate ticket

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/users/components/EditUserDialog.tsx
+++ b/hat/assets/js/apps/Iaso/domains/users/components/EditUserDialog.tsx
@@ -9,6 +9,7 @@ import { Box, Tab, Tabs } from '@mui/material';
 import {
     ConfirmCancelModal,
     IntlMessage,
+    LoadingSpinner,
     makeFullModal,
     useSafeIntl,
 } from 'bluesquare-components';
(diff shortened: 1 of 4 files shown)


quang-le (commented):
Can't we just have the `list` action return the `organization` field?

hugo093:
> Can't we just have the `list` action return the `organization` field?

Yes and no; 

Yes because indeed it would be nice to have it in the table columns

No because it's not just about the organization but also the permissions, org_unit_types, dhis2, etc etc

quang-le:
> > Can't we just have the `list` action return the `organization` field?
> 
> Yes and no;
> 
> Yes because indeed it would be nice to have it in the table columns
> 
> No because it's not just about the organization but also the permissions, org_unit_types, dhis2, etc etc

Ok, what I would then suggest, is to keep using the `retrieve` call for the details/modal, but open a ticket to add a `fields` queryparam to the list API, so we can filter out the fields not shown in the table

quang-le (approved):
Just see prev comment about list API. I suggest another ticket, because it's not super urgent and we have other boiling stuff on the fire

PR #2820, opened 2026-03-16
[IA-4866] replace drf-yasg by drf-spectacular

hugo093:

Our swagger wasn't working and using drf-yasg that is no more maintained.

### Related JIRA tickets

IA-4866

## Changes

* Replace drf-yasg by drf-spectacular
* Made the related changes
* Added a test to always check that our swagger is working

## How to test

* Create a superuser with an account and profile
* Open the urls /swagger, /swagger-ui or /redoc
* Enjoy

## Notes

drf-spectacular, on the opposite to drf-yasg, actually looks at the viewsets and serializer. 

Hence had a lot of errors saying "User has no attribute iaso_profile" or "Profile has no attribute account" because the code assumes everywhere that user.iaso_profile.account exists. 

In order to fix this, instead of modifying the code everywhere, I added a "permission check" to access swagger to actually have that iaso_profile and account. So you'll need a superuser with iaso_profile and account, contrary to drf-yasg

### DIFF
--- a/iaso/api/org_unit_change_request_configurations/views_mobile.py
+++ b/iaso/api/org_unit_change_request_configurations/views_mobile.py
@@ -4,8 +4,8 @@ import django_filters
 
 from django.db.models import Q
 from django.utils import timezone
-from drf_yasg import openapi
-from drf_yasg.utils import swagger_auto_schema
+from drf_spectacular.types import OpenApiTypes
+from drf_spectacular.utils import OpenApiParameter, extend_schema
 from rest_framework import viewsets
 from rest_framework.mixins import ListModelMixin
 from rest_framework.request import Request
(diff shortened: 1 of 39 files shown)


iaso/tests/test_open_api.py
Phil-V, line 1: Good job adding a test for this 👍

iaso/api/org_unit_change_request_configurations/views_mobile.py
```diff
@@ -30,19 +30,19 @@ class MobileOrgUnitChangeRequestConfigurationViewSet(ListModelMixin, viewsets.Ge
...
         name=APP_ID,
-        in_=openapi.IN_QUERY,
+        location=OpenApiParameter.QUERY,
         required=True,
         description="Application id (`Project.app_id`)",
-        type=openapi.TYPE_STRING,
+        type=OpenApiTypes.STR,
     )
-    include_creation_param = openapi.Parameter(
+    include_creation_param = OpenApiParameter(
```
Phil-V, line 40: Not an issue introduced by your PR, but unless I'm mistaken, this param doesn't seem to be included?

Phil-V (approved):
Thanks for making this actually work! 🥳

For some of the endpoints (powerbi, superset, etc.), you removed them from the docs, was it because they're accessible unauthenticated?

iaso/api/org_unit_change_request_configurations/views_mobile.py
hugo093: It's from includeCreationSerializer it seems :)

Phil-V: Or meant to be passed to extend_schema parameters on the list?

hugo093: Yes, IncludeCreationSerializer is used in the list

PR #2825, opened 2026-03-17
feat: backend validation workflow

hugo093:

Basic API operation for validation workflow template general feature

### Related JIRA tickets

IA-4868, IA-4875, IA-4877

## Changes

* Implemented endpoints for validation workflow API
* Added tests


iaso/api/validation_workflows/serializers/update.py
```diff
@@ -0,0 +1,34 @@
+from rest_framework import serializers
+
+from iaso.api.common import ModelSerializer
+from iaso.models import Form, ValidationWorkflow
+
+
+class ValidationWorkflowUpdateSerializer(ModelSerializer):
+    forms = serializers.PrimaryKeyRelatedField(
```
quang-le, line 8: Not sure we'll want to update the forms "from" the workflow. It might cause us headaches, because typically we'd want the workflows to be assignable to any form, but form access can be restricted so that would make our permission management trickier

iaso/api/validation_workflows/filters.py
```diff
@@ -0,0 +1,15 @@
...
+
+from django_filters.rest_framework import FilterSet
+
+from iaso.api.common import NumberInFilter
+from iaso.models import ValidationWorkflow
+
+
+class ValidationWorkflowListFilter(FilterSet):
+    forms = NumberInFilter(label="Related form ids", field_name="form_set__pk", lookup_expr="in", distinct=True)
+    name = django_filters.CharFilter(field_name="name", lookup_expr="icontains")
```
quang-le, line 12: front-end uses "search". Both make sense

iaso/api/common.py
```diff
@@ -531,3 +531,12 @@ class CharInFilter(BaseInFilter, CharFilter):
...
     """
+
+
+class NumberInFilter(BaseInFilter, NumberFilter):
+    """
+    Filter that checks if the model field is in a list of comma-separated numbers.
+
+    Usage (in a FilterSet):
+    >>> filter_name = NumberInFilter(field_name="model_field_name", lookup_expr="in")
+    """
```
quang-le: There's one already in iaso/api/utils/filters.py

quang-le (changes requested):

iaso/api/validation_workflows/serializers/update.py
hugo093: mmmh do you suggest an "exclude forms" field instead ?

quang-le (approved):

PR #2826, opened 2026-03-17
IA-4862: Fix OrgUnitChangeRequest creation fails when UUID already exists.

bmonjoie:

When a OUCR with the same UUID already exists,  the backend errors out. This fixes it by ignoring OUCR with UUIDs which already exist.

### Related JIRA tickets

IA-4862

## Changes

It just tests if the UUID already exists and then insert it normally if not.

## How to test

Create 1 OUCR with the API call and relaunch it afterwards.


tdethier (approved):
LGTM 👍

PR #2827, opened 2026-03-18
IA-4847: Entities list performance

Phil-V:

Improve the entity list performance.

### Related JIRA tickets

IA-4847

## Changes

- Avoid expensive annotations when not necessary and use prefetches instead
- Set default ordering to `-id` to avoid more expensive ordering if not necessary

## How to test

Visit the entities list page and:
- make sure listing, ordering, and csv/xlsx exports are not broken
-  check that entities with pending duplicates correctly have an extra "show duplicates" button in their "actions" column in the entity list
  - you can create a duplicate entry manually in /admin/iaso/entityduplicate/

## Print screen / video

/
## Notes

- Moved some tests from test_entities.py to test_entities_ordering.py. They are unchanged and don't need to be reviewed.
- switched to `-id` instead of `-created_at` as the latter is not indexed (to be potentially addressed in future task) 

## Doc

/


iaso/models/entity.py
```diff
@@ -316,3 +308,24 @@ def soft_delete_with_instances_and_pending_duplicates(self, audit_source, user):
         self.duplicates2.filter(validation_status=ValidationStatus.PENDING).delete()
 
         return self
+
+    @property
+    def pending_duplicate_ids(self):
+        """Retrieve the id list of related pending duplicate entities."""
+        results = set()
```
hugo093, line 317:
```python 
return list(
        self.duplicates1.filter(validation_status=ValidationStatus.PENDING)
        .values_list("id", flat=True)
        .union(
            self.duplicates2.filter(validation_status=ValidationStatus.PENDING)
            .values_list("id", flat=True)
        )
    )
```
 instead ?

iaso/models/entity.py
```diff
@@ -316,3 +308,24 @@ def soft_delete_with_instances_and_pending_duplicates(self, audit_source, user):
...
+
+    @property
+    def latest_instance_created_at(self):
+        """Retrieve the datetime of the last created instance for this entity."""
+        instance_dates = (
+            saved_at
+            for instance in self.instances.all()
+            if (saved_at := instance.source_created_at or instance.created_at) is not None
+        )
+        return max(instance_dates, default=self.created_at)
```
hugo093, line 332: I'd try to turn this into a sql operation (django ORM) as it might be expensive if there are a lot of instances. Moreover if you do that on a list it might trigger N+1 query problem. 

Be also aware that, imho, we should avoid putting `@property` on something that triggers a DB operation. (E.g some debuggers immediately evaluates it and then you don't get why you have more queries than expected)

iaso/api/entities/views.py
```diff
@@ -101,27 +101,39 @@ def pagination_class(self):
             return EntityCursorPagination
         return EntityListPaginator
 
+    @property
+    def _requested_ordering_fields(self):
+        """Access the raw ordering fields from the request."""
+
+        order_param = self.request.query_params.get("order_columns") or self.request.query_params.get("order") or ""
```
hugo093: instead of "order" put api_settings.ORDERING_PARAM. So if we change it, it'll adapt automatically

iaso/api/entities/views.py
```diff
@@ -205,7 +217,7 @@ def list(self, request: Request, *args, **kwargs):
 
         # Handle streaming responses
 
-        queryset = self.filter_queryset(self.get_queryset())
+        queryset = self.filter_queryset(self.get_queryset()).order_by("-id")
```
hugo093: just to be sure, this would erase any other ordering passed, do we want this ?

iaso/tests/api/entities/test_entities_ordering.py
```diff
@@ -0,0 +1,94 @@
...
+        self.entity_2.save()
+
+        self.entity_3 = m.Entity.objects.create(
+            name="Giraffe", entity_type=self.entity_type_2, attributes=inst_3, account=account
+        )
+        self.entity_3.created_at = base_time  # Newest
+        self.entity_3.save()
+
+    def test_ordering_default(self):
+        self.client.force_authenticate(self.yoda)
```
hugo093, line 46: put that line on setup if you do it everytime

hugo093 (changes requested):

iaso/models/entity.py
Phil-V: Good point on the properties, I've moved them to regular methods.

iaso/api/entities/views.py
Phil-V: Added a comment to clarify that this is for performance reasons (as there are few indexed columns available on the Entity model and the exports iterate over potentially a very high number of records).

iaso/models/entity.py
Phil-V: Sadly, using values_list doesn't seem to work with Prefetch. But, I've refined the code to filter on validation_status in the prefetch itself. Let me know if you can think of better ways to architecture this.

hugo093 (approved):
lgtm

PR #2832, opened 2026-03-18

oschvr:

### Related JIRA tickets

IA-4811

## Changes

## How to test

In the new docker-compose.yml file (meant for prod use), you have to make sure to populate the env vars, either through a .env or in your environment itself.

You will need a way to spin up a postgres:postgis instance and pass the connection details as env vars too.

## Print screen / video

Screenshot 2026-03-18 at 15 32 33

Screenshot 2026-03-18 at 14 57 39

## Notes

## Docker Compose

You will need to populate at least these environment variables with your own values in a `.env` file:

```
# PostgreSQL Database connection details
RDS_DB_NAME=
RDS_HOSTNAME=
RDS_PASSWORD=
RDS_USERNAME=
DB_READONLY_PASSWORD=
DB_READONLY_USERNAME=

# Used for encryption and authorisation
ENCRYPTED_TEXT_FIELD_KEY=
SECRET_KEY=

# To interact with Enketo/ODK
ENKETO_API_TOKEN=
ENKETO_SIGNING_SECRET=

# Docker image tag (defaults to "latest" if not set)
IMAGE_TAG=
```

Docker Compose automatically reads a `.env` file in the project root, so no `source` command is needed.

Alternatively, you can uncomment the `env_file:` section in `docker-compose.prod.yml` to pass the variables directly to the container.

Note: for production deployments you need an external PostgreSQL database. The `db` service in the compose file is only intended for local development. Make sure your `RDS_*` variables point to your production database.

Proceed to run docker compose on your server:

```
docker compose -f docker-compose.prod.yml up
```

This will pull the necessary containers (iaso & nginx) and spin up the service at port 80.


.platform/nginx/nginx.conf
```diff
@@ -0,0 +1,72 @@
...
+    error_log /var/log/nginx/error.log debug;
+    access_log /var/log/nginx/access.log detailed;
+
+    # Get allowed origin from environment variable
+    map $http_origin $cors_origin {
+        default "$http_origin";
+        "" "http://localhost";
+    }
+
+    upstream backend {
```
tdethier: Should we call this `iaso` instead of `backend`?

docker/prod/Dockerfile
```diff
@@ -0,0 +1,116 @@
...
+
+WORKDIR /opt/app
+
+# Copy webpack assets
+COPY --from=npmbuilder --chown=appuser:appuser /opt/app/hat/assets /opt/app/hat/assets
+
+# Copy project code (all owned by appuser, no separate chown layer needed)
+COPY --chown=appuser:appuser ../../hat ./hat
+COPY --chown=appuser:appuser ../../iaso ./iaso
+COPY --chown=appuser:appuser ../../setuper ./setuper
```
tdethier, line 95: I'm wondering if we should remove the setuper from this image. I asked the question on the iaso channel, let's see the answers

docker/prod/docker-compose.prod.yml
```diff
@@ -0,0 +1,119 @@
...
+      logging: *iaso_logging
+      ports:
+          - '5433:5432'
+      volumes:
+          - ../pgdata-iaso-16:/var/lib/postgresql/data
+      environment:
+          POSTGRES_DB: ${RDS_DB_NAME}
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
+
```
tdethier, line 146: We should comment this section, otherwise EBS will start a local DB

tdethier (commented):
After struggling a bit, I managed to run this prod image locally 👍 

I ran the setuper on this server, which worked fine

Image

docker/prod/Dockerfile
oschvr: removed

docker/prod/docker-compose.prod.yml
oschvr: commented out :)

.platform/nginx/nginx.conf
oschvr: I'm suspecting there might be an internal network DNS conflict, as Docker already used `iaso` as the DNS to resolve the `iaso` container. 

I'm reading about it and will test

oschvr: So in the `location /` block, the `proxy_pass` directive, already points directly to `iaso`.

[upstream defines a cluster that you can proxy requests to](http://wiki.nginx.org/HttpUpstreamModule).
https://stackoverflow.com/a/5877943 

But it was unused, so I removed it now

docker-compose.yml
```diff
@@ -1,145 +1,119 @@
...
-            WFP_EMAIL_RECIPIENTS_NEW_ACCOUNT:
-            USE_CELERY: ${USE_CELERY:-false}
-            CELERY_BROKER_URL: ${CELERY_BROKER_URL:-}
-            CELERY_RESULT_BACKEND: ${CELERY_RESULT_BACKEND:-}
-        logging: &iaso_logging
-            driver: 'json-file'
-            options:
-                max-size: '5k'
-        command: start_dev
+      image: blsq/iaso:${IMAGE_TAG:-latest}
```
tdethier: You should revert all the changes in this file, so that it can still be used for local development

.ebextensions/00_ami.config
```diff
@@ -1,4 +0,0 @@
-option_settings:
```
tdethier, line 1: As long as we keep this PR open, we can leave these files deleted. But once we merge, we need to make sure that all environments have switched to docker (or at least staging ones) because removing these EBS files will break deployments

oschvr:
If you're migrating to Docker, and you get this error:
> Invalid option specification (Namespace: 'aws:elasticbeanstalk:environment:proxy:staticfiles', OptionName: '/static'): Unknown configuration setting.

This one means that the django.config from .ebextensions already created the static files config, so we need to remove it from the environment. Run this command to remove it:
```
aws elasticbeanstalk update-environment --environment-name "${EB_ENVIRONMENT}" --options-to-remove Namespace=aws:elasticbeanstalk:environment:proxy:staticfiles,OptionName=/static
```

docker-compose.yml
tdethier: Don't forget about this @oschvr

.ebextensions/00_ami.config
mestachs: @oschvr @tdethier I would have tried the approach of the console deploying a subdirectory for docker eb and leaving the "normal" env untouched (so you can keep beanstalk for a moment for other iaso variants)

tdethier: Yeah, it's kinda what @oschvr has done so far, but not exactly

docker/prod/docker-compose.prod.yml
```diff
@@ -0,0 +1,122 @@
...
+        DNS_DOMAIN: ${DNS_DOMAIN}
+        EMAIL_BACKEND: ${EMAIL_BACKEND}
+        ENABLE_ANALYTICS: ${ENABLE_ANALYTICS:-false}
+        ENABLE_GZIP: ${ENABLE_GZIP:-true}
+        ENCRYPTED_TEXT_FIELD_KEY: ${ENCRYPTED_TEXT_FIELD_KEY}
+        ENKETO_API_TOKEN: ${ENKETO_API_TOKEN}
+        ENKETO_DEV: ${ENKETO_DEV:-false}
+        ENKETO_SIGNING_SECRET: ${ENKETO_SIGNING_SECRET}
+        ENKETO_URL: ${ENKETO_URL}
+        FAVICON_PATH: ${FAVICON_PATH}
```
tdethier, line 68: When you check env vars on the server, you see that they are all blank

Image



This is probably the reason why we don't have a favicon or an app title and so on:
- the env var value is not set in terraform (because we're fine with the default from django settings)
- docker compose turns this non-set var into a blank string, because we're explicitly passing it here
- django settings notice that there's a env var and use it instead of using the default value


We've had the same issue on console. We fixed it by [calling a custom method in settings that filters out blank strings](https://github.com/BLSQ/console/blob/6a4087b06ac0e00e6f72cdeff2da586a9ac954c9/console/settings.py#L739). There might be a better way though. Let's discuss this together when you're available!

.ebextensions/00_ami.config
oschvr: I've updated this approach, we are not touching existing files so as not to break anything

docker/prod/Dockerfile
```diff
@@ -0,0 +1,118 @@
...
+    && rm -rf /var/lib/apt/lists/*
+
+# Create virtual environment
+RUN python -m venv /venv
+ENV PATH="/venv/bin:$PATH"
+
+# Install python dependencies + gunicorn
+COPY ../../requirements.txt ./
+RUN pip install --no-cache-dir -r requirements.txt && \
+    pip install --no-cache-dir gunicorn py
```
tdethier: Why do we need to install `py`?

docker/prod/docker-compose.prod.yml
```diff
@@ -0,0 +1,155 @@
...
+        RDS_DB_NAME: ${RDS_DB_NAME}
+        RDS_HOSTNAME: ${RDS_HOSTNAME}
+        RDS_PASSWORD: ${RDS_PASSWORD}
+        RDS_USERNAME: ${RDS_USERNAME}
+        SECRET_KEY: ${SECRET_KEY}
+        SENTRY_ENVIRONMENT: ${SENTRY_ENVIRONMENT}
+        SENTRY_FRONT_ENABLED: ${SENTRY_FRONT_ENABLED}
+        SENTRY_TRACES_SAMPLE_RATE: ${SENTRY_TRACES_SAMPLE_RATE:-"0.1"}
+        SENTRY_ERRORS_SAMPLE_RATE: ${SENTRY_ERRORS_SAMPLE_RATE:-"1.0"}
+        SENTRY_ERRORS_HTTPERROR_SAMPLE_RATE: ${SENTRY_ERRORS_HTTPERROR_SAMPLE_RATE:-"0.8"}
```
tdethier: I think we no longer need quotes, but maybe I'm wrong

.github/workflows/deploy_prod.yml
```diff
@@ -0,0 +1,117 @@
...
+    steps:
+      # Merge the workflow_dispatch config with some default for deployment via the on:push
+      - name: Set configs
+        id: config
+        env:
+            DEFAULT_GIT_REFERENCE: main
+            DEFAULT_ENVIRONMENT: upgrade
+            DEFAULT_SLACK_CHANNEL: cloud-alerts
+        run: |
+          echo "GIT_REFERENCE=${{ github.ref_name || env.DEFAULT_GIT_REFERENCE }}" >> $GITHUB_OUTPUT
```
tdethier: Since this action is only manual, you'll never get a `github.ref_name`, so it's always going to use the other default value (`main`). I think that we should probably have another field to select which branch should be deployed

docker/prod/Dockerfile
oschvr: I don't think there is a particular reason. 
nothing that I see here https://py.readthedocs.io/en/latest/install.html tells me we're using it 

I did ask `claude` to review and it seems the sneaky bastard added this thing.

docker/prod/docker-compose.prod.yml
oschvr:
```
import os
def get_env_var_or_default(env_var, default):
    """
    When an environment variable is not set, docker compose can initialize it with a blank string
    instead of letting it unset, which overrides the default value from os.environ.get().
    This function returns the value if the environment variable is set, or the passed default value.
    """
    value = os.environ.get(env_var)
    if value in ["", None]:
        return default
    return value


def get_env_as_float(variable_name: str, default_str: str) -> float:
    var_str: str = get_env_var_or_default(variable_name, default_str)
    var_str = var_str.strip().strip('"').strip("'")
    try:
        return float(var_str)
    except ValueError:
        raise Exception(f"Error wrong {variable_name} value {var_str}, should be float")


print(get_env_as_float("SENTRY_TRACES_SAMPLE_RATE", "0.1"))
```

I get `0.1` so we should be ok

.github/workflows/deploy_prod.yml
oschvr: good catch. I'll also rename the workflow for now, to avoid confusion

.github/workflows/deploy_manual.yml
```diff
@@ -0,0 +1,120 @@
...
+        run: |
+          docker push blsq/iaso:latest
+          docker push blsq/iaso:${{ steps.version.outputs.VERSION_NAME }}
+
+      - name: Swap to prod docker config
+        run: |
+          # Replace docker-compose with prod version
+          cp docker/prod/docker-compose.prod.yml docker-compose.yml
+          git add -f docker-compose.yml
+
```
tdethier: I think there's some repetition with the deployment script

docker/prod/docker-compose.prod.yml
```diff
@@ -0,0 +1,148 @@
...
+    #       - '5433:5432'
+    #   volumes:
+    #       - ../pgdata-iaso-16:/var/lib/postgresql/data
+    #   environment:
+    #       POSTGRES_DB: ${RDS_DB_NAME}
+    #       POSTGRES_USER: postgres
+    #       POSTGRES_PASSWORD: postgres
+
+volumes:
+  static-files:
```
tdethier: Missing empty line at the end

docker-compose.prod.yml
```diff
@@ -0,0 +1,155 @@
+services:
```
tdethier: We should delete this file

.github/workflows/deploy_manual.yml
```diff
@@ -0,0 +1,120 @@
...
+          SLACK_WEBHOOK: ${{ secrets.TEST_SLACK_WEBHOOK }}
+          SLACK_CHANNEL: ${{ steps.config.outputs.SLACK_CHANNEL }}
+          SLACK_USERNAME: ${{ steps.config.outputs.ENVIRONMENT }}
+          SLACK_COLOR: ${{ job.status }}
+          SLACK_ICON_EMOJI: ":robot_face:"
+          SLACK_TITLE: Deployment failure Iaso ${{ steps.version.outputs.VERSION_NAME }} - ${{ steps.config.outputs.ENVIRONMENT }}
+          SLACK_MESSAGE: ${{  format('<https://github.com/{0}/commit/{1}/checks|🔗{2} logs> | <https://github.com/{0}/commit/{1}/checks|🔗 Commit {1}>', github.repository, github.sha, github.job) }}
+          SLACK_FOOTER: "-"
+          MSG_MINIMAL: true
+        if: failure()
```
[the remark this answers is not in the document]
tdethier: @oschvr I guess we could try this as well

scripts/deploy_prod_docker.sh
```diff
@@ -0,0 +1,88 @@
...
+VERSION_NAME=$(git describe --tags --match "v[[:digit:]]*")
+export "VERSION_NAME=${VERSION_NAME}"
+echo "==> Version name/tag ${VERSION_NAME}"
+
+# Deploy to prod
+echo "==> Deploying ${EB_ENVIRONMENT}"
+python "scripts/eb_deploy.py" "${EB_ENVIRONMENT}"
+
+# Timestamp
+date
```
tdethier: Missing empty line here as well

.github/workflows/deploy_manual.yml
oschvr: This file is no longer valid, we're doing all the CI related stuff on here https://github.com/BLSQ/iaso/blob/IA-4812-clean-and-optimize-c-is-to-build-images-for-deployment/.github/workflows/deploy_docker_manual.yml

docker-compose.prod.yml
oschvr: ah good catch. removed

scripts/deploy_prod_docker.sh
oschvr:
```
-> % shellcheck ./scripts/deploy_prod_docker.sh && echo "Script ok" || echo "Script not ok" 
Script ok
```

the bash validator, I don't see a new line

docker/prod/docker-compose.prod.yml
oschvr: Screenshot 2026-04-22 at 11 34 42

it says no newline at end of file ?

tdethier (approved):
LGTM, nice work 👍

PR #2834, opened 2026-03-18
[IA-4458] fix setuper slow again

Crebert08:

The setuper was slow. This PR is about proposing a fix for that.

### Related JIRA tickets

[IA-4458](https://bluesquare.atlassian.net/browse/IA-4458)

## Changes

- Implemented the serializer that supports fields parameter in `iaso/api/org_units.py`
- sent the patch request with 'id' as a field parameter

## How to test

> Run the setuper: `/home/crebert/Desktop/iaso/setuper/README.md` 
> The duplicates for the patch requests went from 101 to 6
> Total number of queries went from 118 to 28
> Response weight went from ~80Ko to ~10 o

## Notes

## Doc

[IA-4458]: https://bluesquare.atlassian.net/browse/IA-4458?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iaso/api/org_units.py
```diff
@@ -773,6 +770,13 @@ def partial_update(self, request, pk=None):
                     original_copy, org_unit, source=audit_models.ORG_UNIT_API, user=request.user
                 )
 
+                requested_fields = request.query_params.get("fields")
+                if requested_fields:
+                    serializer = OrgUnitSearchSerializer(
+                        org_unit, fields=requested_fields.split(","), context={"request": request}
+                    )
+                    return Response(serializer.data)
+
```
tdethier, line 784: Could you please add a unit test to make sure that you receive only the requested fields as a result of a PATCH?

tdethier (changes requested):
So, I ran the setuper locally with and without this PR in order to compare results - there was a slight improvement:

Without this PR:

Image

With this PR:

Image

-------
But there are still a lot of duplicated queries in multiple calls, which is something we can probably improve. When looking at random times, I still see very long queries, such as
- updating the default picture of an org unit

Image

- other random calls (sorry I did not check what the setuper was doing at that moment)

Image

Image

Image

Image

- this one is way too long, 2 seconds for a GET on org unit types is crazy:
Image

- PATCH on that endpoint is long as well:
Image


---------------------------------
`forms` and `sync` might be too old/difficult to fix, but I'm hoping to improve at least org unit types and profiles

tdethier (commented):
I ran the setuper again locally, it went down from 16 minutes to 13, which is great 👍 

Image

But I'm afraid that the changes that have been introduced might create some other issues. Let's discuss that tomorrow!

iaso/api/teams/serializers.py
```diff
@@ -117,11 +141,17 @@ def save(self, **kwargs):
...
-        Team.objects.bulk_update(teams_to_update, ["path"])
+
+        if set(old_sub_teams_ids) != set(new_sub_teams_ids):
+            team_changed_qs = Team.objects.filter(id__in=new_sub_teams_ids + old_sub_teams_ids)
+            teams_to_update = []
+            for team in team_changed_qs:
+                teams_to_update += team.calculate_paths(force_recalculate=True)
+
+            if teams_to_update:
+                unique_teams = {t.pk: t for t in teams_to_update}.values()
```
tdethier, line 129: I would do it with a set instead of a dict, but it works too

iaso/api/teams/serializers.py
```diff
@@ -53,15 +53,39 @@ class Meta:
...
+                if "users" in active_fields:
+                    active_fields["users"].child_relation.queryset = users_in_account
+
+            team_heavy_fields = {"sub_teams", "parent"}
+            if any(f in active_fields for f in team_heavy_fields):
+                user_teams = Team.objects.filter_for_user(user)
+                if "sub_teams" in active_fields:
+                    active_fields["sub_teams"].child_relation.queryset = user_teams
+                if "parent" in active_fields:
+                    active_fields["parent"].queryset = user_teams
```
tdethier: I would:
- revert these changes
- fetch `Team` data once, store it in a variable and use it for both fields which need it (like `users_in_account`)

iaso/api/teams/serializers.py
```diff
@@ -117,11 +141,17 @@ def save(self, **kwargs):
...
+        if set(old_sub_teams_ids) != set(new_sub_teams_ids):
+            team_changed_qs = Team.objects.filter(id__in=new_sub_teams_ids + old_sub_teams_ids)
+            teams_to_update = []
+            for team in team_changed_qs:
+                teams_to_update += team.calculate_paths(force_recalculate=True)
+
+            if teams_to_update:
+                unique_teams = {t.pk: t for t in teams_to_update}.values()
+                Team.objects.bulk_update(unique_teams, ["path"])
+
```
tdethier, line 131: You can keep this 👍

iaso/api/teams/views.py
```diff
@@ -59,11 +60,50 @@ def get_queryset(self):
...
+            return Response(
+                TeamSerializer(
+                    instance,
+                    fields=requested_fields.split(","),
+                    context={"request": request},
+                ).data,
+                status=status.HTTP_201_CREATED,
+            )
+
+        return Response(serializer.data, status=status.HTTP_201_CREATED)
```
tdethier: I would revert all of this

iaso/api/org_units.py
```diff
@@ -775,6 +775,13 @@ def partial_update(self, request, pk=None):
...
```
tdethier, line 784: You can keep this as well 👍

tdethier (changes requested):
Here's the summary of our call

tdethier (approved):
Somehow it was even faster this time 👍 

Image

I think that's already a nice improvement, thanks!

PR #2837, opened 2026-03-18
[IA-4875] backend crud node templates

hugo093:

Basic CRUD operation for the validation node templates

### Related JIRA tickets

IA-4875

## Changes

* Custom AutoslugField to generate unique slugs before bulk_create
* Remove the always update slug
* Added CRUD endpoints + bulk create and update
* Added drf-extension for nested serializer
* Re organized the common.py file 

## How to test

Run the python tests or use the API endpoints (documented in swagger => Validation workflow node templates)

## Notes

Add to create the BulkAutoSlugField because, for whatever reason, the AutoSlugField pre_save gets triggered after the bulk_create if the slugs aren't provided. So it raises a DB constraint error... 

This PR is based on this [PR](https://github.com/BLSQ/iaso/pull/2825)


iaso/api/common/__init__.py
```diff
@@ -0,0 +1,53 @@
...
+    DynamicFieldsModelSerializer,
+    ModelSerializer,
+    UserSerializer,
+)
+from .serializer_fields import DateTimestampField, HiddenSlugRelatedField, TimestampField
+from .utils import is_field_referenced, parse_comma_separated_numeric_values, safe_api_import
+from .views import DropdownOptionsListViewSet, EtlModelViewset, ModelViewSet
+
+
+__all__ = [
```
tdethier, line 21: Kudos for this refactoring 👍

iaso/tests/api/validation_workflows/test_views.py
```diff
@@ -894,6 +894,6 @@ def test_retrieve(self):
 
     def test_num_queries(self):
         self.client.force_authenticate(self.john_wick)
-        with self.assertNumQueries(7):
+        with self.assertNumQueries(10):
```
tdethier: Nitpick: it would be interesting to write a comment with all queries [like we did here](https://github.com/BLSQ/iaso/blob/7ae9ff12a8db50f2dfdc0dd04b861fa655099618/iaso/tests/api/org_unit_change_requests/test_org_unit_change_requests.py#L83), this way it's easier to understand where the extra queries come from

iaso/api/validation_workflows_node_templates/serializers/bulk_update.py
```diff
@@ -0,0 +1,117 @@
+from django.db import transaction
+from django.db.models import Q
+from rest_framework import serializers
+from rest_framework.exceptions import ValidationError
+
+from iaso.api.common import ModelSerializer
+from iaso.models import UserRole, ValidationNodeTemplate
+
+
+class ValidationNodeTemplateBulkUpdateListSerializer(serializers.ListSerializer):
```
tdethier, line 10: When is this class called? When you bulk update template nodes with more than one value?

iaso/api/validation_workflows_node_templates/serializers/bulk_update.py
```diff
@@ -0,0 +1,117 @@
...
+
+        for data in validated_data:
+            inst = instances[data["slug"]]
+            for field, value in data.items():
+                if field not in ["roles_required", "slug"]:
+                    setattr(inst, field, value)
+
+        update_fields = [f.name for f in ValidationNodeTemplate._meta.fields if f.name not in ["id", "slug"]]
+
+        ValidationNodeTemplate.objects.bulk_update(list(instances.values()), update_fields, batch_size=100)
```
tdethier, line 27: What happens when you don't update a field (_e.g._  `color`) and it's passed here to `bulk_update` in `update_fields`?

iaso/api/validation_workflows_node_templates/serializers/bulk_update.py
```diff
@@ -0,0 +1,117 @@
...
+
+        ordered_instances = [instances[data["slug"]] for data in validated_data]
+
+        through = ValidationNodeTemplate.next_node_templates.through
+
+        # clear current
```
tdethier, line 33: Do you need to redo all objects every time you bulk update? If you just change colors or names, I guess you don't need to do that, right? 

Or maybe you call this endpoint and serializer only if you want to reorder nodes?

iaso/tests/api/validation_workflows_node_templates/test_views.py
```diff
@@ -0,0 +1,1703 @@
...
+        self.assertHasError(res_data, "rolesRequired", 'Invalid pk "222" - object does not exist.')
+
+        res = self.client.post(
+            reverse(
+                "validation-node-templates-list", kwargs={"parent_lookup_workflow__slug": self.validation_workflow.slug}
+            ),
+            data={
+                "name": "Random name",
+                "previousNodeTemplates": [self.outer_workflow_node.slug],
+                "nextNodeTemplates": [self.outer_workflow_node.slug],
```
tdethier, line 129: I realize now that introducing `djangorestframework-camel-case` means we'll have camelCase in all our backend tests... I hate that 😅

iaso/tests/api/validation_workflows_node_templates/test_views.py
```diff
@@ -0,0 +1,1703 @@
...
+            self.second_node.refresh_from_db()
+            self.assertEqual(
+                list(self.second_node.previous_node_templates.values_list("pk", flat=True)), [created_node.pk]
+            )
+            self.assertEqual(
+                list(self.second_node.next_node_templates.values_list("pk", flat=True)), [last_created_node.pk]
+            )
+
+
+class ValidationNodeTemplateAPIBulkCreateTestCase(BaseApiTestCase):
```
tdethier, line 391: I would split this in multiple files, one for each class

iaso/api/validation_workflows_node_templates/views.py
```diff
@@ -0,0 +1,125 @@
...
+        self.perform_create(serializer)
+        headers = self.get_success_headers(serializer.data)
+        return Response(serializer.data, status=status.HTTP_201_CREATED, headers=headers)
+
+    @extend_schema(
+        description="Allow to update an existing list of nodes attached to a workflow (fields and order). Does not allow to delete nodes.",
+        request=ValidationNodeTemplateBulkUpdateSerializer(many=True),
+        responses=ValidationNodeTemplateBulkUpdateSerializer(many=True),
+    )
+    @bulk_create.mapping.put
```
tdethier, line 104: This is a way to say "look at the parameters in `bulk_create`, do the same but it's `PUT` instead"?

iaso/api/validation_workflows_node_templates/views.py
```diff
@@ -0,0 +1,125 @@
...
+
+@extend_schema(tags=["Validation workflow node templates"])
+class ValidationNodeTemplatesView(NestedViewSetMixin, ModelViewSet):
+    lookup_field = "slug"
+    lookup_url_kwarg = "slug"
+    permission_classes = [IsAuthenticated, HasValidationNodeTemplatePermission]
+    model = ValidationNodeTemplate
+    filter_backends = [CamelCaseOrderingFilter]
+    parser_classes = [CamelCaseJSONParser]
+    renderer_classes = [CamelCaseJSONRenderer, CamelCaseBrowsableAPIRenderer]
```
tdethier, line 38: I'm really not a fan of this. I understand that it's easier for frontend development, but our API already lacks uniformity. Adding new endpoints in camelCase is a mistake IMO

iaso/api/validation_workflows_node_templates/views.py
```diff
@@ -0,0 +1,125 @@
...
+            return ValidationNodeTemplateBulkCreateSerializer
+        if self.action == "bulk_update":
+            return ValidationNodeTemplateBulkUpdateSerializer
+        if self.action == "move":
+            return ValidationNodeTemplateMoveSerializer
+        raise NotImplementedError(f"Serializer not implemented for action {self.action}")
+
+    def get_serializer_context(self):
+        ctx = super().get_serializer_context()
+        ctx.update({"workflow_slug": self.kwargs.get("parent_lookup_workflow__slug")})
```
tdethier, line 62: Why do you name this parameter `parent_lookup_workflow__slug` instead of simply `workflow_slug`? Is there some place where you automatically add this as a `Q` filter or something?

Also, why do you search for a workflow based on its `slug` instead of its ID?

iaso/urls.py
```diff
@@ -258,7 +259,12 @@
 router.register(r"mobile/stockledgeritems", StockLedgerItemMobileViewSet, basename="mobilestocklegeritems")
 router.register(r"mobile/stockrulesversions", StockRulesVersionMobileViewSet, basename="mobilestockrulesversions")
 
-router.register(r"validation-workflows", ValidationWorkflowViewSet, basename="validationworkflows")
+router.register(r"validation-workflows", ValidationWorkflowViewSet, basename="validationworkflows").register(
```
tdethier, line 262: We usually use `_` instead of `-` in routes, or stick all words together

iaso/urls.py
```diff
@@ -258,7 +259,12 @@
...
+    r"node-templates",
+    ValidationNodeTemplatesView,
```
tdethier, line 264: I can't see the node templates endpoint when I go to `/api/`, is it because it's nested?

I wanted to check in `/swagger/` but there's no web view anymore, it downloads a `.yaml` file

iaso/tests/api/validation_workflows_node_templates/test_views.py
```diff
@@ -0,0 +1,1703 @@
...
+
+        self.client.force_authenticate(self.john_wick)
+        res = self.client.post(
+            reverse(
+                "validation-node-templates-list", kwargs={"parent_lookup_workflow__slug": self.validation_workflow.slug}
+            )
+        )
+        self.assertJSONResponse(res, status.HTTP_400_BAD_REQUEST)
+
+    def test_validation(self):
```
tdethier, line 90: General comment, it's not just for this endpoint: could you please add some tests where you pass a value for `parent_lookup_workflow__slug` that corresponds to a validation workflow slug of another account?

tdethier (commented):
First of all, kudos for all the nice work 👍 

I did not test this PR, I just read the code. I left some comments here and there, but in general:
- I'm a big fan of the refactor
- I'm really don't like the camelCase stuff
- test coverage looks good, except for testing with slugs from another account
- I would definitely split the huge test file (and maybe move some of these tests to serializer tests, like when you're checking serializer errors or missing fields)
- I found the code was somewhat complex to read, especially some tests and serializers. This probably due to both the business logic of what we want to do, and just code style I guess
- overall, I think this will work and is a good job

I'll let the frontend and mobile integrators weigh in on this

iaso/tests/api/validation_workflows/test_views.py
hugo093: Will try to :)

iaso/api/validation_workflows_node_templates/serializers/bulk_update.py
hugo093: yes, see the related list_serializer_class (from drf doc `For these cases you can modify the class that is used when many=True is passed, by using the list_serializer_class option on the serializer Meta class.`)

iaso/api/validation_workflows_node_templates/serializers/bulk_update.py
hugo093: drf default behavior, it's just not modified and stays to its original state

iaso/api/validation_workflows_node_templates/serializers/bulk_update.py
hugo093: Yes this endpoint is more meant when you want to change the order. Otherwise there's the default update for a single instance

iaso/api/validation_workflows_node_templates/views.py
hugo093: kinda yes, it takes same url-path and name but just changes the method. Only way to make it work in drf (you can ofc change the method behavior)

iaso/api/validation_workflows_node_templates/views.py
hugo093: I would argue there but in the end , it's a matter of "taste". That being said, the API is already inconsistent (mix of camelCase and snake_case) and as our main "client" is react , I think imho this is the way to go so we respect each programming language naming convention. Maybe it should be discussed further more as a plan to move to a "v2" API

iaso/api/validation_workflows_node_templates/views.py
hugo093: See [drf-extension doc about nested routes](https://chibisov.github.io/drf-extensions/docs/#nested-routes). I prefer to use slug as it's more human friendly in URL and it avoid some data leaks

iaso/urls.py
hugo093: try /swagger-ui/ ;) and yes it's because it's nested

PR #2838, opened 2026-03-18
POLIOPM-531: add preparedness scores endpoint

quang-le:

DS need to get the preparedness score for a round at a specific date. This requires a dedicated endpoint since the logic is hard to replicate and the cached spreadsheets hard to parse

### Related JIRA tickets

POLIOPM-531

## Changes

- Add `scores` action to preparedness dashboard endpoint:
   - filters cached sheets by url and date
   - raises if a queryparam is missing (to avoid DB overheating)
   - computes scores for the selected sheet 
- Add tests for the whole viewset 

## How to test

test the endpoint with Django Web API

## Print screen / video

N/A


plugins/polio/api/dashboards/preparedness_dashboard.py
```diff
@@ -1,8 +1,61 @@
...
+from iaso.api.common import HasPermission
 from plugins.polio.models import Campaign, CampaignType
-from plugins.polio.preparedness.summary import get_or_set_preparedness_cache_for_round
+from plugins.polio.models.base import Round, SpreadSheetImport
+from plugins.polio.permissions import POLIO_CONFIG_PERMISSION, POLIO_PERMISSION
+from plugins.polio.preparedness.parser import get_preparedness
+from plugins.polio.preparedness.summary import get_or_set_preparedness_cache_for_round, preparedness_summary
+
+
+class PreparednessScoreFilter(django_filters.rest_framework.FilterSet):
```
hugo093: if possible, put that in a separate file `filters.py`

plugins/polio/api/dashboards/preparedness_dashboard.py
```diff
@@ -19,3 +72,21 @@ def list(self, request):
...
+
+    @action(
+        detail=False,
+        methods=["get"],
+        permission_classes=[HasPermission(POLIO_PERMISSION, POLIO_CONFIG_PERMISSION)],
+    )
+    def score(self, request, **kwargs):
+        missing_params = [p for p in ("url", "date") if p not in request.query_params]
+        if missing_params:
+            raise ValidationError({p: f"Missing mandatory filter query parameter: '{p}'" for p in missing_params})
```
hugo093: use a serializer for that ?

plugins/polio/api/dashboards/preparedness_dashboard.py
```diff
@@ -1,8 +1,61 @@
...
 from plugins.polio.models import Campaign, CampaignType
-from plugins.polio.preparedness.summary import get_or_set_preparedness_cache_for_round
+from plugins.polio.models.base import Round, SpreadSheetImport
+from plugins.polio.permissions import POLIO_CONFIG_PERMISSION, POLIO_PERMISSION
+from plugins.polio.preparedness.parser import get_preparedness
+from plugins.polio.preparedness.summary import get_or_set_preparedness_cache_for_round, preparedness_summary
+
+
+class PreparednessScoreFilter(django_filters.rest_framework.FilterSet):
+    url = django_filters.CharFilter(method="filter_url")  # campaign + round number or spreadsheet url?
```
hugo093: you could do it another way: 

```
class Meta: 
    model = SpreadSheetImport
    fields = ["url"]
```

plugins/polio/api/dashboards/preparedness_dashboard.py
```diff
@@ -1,8 +1,61 @@
...
+        If we pick a later date, we risk showing cache data that meets the objectives
+        """
+
+        queryset = queryset.filter(created_at__lte=value)
+        if queryset.exists():
+            return queryset.order_by("-created_at")[:1]
+        return queryset.none()
+
+
+class PreparednessScoreSerializer(serializers.Serializer):
```
hugo093: if possible, put that in a separate `serializers.py` file

plugins/polio/api/dashboards/preparedness_dashboard.py
```diff
@@ -1,8 +1,61 @@
...
+from plugins.polio.models.base import Round, SpreadSheetImport
+from plugins.polio.permissions import POLIO_CONFIG_PERMISSION, POLIO_PERMISSION
+from plugins.polio.preparedness.parser import get_preparedness
+from plugins.polio.preparedness.summary import get_or_set_preparedness_cache_for_round, preparedness_summary
+
+
+class PreparednessScoreFilter(django_filters.rest_framework.FilterSet):
+    url = django_filters.CharFilter(method="filter_url")  # campaign + round number or spreadsheet url?
+    date = django_filters.DateFilter(
+        method="filter_date", lookup_expr="lte", input_formats=settings.API_DATE_INPUT_FORMATS
```
hugo093: lookup_expr is ignored there and could be removed

plugins/polio/api/dashboards/preparedness_dashboard.py
```diff
@@ -1,8 +1,61 @@
...
+
+    def filter_date(self, queryset, name, value):
+        """This filter returns a quersyet with a single object in it.
+        Ideally we cached data on the reference date. But if for some reason the snapshot failed, we want the closest cache entry BEFORE the date.
+        That's because it's used to see the score of campaigns that missed the preparedness objectives on the reference date.
+        If we pick a later date, we risk showing cache data that meets the objectives
+        """
+
+        queryset = queryset.filter(created_at__lte=value)
+        if queryset.exists():
```
hugo093: do we really need this extra query ? If it doesn't exists [:1] would return an empty "list" anyway

plugins/polio/api/dashboards/preparedness_dashboard.py
```diff
@@ -19,3 +72,21 @@ def list(self, request):
...
+    @action(
+        detail=False,
+        methods=["get"],
+        permission_classes=[HasPermission(POLIO_PERMISSION, POLIO_CONFIG_PERMISSION)],
+    )
+    def score(self, request, **kwargs):
+        missing_params = [p for p in ("url", "date") if p not in request.query_params]
+        if missing_params:
+            raise ValidationError({p: f"Missing mandatory filter query parameter: '{p}'" for p in missing_params})
+        queryset = SpreadSheetImport.objects.all()
```
hugo093: we could reuse what DRF does with list action and just override the get_queryset(), get_serializer_class() method per action 

```python
def get_queryset(self):
      if self.action == "score":
           return SpreadSheetImport.objects.all

def get_serializer_class(self):
      if self.action == "score":
           return PreparednessScoreSerializer

@action(filter_backends=[PreparednessScoreFilter])
def score(....):
     params_serializer = ParamSerializer(data=request.data)
      
     serializer = self.get_serializer(queryset.first())
```

plugins/polio/api/dashboards/preparedness_dashboard.py
```diff
@@ -1,8 +1,61 @@
...
+            return queryset.order_by("-created_at")[:1]
+        return queryset.none()
+
+
+class PreparednessScoreSerializer(serializers.Serializer):
+    campaign_details = serializers.SerializerMethodField()
+    scores = serializers.SerializerMethodField()
+
+    def get_campaign_details(self, obj: SpreadSheetImport):
+        round_qs = Round.objects.filter(preparedness_spreadsheet_url=obj.url).select_related("campaign")
```
hugo093: I know that right now this serializer is used for one instance only, but it  might trigger N+1 query problem later if we use it for multiple items. Try to tweak the prefetch_related in the initial queryset and fetch the attr from there?

plugins/polio/api/dashboards/preparedness_dashboard.py
```diff
@@ -1,8 +1,61 @@
...
+        if round_qs.count() > 1:
+            rounds_list = list(round_qs.values_list("id", flat=True))
+            raise Exception(f"Found more than one round for url:{rounds_list}")
+        if not round_qs.exists():
+            return {}
+        round = round_qs.first()
+        return {"round_id": round.id, "round_number": round.number, "campaign": round.campaign.obr_name}
```
hugo093: We could do as well Round.objects.filter().select_related().only('id', 'number', 'campaign').all() and from there check python wise the list len and so on, without triggering exists(), first(), count()

plugins/polio/api/dashboards/preparedness_dashboard.py
```diff
@@ -1,8 +1,61 @@
...
+
+class PreparednessScoreSerializer(serializers.Serializer):
+    campaign_details = serializers.SerializerMethodField()
+    scores = serializers.SerializerMethodField()
+
+    def get_campaign_details(self, obj: SpreadSheetImport):
+        round_qs = Round.objects.filter(preparedness_spreadsheet_url=obj.url).select_related("campaign")
+        if round_qs.count() > 1:
+            rounds_list = list(round_qs.values_list("id", flat=True))
+            raise Exception(f"Found more than one round for url:{rounds_list}")
```
hugo093: Maybe it's not possible, but imho it should be in the validate process and raise a ValidationError

plugins/polio/tests/api/preparedness_dashboard/test_views.py
```diff
@@ -0,0 +1,303 @@
...
+    "items": PREPAREDNESS_LIST_ITEM_SCHEMA,
+}
+
+
+class PreparednessDashboardListAPITestCase(PreparednessDashboardAPIBase):
+    @mock.patch("plugins.polio.api.dashboards.preparedness_dashboard.get_or_set_preparedness_cache_for_round")
+    def test_list_returns_preparedness_for_polio_campaigns(self, mock_get_cache):
+        mock_get_cache.return_value = {"campaign_obr_name": "test-campaign", "round": "Round1", "score": 80}
+
+        self.client.force_authenticate(self.user_polio)
```
hugo093: repeated in every test, put it in setUp

hugo093 (changes requested):
Good quick job, just a few changes to make it cleaner and more efficient imho

plugins/polio/api/dashboards/preparedness_dashboard.py
quang-le: I opted for a solution where I fetch the round for the single `SpreadsheetImport` and annotate the object with the `Round` fields  I return, so it's tricky to move this to the serializer's validation

hugo093 (approved):
LGTM

PR #2839, opened 2026-03-19
IA-4887: Entity search improvements

Phil-V:

White spaces in the search queries and in the json data itself can make some entities hard to find. This PR seeks to provide some fixes and workarounds.

### Related JIRA tickets

IA-4887 (partially addresses IA-4566)

## Changes

- Trim whitespace on json logic queries inputs
- Change the default entity search behavior to search independently of word order or white spaces.
- Support searching phrases in double quotes for exact search.

## How to test

Go to the entities list page and test:
- Search in submitted fields: pick a form and field and include some extra white spaces into your query, you should still get results. 
- Regular search: Try a mix of single word, multiple words and quoted search queries. The expected behavior should be: 
  - quoted searches are exact (case-insensitive) matches. 
  - multiple words are AND-ed together and should match values in the entity attributes json in different orders. 

## Print screen / video

/

## Notes

What is not fixed by this PR is:
- If the json data itself contains white spaces, the entries will still not be found when using the "search in submitted fields" feature. However, as a workaround, if they're on the entity attributes form, they can still be found with the search field. 

## Doc

/

### DIFF
--- a/iaso/api/entities/filters.py
+++ b/iaso/api/entities/filters.py
@@ -61,9 +61,14 @@ class EntityFilterSet(FilterSet):
                 raise ValidationError(f"Failed parsing uuids in search '{value}'")
             return queryset.filter(uuid__in=uuids)
 
-        return queryset.filter(
-            Q(name__icontains=value) | Q(uuid__icontains=value) | Q(attributes__json__icontains=value)
-        )
+        tokens = re.findall(r'"[^"]*"|\S+', value)
+        tokens = [t.strip('"') for t in tokens]
+        q = Q()
+        for token in tokens:
+            q &= Q(name__icontains=token) | Q(attributes__json__icontains=token)
+        q |= Q(uuid__icontains=value)
+
+        return queryset.filter(q)
 
 
 class EntityDateFilterBackend(filters.BaseFilterBackend):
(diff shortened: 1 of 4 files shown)


iaso/api/entities/filters.py
```diff
@@ -61,9 +61,14 @@ def filter_search(self, queryset, name, value):
                 raise ValidationError(f"Failed parsing uuids in search '{value}'")
             return queryset.filter(uuid__in=uuids)
 
-        return queryset.filter(
-            Q(name__icontains=value) | Q(uuid__icontains=value) | Q(attributes__json__icontains=value)
-        )
+        tokens = re.findall(r'"[^"]*"|\S+', value)
```
hugo093: could it be an issue that we could input single quotes and/or escaped double quotes ? Moreover, as discussed, the front end field do not authorize double quotes, so it won't work

hugo093 (changes requested):
Just need to fix the fact that there s a mismatch between front end and back end logic :)

iaso/api/entities/filters.py
Phil-V: Woops! thanks for catching that. I was hoping to retain the possibility to do an exact search but maybe it's not that necessary (as we have the field search logic for that). Removed that logic.

hugo093 (approved):

PR #2842, opened 2026-03-19
IA-4889: Add change password button to edit user dialog

beygorghor:

We had some complaints of users that couldn’t find the pasword input anymore.

We can add the change password button on the dialog of a profile while editing it

### Related JIRA tickets

IA-4889

## Changes

Adding change password button to edit user dialog

## How to test

Make sure you can change the password of an existing user inside the dialog.
Make sure you can still create a user and specify a password

## Print screen / video

Screenshot 2026-03-19 at 12 00 54

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/users/components/UsersInfos.tsx
+++ b/hat/assets/js/apps/Iaso/domains/users/components/UsersInfos.tsx
@@ -1,14 +1,16 @@
-import React, { FunctionComponent, useCallback, useMemo } from 'react';
+import React, { FunctionComponent, useCallback } from 'react';
 import { Alert, Box, Grid } from '@mui/material';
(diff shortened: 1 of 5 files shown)


hat/assets/js/apps/Iaso/domains/users/components/UsersInfos.tsx
```diff
@@ -82,6 +88,38 @@ export const UsersInfos: FunctionComponent<Props> = ({
     );
 
     const appLocales = useAppLocales();
+    const extraFields = useMemo(() => {
```
quang-le: This should be a separate component, not a Memo

quang-le (changes requested):
just move the extra fields out

quang-le (approved):

PR #2845, opened 2026-03-23
IA-4902: Add mobile feature flag for validation workflow

bmonjoie:

The mobile needs a feature flag to know what calls to make and what screens to show.

### Related JIRA tickets

IA-4902

## Changes

Added a migration to add the feature flag.

## How to test

1. Apply the migration
2. Go to the project's configuration
3. Check for "Mobile: Enable validation workflow" in "Data Validation"


quang-le (changes requested):
We should either add the FF to the front-end, or create a ticket to do so

quang-le (approved):

PR #2851, opened 2026-03-25
[IA-4861]: Proposal smoke healthcheck test

hugo093:

Each release we would manually check and test some env health endpoints instead of doing it automatically

### Related JIRA tickets

IA-4861

## Changes

* Created a playwright test to run automatically on mutliple outside env and check that the health is ok : 
  * Status OK
  * Correct git tag
  
## How to test

If you don't have one yet, create an .env.smoke file with following content 

LOGIN_PASSWORD=whatever
LOGIN_USERNAME=whatever
GIT_TAG=v2026.03.24--main

Run then the following command to test on all envs : 

`HEALTH_ENV=all npm run test:smoke:healthcheck` 

## Notes

Didn't adapt yet the github actions but it could be a post release action

### DIFF
--- a/hat/assets/js/__tests__/playwright/smoke/healthcheck.test.ts
+++ b/hat/assets/js/__tests__/playwright/smoke/healthcheck.test.ts
@@ -1,0 +1,36 @@
+import { test, expect } from '@playwright/test';
+
+const ENV = {
+    prod: 'https://iaso.bluesquare.org',
+    pathways: 'https://collect.projectpathways.org/',
+    playground: 'https://iaso-playground.bluesquare.org',
+    campaigns: 'https://www.poliooutbreaks.com/',
+    'iaso-demo': 'https://demo.openiaso.com',
+};
+
+let envToCheck = process.env?.HEALTH_ENV?.split(',') || [];
+
+if (envToCheck.some(env => env === 'all')) {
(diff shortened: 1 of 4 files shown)


hat/assets/js/__tests__/playwright/smoke/healthcheck.test.ts
```diff
@@ -0,0 +1,36 @@
+import { test, expect } from '@playwright/test';
+
+const ENV = {
+    prod: 'https://iaso.bluesquare.org',
+    pathways: 'https://collect.projectpathways.org/',
+    playground: 'https://iaso-playground.bluesquare.org',
+    campaigns: 'https://www.poliooutbreaks.com/',
+    'iaso-demo': 'https://demo.openiaso.com',
+};
```
beygorghor, line 9: Maybe we can add those as repository variable accessible in the github action ?

beygorghor (approved):
We can probably also add some test to test point 5 on the[ iaso-release repo](https://github.com/BLSQ/iaso-release) (testing polio dashboard on prod then)

hat/assets/js/__tests__/playwright/smoke/healthcheck.test.ts
hugo093: I'll add a ticket with those improvements

PR #2853, opened 2026-03-25
[IA-4898]: make form_set not required , adapt test

hugo093:

In validation workflow, form set was required, which is not convenient for front end and business logic 

### Related JIRA tickets

IA-4898

## Changes

Model and API endpoints were already adapted in a previous branch. 
Adapted the function `is_artifact_allowed` so it doesn't trigger the workflow if no form has been provided. 
Adapted tests

## How to test

Run python tests


tdethier (approved):
LGTM 👍

PR #2858, opened 2026-03-26
[IA-4878]: implement post create to trigger validation workflow, added tests

hugo093:

When creating an instance, we need to trigger a validation workflow if needed

### Related JIRA tickets

IA-4878

## Changes

Added a "post create" trigger to start a validation workflow after instance has been created

Tests

## How to test

Create a validation Workflow with some Nodes
Create a Form, link it to that validation workflow
Create an instance (linked to the form), it should trigger a validation process


tdethier (approved):
LGTM 👍 

Maybe this should be done too on instance bulk upload through zip

PR #2859, opened 2026-03-26
IA-4913: N+1 on /api/mobile/orgunits/changes/

beygorghor:

4 k events in sentry on this api 

### Related JIRA tickets

IA-4913

## Changes

Some api optimisation

## How to test

Call /api/mobile/orgunits/changes/ ?app_id=SOME_APP_ID where the app id is the app of a project with change requests.
Inspect the terminal, you should have 0 duplicates here

## Print screen / video

Screenshot 2026-03-26 at 11 44 32

### DIFF
--- a/iaso/api/org_unit_change_requests/views_mobile.py
+++ b/iaso/api/org_unit_change_requests/views_mobile.py
@@ -1,5 +1,6 @@
 import django_filters
 
+from django.contrib.auth import get_user_model
 from django.db.models import Count, F, Prefetch, Q
 from rest_framework import filters, viewsets
 from rest_framework.mixins import ListModelMixin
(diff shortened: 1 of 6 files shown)


iaso/models/project.py
```diff
@@ -25,14 +25,14 @@ def get_for_user_and_app_id(
 
         if app_id is not None:
             try:
-                project = self.get(app_id=app_id)
+                project = self.select_related("account", "account__default_version").get(app_id=app_id)
```
bmonjoie, line 28: I believe this `select_related` should be added where you use `get_for_user_and_app_id` and not here, otherwise it will be selected even when it's not needed.

bmonjoie (commented):
Nice job!

Just a little change not to impact  other queries.

iaso/models/project.py
beygorghor: i can, it will impact 15 more files

bmonjoie: OK then, leave it there but maybe add a comment explaining that it's needed ;-)

bmonjoie (approved):

iaso/api/org_unit_change_requests/views_mobile.py
```diff
@@ -19,17 +20,34 @@ class MobileOrgUnitChangeRequestViewSet(ListModelMixin, viewsets.GenericViewSet)
...
+        self._cached_user_for_scoped_queries = User.objects.select_related(
+            "iaso_profile__account",
+            "iaso_profile__account__default_version",
+        ).get(pk=self.request.user.pk)
+        return self._cached_user_for_scoped_queries
+
     def get_queryset(self):
+        cached_qs = getattr(self, "_cached_mobile_change_requests_queryset", None)
+        if cached_qs is not None:
+            return cached_qs
```
tdethier: I'm wondering if all the caching stuff is necessary. Have you already tested without?

I'm asking this because I'm a bit scared that you'll keep expired data in the cache (_e.g._ what happens if you cache a user by calling this endpoint, then you change this user's permissions? or if your CR queryset is also updated in the meantime?

Also, I'm not sure how this would behaves when multiple users from multiple accounts try to access this endpoint. What gets cached?

iaso/tests/api/org_unit_change_requests/test_org_unit_change_requests_mobile.py
tdethier, line 93: I'm afraid you'll also need to update this 😅

tdethier (commented):
I tested this on a setuper account, so not with a lot of data. Clearly there's no more repeated queries 👍 


I thought it was a bit slower, so I made some tests, but it's not, unless I add some breakpoint

|  | With the PR | Without the PR |
| --- | ----- | ----- |
| accessing first time on start up| Image | Image |
| refreshing the page a lot (avg) | Image | Image |



I'm just not sure about the caching part. I see that `get_queryset()` is called twice whenever I call this endpoint, and the CR caching is used during the second call (I never see any hit for the User caching). This second call is made because I'm using DRF web view. So, it should always be empty when a standard call is made.

tdethier:
By adding `&format=json` at the end of your call (if you use the DRF web view), you get raw json without the DRF web interface, which does not call `get_queryset()` twice and does not make repeated queries without caching some values. So you could remove all the caching stuff and it will still work fine

PR #2860, opened 2026-03-26
IA-4819: rate limiting on login

Phil-V:

Protect the login endpoints from dictionary attacks by rate-limiting based on username.

### Related JIRA tickets

IA-4819

## Changes

- Integrate `django-axes` to rate-limit for the login form and login api endpoint. 
- Add settings that can be configured with environment variables.

## How to test

- Try to login more than 10 times with a wrong password:
    - on the login page on the dashboard
    - or using the api (`POST /api/token/`, present in our Bruno config)
- Check that you can login again after a cooldown period (30 seconds)

## Print screen / video

/ 

## Notes

- See the specs document in the Jira comments for background information on architectural choices.
- I haven't tested the `wpf-auth` plugin manually, trusting that it shouldn't be affected by these changes.

## Doc

The configuration is documented in the settings file.


requirements.txt
```diff
@@ -12,6 +12,7 @@ django-translated-fields==0.12.0  # https://github.com/matthiask/django-translat
 django-phonenumber-field[phonenumberslite]==7.3.0 # https://django-phonenumber-field.readthedocs.io/en/latest/
 django-qr-code==4.0.1 # https://django-qr-code.readthedocs.io/en/latest/
 django-stubs-ext==4.2.7  # https://github.com/typeddjango/django-stubs/tags
+django-axes==8.0.0 # https://github.com/jazzband/django-axes/tags
```
tdethier, line 15: I'm a bit worried about `django-axes==8.0.0` not officially supporting python 3.9, but if it works just long enough for us to complete the upgrade, I guess it should be fine 🤷‍♂️

tdethier: I just saw that they still have 3.9 in their `setup.py` 😅 

I guess it won't stay for long

iaso/tests/auth/test_rate_limting.py
```diff
@@ -0,0 +1,71 @@
...
+
+        # different username should not be locked out
+        payload_other_user = {"username": "innocent_user", "password": "azerty1234"}
+        resp_diff_user = self.client.post(self.api_url, payload_other_user, REMOTE_ADDR="192.168.1.100")
+        self.assertEqual(resp_diff_user.status_code, status.HTTP_401_UNAUTHORIZED)
+
+    def test_login_rate_limit(self):
+        """Test that /login/ has rate limiting."""
+        self.client.post(self.login_url, self.login_payload, format="multipart")
+        self.client.post(self.login_url, self.login_payload, format="multipart")
```
tdethier: Why do you use `format="multipart"` here?

iaso/tests/auth/test_rate_limting.py
```diff
@@ -0,0 +1,71 @@
...
+
+        resp3 = self.client.post(self.login_url, self.login_payload, format="multipart")
+        self.assertEqual(resp3.status_code, status.HTTP_200_OK)
```
tdethier: Shouldn't it be a `HTTP_429_TOO_MANY_REQUESTS` here as well?

iaso/tests/auth/test_rate_limting.py
```diff
@@ -0,0 +1,71 @@
...
+    def test_admin_login_rate_limit(self):
+        """Test that /admin/login/ has rate limiting"""
+        resp1 = self.client.post(self.admin_login_url, self.login_payload, format="multipart")
+        self.assertEqual(resp1.status_code, status.HTTP_200_OK)
+
+        resp2 = self.client.post(self.admin_login_url, self.login_payload, format="multipart")
+        self.assertEqual(resp2.status_code, status.HTTP_200_OK)
+
+        resp3 = self.client.post(self.admin_login_url, self.login_payload, format="multipart")
+        self.assertEqual(resp3.status_code, status.HTTP_200_OK)
```
tdethier: Same question about `HTTP_429_TOO_MANY_REQUESTS`

hat/settings.py
```diff
@@ -223,6 +223,7 @@
     "drf_spectacular_sidecar",
     "django_json_widget",
     "phonenumber_field",
+    "axes",
```
tdethier, line 226: I see this app added 3 new tables to my database:
- `axes_accessattempt`
- `axes_accessfailurelog`
- `axes_accesslog`

I played a bit with the feature and afterwards all tables were empty. I was expecting to see some stuff in there, especially when my account was blocked. I'm wondering if it's just a configuration issue or standard behavior. Out of curiosity, did you check these tables locally? Do you have some data?

tdethier (approved):
I tested the feature and it works fine - I was blocked for 30 seconds after trying the wrong password 10 times 👍 

Kudos for the clean code and explanations.

I left a few comments and questions, but nice job!

iaso/tests/auth/test_rate_limting.py
Phil-V: Hi @tdethier thanks for the review!

With the DRF test client, the only 2 options are 'json' or 'multipart'. But I'll change the test to use the regular django client with the default `application/x-www-form-urlencoded` so that it's cleaner.

iaso/tests/auth/test_rate_limting.py
Phil-V: I don't think there's a way for ValidationErrors to alter the status code in vanilla Django forms, and it might be Django default behavior for "old fashioned" forms. (I don't have a vanilla Django project on hand to verify this quickly). 
The existing behavior of our login form is that the response is always 200 OK and the error is displayed to the user cleanly in the html response.

hat/settings.py
Phil-V: It's a bit unfortunate that we have these tables as we won't make use of them at all. I've setup django-axes to use its cache-based framework, so it uses the django cache backend. Locally you can see the django-axes keys appearing with `select cache_key from django_cache_table;`, in production this will be with redis.

Phil-V:
- Updated some of the tests to use the django test client instead of the DRF one where it made sense.
- Fixed typo in the test file name.

hat/settings.py
tdethier: Alright I do see them, thanks!

PR #2869, opened 2026-03-30
[IA-4926] Add `app_id` and `app_version` to `APIImport`

bmonjoie:

Allows to filter `APIImport` by `app_id`.

### Related JIRA tickets

IA-4926

## Changes

Retrieve the `app_id` and `app_version` from the query string and add them to the `APIImport`'s fields.

## How to test

Send submissions to the backend via the mobile.
Check that APIImport does have `app_id` and `app_version` filled.


beygorghor (approved):

hat/api_import/models.py
```diff
@@ -50,6 +50,8 @@ class Meta:
     has_problem = models.BooleanField(default=False)
     exception = models.TextField(blank=True, default="")
     file = models.FileField(upload_to=api_import_upload_to, null=True, blank=True)
+    app_id = models.TextField(blank=True, null=True)
+    app_version = models.CharField(max_length=25, blank=True, null=True)
```
tdethier:
```suggestion
    app_id = models.TextField(blank=True)
    app_version = models.CharField(max_length=25, blank=True)
```

iaso/api/common/utils.py
```diff
@@ -29,6 +29,8 @@ def inner(self, request, *args, **kwargs):
             if not request.user.is_anonymous:
                 api_import.user = request.user
             api_import.import_type = key
+            api_import.app_id = request.GET.get("app_id", None)
+            api_import.app_version = request.GET.get("app_version", None)
```
tdethier:
```suggestion
            api_import.app_id = request.GET.get("app_id", "")
            api_import.app_version = request.GET.get("app_version", "")
```

hat/api_import/models.py
tdethier: But that will require you to redo the migration file 😅

hat/api_import/migrations/0007_apiimport_app_id_apiimport_app_version.py
```diff
@@ -0,0 +1,22 @@
+# Generated by Django 4.2.29 on 2026-03-30 10:40
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("api_import", "0006_alter_apiimport_file"),
```
tdethier, line 8: Depending on how fast the migration is, maybe we'll need to run it manually, since `APIImport` is a big table on some servers. We'll need to pay attention to this during the next release

tdethier (dismissed):
Just requesting changes to avoid both `null=True` and `blank=True` on char/text fields, [as recommended by Django](https://docs.djangoproject.com/en/4.2/ref/models/fields/#django.db.models.Field.null). Otherwise it looks good!

bmonjoie:
@tdethier As discussed, I know if it discouraged by DRF to have both because they say that it's basically a way to have two different values for "no values" but I put both on purpose because I never want those values to fail, if you put `None` or `""`, it should never fail. 

I have had the issue in the past that putting `null=False, blank=True` would cause issues when passing `None` to the field. I understand the logic behind's Django's recommendation but I believe it doesn't apply here as we are just trying to store the value as it is, no matter what it is.

hat/api_import/models.py
hugo093: + should we make app_id a textfield ? is it that big ? can't it be a charfield

bmonjoie: @hugo093 I took the same field as the one in `Project` because it's the same information.

hugo093: I understand but well, i'd be curious to see some stats about that field on real DB and check if Textfield is really necessary ... LGTM however

hugo093 (changes requested):
I kind of agree with @tdethier about the null=True, it's a bad practice in django, so better avoid it now. 
While I can live with the TextField because legacy etc. that part I'll request for a proper fix :) FYI on why it's bad : https://docs.astral.sh/ruff/rules/django-nullable-model-string-field/  https://docs.djangoproject.com/en/5.0/ref/models/fields/#django.db.models.Field.null

tdethier (approved):
I agree with @hugo093 it's overkill to have a `TextField` for `app_id`, but since it's already like that on `Project` 🤷‍♂️ We should probably switch both to a `CharField`, but in another PR

LGTM, thanks for making the changes 👍

hugo093 (approved):
LGTM, thanks for the changes

hat/api_import/migrations/0007_apiimport_app_id_apiimport_app_version.py
bmonjoie: It's just adding a column, that should not impact the database. At first, I wanted to extract the information for the existing ones from the `QUERYSTRING` meta data but I thought that it was probably overkill

PR #2870, opened 2026-03-30
IA-4928 Only use async select user

beygorghor:
Fetching all user in a dropdown can lead to performance issues in prod, we should use async select everywhere

### Related JIRA tickets

IA-4928

## Changes

- Fix typings issues in entities
- replace entities filter 
- replace algorithm runs filter
- replace links filter (settings/sources/links/runs/)
- replace duplicate filter
- replace org unit details links filter
- replace create edit team dropdown

## How to test

You should visit all those pages and try to select a user with the dropdown. Check thaht everything is ok 

## Print screen / video

Screenshot 2026-03-30 at 15 42 59
Screenshot 2026-03-30 at 15 42 47
Screenshot 2026-03-30 at 15 42 26


hugo093 (changes requested):
Tested and LGTM, but we really should write integration tests for all those pages : 
* Creating factories for profile API
* Create Integration tests (even though if it's juts for this field / filters right now)

hugo093 (approved):
LGTM, thanks for the tests

PR #2871, opened 2026-03-30
IA-4910: Default month in submissions date filter prevents search

beygorghor:
[239a7a95-17d9-4b75-b180-c7959300711a.webm](https://github.com/user-attachments/assets/e62c7458-5a2d-4ff7-9819-9736b2dbd6fb)

### Related JIRA tickets

IA-4910

## Changes

Updating local state by default was not enough, we have to trigger onChange too to actualise value

## How to test

Open submission pages, try to search by monthly period, select a year, it should not be red,d ecember should be selected by default.
[
](https://github.com/user-attachments/assets/53b8f649-8ae0-4eaf-9f76-1b7065510b8d)

On the same page, filter by a form with a monthly period, try to create a submission, period picker should work as expected
Screenshot 2026-03-30 at 16 03 36

Go to completeness stats, select a form with a period, try to filter per period.
Screenshot 2026-03-30 at 15 57 31

On registry page, select a country, go to the bottom and try to filter submissions per period too

Screenshot 2026-03-30 at 15 59 00

Create a new version of a form with a month period type, period picker should work (here it makes cense not to pre-fill the field)

Screenshot 2026-03-30 at 16 00 53

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/periods/components/PeriodPicker.test.tsx
+++ b/hat/assets/js/apps/Iaso/domains/periods/components/PeriodPicker.test.tsx
@@ -1,0 +1,364 @@
+import React from 'react';
+
+import { screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { IntlProvider } from 'react-intl';
+import { renderWithTheme } from '../../../../../tests/helpers';
+
+import {
+    NO_PERIOD,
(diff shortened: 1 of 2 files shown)


hugo093 (changes requested):
LGTM but please write tests (at least basic unit tests) for this component

hugo093 (approved):
LGTM, tx for the tests

PR #2877, opened 2026-04-01
[IA-4927]: fix backend integrity errors raised on API

hugo093:

Integrity errors were raised while creating/iupdating validation workflow and node templates
 
### Related JIRA tickets

IA-4927

## Changes

* Adapted DB constraints on workflow template to include deleted_at
* Adapted serializers
* Added tests

## How to test

See tests written but : 
* Call the validation workflow API POST/PUT/PATCH with an already existing name in body payload. It should raise a 400 and not a 500. 
* Call the validation workflow API POST/PUT/PATCH with an already existing name that belongs to a deleted VF in body payload. It should go through
* Call the validation node templates API calls (bulk update, bulk create, update, create) and check that when using an existing name, it raises a 400 and not a 500.


tdethier (approved):
LGTM, thanks for all the tests 👍

PR #2879, opened 2026-04-01
[IA-4784]: improve json_schema_field + tests

hugo093:

Improving JsonSchemaField for DRF

### Related JIRA tickets

IA-4784

## Changes

Moved logic to a validator
Added some checks (not null, not blank)
Added swagger proper typing
Added tests

## How to test

Run python tests


iaso/api/metrics/serializers.py
```diff
@@ -3,11 +3,11 @@
 from django.utils.translation import gettext_lazy as _
 from rest_framework import serializers
 
+from iaso.api.common.serializer_fields import JSONSchemaField
```
tdethier, line 6: I'll need to make a PR in SNT to fix this as well then

iaso/api/common/serializer_fields.py
```diff
@@ -45,3 +47,14 @@ def run_validation(self, data=serializers.empty):
...
+
+
+@extend_schema_field(OpenApiTypes.OBJECT)
+class JSONSchemaField(serializers.JSONField):
+    def __init__(self, schema, *args, **kwargs):
+        # for swagger
+        kwargs.setdefault("help_text", f"JSON object matching schema: {schema}")
+        super().__init__(*args, **kwargs)
+        self.allow_null = False
+        self.allow_blank = False
```
tdethier: I think we should fetch that from parameters rather than enforcing it. In SNT, we use a `JSONSchemaField` that could receive a null value

tdethier (commented):
This looks good, I'm just not sure about the `allow_null` and `allow_blank` part. I'll make a PR to fix imports in SNT in the meantime

iaso/api/common/serializer_fields.py
hugo093: fine by me, I can maybe make a kwargs.setdefault instead ?

tdethier (approved):
LGTM 👍

PR #2882, opened 2026-04-02
IA-4905: remove double click search org unit

beygorghor:
Now that api is using fields, we can remove the two step search.

Also map icon is not displayed anymore in the org units list, probably due tu has_geo_json / lat, long field missing
### Related JIRA tickets

IA-4905

## Changes

- remove searchactive param
- fix missing map icon on org unit row

## How to test

Visit org units page, it should search directly, org unit wiht a shape or a location should have a map icon on the row

## Print screen / video

Screenshot 2026-04-02 at 10 25 51

https://github.com/user-attachments/assets/f57156cf-e880-4daf-a7b9-a3456b7f8a3d


hat/assets/js/apps/Iaso/domains/orgUnits/hooks/requests/useGetOrgUnits.ts
```diff
@@ -104,7 +96,7 @@ export const useGetOrgUnitsLocations = ({
             enabled,
             staleTime: Infinity,
             select: data => {
-                if (isSearchActive) {
+                if (data) {
```
Bewi, line 99: Meh, not from you, but for short stuff like this, I always prefer :
`data => data && mapOrgUnitByLocation(data, searchers)`

Bewi (approved):
WaE & LGTM

PR #2883, opened 2026-04-02
[IA-4899]: backend handle resubmission mechanism

hugo093:

Handling the resubmission mechanism for validation workflow

### Related JIRA tickets

IA-4899

## Changes

* Implemented a mechanism that creates a fake validation node, attached to the starting node template, that would contain either status SUBMISSION or NEW_VERSION
* Re-submission will trigger a new validation workflow only if the previous one was pending or rejected. 
* Adapted engine, models, API and tests
* Front-end API endpoints will not receive the new_version and submission validation nodes, they have been filtered out. 

## How to test

Run python tests
Try to resubmit an instance, if there was no validation workflow and/or the previous one was PENDING or REJECTED. It should trigger a new one. 
Otherwise, nothing will happen.


iaso/api/validation_workflow_instances/serializers.py
```diff
@@ -61,7 +61,10 @@ class Meta:
     @extend_schema_field(NestedHistorySerializer(many=True))
     def get_history(self, obj):
         return NestedHistorySerializer(
-            obj.get_all_validation_nodes().select_related("created_by", "updated_by", "node"), many=True
+            obj.get_all_validation_nodes()
+            .exclude(status__in=[ValidationNodeStatus.NEW_VERSION, ValidationNodeStatus.SUBMISSION])
```
bmonjoie: Why do you exclude those here?

hugo093: cause the front end don't display those, so no need to change front end code

bmonjoie: The frontend web is also supposed to show those :sweat_smile:

hugo093: Oh, then i need to update the front as well :D

bmonjoie (approved):

PR #2888, opened 2026-04-02
[IA-4942] Add APIImport UI

bmonjoie:

Add a UI for APIImport and fix some issues

### Related JIRA tickets

IA-4942

## Changes

Expose APIImport in an API
Add a new UI to filter through them
Add a notification when there are failed APIImport within the last 30 days.

## How to test

With failed APIImports, go to the IASO frontend web as a super user or staff user. 
You should see the notification badge.
Click on it and you'll be redirected to new APIImport page.


iaso/api/api_import/views.py
```diff
@@ -0,0 +1,42 @@
...
+    page_size = 20
+
+
+class IsAdminOrSuperUserPermission(BasePermission):
+    """
+    Allows access only to admins and superusers.
+    """
+
+    def has_permission(self, request, view):
+        return bool(request.user and (request.user.is_staff or request.user.is_superuser))
```
quang-le: There's already reusable code for that: https://github.com/BLSQ/iaso/blob/374339f700bde9f4afe1200c5a24fdbb05c714bc/iaso/api/common/permissions.py#L56

hat/assets/js/apps/Iaso/domains/apiimports/components/Filters.tsx
```diff
@@ -0,0 +1,133 @@
...
+    const handleChange = useCallback(
+        (key: string, value: any) => {
+            setFiltersUpdated(true);
+            setFilters({
+                ...filters,
+                [key]: value,
+            });
+        },
+        [filters],
+    );
```
quang-le, line 50: Why not `useFilterState`?

entrypoint.sh
```diff
@@ -100,8 +100,8 @@ case "$1" in
     python "${@:2}"
   ;;
   * )
-    if [[ $2 == /opt/.pycharm_helpers/* ]]; then
-      ${@}
+    if [[ $2 == /opt/.pycharm_helpers/* || $2 == /opt/project/manage.py ]]; then
+      "${@}"
```
quang-le, line 104: how is that related to the ticket? 🤔

hat/assets/js/apps/Iaso/domains/apiimports/types/filters.ts
```diff
@@ -0,0 +1,14 @@
...
+    importType?: string;
+    hasProblem?: boolean;
+    appId?: string;
+    appVersion?: string;
+};
+
+export type Params = Filters & {
+    order: string;
+    page: string;
+    pageSize: string;
```
quang-le: These are `PaginationParams`: https://github.com/BLSQ/iaso/blob/374339f700bde9f4afe1200c5a24fdbb05c714bc/hat/assets/js/apps/Iaso/types/general.ts#L56

hat/assets/js/apps/Iaso/domains/apiimports/components/Filters.tsx
```diff
@@ -0,0 +1,133 @@
...
+                    onEnterPressed={handleSearch}
+                    options={[
+                        {
+                            value: 'bulk',
+                            label: 'Bulk Org Units and Instances',
+                        },
+                        { value: 'instance', label: 'Form instance' },
+                        { value: 'orgUnit', label: 'Org Unit' },
+                        { value: 'storageLog', label: 'Storage Logs' },
+                    ]}
```
quang-le: Please move the array to a variable and translate labels (probably requires custom hook w/ useMemo)

hat/assets/js/apps/Iaso/constants/routes.tsx
```diff
@@ -473,6 +474,12 @@ export const pipelineDetailsPath = {
     permissions: [],
     element: <PipelineDetails />,
 };
+export const adminApiImportPath = {
+    baseUrl: baseUrls.adminApiImport,
+    routerUrl: `${baseUrls.adminApiImport}/*`,
+    permissions: [],
+    element: <Apiimports />,
```
quang-le: ApiImports?

hat/assets/js/apps/Iaso/components/nav/NotificationBadge/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+});
+
+const useStyles = makeStyles(styles);
+
+export const NotificationBadge: FunctionComponent = () => {
+    const [open, setOpen] = useState(false);
+    const anchorRef = useRef<HTMLDivElement>(null);
+    const classes = useStyles();
+    const { data, isFetching } = useGetNotifications();
+    const redirect = useRedirectToReplace();
```
quang-le:
```suggestion
    const redirectToReplace = useRedirectToReplace();
```

there's also a `useRedirect` they don't have the same behaviour, so we try to keep matching the var name with the hook for clarity

hat/api_import/management/commands/remove_authorization_headers_from_api_imports.py
```diff
@@ -0,0 +1,73 @@
...
+        batch_size = int(options[BATCH_SIZE_ARG])
+
+        # Use a list to collect objects for bulk writing
+        objs_to_update = []
+        total_modified = 0
+
+        self.stdout.write(self.style.SUCCESS(f"Starting clean-up with a batch size of {batch_size}..."))
+
+        # .iterator() fetches records in chunks to save memory
+        iterator = APIImport.objects.filter(headers__isnull=False).all().iterator(chunk_size=batch_size)
```
quang-le: .all() is redundant

hat/assets/js/apps/Iaso/domains/apiimports/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+                    data={data?.results ?? []}
+                    pages={data?.pages ?? 1}
+                    defaultSorted={[{ id: 'created_at', desc: true }]}
+                    columns={columns}
+                    count={data?.count ?? 0}
+                    baseUrl={baseUrl}
+                    params={params}
+                    onTableParamsChange={p => redirectTo(baseUrl, p)}
+                />
+            </Box>
```
quang-le: You should use `TableWithDeepLink` or `SimpleTableWithDeepLink` to avoid the boilerplate with `onTableParamsChange`, and the null checks on `data`

quang-le (changes requested):
Some minor things in the code like "use existing component for this"

Feature wise, the I don't see the badge in the top bar, though I see the API call is made

entrypoint.sh
bmonjoie: It's not, but I needed it to run IASO in debug in PyCharm; and I was not going to make a PR just for that

hat/assets/js/apps/Iaso/domains/apiimports/components/Filters.tsx
bmonjoie: Because the code I copy/pasted elsewhere was not using that :upside_down_face:

hat/assets/js/apps/Iaso/domains/apiimports/components/Filters.tsx
```diff
@@ -21,48 +18,27 @@ type Props = {
...
+    const { filters, handleSearch, handleChange, filtersUpdated } =
+        useFilterState({ baseUrl: baseUrl, params });
+    const typeOptions = useMemo(
+        () => [
+            { value: 'bulk', label: 'Bulk Org Units and Instances' },
+            { value: 'instance', label: 'Form instance' },
+            { value: 'orgUnit', label: 'Org Unit' },
+            { value: 'storageLog', label: 'Storage Logs' },
+        ],
+        [],
```
quang-le: Boooo! Please add translations, see example: https://github.com/BLSQ/iaso/blob/9f0446be252fa4b4c5ceb817be333df06eb20511/plugins/polio/js/src/domains/LQAS-IM/LQAS/LqasAfroOverview/utils.ts#L9

quang-le (changes requested):
Just needs proper translation for the dropdown list

quang-le (approved):

bmonjoie:
@tdethier @hugo093 I would appreciate it if you could check the command in this PR.

iaso/api/api_import/views.py
```diff
@@ -0,0 +1,33 @@
...
+    permission_classes = [IsAdminOrSuperUser]
+    serializer_class = APIImportSerializer
+    filterset_class = APIImportFilterSet
+    pagination_class = APIImportPagination
+    ordering_fields = ["created_at", "import_type", "user_id", "app_id", "app_version", "has_problem"]
+
+    def get_queryset(self):
+        queryset = APIImport.objects.all()
+        user = self.request.user
+        if not user.is_staff:
```
hugo093, line 25: shoudn't really happen considering your permission class only allows staff and/or superuser users. So this condition is only valid for superusers ?

iaso/api/api_import/views.py
```diff
@@ -0,0 +1,33 @@
...
+
+from hat.api_import.models import APIImport
+from iaso.api.api_import.filters import APIImportFilterSet
+from iaso.api.api_import.serializers import APIImportSerializer
+from iaso.api.common import IsAdminOrSuperUser, Paginator
+from iaso.models import Project
+
+
+class APIImportPagination(Paginator):
+    page_size = 20
```
hugo093: detail but put that in a separate pagination.py file

iaso/api/api_import/serializers.py
```diff
@@ -0,0 +1,34 @@
+from rest_framework import serializers
+
+from hat.api_import.models import APIImport
+from iaso.api.common import TimestampField
+from iaso.api.form_versions.serializers import UserNestedSerializer
+
+
+class APIImportSerializer(serializers.ModelSerializer):
+    user = UserNestedSerializer(read_only=True)
```
hugo093, line 17: i understand the purpose, but we should either : 
Move that UserNestedSerializer to common package so we know it has side effects when modifying it
Recreate in this file the serializer

iaso/api/api_import/serializers.py
```diff
@@ -0,0 +1,34 @@
+from rest_framework import serializers
+
+from hat.api_import.models import APIImport
+from iaso.api.common import TimestampField
+from iaso.api.form_versions.serializers import UserNestedSerializer
+
+
+class APIImportSerializer(serializers.ModelSerializer):
```
hugo093, line 16: can you make sure this is swagger compliant?

iaso/api/api_import/views.py
```diff
@@ -0,0 +1,33 @@
...
+    filterset_class = APIImportFilterSet
+    pagination_class = APIImportPagination
+    ordering_fields = ["created_at", "import_type", "user_id", "app_id", "app_version", "has_problem"]
+
+    def get_queryset(self):
+        queryset = APIImport.objects.all()
+        user = self.request.user
+        if not user.is_staff:
+            queryset = queryset.filter(
+                app_id__in=Project.objects.filter(account=user.iaso_profile.account)
```
hugo093, line 27: based on the permission class, it could be that we have an authenticated user without any iaso_profile nor account, so better check for that first

iaso/api/api_import/views.py
bmonjoie: Yes, this applies to superusers who are not staff.

iaso/api/notifications/views.py
```diff
@@ -0,0 +1,31 @@
...
+from iaso.api.notifications.models import Notification, NotificationLevel, NotificationType
+from iaso.api.notifications.serializers import NotificationSerializer
+
+
+class NotificationViewSet(viewsets.ViewSet):
+    permission_classes = [permissions.IsAuthenticated]
+
+    def list(self, request):
+        notifications = []
+        if request.user.is_staff or request.user.is_superuser:
```
hugo093: shouldn't we put the permissions to IsAdminOrSuperUser then ? or do we really want to return an empty response to unauthenticated users ?

bmonjoie: We want to return an empty response to unauthenticated users because they might have notifications one day.

iaso/api/notifications/views.py
```diff
@@ -0,0 +1,31 @@
...
+                            singular="There is %(count)d failing import needing your attention.",
+                            plural="There are %(count)d failing imports needing your attention.",
+                            number=apiimport_count,
+                        )
+                        % {"count": apiimport_count},
+                        level=NotificationLevel.ERROR,
+                        type=NotificationType.APIIMPORT,
+                    )
+                )
+        serializer = NotificationSerializer(notifications, many=True)
```
hugo093, line 31: you put many=true and use a list but you actually return one Notification result whatever happens ?

iaso/tests/api/api_import/test_views.py
```diff
@@ -0,0 +1,108 @@
+from datetime import datetime
+
+from hat.api_import.models import APIImport
+from iaso.models import Account, Project
+from iaso.test import APITestCase
+
+
+class APIImportViewSetTest(APITestCase):
+    def setUp(self):
```
hugo093, line 14: add super().setUp() just in case

iaso/tests/api/api_import/test_views.py
```diff
@@ -0,0 +1,108 @@
...
+        self.account1 = account1 = Account.objects.create(name="account1")
+        self.account2 = account2 = Account.objects.create(name="account2")
+
+        self.project1 = Project.objects.create(name="project1", app_id="P1", account=account1)
+        self.project2 = Project.objects.create(name="project2", app_id="P2", account=account2)
+
+        self.staff = staff = self.create_user_with_profile(username="staff", account=account1, permissions=[])
+        staff.is_staff = True
```
hugo093: you can put is_staff=True directly in create_user_with_profile

iaso/tests/api/api_import/test_views.py
```diff
@@ -0,0 +1,108 @@
...
+        staff.save()
+        self.superuser = superuser = self.create_user_with_profile(
+            username="superuser", account=account1, permissions=[]
+        )
+        superuser.is_superuser = True
```
hugo093: same than is_staff

iaso/tests/api/api_import/test_views.py
```diff
@@ -0,0 +1,108 @@
...
+        self.client.force_authenticate(self.staff)
+        response = self.client.get("/api/api_import/?import_type=bulk")
+        json_response = self.assertJSONResponse(response, expected_status_code=200)
+        self.assertEqual(json_response["count"], 2)
+
+    def test_get_filter_user_id(self):
+        self.client.force_authenticate(self.staff)
+        response = self.client.get(f"/api/api_import/?user_id={self.user.id}")
+        json_response = self.assertJSONResponse(response, expected_status_code=200)
+        self.assertEqual(json_response["count"], 2)
```
hugo093, line 128: can you check for the content ? at least the shape?

iaso/tests/api/notifications/test_views.py
```diff
@@ -0,0 +1,81 @@
+from datetime import datetime, timedelta
+
+from hat.api_import.models import APIImport
+from iaso import test
+from iaso.models import Account, Project
+
+
+class TestNotificationsViewSet(test.APITestCase):
+    def setUp(self):
```
hugo093, line 13: super().setUp()

iaso/tests/api/notifications/test_views.py
```diff
@@ -0,0 +1,81 @@
...
+        self.account1 = account1 = Account.objects.create(name="account1")
+        self.account2 = account2 = Account.objects.create(name="account2")
+
+        self.project1 = Project.objects.create(name="project1", app_id="P1", account=account1)
+        self.project2 = Project.objects.create(name="project2", app_id="P2", account=account2)
+
+        self.staff = staff = self.create_user_with_profile(username="staff", account=account1, permissions=[])
```
hugo093: same remark

iaso/tests/api/notifications/test_views.py
```diff
@@ -0,0 +1,81 @@
...
+        self.assertEqual(len(json_response), 1)
+
+    def test_get_notifications_staff_problem_not_older_than_30_days(self):
+        apiimport = APIImport.objects.create(app_id="P1", has_problem=True, json_body={})
+        apiimport.created_at = datetime.now() - timedelta(days=28)
+        apiimport.save()
+        self.client.force_authenticate(user=self.staff)
+        response = self.client.get("/api/notifications/")
+        json_response = self.assertJSONResponse(response, 200)
+        self.assertEqual(len(json_response), 1)
```
hugo093, line 76: can you check the content and/or the shape of the response as well?

iaso/locale/fr/LC_MESSAGES/django.po
```diff
@@ -183,6 +187,11 @@ msgstr "Aucune assignation à effacer"
 msgid "Successfully deleted %(count)s assignments"
 msgstr "%(count)s assignations ont été supprimées avec succès"
 
+#: iaso/api/notifications/views.py
+#, python-format
+msgid "There is %(count)d failing import needing your attention."
```
hugo093, line 200: plural isn't translated, i think your code as is won't work then :/ see https://docs.djangoproject.com/en/6.0/topics/i18n/translation/#marking-strings-as-no-op

hat/api_import/management/commands/remove_authorization_headers_from_api_imports.py
```diff
@@ -0,0 +1,73 @@
...
+    help = "Clean-up duplicate submissions"
+
+    def add_arguments(self, parser):
+        parser.add_argument(f"--{DRY_RUN_ARG}", default=False, action=argparse.BooleanOptionalAction)
+        parser.add_argument(f"--{BATCH_SIZE_ARG}", default=1000)
+
+    @transaction.atomic
+    def handle(self, *args, **options):
+        dry_run = options[DRY_RUN_ARG]
+        batch_size = int(options[BATCH_SIZE_ARG])
```
hugo093: no need for this, in your add_arguments => parser.add_argument(f"--{BATCH_SIZE_ARG}", type=int, default=1000)

hat/api_import/management/commands/remove_authorization_headers_from_api_imports.py
```diff
@@ -0,0 +1,73 @@
...
+                objs_to_update.append(obj)
+
+            # When we hit the batch limit, write to DB and clear list
+            if len(objs_to_update) >= batch_size:
+                self.bulk_update(objs_to_update, dry_run)
+                objs_to_update = []
+
+        # Final batch update for remaining records
+        if objs_to_update:
+            self.bulk_update(objs_to_update, dry_run)
```
hugo093, line 63: detail but that's actually the only part that needs to be wrapped in transaction.atomic(), not the whole code

hat/api_import/management/commands/remove_authorization_headers_from_api_imports.py
```diff
@@ -0,0 +1,73 @@
...
+            modified = False
+
+            # Process Headers
+            if "HTTP_AUTHORIZATION" in obj.headers:
+                obj.headers.pop("HTTP_AUTHORIZATION", None)
+                modified = True
+
+            # Process App ID/Version from Query String
+            if obj.app_id == "":
+                querystring = obj.headers.get("QUERY_STRING", None)
```
hugo093, line 45: i wonder if we could put those two conditions "HTTP_AUTHORIZATION" and "QUERY_STRING" in the initial filter already, so it would even narrow down the initial queryset

hugo093 (changes requested):
good job overall, just need a few improvements

iaso/api/notifications/views.py
bmonjoie: Because we might have more notifications in the future.

hat/api_import/management/commands/remove_authorization_headers_from_api_imports.py
bmonjoie: I have thought about it, and I don't think so because we are doing two things. The first is to remove the header, and the second is to fill in the AppId and AppVersion; they are not related.

hat/api_import/management/commands/remove_authorization_headers_from_api_imports.py
```diff
@@ -0,0 +1,73 @@
...
+                    obj.app_version = parsed.get("app_version", [""])[0]
+                    modified = True
+
+            if modified:
+                total_modified += 1
+                objs_to_update.append(obj)
+
+            # When we hit the batch limit, write to DB and clear list
+            if len(objs_to_update) >= batch_size:
+                self.bulk_update(objs_to_update, dry_run)
```
tdethier, line 58: This is not part of the transaction?

hat/api_import/management/commands/remove_authorization_headers_from_api_imports.py
tdethier: I guess it makes sense to already add them in the initial filters if we do have a lot of objects without these values. 

I just checked on prod, there are only ~1.54% without `QUERY_STRING`, ~18.25% without `HTTP_AUTHORIZATION`. I'd say we can leave things as they are, but I don't have a strong opinion 🤷‍♂️

iaso/api/api_import/pagination.py
```diff
@@ -0,0 +1,5 @@
+from iaso.api.common import Paginator
```
tdethier, line 1: Nitpick: usually the file name is in singular - `pagination.py`

iaso/locale/fr/LC_MESSAGES/django.po
```diff
@@ -183,6 +195,13 @@ msgstr "Aucune assignation à effacer"
...
+msgid_plural "There are %(count)d failing imports needing your attention."
+msgstr[0] "Il y a %(count)d importation en échec nécessitant votre attention."
+msgstr[1] "Il y a %(count)d importation en échec nécessitant votre attention."
```
tdethier: Both lines have the exact same translation?

iaso/api/api_import/views.py
```diff
@@ -0,0 +1,30 @@
...
+    permission_classes = [IsAdminOrSuperUserWithIasoProfile]
+    serializer_class = APIImportSerializer
+    filterset_class = APIImportFilterSet
+    pagination_class = APIImportPagination
+    ordering_fields = ["created_at", "import_type", "user_id", "app_id", "app_version", "has_problem"]
+
+    def get_queryset(self):
+        queryset = APIImport.objects.all()
+        user = self.request.user
+        if not user.is_staff:
```
tdethier, line 25: I don't really understand why we make a distinction between `is_staff` and `is_superuser`, do we have some users that are only one of them?

hat/assets/js/apps/Iaso/components/nav/TopBarComponent.tsx
```diff
@@ -154,26 +155,38 @@ const TopBar: FunctionComponent<Props> = ({
                     </Grid>
                     {currentUser && !isMobileLayout && (
                         <Grid container item xs={5} justifyContent="flex-end">
+                            <Box
+                                display="flex"
+                                alignItems="center"
+                                justifyContent="flex-end"
+                            >
+                                <NotificationBadge />
+                            </Box>
```
tdethier, line 164: If we put this on the top bar, it means that we'll query again and again the notifications endpoint and put a huge load on the DB, right? I'm a bit afraid that the DB will not appreciate all these queries on this huge table

iaso/api/notifications/checks.py
```diff
@@ -0,0 +1,22 @@
...
+def check_apiimport_for_user(user: User) -> int:
+    account = user.iaso_profile.account if not user.is_staff else None
+    queryset = APIImport.objects.filter(has_problem=True).filter(
+        created_at__gte=timezone.now() - timedelta(days=DAYS_TO_CHECK)
+    )
+    if account:
+        queryset = queryset.filter(
+            app_id__in=Project.objects.filter(account=account).only("app_id").values_list("app_id")
+        )
+    return queryset.count()
```
tdethier, line 22: My notification says that there are 2 failing imports, but when I check that page I see 4 failures. I'm not sure if this is an issue with multi account (my user is a multi account) or with the user who made the failing import

tdethier (commented):
Some comments in any order:
- having a view for API imports was very necessary, thank you 👍 
- I'm not sure about putting the notification (and therefore fetching info about failing imports) on each page, at least without a cache; I'm a bit scared about the added load on the DB
- I'm not sure why there's a difference in behavior between `is_staff` and `is_superuser` 
- I think some part of the new command should be included in a transaction and is currently not

iaso/locale/fr/LC_MESSAGES/django.po
tdethier: Also I would talk about `import` instead of `importation`

scripts/translations/check.py
```diff
@@ -51,6 +51,10 @@ def check_po_file(po_path):
             in_msgid = False
             in_msgstr = True
             current_msgstr = [line[8:-1]]
+        elif line.startswith(('msgstr[0] "', 'msgstr[1] "')):
```
hugo093, line 54: 👍

hugo093 (approved):

hat/api_import/management/commands/remove_authorization_headers_from_api_imports.py
bmonjoie: I changed after @hugo093 's comment but I'll revert and put everything back into one transaction.

iaso/locale/fr/LC_MESSAGES/django.po
bmonjoie: One had an "s", I don't know where it went... I'll fix that

iaso/api/api_import/views.py
bmonjoie: I don't know, but my understanding is that `superusers` can be limited to one account and `staff` is not. Therefore, I make the distinction to only show stuff related to their account to `superusers` who are not `staff`.

hat/assets/js/apps/Iaso/components/nav/TopBarComponent.tsx
bmonjoie: No, because the hook has a `scaleTime` and `cacheTime` of 1 minute. Also, it's only admins who are going to query the database, not all the users.

iaso/api/notifications/checks.py
bmonjoie: The notification is about failures that happened in the past 30 days. There could be more than that.

tdethier:
I'm finishing up a few things on this PR since @bmonjoie is off for a week

tdethier (approved):
LGTM 👍

PR #2891, opened 2026-04-06
[IA-4947] Form copilot

madewulf:

IASO users need to create and modify ODK XLSForms, which currently requires manual spreadsheet editing — this PR adds an AI-powered chat interface ("Form Copilot") that generates and iteratively refines XLSForms
  from plain-language descriptions.

  ### Related JIRA tickets

  IA-4947

  ## Changes

  ### Backend

  **New module: `iaso/api/form_copilot/`**
  - `agent.py` — AI agent built on the Anthropic SDK. `generate_form` calls Claude and returns a parsed `GeneratedForm` Pydantic object. 
  - `views.py` — Four endpoints gated by the existing `iaso_forms` permission (no new permission introduced):
    - `POST /api/form_copilot/` — chat, generates a form and stores it in `TemporaryForm`
    - `GET /api/form_copilot/load/<form_id>/` — load an existing form's XLS into JSON for the AI context
    - `GET /api/form_copilot/download/<uuid>/` — authenticated download of a generated XLS (user-scoped)
    - `POST /api/form_copilot/save/` — promote a `TemporaryForm` to a real `FormVersion` via the existing `FormVersionSerializer`, which also enforces per-form access control

  **New model: `iaso/models/form_copilot.py`**
  - `TemporaryForm` — stores AI-generated XLS files with UUID as primary key (no redundant integer PK), scoped to user and account. Cleaned up via Django cascade on user/account deletion. Should play well with account deletion scripts.

  **Migration `0378_form_copilot`** — adds `anthropic_api_key` (nullable TextField) to `Account`, adds `FORM_COPILOT` to the `modules` choices array, and creates `TemporaryForm`. 

  **`iaso/modules.py`** — `MODULE_FORM_COPILOT` added. No linked permission; the existing `iaso_forms` permission is reused.

  **`iaso/admin/base.py`** — `TemporaryFormAdmin` registered with `raw_id_fields` on user to avoid loading the full user list.

  **API key** — stored per account (`Account.anthropic_api_key`), never falls back to an environment variable. Returns HTTP 503 with an explicit message if not configured.

  ### Frontend

  **New page: `hat/assets/js/apps/Iaso/domains/formCopilot/`**
  - Split layout: chat panel (40%) + live ODK Web Forms preview (60%)
  - `ChatPanel` — scrollable message thread with a text input
  - `FormPreview` — embeds the ODK preview app in an iframe and uses `postMessage` to inject XForm XML on each generation
  - `SaveFormDialog` — handles two save flows: "new version of an existing form" and "brand-new form" (with name, project, and optional ODK form ID fields)
  - Conversation history is kept client-side and sent with every request so the AI has full context for iterative edits (there is no compression of context here yet)
  - Loading an existing form injects its full XLSForm JSON into the conversation as a system context message (this will perform poorly on very large forms)

  **ODK Web Forms preview: `docker/odk-preview/`**
  - Small standalone Vue 3 app using `@getodk/web-forms`, built with Vite
  - Compiled output committed to `iaso/static/odk-preview/` so it is served directly as Django static files — no extra container or service needed
  - Receives XForm XML via `postMessage` (`load-form-xml`) from the React parent

  **Routing/menu** — Form Copilot added to the sidebar under Forms, visible only to users with the `iaso_forms` permission. Route added to `urls.ts` and `routes.tsx`.

  **i18n** — English and French translations added for all UI strings and the module name.

  ### Tests

  `iaso/tests/api/test_form_copilot.py` — 25 tests covering all four endpoints: authentication (401), permission (403), missing API key (503), 404 for wrong UUIDs/missing forms, successful generation with
  `TemporaryForm` creation, conversational responses without XLS output, form load with full response structure, file download scoped to the generating user, version save with and without `form_odk_id` patching.

  ## How to test

  1. In the Django admin, open your account and set an **Anthropic API key** (`anthropic_api_key` field). There is one in Bluesquare's 1password called "Claude API Key for dev testing". 
  2. Enable the **Form Copilot** module for the account via **Settings → Modules**.
  3. Log in as a user with the **Forms management** (`iaso_forms`) permission.
  4. Navigate to **Forms → Form Copilot**.
  5. Type: *"Create a household survey with name, age, gender, and GPS location"* and click Send.
  6. Verify the form preview appears on the right and a Download button is available.
  7. Send a follow-up: *"Make the age field required and add a phone number question"* — verify the preview updates.
  8. Click **Save form**, fill in name and project, and confirm the form appears in **Forms → Form list**.
  9. Load an existing form using the dropdown, ask for a modification, and save it as a new version.
  10. Test without an API key configured — expect a 503 error message in the UI.

  ## Print screen / video

https://github.com/user-attachments/assets/add13651-da18-4d45-bea1-73d9e366f49a

  ## Notes

  - The ODK preview bundle (`iaso/static/odk-preview/`) is committed to the repo to avoid adding build steps to the project. When `@getodk/web-forms` needs to be updated, rebuild with `cd docker/odk-preview && npm install && npm run build` and copy the
  `dist/` output to `iaso/static/odk-preview/`.
  - Conversation history is managed client-side only — refreshing the page loses the current draft. Users should save or download before leaving. 
  - Formatting (cell colours, custom styles) of an existing XLSForm is lost when it is modified through the Copilot, since the agent regenerates the full file from scratch.
  - The `TemporaryForm` table accumulates generated files over time. A periodic cleanup task is out of scope for this PR.

Possible improvements later on: 
- adding support for sockets and push the conversation handling to workers. Currently the calls to the AI are made synchronously.
- Adding a spreadsheet like editor as an alternative to the AI chatbot, to benefit from the webform preview that is awesome, but in a more deterministic manner than through an ai chatbot, which is fun and flashy, but sometimes frustrating.  I made a quick demo here, but this is beyond the scope of this already large PR: 
CleanShot 2026-04-19 at 14 03 11@2x
- allowing to pick various LLMs providers, in particular a self hosted one would be useful
- allowing to keep xls file formatting would be great when editing

  ## Doc

  User documentation (EN + FR) in `docs/pages/users/how_to/use_form_copilot/`.


iaso/api/form_copilot/agent.py
```diff
@@ -0,0 +1,299 @@
...
+
+
+def build_xlsform(form: GeneratedForm) -> io.BytesIO:
+    """Build an XLSForm Excel file from the generated form data."""
+    wb = openpyxl.Workbook()
+
+    # Survey sheet
+    survey_ws = wb.active
+    survey_ws.title = "survey"
+    survey_headers = [
```
mestachs: I think there's more columns than that (ex media::image, parameters, choice_filter, readonly, repeat_count)

if it's an update it means you loose all "color", extra blank, lines and other non xlsform related columns (like xlsx formula checking for uniqueness coding from the initial workbook and all other columns ? 

internationalized labels should be supported

Image

madewulf: Good point, the excel file should stay well formatted, and generated new forms should also be well formatted in fact. I'll look at that

iaso/urls.py
```diff
@@ -317,6 +323,10 @@ def append_datasources_subresource(viewset, resource_name, urlpatterns):
     path("workflows/export/<workflow_id>/", export_workflow, name="export_workflow"),
     path("workflows/import/", import_workflow, name="import_workflow"),
     path("colors/", colors_list, name="colors"),
+    path("form_copilot/", form_copilot_chat, name="form_copilot_chat"),
+    path("form_copilot/xform_xml/", form_copilot_xform_xml, name="form_copilot_xform_xml"),
+    path("form_copilot/load/<int:form_id>/", form_copilot_load_form, name="form_copilot_load_form"),
```
mestachs: this doesn't look really "rest"

docs/pages/dev/how_to/build_odk_preview/build_odk_preview.en.md
```diff
@@ -0,0 +1,55 @@
...
+## Prerequisites
+
+- Node.js ≥ 18
+- npm
+
+## Build
+
+```bash
+cd docker/odk-preview
+npm install
```
beygorghor: i would use `npm ci` instead, to use same version of dependencies from the package-lock.json file

docs/pages/dev/how_to/build_odk_preview/build_odk_preview.en.md
```diff
@@ -0,0 +1,55 @@
...
+git commit -m "rebuild odk-preview bundle"
+```
+
+## Run locally (dev server)
+
+During development of the Vue app itself you can run the Vite dev server instead of rebuilding on every change:
+
+```bash
+cd docker/odk-preview
+npm install
```
beygorghor: same here

docs/pages/dev/how_to/build_odk_preview/build_odk_preview.en.md
```diff
@@ -0,0 +1,55 @@
...
+
+Remember to revert this change before committing.
+
+## Updating `@getodk/web-forms`
+
+`package.json` currently pins `@getodk/web-forms` to `latest`. To update:
+
+```bash
+cd docker/odk-preview
+npm install @getodk/web-forms@latest
```
beygorghor, line 51: same

docker/odk-preview/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM node:20-alpine
+
+WORKDIR /app
+
+COPY package.json ./
```
beygorghor: Same remark here, we have a package-lock.json file, we can copy it and reuse it with `npm ci` just to be sure we are using the same version of dependencies

hat/assets/js/apps/Iaso/domains/formCopilot/components/ChatPanel.tsx
```diff
@@ -0,0 +1,224 @@
...
+                        }}
+                    >
+                        <Typography variant="body1">
+                            {formatMessage(MESSAGES.placeholder)}
+                        </Typography>
+                    </Box>
+                )}
+                {messages.map((msg, index) => (
+                    <Box
+                        key={`msg-${index}`}
```
beygorghor: msg has probably a unique id like `msg.id` or something no ?

hat/assets/js/apps/Iaso/domains/formCopilot/components/ChatPanel.tsx
```diff
@@ -0,0 +1,224 @@
...
+    const [inputValue, setInputValue] = useState('');
+    const messagesEndRef = useRef<HTMLDivElement>(null);
+
+    const scrollToBottom = useCallback(() => {
+        messagesEndRef.current?.scrollIntoView({ behavior: 'smooth' });
+    }, []);
+
+    useEffect(() => {
+        scrollToBottom();
+    }, [messages, scrollToBottom]);
```
beygorghor:
```suggestion
    useEffect(() => {
        messagesEndRef.current?.scrollIntoView({ behavior: 'smooth' });
    }, []);
```

hat/assets/js/apps/Iaso/domains/formAI/components/ChatPanel.tsx
beygorghor, line 1: Next time if you can take out all `sx` props into a separate variable outside of the component
`const styles: SxStyles = {
    table: {
       ...
    },
    ...,
};
` and reuse it in the JSX 🙏

iaso/api/form_copilot/views.py
```diff
@@ -0,0 +1,254 @@
...
+@api_view(["GET"])
+@permission_classes([HasFormCopilotPermission])
+def form_copilot_load_form(request, form_id):
+    """Load an existing form's latest version for editing in the copilot.
+
+    Returns the form's XLSForm structure as JSON (survey, choices, settings)
+    plus the XForm XML for preview.
+    """
+    try:
+        form = Form.objects.filter_for_user_and_app_id(request.user, None).get(id=form_id)
```
beygorghor: Maybe add a distinct here, i had some weird issue while loading a form

Image

hat/assets/js/apps/Iaso/domains/formCopilot/hooks/requests/useGetFormsList.ts
beygorghor, line 1: We have already a hook doing this

hat/assets/js/apps/Iaso/domains/formAI/components/SaveFormDialog.tsx
```diff
@@ -0,0 +1,234 @@
...
+    xlsformUuid,
+    selectedFormId,
+    selectedFormName,
+    onSaveNewVersion,
+    onSaveNewForm,
+}) => {
+    const { formatMessage } = useSafeIntl();
+    const [tab, setTab] = useState(selectedFormId ? 0 : 1);
+    const [formName, setFormName] = useState('');
+    const [formOdkId, setFormOdkId] = useState('');
```
beygorghor, line 63: Why not using `undefined`, also you can type the state with useState like on the next line.

hat/assets/js/apps/Iaso/domains/formCopilot/components/SaveFormDialog.tsx
```diff
@@ -0,0 +1,234 @@
...
+                            required
+                        />
+                        <TextField
+                            label={formatMessage(MESSAGES.formOdkId)}
+                            value={formOdkId}
+                            onChange={e =>
+                                setFormOdkId(
+                                    e.target.value
+                                        .toLowerCase()
+                                        .replace(/[^a-z0-9_]/g, '_'),
```
beygorghor: We can probably use it has a utility function outside of the component...

hat/assets/js/apps/Iaso/domains/formCopilot/index.tsx
```diff
@@ -0,0 +1,301 @@
...
+        boxSizing: 'border-box',
+    },
+    toolbar: {
+        display: 'flex',
+        gap: '8px',
+        alignItems: 'center',
+        padding: '8px',
+        flexShrink: 0,
+    },
+}));
```
beygorghor: Classes and makeStyles is the old way to proceed, we should use sx props like the rest

beygorghor (approved):
Nothing really blocking on the frontend side. 
I would make sure in the future that everything is translated, Vue app too.

quang-le:
I just want to say we should find and use another name

mestachs:
from backend point of view

- not sure we will keep return 500-503 when llm api is not ok, this might trigger either unhealthy servers (beantalk/k8s)
   - it feels a bit like throwing intentionally NullPointerException from our code ;)
   - @madewulf @beygorghor do you agree to change to one of the 2
     - 400 (Bad Request) → malformed request / validation basic
     - 422 (Unprocessable Entity) → request is valid JSON but semantically wrong (often better for APIs)
- while using DRF, it's perhaps not super drf/rest but I guess if we want a first version soon, changes will come after the first usage.
- in the things I don't like, we create 2 path for creating a formVersion
   - from what I saw we use the FormVersionSerializer and so hope we call create_for_form_and_survey that will copy the previous version dhis2 mappings 
   - but it opens a road where we might have some validation in a part of the app and not the other.

beygorghor:
> from backend point of view
> 
> * not sure we will keep return 500-503 when llm api is not ok, this might trigger either unhealthy servers (beantalk/k8s)
>   
>   * it feels a bit like throwing intentionally NullPointerException from our code ;)
>   * @madewulf @beygorghor do you agree to change to one of the 2
>     
>     * 400 (Bad Request) → malformed request / validation basic
>     * 422 (Unprocessable Entity) → request is valid JSON but semantically wrong (often better for APIs)
> * while using DRF, it's perhaps not super drf/rest but I guess if we want a first version soon, changes will come after the first usage.
> * in the things I don't like, we create 2 path for creating a formVersion
>   
>   * from what I saw we use the FormVersionSerializer and so hope we call create_for_form_and_survey that will copy the previous version dhis2 mappings
>   * but it opens a road where we might have some validation in a part of the app and not the other.

@mestachs I agree with this, we can already merge this and make other tickets to improve form version creation.
We can maybe already change the 500 error... 400 or 422 i don't really have a preference i'd say

madewulf:
@mestachs You're right about the return codes: this feels ok to me
400 (Bad Request) for most chat errors
I would keep 503 as 503 when it's in fact Anthropic that is giving us a 503 OTOH: this will warn us if our provider is problematic, while the first error is more of a configuration error. I updated the PR for that.

madewulf:
@mestachs for the fact that there are two ways to create form versions now, I don't think there is a better option. I could move the logic of rewriting odk_form_id to the main form version endpoint, but I think that's not clean either. The fact that the new endpoint is using the same form version serializer implies that the same validation is done, and also that dhis2 mappings are copied too to the new form versions. I think it's the optimal way to do it, frankly, even if I do see the risk of forgetting some validation on this endpoint in the future.

madewulf:
@quang-le I renamed the feature Form AI and all the references in the code. @mestachs had the same question on the existing name, so I did rename it.

PR #2892, opened 2026-04-07
IA-4950 : Add validation workflow feature flag for web

quang-le:

Feature needs to be behind feature flag until full release + we might want to be able to toggle it anyway

### Related JIRA tickets

IA-4950

## Changes

- Add `AccountFeatureFlag` to block web endpoints
- Add migration
- Add translations + tests

## How to test

With an account without the FF active, try to use any part (excpet mobile) of the validation workflows API --> should get 403
Do the reverse with an account with the FF active
Check the Projects UI: the feature flag should be visible in the list (data validation section). name +tooltip should be translated

## Print screen / video

Screenshot 2026-04-07 at 15 50 48


hugo093 (approved):
lgtm

PR #2899, opened 2026-04-09
[IA-4951]Setting the updated name on the group object before saving

Crebert08:

When updating the name of the user role, IASO showed a sucess message but the name didn't update. This PR is about fixing it

### Related JIRA tickets

[IA-4951](https://bluesquare.atlassian.net/browse/IA-4951)

## Changes

Added one line to set the update name on the group object.

## How to test

> Go to User role page
> Update the name of an existing user role
> The name should be updated correctly

## Print screen / video

https://github.com/user-attachments/assets/1e1b7908-d059-4119-ac2e-c3c670bfb0d9

## Notes

While resolving the ticket, I found these duplicates issues: 
image

This performance issue could be resolved by removing the `get_object_or_404` that fetchs each permission one by one and implement a check that fetchs all the permissions at once. The same issue is present in the Post Request. 
Let me know if it should stay like this or needs to be optimized knowing that the invalid permissions would return 400 listing all invalid values and could be dangerous for any client relying on a 404 response

The result of the optimization would look like this: 
image

## Doc

[IA-4951]: https://bluesquare.atlassian.net/browse/IA-4951?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iaso/api/user_roles.py
+++ b/iaso/api/user_roles.py
@@ -92,6 +92,7 @@ class UserRoleSerializer(serializers.ModelSerializer):
                 permission = get_object_or_404(Permission, codename__startswith="iaso_", codename=permission_codename)
                 group.permissions.add(permission)
 
+        group.name = group_name
         group.save()
         user_role.save()
         user_role.editable_org_unit_types.set(editable_org_unit_types)
--- a/iaso/tests/api/test_user_roles.py
+++ b/iaso/tests/api/test_user_roles.py
@@ -169,6 +169,7 @@ class UserRoleAPITestCase(APITestCase):
         r = self.assertJSONResponse(response, 200)
         expected_name = self.user_role.group.name.removeprefix(f"{self.account.id}_")
         self.assertEqual(r["name"], expected_name)
+        self.assertEqual(self.group.name, f"{self.account.id}_user role modified")
 
     def test_partial_update_permissions_modification(self):
         self.client.force_authenticate(self.user)


iaso/api/user_roles.py
```diff
@@ -92,6 +92,7 @@ def update(self, user_role, validated_data):
                 permission = get_object_or_404(Permission, codename__startswith="iaso_", codename=permission_codename)
                 group.permissions.add(permission)
 
+        group.name = group_name
```
tdethier, line 95: Could you please add a unit test where you call the endpoint with a PATCH call in order to update the user role name?

tdethier (changes requested):
I'm requesting changes because there's no test that shows that the endpoint works properly now.

>This performance issue could be resolved by removing the get_object_or_404 that fetchs each permission one by one and implement a check that fetchs all the permissions at once. The same issue is present in the Post Request.
Let me know if it should stay like this or needs to be optimized knowing that the invalid permissions would return 400 listing all invalid values and could be dangerous for any client relying on a 404 response

Good catch! It would indeed be better to check all permissions in a single call. The new profile API checks this in `ProfileCreateSerializer` for instance. However, you should probably create another ticket for this other issue and fix it in another PR.

iaso/api/user_roles.py
Crebert08: Allowed endpoints don't include the patch method, the unit test will fail with a 405 error if this method is not added. Is the intention of the comment adding the new endpoint or test to see if it fails correctly ?

@tdethier

tdethier: My bad I thought that it was a PATCH instead of PUT. Please replace `PATCH` with `PUT` in my previous comment

tdethier: Basically I want a test that shows that your fix is working 🙂

Crebert08: There is a test unit already covering this feature. 
But Because the update method didn't have the line I added, the test was comparing 'old name' vs 'old name'. I modified the test to assess if the name in the `put` payload actually corresponds to the db name that was modified

tdethier (approved):
LGTM 👍

PR #2900, opened 2026-04-09
[SNT-298] SNT account setup - backend

tdethier:

Update requirements in order to allow SNT account setups

### Related JIRA tickets

SNT-298

## Changes

- update requirements
- fix gpkg export after geopandas update

## How to test

- run tests, make sure that nothing is broken
- export any source version as gpkg, reimport it and make sure that it works properly
  - optional: open it in QGIS and make sure that there's no CRS issue

## Print screen / video

/

## Notes

- This is required to add a new endpoint in SNT, which is in [a dedicated SNT PR](https://github.com/BLSQ/snt-malaria/pull/256)
- Since geopandas swiched their engine in `1.0.0`, passing directly `crs=` in `geodataframe.to_file()` [is no longer allowed because the new engine does not support it](https://geopandas.org/en/v1.0.1/docs/reference/api/geopandas.GeoDataFrame.to_file.html)  

## Doc

/

### DIFF
--- a/iaso/gpkg/export_gpkg.py
+++ b/iaso/gpkg/export_gpkg.py
@@ -128,6 +128,7 @@ def export_org_units_to_gpkg(filepath, orgunits: "QuerySet[OrgUnit]") -> None:
     # Convert to GeoDataframe, and group by org unit type
     # as we want one layer per OrgUnitType
     ou_gdf = gpd.GeoDataFrame(df, geometry="geography")
+    ou_gdf.set_crs(crs="EPSG:4326", inplace=True)
     ou_gdf["layer_name"] = "level-" + ou_gdf["depth"].astype(str) + "-" + ou_gdf["type"]
 
     ou_gdf_by_type = ou_gdf.groupby("layer_name")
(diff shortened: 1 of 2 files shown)


beygorghor (approved):

iaso/gpkg/export_gpkg.py
```diff
@@ -128,6 +128,7 @@ def export_org_units_to_gpkg(filepath, orgunits: "QuerySet[OrgUnit]") -> None:
     # Convert to GeoDataframe, and group by org unit type
     # as we want one layer per OrgUnitType
     ou_gdf = gpd.GeoDataFrame(df, geometry="geography")
+    ou_gdf.set_crs(crs="EPSG:4326", inplace=True)
```
Bewi, line 131: Ah tes souhaits, j'ai l'impression qu'on parle d'une unité CRS d'élite.

requirements.txt
```diff
@@ -34,7 +34,7 @@ django-redis==4.12.1  # https://github.com/jazzband/django-redis/tags
 # Pandas.
 # ------------------------------------------------------------------------------
 pandas==2.0.3  # https://github.com/pandas-dev/pandas/tags
-geopandas==0.14.2  # https://github.com/geopandas/geopandas/tags
+geopandas==1.0.1  # https://github.com/geopandas/geopandas/tags
```
Bewi, line 37: Sure we don't have breaking changes with this ?

Bewi (approved):
LGTM

requirements.txt
tdethier: We should be good, we have tests that cover gpkg import and export (I think we even have one that imports the result of an export). When I tested manually it went fine

PR #2901, opened 2026-04-09
[IA-4948] Adding the staletime option to avoid repeting the call

Crebert08:

When going back from the orgunit details to the orgunits list, an additional call was being made to `api/validationstatus/`. This additional call was also being made when fetching the org units list. This PR is about fixing it.

### Related JIRA tickets

[IA-4948](https://bluesquare.atlassian.net/browse/IA-4948)

## Changes

added the staletime option and set it to 1 hour

## How to test

>Go to org units list
>Go to details page
>Go back

Expected: the call to api/validationstatus/ is not repeated 

## Print screen / video

## Notes

## Doc

[IA-4948]: https://bluesquare.atlassian.net/browse/IA-4948?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/orgUnits/hooks/utils/useGetOrgUnitValidationStatus.ts
+++ b/hat/assets/js/apps/Iaso/domains/orgUnits/hooks/utils/useGetOrgUnitValidationStatus.ts
@@ -39,7 +39,7 @@ export const useGetOrgUnitValidationStatus = (
             retry: false,
             enabled,
             keepPreviousData: true,
-            staleTime: 1000 * 60 * 60,
+            staleTime: Infinity,
             select: (data: OrgUnitStatus[]) => {
                 const options: DropdownOptions<string>[] = data.map(
                     (status: OrgUnitStatus) => ({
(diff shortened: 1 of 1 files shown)


hat/assets/js/apps/Iaso/domains/orgUnits/hooks/utils/useGetOrgUnitValidationStatus.ts
```diff
@@ -39,6 +39,7 @@ export const useGetOrgUnitValidationStatus = (
             retry: false,
             enabled,
             keepPreviousData: true,
+            staleTime: 1000 * 60 * 60,
```
quang-le: You can set it to `Infinity`, this data doesn't move much

quang-le (changes requested):
nothing much, you found the right change

quang-le (approved):

PR #2902, opened 2026-04-10
[IA-4958]: Permission issue in planning API

hugo093:

When an anonymous user would try to access planning API, it wouldn't get a 401 as expected
### Related JIRA tickets

IA-4958

## Changes

Changed permissions on the viewset + tests

## How to test

call planning API endpoints (e.g /api/microplanning/plannings/ ) while unauthenticated, you should get a 401 and not a 500.


quang-le (approved):
LGTM

PR #2903, opened 2026-04-10
[IA-4821] Enforce password validation

tdethier:

Enforce password validation

### Related JIRA tickets

IA-4821

## Changes

- Added `Account.enforce_password_validation`
- Added password validation on:
  - Profile creation API
  - Profile update API
  - User bulk create API
  - setup account API
- Refactors:
  - moved `Account` and `AccountFeatureFlag` to `account.py`
  - created directory for API tests of `/setupaccount/` and moved existing tests there

## How to test

- run the new migration, all existing accounts should not have password validation
  - create a new user through the web interface with a stupid password (_e.g._ `1234`) :arrow_right: it should still be possible
- create a new account with the form in the admin section
  - if you use a stupid password (_e.g._ `1234`), you should get a validation error
  - if you use a proper password (at least 8 chars, not only numbers, not similar to the user name/email/...), the new account should be created
- log into your new account and create a new user through the web interface: same here, validation should prevent you from creating bad password
- run the setuper on your local instance: it should finish successfully and you should still be able to login with the same string for username and password (which is disabled otherwise)

## Print screen / video

image

image

image

image

## Notes

- password validation rules are the Django default ones - see `AUTH_PASSWORD_VALIDATORS` in `settings.py`
- once this is released, we're going to need to go through all accounts, one by one and decide whether password validation should be active or not

## Doc

/


iaso/migrations/0378_account_enforce_password_validation.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+def migrate_data_forward(apps, schema_editor):
+    Account = apps.get_model("iaso", "Account")
+    accounts = Account.objects.all()
+
+    for account in accounts:
+        account.enforce_password_validation = False
+        account.save()
+
```
madewulf, line 13: Not a big deal at all and this can be kept as is, but this feels like it could have been done through Accounts.objects.all().update(enforce_password_validation=False)

madewulf:
The double error message : popup + red around the fields is a bit unelegant, but this could be treated in a second pass.

setuper/setuper.py
```diff
@@ -75,6 +75,7 @@ def setup_account(account_name, server_url, username, password, create_main_org_
         ],
         "create_main_org_unit": create_main_org_unit,
         "create_demo_form": True,
+        "enforce_password_validation": False,
```
madewulf, line 78: I would advise to enforce it by default.

madewulf (dismissed):

setuper/setuper.py
tdethier: Then we have to redo the setuper, because by design all passwords are the same as the username. How should passwords be generated and communicated to the user for a setuper account? And/or how should the user input password for all users generated by the setuper?

beygorghor (approved):

beygorghor:
⚠️ This should be communicate to the rest of the team when released. New account will have this by default and old accounts require a manual activation.

beygorghor:
@madewulf we will take care of setuper issue in another ticket but this needs a bit more specifications

tdethier:
Setuper specification will be done in IA-5010

PR #2911, opened 2026-04-14
[IA-4966] Fixing the n+1 write issues on the put endpoint and adapting the tests

Crebert08:

N+1 issues on the user roles PUT endpoint. This PR is about fixing it

### Related JIRA tickets

[IA-4966](https://bluesquare.atlassian.net/browse/IA-4966)

## Changes

Removed the get_object_or_404 method that was causing extra computations

## How to test

Go to user roles page
Update a role
the docker logs should display a diminition in the number of duplicates

## Print screen / video

image

## Notes

Same performance issue appear on the Post endpoint

## Doc

[IA-4966]: https://bluesquare.atlassian.net/browse/IA-4966?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iaso/api/user_roles.py
+++ b/iaso/api/user_roles.py
@@ -1,6 +1,5 @@
 from django.contrib.auth.models import Group, Permission
 from django.db.models import Q, QuerySet
-from django.shortcuts import get_object_or_404
 from drf_spectacular.utils import extend_schema
 from rest_framework import permissions, serializers, status
 from rest_framework.response import Response
--- a/iaso/api/user_roles.py
+++ b/iaso/api/user_roles.py
@@ -90,8 +90,8 @@ class UserRoleSerializer(serializers.ModelSerializer):
             permission_objects = Permission.objects.filter(codename__startswith="iaso_", codename__in=permissions)
             found_codenames = set(permission_objects.values_list("codename", flat=True))
             invalid = [p for p in permissions if p not in found_codenames]
-            invalid_str = ", ".join(invalid)
             if invalid:
+                invalid_str = ", ".join(invalid)
                 raise serializers.ValidationError({"permissions": f"Invalid permission codenames: {invalid_str}"})
             group.permissions.set(permission_objects)
 
--- a/iaso/api/user_roles.py
+++ b/iaso/api/user_roles.py
@@ -87,10 +87,13 @@ class UserRoleSerializer(serializers.ModelSerializer):
             raise serializers.ValidationError({"name": "User role already exists"})
 
         if permissions is not None:
-            group.permissions.clear()
-            for permission_codename in permissions:
-                permission = get_object_or_404(Permission, codename__startswith="iaso_", codename=permission_codename)
-                group.permissions.add(permission)
+            permission_objects = Permission.objects.filter(codename__startswith="iaso_", codename__in=permissions)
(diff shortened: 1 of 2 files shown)


iaso/tests/api/test_user_roles.py
```diff
@@ -189,17 +192,18 @@ def test_partial_update_permissions_modification(self):
     def test_partial_update_not_allowable_permissions_modification(self):
         self.client.force_authenticate(self.user)
 
+        invalid_perms = self.permission_not_allowable.codename, self.permission_not_allowable2.codename
```
tdethier: FYI, this is a tuple, not a list

iaso/tests/api/test_user_roles.py
```diff
@@ -41,6 +41,9 @@ def setUpTestData(cls):
         cls.permission_not_allowable = Permission.objects.create(
             name="admin permission", content_type_id=1, codename="admin_permission1"
         )
+        cls.permission_not_allowable2 = Permission.objects.create(
+            name="admin permission", content_type_id=2, codename="admin_permission2"
```
tdethier: Why did you set `content_type_id=2`?

iaso/tests/api/test_user_roles.py
```diff
@@ -189,17 +192,18 @@ def test_partial_update_permissions_modification(self):
...
+            "permissions": invalid_perms,
         }
         response = self.client.put(f"/api/userroles/{self.user_role.id}/", data=payload, format="json")
 
-        r = self.assertJSONResponse(response, 404)
-        self.assertEqual(
-            r["detail"],
-            "Not found.",
-        )
+        r = self.assertJSONResponse(response, 400)
```
tdethier: Nitpicks (even if I know that you did not write these lines, you just edited them, but FYI):
- 99% of times, I don't like single letter variables like `r`, because it's not explicit - IMO `result` would be much better
- I don't like when `400` is used directly, using a const value is clearer (for example from `rest_framework.status`), even if the 400 error is a familiar one

iaso/api/user_roles.py
```diff
@@ -87,10 +87,13 @@ def update(self, user_role, validated_data):
...
 
         if permissions is not None:
-            group.permissions.clear()
-            for permission_codename in permissions:
-                permission = get_object_or_404(Permission, codename__startswith="iaso_", codename=permission_codename)
-                group.permissions.add(permission)
+            permission_objects = Permission.objects.filter(codename__startswith="iaso_", codename__in=permissions)
+            found_codenames = set(permission_objects.values_list("codename", flat=True))
+            invalid = [p for p in permissions if p not in found_codenames]
+            invalid_str = ", ".join(invalid)
```
tdethier: I would place this line inside the `if invalid:`, because you don't need to build this string if there's no issue

tdethier (commented):
I'm updating the same user role (changing only its name)

Before:
Image

After:

Image

So clearly, it's better 👍 

I left a few comments and nitpicks, please feel free to implement them if you think they make sense.
Regarding the POST endpoint, I think you can fix it in this PR as well, good catch.

tdethier (approved):
LGTM 👍

PR #2912, opened 2026-04-14
[IA-4831]: backend review list and retrieve payloads

hugo093:

Profile API list payload was too big for our current usage

### Related JIRA tickets

IA-4831

## Changes

* Reduce payload of profile list api 
* Adapt tests and TS definitions

## How to test

Go to users management UI, the table should still work and display all needed data


quang-le (changes requested):
Can you add tests for the response shape for both list and retrieve, since they're now diverging?

hugo093:
> Can you add tests for the response shape for both list and retrieve, since they're now diverging?

Front-end or back-end ? (or both)

iaso/api/profiles/serializers/list.py
```diff
@@ -17,19 +18,21 @@ class Meta:
...
-class ProfileListSerializer(ModelSerializer):
+class ProfileListSerializer(DynamicFieldsModelSerializer):
     first_name = serializers.SerializerMethodField(read_only=True)
     user_name = serializers.SerializerMethodField(read_only=True)
     last_name = serializers.SerializerMethodField(read_only=True)
     email = serializers.SerializerMethodField(read_only=True)
     phone_number = serializers.CharField(source="phone_number.as_e164", read_only=True)
     user_roles = NestedUserRoleSerializer(many=True, read_only=True)
     projects = RelatedProjectSerializer(many=True, read_only=True)
+    user_display = serializers.SerializerMethodField(read_only=True)
```
quang-le, line 35: I think we can remove that one no? Isn't it the one I removed from the table?

quang-le (approved):

iaso/api/profiles/serializers/list.py
hugo093: used in dropdown to avoid the big function in front end

PR #2914, opened 2026-04-14
[IA-4840]: Profile search UI: dynamic columns

hugo093:

Implement dynamic columns on the profiles search UI

### Related JIRA tickets

IA-4840

## Changes

* Adapt backend:
  * Added DynamicFieldsModelSerializer
  * Added user_display field to help frontend component `UsersSelect`
* Backend tests
* Modify `DynamicFieldsModelSerializer` so drf-spectacular picks automatically all the fields and not the default ones
* Adapt frontend 

## How to test

Go to Admin -> Users -> Management. On that search view, you can toggle the columns with the "Select visible columns" button and play with it


quang-le (changes requested):
I guess you took inspiration from the forms or instances table, but those are old, and have some legacy code.

`TableWithDeepLink` has a props:

```typescript
    columnSelectorEnabled?: boolean;
    columnSelectorButtonDisabled?: boolean;
    columnSelectorButtonType?: 'button' | 'icon';
```

That you can use to enable the `ColumnDrawer`. Unless you tried it and it didn't work for some reason, you should use that instead. Sorry for the wasted work

quang-le (approved):

PR #2915, opened 2026-04-14
[IA-4970]: make swagger accessible to non superuser

hugo093:

Make our swagger endpoints accessible to authenticated user , and not only the superusers

### Related JIRA tickets

IA-4970

## Changes

Change settings + adapt tests

## How to test

Try to access /swagger-ui or /swagger with a normal authenticated user, should work


tdethier (approved):
LGTM 👍

PR #2917, opened 2026-04-15
[IA-4843]: add org unit write type panel

hugo093:

In user detail view, adding a org unit write types panel

### Related JIRA tickets

IA-4843

## Changes

* Adapt backend to return id and name for editable org unit write types
* Add a paper panel on the front end side + tests
* Adapt edit user dialog with initial data

## How to test

Go to any user detail view, you should see a new org unit write type panel with data

## Print screen / video

screencapture-localhost-8000-dashboard-settings-users-management-details-accountId-1-userId-2-2026-04-15-10_55_01


hat/assets/js/apps/Iaso/domains/users/components/UserDetailViewComponents/OrgUnitWriteTypePaper.tsx
```diff
@@ -0,0 +1,54 @@
...
+                        pb: 2,
+                    }}
+                >
+                    {profile?.editable_org_unit_types?.map(orgUnitWriteType => (
+                        <Box key={`orgUnitWriteType-${orgUnitWriteType.id}`}>
+                            <Chip
+                                sx={{
+                                    backgroundColor: '#EA8418',
+                                    color: 'white',
+                                    fontSize: '0.8rem',
```
quang-le: Can you use the theme's secondary color, or some other theme based color instead?

quang-le (changes requested):
small change

quang-le (approved):

PR #2921, opened 2026-04-16
[IA-4971] fix(forms) : fix differentiate showDeleted and only_deleted semantics

62496:

This PR fixes the deleted forms filters semantics in the forms endpoint and add onlyDeleted filter

### Related JIRA tickets

IA-4971

## Changes

- This PR updates the backend forms queryset logic so that only_deleted and showDeleted no longer behave the same way.
- Before this change, both filters were switching the queryset to deleted forms only. With this PR:
- only_deleted returns only deleted forms
- showDeleted returns both deleted and non-deleted forms
- a regular forms request still returns only non-deleted forms

## How to test

On the backend:

    - call /api/forms/ and verify that only non-deleted forms are returned
    - call /api/forms/?only_deleted=true and verify that only deleted forms are returned
    - call /api/forms/?showDeleted=true and verify that both deleted and non-deleted forms are returned

## Doc

The main documentation for this change is currently in:

    - the backend queryset logic for forms
    - related tests

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/forms/components/Filters.tsx
+++ b/hat/assets/js/apps/Iaso/domains/forms/components/Filters.tsx
@@ -11,14 +11,17 @@ import {
 
 import { SearchButton } from 'Iaso/components/SearchButton';
 import { baseUrls } from 'Iaso/constants/urls';
+import { PlanningsDropdown } from 'Iaso/domains/plannings/components/PlanningsDropdown';
+import { userHasOneOfPermissions } from 'Iaso/domains/users/utils';
 import { useQueryString } from 'Iaso/hooks/useApiParams';
 import * as Permission from 'Iaso/utils/permissions';
+import { PLANNING_READ, PLANNING_WRITE } from 'Iaso/utils/permissions';
+import { useCurrentUser } from 'Iaso/utils/usersUtils';
 import { DisplayIfUserHasPerm } from '../../../components/DisplayIfUserHasPerm';
 import DownloadButtonsComponent from '../../../components/DownloadButtonsComponent';
 import InputComponent from '../../../components/forms/InputComponent';
 import { useFilterState } from '../../../hooks/useFilterState';
 import { useGetOrgUnitTypesDropdownOptions } from '../../orgUnits/orgUnitTypes/hooks/useGetOrgUnitTypesDropdownOptions';
-import { useGetPlanningsOptions } from '../../plannings/hooks/requests/useGetPlannings';
 import { useGetProjectsDropdownOptions } from '../../projects/hooks/requests';
 import { baseUrl } from '../config';
 import { FormResponse, tableDefaults } from '../hooks/useGetForms';
@@ -57,7 +60,6 @@ const Filters: FunctionComponent<Props> = ({
         },
         [handleChange],
     );
-    const { data: planningsDropdownOptions } = useGetPlanningsOptions();
     const { data: orgUnitTypes, isFetching: isFetchingOuTypes } =
         useGetOrgUnitTypesDropdownOptions();
     const { data: allProjects, isFetching: isFetchingProjects } =
@@ -78,6 +80,12 @@ const Filters: FunctionComponent<Props> = ({
     );
     const csvUrl = `${dwnldBaseUrl}/?${downloadQueryString}&csv=true`;
     const xlsxUrl = `${dwnldBaseUrl}/?${downloadQueryString}&xlsx=true`;
+
+    const currentUser = useCurrentUser();
+    const hasPlanningPermission = userHasOneOfPermissions(
+        [PLANNING_READ, PLANNING_WRITE],
+        currentUser,
+    );
     return (
         <Grid container>
             <Grid container item xs={12} spacing={2}>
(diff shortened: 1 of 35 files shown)


iaso/tests/api/test_forms.py
```diff
@@ -232,13 +232,60 @@ def test_form_return_only_deleted(self):
...
+        self.assertEqual(response.json()["forms"][0]["id"], self.form_1.id)
+
+    def test_forms_list_show_deleted(self):
+        """GET /forms/?showDeleted=true should return both deleted and non-deleted forms."""
+        self.client.force_authenticate(self.yoda)
+
+        self.form_1.delete()
+
+        response = self.client.get(
+            "/api/forms/?showDeleted=true&limit=50",
```
hugo093: for get request, you can pass query params through the data;
Like : 
```
response = self.client.get('/api/forms/', {showDeleted: True, limit: 50}) 

```

also no need to add headers content type application/json => see in settings.py 

```   
 "TEST_REQUEST_DEFAULT_FORMAT": "json",  # The default format that should be used when making test requests.
```

hat/assets/js/apps/Iaso/domains/forms/components/Filters.tsx
```diff
@@ -50,7 +50,7 @@ const Filters: FunctionComponent<Props> = ({
         searchAlwaysEnabled: true,
     });
     const [textSearchError, setTextSearchError] = useState<boolean>(false);
-    const handleShowDeleted = useCallback(
+    const handleonlyDeleted = useCallback(
```
hugo093: naming convention , should be handleOnlyDeleted

hat/assets/js/apps/Iaso/domains/forms/hooks/useRestoreForm.tsx
```diff
@@ -4,7 +4,7 @@ import { useSnackMutation } from '../../../libs/apiHooks';
 import MESSAGES from '../messages';
 
 const restoreForm = (id: number) => {
-    return restoreRequest(`/api/forms/${id}/?only_deleted=1`);
+    return restoreRequest(`/api/forms/${id}/?showDeleted=true`);
```
hugo093: I know it's a bit annoying, but for the moment we try to keep everything as snake_case when it's related to backend, so it should be show_deleted (same goes for query params in backend). 

Also =1 instead of =true ?

iaso/api/forms.py
```diff
@@ -281,12 +281,12 @@ class FormsViewSet(ModelViewSet):
 
     def get_queryset(self, mobile=False):
         form_objects = Form.objects
-        if self.request.query_params.get("only_deleted", None):
+        if self.request.query_params.get("onlyDeleted", None):
```
hugo093: should be only_deleted

iaso/tests/api/test_forms.py
```diff
@@ -232,13 +232,60 @@ def test_form_return_only_deleted(self):
...
+
+        self.form_1.delete()
+
+        response = self.client.get(
+            "/api/forms/?&limit=50",
+            headers={"Content-Type": "application/json"},
+        )
+
+        self.assertJSONResponse(response, 200)
+        self.assertEqual(response.json()["count"], 1)
```
hugo093: assertJSONResponse returns the response.json() ;) so you can do res_json = self.assertJSONResponse(response, 200) and use it later

iaso/api/forms.py
```diff
@@ -281,12 +281,12 @@ class FormsViewSet(ModelViewSet):
...
             form_objects = Form.objects_only_deleted
 
         show_deleted = self.request.query_params.get("showDeleted", "false")
         if show_deleted == "true":
-            form_objects = Form.objects_only_deleted
+            form_objects = Form.objects_include_deleted
```
hugo093: while it works, I'm not really a fan of this for those reasons : 

* No way to validate the query input
* I won't appear in the openAPI file automatically (drf-spectacular)

I think we should investigate to create a custom filter for that ? (ping @tdethier )
But that's way out of scope so no worries

hat/assets/js/apps/Iaso/domains/forms/hooks/useRestoreForm.test.ts
```diff
@@ -0,0 +1,31 @@
...
+vi.mock('../../../libs/apiHooks', () => ({
+    useSnackMutation: mockUseSnackMutation,
+}));
+
+describe('useRestoreForm', () => {
+    beforeEach(() => {
+        vi.clearAllMocks();
+    });
+
+    it('uses showDeleted so a deleted form can be restored', async () => {
```
hugo093, line 19: 👍

hugo093 (changes requested):
Just read the code, some minor changes here and there, but good job overall and thanks for writing tests :) I'll test it really after ;)

iaso/api/forms.py
62496: Quick question on naming convention: would you prefer us to use snake_case end-to-end in the frontend as well (only_deleted, show_deleted), or keep camelCase in frontend state/types and convert only when calling the API?

hat/assets/js/apps/Iaso/domains/forms/components/FormActions.tsx
```diff
@@ -165,17 +185,17 @@ export const FormActions: FunctionComponent<Props> = ({
...
+                                        }).then(closeDialog)
                                     }
                                 />
                             </DisplayIfUserHasPerm>
                         </>
                     )}
                 </>
             )}
-            {settings.row.original.latest_form_version !== null && (
+            {settings.row.original.latest_form_version != null && (
```
hugo093: generally speaking, you should avoid using != and == in javascript. Behind the scene , it does some type conversion and we all know javascript is good and consistent at this (sarcasm) 

So fix it to use !== instead

hugo093 (changes requested):
a small detail, otherwise lgtm, gj 👍

hat/assets/js/apps/Iaso/domains/forms/components/FormActions.tsx
62496: Yes, thanks for the feedback. I was wondering which one I should use; I'd been told before not to use !=, but I didn't know why. Now I do, thanks!

hugo093 (approved):
lgtm 👍

PR #2922, opened 2026-04-16
[IA-4622] Payments tables do not update data

Crebert08:

On the payment lot page, the manual change of page number, results by page, don't result in a results list refresh. This PR is about fixing it

### Related JIRA tickets

[IA-4622](https://bluesquare.atlassian.net/browse/IA-4622)

## Changes
===== Frontend====
Added the defaultPageSize and isLoading props accuratly

===== Backend =======

`Views.py`
- Added `select_related("user__iaso_profile")` on the Payments Prefetch queryset to avoid 1 query per payment when getting the phone_number

- Added `select_related("org_unit")` on the OrgUnitChangeRequest queryset queryset to eliminate 1 query per change_request while checking who can see the change_requests

- Also added `select_related("created_by__iaso_profile", "task")` on the lot queryset to avoid 1 query per lot for each

`Serializers`
- In the `get_change_requests` function: changed `obj.change_requests.all()` to read from
  `prefetched_change_requests` (the in-memory list set by Django's prefetch machinery
  via `to_attr`). The previous call ignored the prefetch and was making 1 call per payment

- in the `NestedPaymentSerializer.get_can_see_change_requests`: removed the double serialization of the `change_requests`, and the double querying of `OrgChangeRequestNestedSerializer`. It now reads directly from the prefetched instances

- Finally in the PaymentLotSerializer's get_can_see_change_requests, I implemented a for loop that itirates the already fetched `obj.payments` and I removed the `values_list()` ignoring the prefetch

`Tests`
- Added a test to prove the 0(1) query count
- Added another one to confirm the API contract

## How to test

> Go to Payment lots table
> Change the number of results displayed in the tables
> It should work correctly

## Print screen / video

https://github.com/user-attachments/assets/7f517e3b-0e7e-45c4-97a7-c37af67df62c

Endpoint Calls went from this:
image

To this:
image

## Notes

## Doc

[IA-4622]: https://bluesquare.atlassian.net/browse/IA-4622?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/iaso/api/payments/serializers.py
+++ b/iaso/api/payments/serializers.py
@@ -64,8 +64,8 @@ class NestedPaymentSerializer(serializers.ModelSerializer):
         return OrgChangeRequestNestedSerializer(change_requests, many=True, context=self.context).data
 
     def get_can_see_change_requests(self, obj):
-        user = self.context.get("request").user
-        if user.is_superuser:
+        user = getattr(self.context.get("request", None), "user", None)
+        if getattr(user, "is_superuser", False):
             return True
         user_org_units = self.context.get("user_org_units")
         change_requests = getattr(obj, "prefetched_change_requests", None)
@@ -110,11 +110,7 @@ class PaymentLotSerializer(serializers.ModelSerializer):
             return True
         org_unit_ids = []
         for payment in obj.payments.all():
-            change_requests = getattr(payment, "prefetched_change_requests", None)
-            if change_requests is None:
-                change_requests = payment.change_requests.all()
-            for change_request in change_requests:
-                org_unit_ids.append(change_request.org_unit.id)
+            org_unit_ids.extend([cr.org_unit_id for cr in payment.prefetched_change_requests])
         return set(org_unit_ids).issubset(set(self.context["user_org_units"]))
 
     def get_payments(self, obj):
--- a/iaso/api/payments/serializers.py
+++ b/iaso/api/payments/serializers.py
@@ -4,7 +4,6 @@ from rest_framework import serializers
 from hat.audit.audit_logger import AuditLogger
 from hat.audit.models import PAYMENT_API, PAYMENT_LOT_API
 from iaso.api.payments.filters.potential_payments import filter_by_dates, filter_by_forms, filter_by_parent
-from iaso.api.payments.pagination import PaymentPagination
 from iaso.models import OrgUnitChangeRequest, Payment, PaymentLot, PotentialPayment
 from iaso.models.payments import PaymentStatuses
 from iaso.models.task import Task
@@ -100,7 +99,6 @@ class PaymentLotSerializer(serializers.ModelSerializer):
         ]
         read_only_fields = ["id", "created_at"]
 
-    pagination_class = PaymentPagination
     created_by = UserNestedSerializer()
     created_at = TimestampField(read_only=True)
 
@@ -127,7 +125,6 @@ class PotentialPaymentSerializer(serializers.ModelSerializer):
         fields = ["id", "user", "change_requests", "payment_lot", "can_see_change_requests"]
         read_only_fields = ["id", "payment_lot", "can_see_change_requests"]
 
-    pagination_class = PaymentPagination
     user = UserNestedSerializer()
 
     def get_change_requests(self, obj):
@@ -178,8 +175,6 @@ class PaymentSerializer(serializers.ModelSerializer):
             "payment_lot",
             "updated_by",
         ]
-
-    pagination_class = PaymentPagination
 
     def validate_status(self, status):
         if status not in PaymentStatuses:
--- a/iaso/api/payments/serializers.py
+++ b/iaso/api/payments/serializers.py
@@ -58,18 +58,20 @@ class NestedPaymentSerializer(serializers.ModelSerializer):
         read_only_fields = ["id"]
 
     def get_change_requests(self, obj):
-        change_requests = obj.change_requests.all()
+        change_requests = getattr(obj, "prefetched_change_requests", None)
+        if change_requests is None:
+            change_requests = obj.change_requests.all()
         return OrgChangeRequestNestedSerializer(change_requests, many=True, context=self.context).data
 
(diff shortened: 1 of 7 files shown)


beygorghor (approved):

iaso/api/payments/serializers.py
```diff
@@ -106,10 +108,14 @@ def get_can_see_change_requests(self, obj):
...
-        ).distinct()
-        return set(change_requests_org_units_for_lot).issubset(set(self.context["user_org_units"]))
+        org_unit_ids = []
+        for payment in obj.payments.all():
+            change_requests = getattr(payment, "prefetched_change_requests", None)
+            if change_requests is None:
+                change_requests = payment.change_requests.all()
+            for change_request in change_requests:
+                org_unit_ids.append(change_request.org_unit.id)
+        return set(org_unit_ids).issubset(set(self.context["user_org_units"]))
```
hugo093, line 112: i think you're a bit missing the point with your queryset here, reusing the `change_requests_prefetch` is not the way to go imho. 

Your optimization is tied to the payments, so better to keep the logic in that Prefetch : 

```python 
payments_prefetch = Prefetch(
            "payments",
            queryset=Payment.objects.select_related("user__iaso_profile").prefetch_related(
                Prefetch(
                    "change_requests",
                    queryset=OrgUnitChangeRequest.objects.select_related("org_unit"),
                    to_attr="prefetched_change_requests"
                )
            ),
        )

# remove change_requests_prefetch
queryset = queryset.prefetch_related(payments_prefetch)
```

and then instead of that block of code : 

```python 
  for payment in obj.payments.all():
      org_unit_ids.extend([change_request.org_unit_id for change_request in payment.prefetched_change_requests])
  return set(org_unit_ids).issubset(set(self.context["user_org_units"]))
```

Unfortunately that will imply to modify the other serializers

iaso/api/payments/serializers.py
```diff
@@ -58,18 +58,20 @@ class Meta:
...
-        change_requests = self.get_change_requests(obj)
-
-        blocked_change_requests = [
-            change_request for change_request in change_requests if change_request["can_see_change_request"] == False
-        ]
-        if blocked_change_requests:
-            return False
-        return True
+        user = self.context.get("request").user
+        if user.is_superuser:
```
hugo093: better be bulletproof there, just in case : 

```python 

user = getattr(self.context.get('request', None), "user", None)

if getattr(user, "is_superuser", False): 
   ...
```

hugo093 (changes requested):
A few changes to do otherwise good job debunking this and thanks for the tests👍 

A few small things that can improve this PR:

* change some serializer fields  (like the get_payments method that shouldn't be there) 
* try to be as swagger compliant as possible 
* remove unused / wrong code. Like pagination_class in a serializer ? 
* try to split the serializers by actions and put them in separate files
* try to split the get_queryset by action. 
* there are three viewsets in this file that are related to three endpoints, should be splitted in three files. 

I know it's not possible to always do all those, but even a small extra thing can improve the code and help the others that will work on this after.

iaso/api/payments/views.py
```diff
@@ -589,6 +601,7 @@ class PaymentsViewSet(ModelViewSet):
     http_method_names = ["patch", "get", "options"]
     results_key = "results"
     serializer_class = PaymentSerializer
+    pagination_class = PaymentPagination
```
hugo093, line 604: thanks 👍

iaso/api/payments/views.py
```diff
@@ -461,6 +472,7 @@ class PotentialPaymentsViewSet(ModelViewSet, AuditMixin):
     ordering = ["user__last_name"]
 
     serializer_class = PotentialPaymentSerializer
+    pagination_class = PaymentPagination
```
hugo093, line 475: thanks 👍

iaso/api/payments/views.py
```diff
@@ -102,6 +103,7 @@ class PaymentLotsViewSet(ModelViewSet):
     ordering = ["updated_at"]
     serializer_class = PaymentLotSerializer
     http_method_names = ["get", "post", "patch", "head", "options", "trace"]
+    pagination_class = PaymentPagination
```
hugo093, line 106: thanks 👍

iaso/api/payments/serializers.py
```diff
@@ -98,18 +99,17 @@ class Meta:
         ]
         read_only_fields = ["id", "created_at"]
 
-    pagination_class = PaymentPagination
```
hugo093, line 101: ❤️

iaso/api/payments/serializers.py
```diff
@@ -125,7 +125,6 @@ class Meta:
         fields = ["id", "user", "change_requests", "payment_lot", "can_see_change_requests"]
         read_only_fields = ["id", "payment_lot", "can_see_change_requests"]
 
-    pagination_class = PaymentPagination
```
hugo093, line 128: ❤️

iaso/api/payments/serializers.py
```diff
@@ -177,8 +176,6 @@ class Meta:
             "updated_by",
         ]
 
-    pagination_class = PaymentPagination
```
hugo093, line 180: ❤️

iaso/tests/api/payments/test_payment_lots.py
```diff
@@ -207,7 +207,7 @@ def test_retrieve_payment_lot_to_xlsx(self):
         )
         extra_change_request.new_reference_instances.set([self.instance1, self.instance2, self.instance3])
 
-        with self.assertNumQueries(10):
+        with self.assertNumQueries(11):
```
hugo093, line 210: this is likely due to the way retrieve queryset is coded, but out of scope so good enough

hugo093 (approved):
LGTM, thanks for the changes

PR #2923, opened 2026-04-17
Update of mobile app onboarding

lpontis:

Explain here in one sentence.

### Related JIRA tickets

IA-XXX, WC2-XXX, POLIO-XXX, POLIOPM-XXX SLEEP-XXX, SNT-XXX, CONSOLE-XXX

## Changes

## How to test

Explain how to test your PR.

## Print screen / video

## Notes

## Doc


quang-le:
https://github.com/BLSQ/iaso/pull/2926

PR #2924, opened 2026-04-17
[IA-4969]: new DynamicModelFieldSerializer

hugo093:

Our current DynamicFieldsModelSerializer was not swagger compliant and doing a bit of shadow IT by modifying the fields in the `__init__` method

### Related JIRA tickets

IA-4969

## Changes

* Implement new dynamic_field serializer and filters
* fields parameters now accepts a list, and no more a string field with comma separated values
* Adapted current backend endpoints
* Adapted related front ends

## How to test

The endpoints affected are : 

* /api/forms/
* /api/formversions/
* /api/datasources/sync/
* /api/group_sets/
* /api/orgunittypes/
* /api/v2/orgunittypes/
* /api/mappingversions/
* /api/orgunits/
* /api/polio/chronograms/tasks/
* /api/polio/chronograms/template_tasks/
* /api/polio/chronograms/
* /api/polio/budget/

They should be tested and checked that it's swagger compliant (go to /swagger-ui/ , check those endpoints, the fields parameters should be outputed as a multi choice field)

The impact on the front end related on those endpoints should be tested as well

## Notes

* I pushed to the wrong branch i know
* I feel like the dynamic_fields folder could be a library / another repo in itself
 
## Doc

Heavily inspired by [drf-flex-fields](https://github.com/openbook-education/drf-flex-fields2)


hat/assets/js/apps/Iaso/domains/forms/requests.ts
```diff
@@ -13,22 +14,19 @@ import { Form } from './types/forms';
...
-        }
-    } else if (appId) {
-        url += `/?app_id=${appId}`;
-    }
+
+    const params = {
+        ...(appId ? { app_id: appId } : {}),
+        ...(fields ? { fields: fields } : {}),
+    };
+    const url = `/api/forms/${formId}?${createSearchParamsWithArray(params).toString()}`;
```
tdethier: Shouldn't there be an extra `/` after `formId`? 
```suggestion
    const url = `/api/forms/${formId}/?${createSearchParamsWithArray(params).toString()}`;
```

iaso/tests/api/stocks/test_serializers_mobile.py
```diff
@@ -16,20 +16,20 @@
...
-        cls.account_2 = account_2 = m.Account.objects.create(name="Forbidden Account")
-        cls.org_unit_type_1 = org_unit_type_1 = m.OrgUnitType.objects.create(name="Org unit type 1")
-        cls.org_unit_type_2 = org_unit_type_2 = m.OrgUnitType.objects.create(name="Org unit type 2")
-        cls.org_unit_type_3 = m.OrgUnitType.objects.create(name="Org unit type 3")
-        cls.project_1 = project_1 = m.Project.objects.create(name="Project 1", account=account_1)
-        cls.project_2 = project_2 = m.Project.objects.create(name="Project 2", account=account_1)
-        cls.project_3 = m.Project.objects.create(name="Project 3", account=account_2)
+    def setUp(self):
+        super().setUp()
+        self.account_1 = account_1 = m.Account.objects.create(name="Test Account")
```
tdethier, line 21: Oh thank you for this 👍

plugins/polio/js/src/domains/Chronogram/api/useOptionChronogram.ts
```diff
@@ -1,7 +1,7 @@
 import { UseQueryResult } from 'react-query';
 
-import { optionsRequest } from '../../../../../../../hat/assets/js/apps/Iaso/libs/Api';
-import { useSnackQuery } from '../../../../../../../hat/assets/js/apps/Iaso/libs/apiHooks';
+import { optionsRequest } from 'Iaso/libs/Api';
+import { useSnackQuery } from 'Iaso/libs/apiHooks';
```
tdethier, line 4: 👍

plugins/polio/js/src/utils/index.tsx
```diff
@@ -176,3 +176,13 @@ export const findCampaignRound = (
 ): Round | undefined => {
     return campaign.rounds.find(rnd => rnd.number === round);
 };
+
+export const createSearchParamsWithArray = (params: Record<string, string | Array<any>>) => {
```
tdethier, line 180: Why don't you use the one in core?

iaso/api/forms.py
```diff
@@ -272,9 +275,16 @@ class FormsViewSet(ModelViewSet):
...
+    filter_backends = [DjangoFilterBackend, DynamicFieldsFilterBackend]
     EXPORT_ADDITIONAL_SERIALIZER_FIELDS = ("instance_updated_at", "instances_count")
     FORM_PK = "form_pk"
 
+    # @property
+    # def filter_backends(self):
+    #     if self.action in ["list"]:
+    #         return [DjangoFilterBackend, DynamicFieldsFilterBackend]
+    #     return [DjangoFilterBackend]
+
```
tdethier: I guess you can delete this?

dynamic_fields/filter_backends.py
```diff
@@ -0,0 +1,55 @@
+from django.conf import settings
+from django.core.exceptions import ValidationError
+from rest_framework.filters import BaseFilterBackend
+
+from dynamic_fields.serializer import DynamicFieldsModelSerializerMixin
+
+
+class DynamicFieldsFilterBackend(BaseFilterBackend):
```
tdethier, line 8: As discussed, I would maybe add a comment here to explain why this filtering class does not filter anything, for future devs 😅

tdethier (changes requested):
Some comments in any order:
- the feature works mostly fine, I can still select fields where the feature has been implemented AND I can see the available fields in swagger 👍 

Image

- except for the profile list endpoint: there's no mention of the `fields` param in swagger. The validation does not work either, you can enter any value you want and just don't get results if something is not valid

Image

Image

- the same goes for org units list: no mention of that field in swagger, but considering everything on that endpoint, I guess it's normal. However I was expecting an error if I pass invalid fields and it behaved like profiles.

- queries can be really long now (for instance, this is triggered when you go to the forms page: `http://localhost:8081/api/forms/?order=name&limit=50&page=1&fields=created_at&fields=name&fields=org_unit_types&fields=projects&fields=updated_at&fields=id&fields=form_id&fields=device_field&fields=location_field&fields=org_unit_type_ids&fields=project_ids&fields=period_type&fields=single_per_period&fields=periods_before_allowed&fields=periods_after_allowed&fields=latest_form_version&fields=updated_at&fields=deleted_at&fields=derived&fields=label_keys&fields=possible_fields&fields=legend_threshold&fields=has_mappings`) - we still have quite some margin before reaching the URL max size, but I can see us reaching it one day 😅 
- thanks for the cleanup, I appreciate it

--------------------------
Overall, I'm requesting changes because of the profiles issue, I think something is wrong there.

plugins/polio/js/src/utils/index.tsx
hugo093: cause if we move polio out as a separate app, i don't want to loose this :)

hat/assets/js/apps/Iaso/domains/forms/requests.ts
hugo093: not needed but i'll add it

tdethier (approved):
LGTM, let's merge this after the release 👍

PR #2925, opened 2026-04-17
SNT-408: Use delimiter from CSV when importing metric values

Bewi:

Sniff delimiter from input CSV when importing metric values

### Related JIRA tickets

SNT-408

## Changes

Define delimiter when reading CSV by first sniffing the CSV.

## How to test

Go to SNT Malaria, go to DataLayers, create at least one custom layer.
Export the template CSV which uses "," delimiter and try to import it directly, it is empty but it should work.
Now, modify it to use ";" delimiter and try to import it, it should work as well.

## Print screen / video

N/A

## Notes

N/A

## Doc

N/A

### DIFF
--- a/iaso/api/metrics/serializers.py
+++ b/iaso/api/metrics/serializers.py
@@ -1,3 +1,5 @@
+import csv
+
 import pandas as pd
 
 from django.utils.translation import gettext_lazy as _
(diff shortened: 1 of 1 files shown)


iaso/api/metrics/serializers.py
```diff
@@ -136,7 +138,9 @@ def validate_file(self, value):
         if not value.name.endswith(".csv"):
             raise serializers.ValidationError(_("The file must be a CSV."))
 
-        df = pd.read_csv(value)
+        dialect = csv.Sniffer().sniff(value.read(1024).decode("utf-8", errors="ignore"))
+        value.seek(0)  # Reset file pointer after reading for dialect inference
+        df = pd.read_csv(value, sep=dialect.delimiter)
```
michiwend, line 143: shouldn't this block be error-handled?

michiwend (changes requested):

michiwend (approved):

PR #2927, opened 2026-04-17
[IA-4108] Prepare .env.example + clean up

tdethier:

Prepare `.env.example` + clean up

### Related JIRA tickets

IA-4108

## Changes

- Gathered and organized env vars in `.env.example`
- Did some light clean up of old env vars that were no longer used

## How to test

- Clone the repository elsewhere, with another name `git clone <....ssh_or_https_link> <new_name_for_the_directory`
- Copy the new `.env.example`: `cp .env.example .env`
- Start your local setup `docker compose up`
- Everything should be working and the required env vars should be prepopulated
  - except maybe for the database creation

## Print screen / video

/

## Notes

- For some reason, IASO it still not reading directly `.env` files, I really want to introduce `dotenv` or `environs`

## Doc

/


.env.example
```diff
@@ -0,0 +1,152 @@
...
+# STATIC_URL=
+
+## AUTHENTICATION & SECURITY
+AUTHENTICATION_ENFORCED="false"
+CSRF_COOKIE_HTTPONLY=
+CSRF_COOKIE_SECURE=
+DISABLE_PASSWORD_LOGINS=
+DISABLE_SSL_REDIRECT="false"
+# ENABLE_CORS=
+ENCRYPTED_TEXT_FIELD_KEY="bNqPWD4KdouaH-yxfl7PPdJLbbNt31mnJDwg9IHoGDM="  # run `python -c "from cryptography.fernet import Fernet; print(Fernet.generate_key().decode())"` to generate a new key
```
hugo093: not a fan of having those keys in the .env.example

Even though it's some dummy key, the value shouldn't be there

.env.example
```diff
@@ -0,0 +1,152 @@
...
+# ============================================================
+# INTEGRATIONS
+# ============================================================
+## ClamAV (antivirus)
+CLAMAV_ACTIVE="false"
+# CLAMAV_FQDN=
+# CLAMAV_PORT=
+
+## Enketo
+ENKETO_API_TOKEN="AZE78974654azeAZE"
```
hugo093: not a fan of having those keys in the .env.example

Even though it's some dummy key, the value shouldn't be there

hugo093 (changes requested):
a few minor changes, otherwise works 

> I really want to introduce dotenv or environs

+1000

.env.example
tdethier: Yeah I just fetched the values that already exist in `docker-compose.yml`.

Considering that it's only for local development, I don't mind having them there, but I understand why you don't.

We could remove them, but it will require a bit more work when setting up IASO + updating the README

.env.example
```diff
@@ -0,0 +1,152 @@
...
+# AZURE_CUSTOM_DOMAIN=
+# AZURE_STORAGE_ACCOUNT_KEY=
+# AZURE_STORAGE_ACCOUNT_NAME=
+
+# ============================================================
+# UI & THEMING
+# ============================================================
+APP_TITLE="IASO"
+AVAILABLE_LANGUAGES="en,fr"
+# FAVICON_PATH=
```
quang-le, line 112: This is actually used. We use it for polio

.env.example
```diff
@@ -0,0 +1,152 @@
...
+FORUM_PATH=
+# HIDE_BASIC_NAV_ITEMS=
+# LEARN_MORE_URL=
+# LOGIN_LOGO_PATH=
+# LOGO_PATH=
```
quang-le, line 117: Same

.env.example
```diff
@@ -0,0 +1,152 @@
...
+# SHOW_NAME_WITH_LOGO=
```
quang-le, line 118: Used to show/hide app name in menu Drawer. defaults to True somehow, I think

.env.example
```diff
@@ -0,0 +1,152 @@
...
+# THEME_PRIMARY_BACKGROUND_COLOR=
+# THEME_PRIMARY_COLOR=
+# THEME_SECONDARY_COLOR=
```
quang-le, line 121: Used to override theme colors. used in polio

.env.example
```diff
@@ -0,0 +1,152 @@
...
+# CLAMAV_PORT=
+
+## Enketo
+ENKETO_API_TOKEN="AZE78974654azeAZE"
+ENKETO_DEV="true"
+ENKETO_SIGNING_SECRET="supersecret"
+ENKETO_URL="http://enketo:8005/"
+
+## Product Fruits (in-app guidance)
+PRODUCT_FRUITS_WORKSPACE_CODE=
```
quang-le, line 139: Is this still a thing?

quang-le (commented):
dunno if you want to document here what each var does, but left comments in that vein. All for adding dotenv as well

.env.example
tdethier: I have no idea, but we still have code that uses it, so I kept it

hugo093 (approved):
as discussed , lgtm 👍

PR #2929, opened 2026-04-20
[IA-4977]: Form create/edit view: disable API call to validation workflows if user don't have the required perms or FF

hugo093:

Form create/edit view: an API call to validation workflow would trigger whatever the users permissions or feature flag are and return potentially a 403

### Related JIRA tickets

IA-4977

## Changes

Created a custom validation workflow dropdown component that checks for users permissions and feature flag and trigger the call if so. 
Outputs a disabled combobox with an info tooltip next to it if users don't have the permissions or he doesn't have the FF.
Added tests

## How to test

Go to forms edit/create views with a user that doesn't have the validation workflow permissions or the FF, it shouldn't call the api endpoint and not raise a front end error.


hat/assets/js/apps/Iaso/domains/instances/validationWorkflow/components/ValidationWorkflowDropdown.tsx
```diff
@@ -0,0 +1,66 @@
...
+    const isDisabled = disabled || !hasPermission || !userHasFeatureFlag;
+
+    const inputComponent = (
+        <InputComponent
+            dataTestId={'validation-workflow-dropdown-input'}
+            type="select"
+            options={workflowOptions || []}
+            loading={isLoading}
+            disabled={isDisabled}
+            value={hasPermission && userHasFeatureFlag ? value : null}
```
quang-le: This is not ideal, we'd still want the user to see the selected workflow even if they can't edit it. Or if we're going to leave the dropdown empty, we might as well hide it

quang-le (changes requested):

quang-le (approved):

PR #2930, opened 2026-04-21
[IA-4988]: validation workflows: remove permission check on mobile

hugo093:

Validation workflow mobile endpoint was checking against the permissions when it shouldn't.

### Related JIRA tickets

IA-4988

## Changes

* Remove permission check
* Added feature flag check + custom message
* Added tests

## How to test

Call the endpoint mobile/validation-workflows with a user without perm and the right feature flag (SUBMISSION_VALIDATION_WORKFLOW), it shouldn't raise a 403.

### DIFF
--- a/iaso/api/validation_workflows/permissions.py
+++ b/iaso/api/validation_workflows/permissions.py
@@ -1,3 +1,4 @@
+from django.utils.translation import gettext_lazy as _
 from rest_framework.permissions import BasePermission
 
 from iaso.permissions.core_permissions import CORE_VALIDATION_WORKFLOW_PERMISSION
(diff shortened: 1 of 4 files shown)


iaso/api/validation_workflows/permissions.py
```diff
@@ -15,6 +16,8 @@ def has_permission(self, request, view):
 
 
 class HasAccountFeatureFlag(BasePermission):
+    message = _("This feature is disabled for your account.")
```
bmonjoie, line 19: ❤️

bmonjoie (approved):
LGTM

PR #2931, opened 2026-04-21
[IA-4984]: allow empty first name , last name and email on bulk create user

hugo093:

Bulk create users from csv would crash if a row would not contain first_name, last_name or email
### Related JIRA tickets

IA-4984

## Changes

* Allowing blank fields on user model by going from `validated_data[field]` to `validated_data.get(field, "") or ""`
* Added tests

## How to test

From users management screen, bulk create users with a csv that has no first_name, last_name or email in its rows.


Phil-V (approved):
LGTM! 👍

Small stylistic note: I would use `validated_data.get("last_name") or ""`  instead of  `validated_data.get("last_name", "") or ""`

PR #2934, opened 2026-04-21
hotfix: hide wf validation if no FF

quang-le:

Validation workflow menu entry should be hidden when feature flag not active

## Changes

- Add check on FF when building menu
- move the entry up one level

## How to test
- Load iaso with an account with or w/o the FF `SUBMISSION_VALIDATION_WORKFLOW` active, and see if the menu entry is hidden as expected

## Print screen / video
without FF
Screenshot 2026-04-21 at 15 03 18

with FF
Screenshot 2026-04-21 at 15 03 01

## Note 
can be cherry-picked on main for hotfix


Bewi (approved):
LGTM and WAE

PR #2935, opened 2026-04-21
[IA-5001]: fix csv detection in FileTypeValidator for bulk create user

hugo093:

Using a csv file without \n at the end would cause the bulk create user API to raise a 400 error with "Invalid file type"

### Related JIRA tickets

IA-5001

## Changes

Added plain/text as valid file types.
python-magic checks the signature of the file and .csv files don't have any. So to avoid any unreliable checks, I added this file type.

## How to test

Go to users management and try to create from a file without eol  like the one attached. It shouldn't raise an "Invalid file type" error
[bulk_create_user_csv_no_eof.csv](https://github.com/user-attachments/files/26934153/bulk_create_user_csv_no_eof.csv)

## Notes

Didn't put too much effort into the tests cause : 
* I couldn't make it work properly after 2 hours
* we might ditch .csv and switch to xlsx

### DIFF
--- a/iaso/api/bulk_create_users/serializers.py
+++ b/iaso/api/bulk_create_users/serializers.py
@@ -308,7 +308,7 @@ class BulkCreateUserSerializer(ModelSerializer):
             "file": {
                 "write_only": True,
                 "validators": [
-                    FileTypeValidator(allowed_mimetypes=["text/csv"]),
+                    FileTypeValidator(allowed_mimetypes=["text/csv", "text/plain"]),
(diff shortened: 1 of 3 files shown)


Phil-V (approved):
Can confirm the fix works 👍

Just as a note to be thorough (although it's not important, especially if we ditch csv), but FYI I couldn't quite fully replicate the original issue, except when using the file you provided (which did trigger the "Invalid file type"). I tried one of my own csv files and removed the trailing end characters, double-checked with a hex editor, and my file was accepted 🤷‍♂️ even before the fix.

PR #2939, opened 2026-04-22
POLIO-2131: Embedded links: user dropdown list is incomplete

beygorghor:

The dropdown list uses a paginated endpoint that doesn’t return the full list of available users

### Related JIRA tickets

POLIO-2131

## Changes

- Use of async select to display list of users
- Some tests on `UserAsyncSelect` and `UsersSelect`

## How to test

Edit or create a embedded page, you user should be preselected, you should be able to select any other user.

⚠️ There is a small bug on the `asyncSelect' from bluesquare-component, fixed with this [PR](https://github.com/BLSQ/bluesquare-components/pull/214)

## Print screen / video

https://github.com/user-attachments/assets/8fa12eb3-e362-4e64-a8b2-b9d1b2038b8e

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/pages/components/UsersSelect.tsx
+++ b/hat/assets/js/apps/Iaso/domains/pages/components/UsersSelect.tsx
@@ -76,7 +76,7 @@ export const UsersSelect = ({
                     <Chip
                         {...restTagProps}
                         key={String(option?.value ?? option?.label ?? index)}
-                        disabled
+                        disabled={isCurrentUser}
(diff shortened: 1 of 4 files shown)


hat/assets/js/apps/Iaso/domains/pages/components/UsersSelect.tsx
```diff
@@ -20,74 +23,80 @@ type Props = {
 export const UsersSelect = ({
     field,
     form,
-    label = '',
+    label: _label,
```
quang-le, line 26: This isn't defined anywhere

quang-le (approved):

PR #2940, opened 2026-04-22
SNT-436: Flag to show menu item if no permissions on menu item

Bewi:

If a menu entry has an empty array for permissions, it will never be shown (except for admins), accept a flag to override this behavior.

### Related JIRA tickets

SNT-436

## Changes

Add a ignorePermissions flag on menuItem to always show the menu item.

## How to test

With an account which is not Admin, take one of the nav that you see and modify the permission list of the menu item to be an empty array. 
The menu should disappear. 
add ignorePermissions: true to menuItem and verify that it nows appears.

## Print screen / video

N/A

## Notes

N/A

## Doc

N/A

### DIFF
--- a/hat/assets/js/apps/Iaso/constants/menu.tsx
+++ b/hat/assets/js/apps/Iaso/constants/menu.tsx
@@ -507,11 +507,11 @@ export const useMenuItems = (): MenuItems => {
             menuItemsTemp.push(admin as MenuItem);
         }
         const authorizedItems = menuItemsTemp.filter(menuItem => {
-            if (menuItem.ignorePermissions) {
+            const permissionsList = listMenuPermission(menuItem);
+            // If not permission set on the menuItem, we consider that everyone has access to it
+            if (permissionsList.length === 0) {
                 return true;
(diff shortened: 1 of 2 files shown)


beygorghor (approved):
We need to be carefull with this, it means now user with restricted permissions can access pipelines pages (only using the url) see hat/assets/js/apps/Iaso/constants/routes.tsx.
We have other empty list of permissions, for error page, not a problem, and on setupaccount path, but those are protected by a is staff or is admin. 
Can you create a ticket to put pipelines paths behind a specific permission please ?

Bewi:
> We need to be carefull with this, it means now user with restricted permissions can access pipelines pages (only using the url) see hat/assets/js/apps/Iaso/constants/routes.tsx. We have other empty list of permissions, for error page, not a problem, and on setupaccount path, but those are protected by a is staff or is admin. Can you create a ticket to put pipelines paths behind a specific permission please ?

Done here @beygorghor : https://bluesquare.atlassian.net/browse/IA-5017

PR #2943, opened 2026-04-22
IA-5009 prevent xlsform accident by previewing structural changes

mestachs:

If you make too big changes to xlsform between versions, it's creating more troubles to actually use exploit the data afterwards (pipeline, superset,...). 

We had several incident where a power user uploaded an "old" version without the previous changes (ex new questions that are lost)

### Related JIRA tickets

IA-5009

## Changes

add a preview endpoint to get info about structural diff between the version to be created and latest version and display the changes to the user for confirmation

## How to test

1. Seed command or launch setuper
2. Create a new form
3. Create a form version (see xlsform in the jira)
4. Create a second version with structural changes 
   - the dialog will show up with the differences
   - you can cancel 
   - the version isn't created
   - re-create a version with the same form
   - you click confirm
   - the version should be created
4. Create a third version without structural changes (just labels or new questions)
   - the dialog will not show up
   - the version is created
   

## Print screen / video

image


iaso/api/form_versions/views.py
```diff
@@ -93,3 +95,10 @@ def get_queryset(self):
         queryset = queryset.order_by(*orders)
 
         return queryset
+
+    @action(detail=False, methods=["post"], url_path="preview")
+    def preview(self, request, *args, **kwargs):
+        """Return a diff of questions added/removed compared to the latest version without saving."""
+        serializer = FormVersionPreviewSerializer(data=request.data, context=self.get_serializer_context())
```
hugo093: in order to be more swagger compliant, we should define it like so : 
```python 
serializer = self.get_serializer(data=request.data, context=self.get_serializer_context())
```

and at the top of your API viewset : 

```python
def get_serializer(self):
      if self.action == "preview":
           return FormVersionPreviewSerializer
return FormVersionSerializer
```

doing it like so will tell django-spectacular to use those serializer for request/response for the openAPI file :)

iaso/api/form_versions/views.py
hugo093, line 35: just to be sure, when i read this, i assume that the DELETE operation is not possible. However the ModelViewSet tells the contrary, can you confirm?

iaso/api/form_versions/serializers.py
```diff
@@ -160,3 +165,85 @@ def update(self, form_version, validated_data):
         form_version.updated_by = self.context["request"].user
         form_version.save()
         return form_version
+
+
+class FormVersionPreviewSerializer(serializers.Serializer):
+    """Validates the preview request and computes the diff without persisting anything."""
+
+    form_id = serializers.PrimaryKeyRelatedField(source="form", queryset=Form.objects.all())
+    xls_file = serializers.FileField(allow_empty_file=False)
```
hugo093: I just created some validator that you can use to check upon the filetype : FileTypeValidator. you can also use the builtin `from django.core.validators import FileExtensionValidator`

Also add required=True and write_only=True

iaso/api/form_versions/serializers.py
```diff
@@ -160,3 +165,85 @@ def update(self, form_version, validated_data):
         form_version.updated_by = self.context["request"].user
         form_version.save()
         return form_version
+
+
+class FormVersionPreviewSerializer(serializers.Serializer):
+    """Validates the preview request and computes the diff without persisting anything."""
+
+    form_id = serializers.PrimaryKeyRelatedField(source="form", queryset=Form.objects.all())
```
hugo093: our general codestyle is to set the queryset to Form.objects.none() and then restrict it in the __init__
Also the source seems unecessary here

also, it should have attributes required=True and write_only=True

iaso/api/form_versions/serializers.py
```diff
@@ -160,3 +165,85 @@ def update(self, form_version, validated_data):
...
+    xls_file = serializers.FileField(allow_empty_file=False)
+
+    def validate(self, data):
+        form = data["form"]
+
+        permission_checker = HasFormPermission()
+        if not permission_checker.has_object_permission(self.context["request"], self.context["view"], form):
+            raise serializers.ValidationError({"form_id": "Invalid form id"})
```
hugo093: this should be in a separate validate_form_id() method 
But it could also be automatically detected by providing the proper queryset in the __init__ method for that field

iaso/api/form_versions/serializers.py
```diff
@@ -160,3 +165,85 @@ def update(self, form_version, validated_data):
...
+
+        xls_file = data["xls_file"]
+        validation_errors = validate_xls_form(xls_file)
+        if validation_errors:
+            raise serializers.ValidationError({"xls_file": "", "xls_file_validation_errors": validation_errors})
```
hugo093: "xls_file": "" is unecessary here

hat/assets/js/apps/Iaso/domains/forms/requests.ts
```diff
@@ -49,6 +49,26 @@ export const updateForm = (formId, formData) =>
         openSnackBar(errorSnackBar('updateFormError', null, error));
     });
 
+export type FormVersionDiff = {
+    previous_version_id: string | null;
```
hugo093, line 53: use `previous_version_id?: string`

iaso/tests/api/form_versions/test_form_versions_preview.py
```diff
@@ -0,0 +1,298 @@
...
+from django.test import override_settings
+
+from iaso import models as m
+from iaso.permissions.core_permissions import CORE_FORMS_PERMISSION
+from iaso.test import APITestCase
+
+
+PREVIEW_URL = "/api/formversions/preview/"
+FIXTURE_V1 = "iaso/tests/fixtures/form_preview_diff_v1.xlsx"
+FIXTURE_V2 = "iaso/tests/fixtures/form_preview_diff_v2.xlsx"
```
hugo093: can you create a separate folder in fixtures for those files ? so we try to start organizing that folder a bit :)

iaso/tests/api/form_versions/test_form_versions_preview.py
```diff
@@ -0,0 +1,298 @@
...
+        """v2 vs v1 produces exactly 1 removed, 1 added, 1 modified question."""
+        self.client.force_authenticate(self.yoda)
+        with open(FIXTURE_V2, "rb") as xls_file:
+            response = self.client.post(
+                PREVIEW_URL,
+                data={"form_id": self.form_with_version.id, "xls_file": xls_file},
+                format="multipart",
+            )
+        self.assertJSONResponse(response, 200)
+        data = response.json()
```
hugo093:
```suggestion
        data = self.assertJSONResponse(response, 200)
```

iaso/tests/api/form_versions/test_form_versions_preview.py
```diff
@@ -0,0 +1,298 @@
...
+        """full_name and instanceID are identical in v1 and v2 → absent from all diff lists."""
+        self.client.force_authenticate(self.yoda)
+        with open(FIXTURE_V2, "rb") as xls_file:
+            response = self.client.post(
+                PREVIEW_URL,
+                data={"form_id": self.form_with_version.id, "xls_file": xls_file},
+                format="multipart",
+            )
+        self.assertJSONResponse(response, 200)
+        data = response.json()
```
hugo093:
```suggestion
        data = self.assertJSONResponse(response, 200)
```

iaso/tests/api/form_versions/test_form_versions_preview.py
```diff
@@ -0,0 +1,298 @@
...
+
+
+PREVIEW_URL = "/api/formversions/preview/"
+FIXTURE_V1 = "iaso/tests/fixtures/form_preview_diff_v1.xlsx"
+FIXTURE_V2 = "iaso/tests/fixtures/form_preview_diff_v2.xlsx"
+INVALID_XLS = "iaso/tests/fixtures/odk_form_blatantly_invalid.xlsx"
+
+
+@override_settings(MEDIA_ROOT=tempfile.mkdtemp())
+class FormVersionPreviewAPITestCase(APITestCase):
```
hugo093, line 45: can you add a test to see the  number of queries ? and maybe optimize if needed

iaso/api/form_versions/serializers.py
```diff
@@ -160,3 +165,85 @@ def update(self, form_version, validated_data):
...
+                                "old_type": old_type,
+                                "new_type": new_type,
+                            }
+                        )
+
+        return {
+            "previous_version_id": previous_version_id,
+            "removed_questions": removed_questions,
+            "added_questions": added_questions,
+            "modified_questions": modified_questions,
```
hugo093: to be more DRF idiomatic : those fields should be in the serializers itself, and the whole logic of compute diff should be in the related methods or the validate_x methods
You'll have to switch the logic a bit, but it will be more DRF standard
OR
You put all of that in the view action itself, since it's a process that depends on some input data , transforms it and output data. But you'll have to define two serializers (input and output) and make sure that the swagger generated is good enough. I think it's the cleanest option here :)

hugo093 (changes requested):
I only reviewed the backend code so far, I think it's already good to go through those comments before i dig further :) feel free to ask if you have any questions

iaso/tests/api/form_versions/test_form_versions_preview.py
mestachs: see test_preview_query_count

iaso/api/form_versions/views.py
mestachs: not sure what you mean here

https://github.com/BLSQ/iaso/blob/develop/iaso/api/form_versions/views.py#L36 
no delete http method name

haven't seen a delete/destroy or perhaps in the base one ?

iaso/api/form_versions/serializers.py
mestachs: I think it's relicat of the front end way of showing highlighted field
@beygorghor no ?

mestachs:
@hugo093 I think I made the changes required but perhaps good to look at it. Some comments I'm unsure if really did how you wanted it to be.

iaso/api/form_versions/serializers.py
mestachs: Finally understood a bit what you wanted, added some serializer to document the returned endpoint.

image

iaso/api/form_versions/views.py
hugo093: No my bad you're right :) since the delete method is not listed in http_method_names, drf doesn't generate the endpoint from the action destroy that is in the base class ModelViewSet. So all good

hugo093 (approved):
Front-end code could be improved a bit imho, a lot of useState, no tests. But maybe in another ticket as I think you're quite busy ;) Good job, thanks for the changes on the backend.

PR #2947, opened 2026-04-24
[IA-5000]: fix the impossibility to clear projects when editing a user

hugo093:

It was impossible to clear projects data when updating an user

### Related JIRA tickets

IA-5000

## Changes

Within the field, set the value to an empty array instead of undefined when empty field. 
Removed the allow_null=True from api as it made no  sense
JS tests

## How to test

Go to users management view, try to update a user with existing projects. Clear the project field, it should work and update the user so it doesn't have projects anymore


Bewi (approved):
LGTM

PR #2951, opened 2026-04-24
Fix/IA-4981 rename validation nodes to steps 

62496:

This PR makes the submission validation workflow UI less technical for end users by replacing “node” with “step”, and ensures the feature is only accessible when the related feature flag is enabled.

### Related JIRA tickets

IA-4981

## Changes

This PR updates the submission validation workflow UI wording in the front-end so users see “step” instead of “node” in the validation workflow details view and related actions.

It also aligns feature-flag protection across the feature:
- the validation submenu entry is displayed only when `SUBMISSION_VALIDATION_WORKFLOW` is enabled
- the submission validation routes are now also protected by the same feature flag, so direct URL access is blocked when the flag is disabled

The main changes are grouped as follows:
- validation workflow detail UI updates in `validationWorkflow/details/WorkflowConfiguration.tsx`
- shared message updates in `domains/instances/messages.ts`
- translation updates in `domains/app/translations/en.json` and `domains/app/translations/fr.json`
- route protection updates in `constants/routes.tsx`

## How to test

Use an account with the `VALIDATION_WORKFLOWS` permission.

1. Enable the `SUBMISSION_VALIDATION_WORKFLOW` feature flag on the account.
2. Go to `Forms > Submission > Validation > details`.
3. Check that the table section uses “step(s)” instead of “node(s)`.
4. Check that related UI labels also use “step”, for example:
   - delete confirmation
   - “Can skip previous steps”
5. Confirm that the `Validation > Submissions` menu entry is visible when the feature flag is enabled.
6. Disable the `SUBMISSION_VALIDATION_WORKFLOW` feature flag.
7. Confirm that the menu entry disappears.
8. Try opening the submission validation URLs directly and confirm access is blocked.


hugo093 (approved):
LGTM

PR #2952, opened 2026-04-24
IA-5012: Planning dropdowns should be displayed if user doesn't have the correct permissions

beygorghor:

Recently api has been changed to check planning permission. Frontend is still trying to fetch planning a bit everywhere in the app, we should hide those filters if user doesn’t have the permission

### Related JIRA tickets

IA-5012

## Changes

- generic component for dropdown planning
- frontend tests

## How to test

Make sure you have one user with a planning permission and one without.

Visit those pages and make sure:
- user with permissions can filter per planning
- user without doesn't have the planing dropdown AND the api call is not done

/dashboard/forms/stats/completenessStats
/dashboard/forms/list
/dashboard/forms/submissions/list

## Print screen / video

Screenshot 2026-04-24 at 16 06 00
Screenshot 2026-04-24 at 16 05 46
Screenshot 2026-04-24 at 15 53 45

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/plannings/hooks/requests/useGetPlannings.ts
+++ b/hat/assets/js/apps/Iaso/domains/plannings/hooks/requests/useGetPlannings.ts
@@ -81,12 +81,14 @@ const getPlanningsOptions = async (formIds?: string): Promise<Planning[]> => {
 };
 export const useGetPlanningsOptions = (
     formIds?: string,
+    enabled = true,
 ): UseQueryResult<DropdownOptions<number>[], Error> => {
     const queryKey: any[] = ['planningsList', formIds];
     return useSnackQuery({
         queryKey,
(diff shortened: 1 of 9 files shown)


hat/assets/js/apps/Iaso/domains/plannings/hooks/requests/useGetPlannings.ts
```diff
@@ -81,12 +81,14 @@ const getPlanningsOptions = async (formIds?: string): Promise<Planning[]> => {
 };
 export const useGetPlanningsOptions = (
     formIds?: string,
+    enabled = true,
```
quang-le: I would set the default to `false`, since we're using the permission to trigger the call, we should be restrictive by default

quang-le (changes requested):
small change in the default params of the api hook

quang-le (approved):

PR #2954, opened 2026-04-27
[IA-4882] Fixing the findIndex method check

Crebert08:

The Form submissions filter on the map tab in the org units details was returning duplicate. This PR is about fixing it.

### Related JIRA tickets

[IA-4882](https://bluesquare.atlassian.net/browse/IA-4882)

## Changes

findIndex() method returns `0` when the element is found at index `0` which  accidentally works and pushes a new instance. But when findIndex returns in index > 0, instead of pushing a new instance because the form already exists, it created a new form thus creating the duplicates. 
To fix this, I compared to `-1`. 

## How to test

> Go to org unit list,
> Go to details
> Go to map tab
> Filter the form submissions
> No duplicates should appear

## Print screen / video

## Notes

## Doc

[IA-4882]: https://bluesquare.atlassian.net/browse/IA-4882?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/forms/components/FormsFilterComponent.test.tsx
+++ b/hat/assets/js/apps/Iaso/domains/forms/components/FormsFilterComponent.test.tsx
@@ -1,0 +1,535 @@
+import React from 'react';
+import { screen, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { describe, it, expect, vi, beforeEach } from 'vitest';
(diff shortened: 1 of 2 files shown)


beygorghor (changes requested):
Well spotted !
Can you take the time to fix typings issues on this file please ? 
Also if you find th time to write some tests with vitest it can be a plus

beygorghor (approved):

PR #2956, opened 2026-04-27
POLIO-2125: Temporary Form A (backend)

quang-le:
Users need to be able to temporary fill in form A data, for when the campaign starts and the form hasn't been received yet.

They also need to switch the status and correct the numbers once the document has been received

### Related JIRA tickets

POLIO-2087, POLIO-2125
## Changes

- Updated Form A model in `plugins/polio/models/base.py`:
  - added `status` (`temporary` / `received`)
  - made `form_a_reception_date` nullable to support temporary forms
  - kept delayed lock logic derived from `created_at` + edit-window rules (no extra lock timestamp field)
- Added migration in `plugins/polio/migrations/0253_outgoingstockmovement_status_and_more.py`:
  - schema changes for `status` and nullable `form_a_reception_date`
  - data migration to mark existing records as `received`
- Updated Form A serializer/viewset logic in `plugins/polio/api/vaccines/stock_management.py`:
  - lifecycle compatibility validation:
    - temporary rejects `form_a_reception_date` and `file`
    - received requires `form_a_reception_date`
  - temporary edit-window allowlist enforcement (outside window only `status`, `form_a_reception_date`, `file`, `comment`)
  - temporary vials immutability enforcement with transition-aware behavior
  - transaction + pessimistic row locking (`select_for_update`) for race-safe updates
- Added/updated backend tests:
  - `plugins/polio/tests/test_vaccine_stock_management.py`
  - `plugins/polio/tests/api/vaccine_stock_management/test_outgoing_stock_movement_virus_scan.py`
  - lifecycle validation, edit-window behavior, transition rules, and null-date ordering regressions

## How to test

1. Apply migrations.
2. Call Form A endpoints under `/api/polio/vaccine/stock/outgoing_stock_movement/`.
3. **Create received Form A**:
   - omit `form_a_reception_date` -> expect `400`
   - provide `form_a_reception_date` -> expect `201`
4. **Create temporary Form A**:
   - include `file` or `form_a_reception_date` -> expect `400`
   - exclude both -> expect `201`
5. **Update temporary Form A (within edit window)**:
   - change `usable_vials_used` while staying temporary -> expect `400`
   - switch to received with required date (+ optional file) -> expect success
6. **Update temporary Form A (outside edit window)**:
   - update non-allowlisted field -> expect `400`
   - update only allowlisted completion fields -> expect success
7. **Concurrency check**:
   - simulate concurrent updates to same temporary record; verify row-lock path prevents inconsistent transition/save behavior.
8. Run backend tests for touched modules:
   - `plugins/polio/tests/test_vaccine_stock_management.py`
   - `plugins/polio/tests/api/vaccine_stock_management/test_outgoing_stock_movement_virus_scan.py`

If needed, test with users covering write/read-only permissions and records inside/outside `MANAGEMENT_DAYS_OPEN`.

## Print screen / video

N/A (backend-focused PR).

## Notes

- Backend remains the source of truth; frontend constraints are advisory only.
- Existing Form A consumers must tolerate nullable `form_a_reception_date` for temporary records.
- This PR does not add a dedicated lock timestamp field; lock timing remains deterministic from `created_at`.

## Doc

- Behavior is documented in code and tests:
  - `plugins/polio/models/base.py`
  - `plugins/polio/api/vaccines/stock_management.py`
  - `plugins/polio/tests/test_vaccine_stock_management.py`
  - `plugins/polio/tests/api/vaccine_stock_management/test_outgoing_stock_movement_virus_scan.py`


plugins/polio/api/vaccines/stock_management.py
```diff
@@ -432,6 +561,21 @@ def get_queryset(self):
...
+        with transaction.atomic():
+            locked_instance = self.get_queryset().select_for_update().get(**{self.lookup_field: lookup_value})
+            self.check_object_permissions(request, locked_instance)
+
+            serializer = self.get_serializer(locked_instance, data=request.data, partial=partial)
+            serializer.is_valid(raise_exception=True)
+
+            self.perform_update(serializer)
+
+        return Response(serializer.data)
```
hugo093: this is basically rewriting the inherited update just to add select_for_update in the queryset.
While it works, it adds code that is done in a better way in inherited update method of drf viewset. 

See the super().update() code : 

```python 

    def update(self, request, *args, **kwargs):
        partial = kwargs.pop('partial', False)
        instance = self.get_object()
        serializer = self.get_serializer(instance, data=request.data, partial=partial)
        self.perform_update(serializer)

        if getattr(instance, '_prefetched_objects_cache', None):
            # If 'prefetch_related' has been applied to a queryset, we need to
            # forcibly invalidate the prefetch cache on the instance.
            instance._prefetched_objects_cache = {}

```

Solution is to adapt your get_queryset method like : 

```python
def get_queryset(self):
      qs = base_qs
      if self.action in ["update", "partial_update"]:
            qs = qs.select_for_update()
      return qs
```

If you really want to keep the transaction atomic : 

```python
@transaction.atomic
def update(...):
       return super().update()
```

plugins/polio/api/vaccines/stock_management.py
```diff
@@ -369,6 +488,16 @@ def update(self, instance, validated_data):
...
         if campaign:
             instance.campaign = campaign
+        next_status = validated_data.get("status", instance.status)
+        is_received_to_temporary = (
+            instance.status == OutgoingStockMovement.StatusChoices.RECEIVED
+            and next_status == OutgoingStockMovement.StatusChoices.TEMPORARY
+        )
+        if is_received_to_temporary:
+            if instance.file:
+                instance.file.delete(save=False)
```
hugo093: curious there to know why you need to explicitly delete it ? is it because you want to delete the file in the media storage ?

plugins/polio/api/vaccines/stock_management.py
```diff
@@ -369,6 +488,16 @@ def update(self, instance, validated_data):
...
+            validated_data["file"] = None
+            validated_data["form_a_reception_date"] = None
         self.scan_file_if_exists(validated_data, instance)
```
hugo093: should be wrapped in an else: statement. Otherwise you might scan for a file that is sure to not exist

plugins/polio/api/vaccines/stock_management.py
```diff
@@ -295,17 +298,28 @@ def compute_category_from_campaign(campaign: Optional[Campaign], round: Optional
 
 
 class OutgoingStockMovementSerializer(ModelWithFileSerializer):
+    EDIT_ACCESS_NONE = "none"
```
hugo093: on a general note, could you move this to a serializers.py file ? it would increase readability

plugins/polio/api/vaccines/stock_management.py
```diff
@@ -369,6 +488,16 @@ def update(self, instance, validated_data):
         campaign = self.extract_campaign_data(validated_data)
```
hugo093: seeing the code there, why didn't we make campaign as a SlugRelatedField(slug_field="obr_name") while restricitng the queryset in the __init__

plugins/polio/api/vaccines/stock_management.py
```diff
@@ -318,31 +332,136 @@ class Meta:
...
+            return self.EDIT_ACCESS_FULL
+        # Completion-only: temporary Form A past the edit window stays editable for
+        # completion (TEMPORARY_FORM_A_COMPLETION_FIELDS only; enforced server-side).
+        if obj.status == OutgoingStockMovement.StatusChoices.TEMPORARY and user.has_perm(
+            POLIO_VACCINE_STOCK_MANAGEMENT_READ_PERMISSION.full_name()
+        ):
+            return self.EDIT_ACCESS_COMPLETION_ONLY
+        return self.EDIT_ACCESS_NONE
+
+    def get_within_edit_window(self, obj):
```
hugo093: add drf spectacular annotation to tell that this field is boolean 

```python 
 @extend_schema_field(serializers.BooleanField(read_only=True))
```

plugins/polio/api/vaccines/stock_management.py
```diff
@@ -318,31 +332,136 @@ class Meta:
...
         validated_data = super().validate(data)
         return validated_data
 
     def get_round_number(self, obj):
         return obj.round.number if obj.round else None
 
-    def get_can_edit(self, obj):
-        return can_edit_helper(
-            self.context["request"].user,
+    def get_edit_access(self, obj):
```
hugo093: same, add some annotation with drf spectacular to tell openAPI the type

plugins/polio/api/vaccines/stock_management.py
```diff
@@ -318,31 +332,136 @@ class Meta:
...
+
     def validate(self, data):
         # The `source` attribute is used as the key in `data` instead of the name of the serializer field.
-        if data.get("campaign", None) and data.get("non_obr_name", None):
+        current_campaign = self.instance.campaign if self.instance else None
+        current_non_obr_name = self.instance.non_obr_name if self.instance else None
+        campaign_value = data.get("campaign", current_campaign)
+        non_obr_name_value = data.get("non_obr_name", current_non_obr_name)
+        if campaign_value and non_obr_name_value:
             raise serializers.ValidationError({"error": "campaign and alternative campaign cannot both be defined"})
```
hugo093: this is kind of weird as you might have errors triggered on existing data and not the input data. The user might get confused ? 
Also why would this happen ? Shouldn't we clean up the old data

plugins/polio/api/vaccines/stock_management.py
```diff
@@ -318,31 +332,136 @@ class Meta:
...
             "alternative_campaign",
             "campaign_category",
             "doses_per_vial",
         ]
 
+    def _should_validate_lifecycle(self):
+        view = self.context.get("view")
+        action = getattr(view, "action", None)
+        write_actions = {"create", "update", "partial_update"}
+        return action in write_actions or action is None
```
hugo093: this is viewsets logic, so it should be in the viewsets using get_serializer_context

plugins/polio/api/vaccines/stock_management.py
```diff
@@ -318,31 +332,136 @@ class Meta:
...
+            current_status = self.instance.status if self.instance else None
+            current_form_a_reception_date = self.instance.form_a_reception_date if self.instance else None
+            current_file = self.instance.file if self.instance else None
+
+            status_value = data.get("status", current_status)
+            is_received_to_temporary = (
+                self.instance is not None
+                and current_status == OutgoingStockMovement.StatusChoices.RECEIVED
+                and status_value == OutgoingStockMovement.StatusChoices.TEMPORARY
+            )
```
hugo093: seeing this, we should split this serializer into multiple ones: createSerializer, updateSerializer , ... so it will be clearer which one needs which logic (in that case self.instance is not None means that only the update serializer will be affected)

plugins/polio/api/vaccines/stock_management.py
```diff
@@ -318,31 +332,136 @@ class Meta:
...
+                uploaded_file = None
+            # Validation precedence:
+            # 1) lifecycle compatibility (temporary vs received field rules)
+            # 2) temporary vials immutability
+            # 3) temporary post-window completion-field allowlist
+            self._validate_editable_fields_based_on_status(status_value, form_a_reception_date, uploaded_file)
+            self._enforce_temporary_vials_immutability(data, status_value)
+            self._validate_temporary_after_window_allowed_fields(status_value)
+
         validated_data = super().validate(data)
```
hugo093: it should be at the top ?

plugins/polio/api/vaccines/stock_management.py
```diff
@@ -318,31 +332,136 @@ class Meta:
...
+                # Transition normalization: received -> temporary clears reception metadata.
+                form_a_reception_date = None
+                uploaded_file = None
+            # Validation precedence:
+            # 1) lifecycle compatibility (temporary vs received field rules)
+            # 2) temporary vials immutability
+            # 3) temporary post-window completion-field allowlist
+            self._validate_editable_fields_based_on_status(status_value, form_a_reception_date, uploaded_file)
+            self._enforce_temporary_vials_immutability(data, status_value)
+            self._validate_temporary_after_window_allowed_fields(status_value)
```
hugo093: same remark, most of this logic is only valid if we have an instance (and so update action only)

plugins/polio/tests/test_vaccine_stock_management.py
```diff
@@ -118,6 +118,500 @@ def test_vaccine_stock_management_permissions_outgoing_stock_movement(self):
...
+            "doses_per_vial": 20,
+        }
+
+        response = self.client.post(
+            f"{BASE_URL_SUB_RESOURCES}outgoing_stock_movement/",
+            payload_temp_form_a,
+            format="json",
+        )
+        data = self.assertJSONResponse(response, 201)
+        self.assertEqual(data["status"], "temporary")
```
hugo093: should use your textchoice value

plugins/polio/tests/test_vaccine_stock_management.py
```diff
@@ -118,6 +118,500 @@ def test_vaccine_stock_management_permissions_outgoing_stock_movement(self):
...
+            **payload_temp_form_a,
+            "form_a_reception_date": "2024-01-02",
+        }
+        response = self.client.post(
+            f"{BASE_URL_SUB_RESOURCES}outgoing_stock_movement/",
+            payload_temp_form_a_with_date,
+            format="json",
+        )
+        data = self.assertJSONResponse(response, 400)
+        self.assertIn("form_a_reception_date", data)
```
hugo093: might be worth checking the message error with self.assertHasError

plugins/polio/tests/test_vaccine_stock_management.py
```diff
@@ -118,6 +118,500 @@ def test_vaccine_stock_management_permissions_outgoing_stock_movement(self):
...
+        with open("plugins/polio/tests/fixtures/virus_scan/safe_file.pdf", "rb") as safe_file:
+            response = self.client.post(
+                f"{BASE_URL_SUB_RESOURCES}outgoing_stock_movement/",
+                {
+                    **payload_temp_form_a,
+                    "file": safe_file,
+                },
+                format="multipart",
+            )
+        data = self.assertJSONResponse(response, 400)
```
hugo093, line 184: same remark

plugins/polio/api/vaccines/stock_management.py
```diff
@@ -318,31 +332,136 @@ class Meta:
...
+        )
+
+    def _enforce_temporary_vials_immutability(self, data, resulting_status):
+        if not self.instance or self.instance.status != OutgoingStockMovement.StatusChoices.TEMPORARY:
+            return
+
+        if resulting_status != OutgoingStockMovement.StatusChoices.TEMPORARY:
+            return
+
+        if "usable_vials_used" in data and data["usable_vials_used"] != self.instance.usable_vials_used:
```
hugo093: if data.get(key, default) != ..

plugins/polio/api/vaccines/permissions.py
```diff
@@ -91,8 +112,13 @@ def has_object_permission(self, request, view, obj):
                     return (
                         True  # There are multiple objects in one request for those so this is checked in the serializer
                     )
-                one_week_ago = self.datetime_now_today() - datetime.timedelta(days=self.days_open)
-                return getattr(obj, self.datetime_field) >= one_week_ago
+                if is_within_edit_window(
+                    getattr(obj, self.datetime_field),
+                    days_open=self.days_open,
+                    now=self.datetime_now_today(),
```
hugo093: is there really any useful thing about passing this datetime_now_today into the function ? searchin into the code, it seems that VaccineStockPermission is never initialized with a special datetime_now_today, so it defaults to timezone.now() which is also the default in is_within_edit_window

plugins/polio/api/vaccines/permissions.py
```diff
@@ -7,6 +7,31 @@
...
+        now = timezone.now()
+    return the_date >= now - datetime.timedelta(days=days_open)
+
+
+def is_temporary_form_a_completion_edit(request, obj):
+    if request.method not in ["PUT", "PATCH"]:
+        return False
+
+    if obj.__class__.__name__ != "OutgoingStockMovement":
+        return False
```
hugo093: 100% we will forget to update this if we change the class name :) 
If it's really specific  to OutgoingStockMovement and its related viewset, maybe it would be worth creating a separate permission just for it ? 

Overall seems like this whole code started as something common where we added layers upon layers. Now it's the time to split it :)

plugins/polio/api/vaccines/permissions.py
```diff
@@ -7,6 +7,31 @@
...
+
+    if getattr(obj, "status", None) != "temporary":
```
hugo093: general remark : use the textchoices you defined in your model instead of temporary => if we change the value , it will change everywhere automatically

plugins/polio/tests/test_vaccine_stock_management.py
```diff
@@ -118,6 +118,500 @@ def test_vaccine_stock_management_permissions_outgoing_stock_movement(self):
         response = self.client.delete(f"{BASE_URL_SUB_RESOURCES}outgoing_stock_movement/{osm_id}/")
         self.assertEqual(response.status_code, 204)
 
+    def test_outgoing_stock_movement_temporary_lifecycle_validation(self):
```
hugo093, line 122: no test for assertNumQueries ?

hugo093 (changes requested):
Didn't test it yet, just read the code. Overall lot of reorganization and changes are needed. I also find the code hard to read, but it's certainly because the business needs are complex.

plugins/polio/api/vaccines/stock_management.py
quang-le: yes

plugins/polio/api/vaccines/stock_management.py
quang-le: TBH I don't know and I don't dare changing it

plugins/polio/api/vaccines/stock_management/destructions/serializers.py
```diff
@@ -0,0 +1,49 @@
...
+
+class DestructionReportSerializer(ModelWithFileSerializer):
+    file = serializers.FileField(required=False)
+    can_edit = serializers.SerializerMethodField()
+
+    class Meta:
+        model = DestructionReport
+        exclude = ["file_last_scan", "file_scan_status"]
+
+    def get_can_edit(self, obj):
```
hugo093, line 23: add drf-spectacular annotation to say it's a boolean (@extend_schema_field(serializers.BooleanField)

Same remark goes for the whole code and other get_xxx() methods

plugins/polio/api/vaccines/stock_management/destructions/serializers.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def to_representation(self, instance):
+        ret = super().to_representation(instance)
+        if instance.file:
+            ret["file"] = {
+                "path": instance.file.url,
+                "name": os.path.basename(instance.file.name),
+            }
+        else:
+            ret["file"] = None
```
hugo093, line 48: if you aim for that, it should be a different file serializer field than the native one. (or another nested serializer )

Same remarks goes for the other serializers

plugins/polio/api/vaccines/stock_management/destructions/views.py
```diff
@@ -0,0 +1,59 @@
...
+        )
+    ]
+    ordering_fields = ["doses_per_vial"]
+
+    @action(detail=False, methods=["GET"])
+    def check_duplicate(self, request):
+        vaccine_stock_id = request.query_params.get("vaccine_stock")
+        destruction_report_date = request.query_params.get("destruction_report_date")
+        unusable_vials_destroyed = request.query_params.get("unusable_vials_destroyed")
+        destruction_report_id = request.query_params.get("destruction_report_id")
```
hugo093, line 35: should be a filter ( from django_filters.rest_framework import DjangoFilterBackend ) so you can move all the logic and queryset filtering there

plugins/polio/api/vaccines/stock_management/outgoingstockmovement/serializers/base.py
```diff
@@ -0,0 +1,118 @@
...
+            "round_number",
+            "edit_access",
+            "within_edit_window",
+            "alternative_campaign",
+            "campaign_category",
+            "doses_per_vial",
+        ]
+
+    def get_round_number(self, obj):
+        return obj.round.number if obj.round else None
```
hugo093, line 73: can be replaced by a proper serializer ? with source="round.number"

plugins/polio/api/vaccines/stock_management/earmarks/serializers.py
```diff
@@ -0,0 +1,95 @@
...
+        round_obj = None
+
+        campaign_data = self.initial_data.get("campaign", None)
+        round_number = self.initial_data.get("round_number", None)
+
+        if campaign_data:
+            campaign = Campaign.objects.get(
+                obr_name=campaign_data,
+                account=self.context["request"].user.iaso_profile.account,
+            )
```
hugo093, line 88: the more i read that and the get_campaign method above, the more i think we should clean this up someday

hugo093 (approved):
overall lgtm, maybe create a ticket for the detail comments i have here and there

PR #2962, opened 2026-04-28
[IA-4987] bulk import user: add xlsx file

hugo093:

Bulk import user was only accepting .csv files, excel support was asked.

### Related JIRA tickets

IA-4987

## Changes

* Adapt backend with a clean split into mixin to handle multiple file types
* Refactor the test base to avoid writing same tests for csv and xlsx
* Implement front end , moving the modal to another page
* Added tests

## How to test

Go to users management, click the create from file button. 
You should end up on a new page where you can fill the form and submit. 
It should work both  with csv and xlsx
Upon successful submit, you should be redirected to user management screen


iaso/migrations/0384_rename_bulkcreateusercsvfile_bulkcreateuserfile.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 4.2.28 on 2026-04-27 12:21
+
+from django.conf import settings
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
```
tdethier, line 7: I'm afraid you'll need to regenerate your migration file, many were added in the meantime 😅

iaso/migrations/0381_merge_20260428_1426.py
```diff
@@ -0,0 +1,12 @@
+# Generated by Django 4.2.28 on 2026-04-28 14:26
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
```
tdethier, line 4: And this will allow you to get rid of this one

tdethier (approved):
I managed to create users with CSV and XLSX files 👍 

Some comments:
- I really like the new page, it's better than a modal
- I would rebase and recreate migration files in order to reduce the number of migrations
- it worked, but it was quite slow, it took me 25 seconds to create 1 user from CSV, I can't imagine how long it would take for dozens or hundreds of users 😅 maybe we'll need to put this in an async task in the future

Image

hugo093:
> * it worked, but it was quite slow, it took me 25 seconds to create 1 user from CSV, I can't imagine how long it would take for dozens or hundreds of users 😅 maybe we'll need to put this in an async task in the future

I agree, we should move this to an async task

PR #2963, opened 2026-04-28
IA-4979 fix: remove id in user role

62496:

This PR removes the account ID prefix from role names in the `Roles required` column on the validation workflow detail screen.

### Related JIRA tickets

IA-4979

## Changes

This change is limited to the validation workflow detail retrieve endpoint.

- Replaced the direct `group.name` serialization with a computed value that removes the `{account_id}_` prefix.

Expected behavior:
- Before: `1_data manager`
- After: `data manager`

## How to test

1. Start the app locally.
2. Open `/dashboard/forms/submissions/validation/detail/`.
3. Check the `Roles required` column in the steps table.
4. Confirm that the role name is displayed without the account ID prefix.

## Print screen / video
image

## Notes

- Scope intentionally kept limited to the workflow detail screen API.
- Other endpoints that may expose raw `group.name` are out of scope for this PR.

## Doc

No additional documentation required.

### DIFF
--- a/iaso/api/validation_workflows/serializers/retrieve.py
+++ b/iaso/api/validation_workflows/serializers/retrieve.py
@@ -7,11 +7,12 @@ from iaso.models import Form, UserRole, ValidationNodeTemplate, ValidationWorkfl
 
 
 class NestedRolesRequiredSerializer(ModelSerializer):
-    name = serializers.CharField(read_only=True, source="group.name")
-
+    name = serializers.SerializerMethodField()
     class Meta:
         model = UserRole
         fields = ["name", "id"]
+    def get_name(self, obj):
+        return obj.group.name.removeprefix(f"{obj.account_id}_")
 
 
 class NestedValidationNodeTemplateSerializer(ModelSerializer):
--- a/iaso/api/validation_workflows/serializers/retrieve.py
+++ b/iaso/api/validation_workflows/serializers/retrieve.py
@@ -8,9 +8,11 @@ from iaso.models import Form, UserRole, ValidationNodeTemplate, ValidationWorkfl
 
 class NestedRolesRequiredSerializer(ModelSerializer):
     name = serializers.SerializerMethodField()
+
     class Meta:
         model = UserRole
         fields = ["name", "id"]
+
     def get_name(self, obj):
         return obj.group.name.removeprefix(f"{obj.account_id}_")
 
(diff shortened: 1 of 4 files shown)


iaso/api/validation_workflows/serializers/retrieve.py
```diff
@@ -7,12 +7,15 @@
...
 class NestedRolesRequiredSerializer(ModelSerializer):
-    name = serializers.CharField(read_only=True, source="group.name")
+    name = serializers.SerializerMethodField()
 
     class Meta:
         model = UserRole
         fields = ["name", "id"]
 
+    def get_name(self, obj):
+        return obj.group.name.removeprefix(f"{obj.account_id}_")
```
hugo093, line 17: it works but : 

* it's always better in the case of SerializerMethodField to tell drf-spectacular the exact type for openAPI. You can annotate the method like : 
```python 
@extend_schema_field(serializers.CharField())
    def get_name(self, obj):
```

Also, there s the same issue in : 
* `api/validation_workflows_node_templates/serializers/retrieve.py`
* `api/validation_workflows_node_templates/serializers/list.py`
* `api/validation_workflows_instances/serializers.py`

From there you have two options : 

1. Copy paste your code in those serializers (quick fix that we will do now)

2. be more modular and future proof oriented :) (tbc with @tdethier  and create a ticket for this ? )

* Creating a `GroupNameSerializerField` that does what you've done there. putting it under `api/common/serializer_fields.py`  so we can reuse it everywhere if needed and have only one place to change (you could even be more generic by using a different name like `SerializerWithoutPrefixField` or whatever)
* Creating a UserRoleSerializer under `api/common/serializer.py` which is that serializer but using the above GroupNameSerializer. 
* Adding tests for those two new classes :) 
 
That way we avoid duplicated code (and different errors) and we have one source of "truth". E.g there s the same kind of code in `profiles/serializers/list.py`

hugo093 (changes requested):
few other places to change

iaso/api/validation_workflows/serializers/retrieve.py
62496: Thanks for the feedback, I appreciate it 👍. I’ll just copy-paste it for now, as suggested.

hugo093 (approved):
LGTM 👍 can you create a ticket about unifying this kind of behavior ? (cfr discussion)

62496:
okay, I'll do it 👍

PR #2966, opened 2026-04-29
[IA-4996]: validation workflows, fix next bypasses steps in case of resubmit

hugo093:

In validation workflows, when resubmitting, the next bypass steps were not displayed correctly again. 

### Related JIRA tickets

IA-4996 , IA-5007

## Changes

After discussing with @quang-le , we agreed that we should redo the backend as there was way too much post processing on the front end. 

So backend has been redone to more fit the needs of the front. At the same time, work related to IA-5007 has been done (props to Elie for the great mockup), hence a new UI for the timeline

* Redo backend + optimization
* Backend tests
* Redo front end + tests
* Fixed a small issue in color picker in case of null initial value

## How to test

Create a validation workflows 
Submit an instance that triggers that VF 
Do whatever you like next, everything should work.

You can use this python script to help you 

```python 

import os
import django
os.environ.setdefault("DJANGO_SETTINGS_MODULE", "hat.local_settings")


django.setup()

from iaso.engine.validation_workflow import ValidationWorkflowEngine
from iaso.models import ValidationWorkflow, Instance
import time


workflow_template_slug = "test"
user = get_user_model().objects.get(username="axkoefm")
instance = Instance.objects.get(id="259")

# uncomment if you want a fresh new data, otherwise it will resubmit if possible
# instance.validationnode_set.all().delete()

workflow_template = ValidationWorkflow.objects.get(slug=workflow_template_slug)

ValidationWorkflowEngine.start(workflow_template, user, instance)

```
## Print screen / video

image


iaso/api/validation_workflow_instances/serializers.py
```diff
@@ -74,26 +74,59 @@ class Meta:
...
+                    .filter(can_skip_previous_nodes=True, workflow=obj.form.validation_workflow)
+                    .filter(Q(active_nodes__isnull=True) | Q(active_nodes__status=ValidationNodeStatus.UNKNOWN))
+                )
+            else:
+                qs = (
+                    ValidationNodeTemplate.objects.prefetch_related(
+                        "validationnode_set", Prefetch("roles_required", UserRole.objects.select_related("group"))
+                    )
+                    .filter(can_skip_previous_nodes=True, workflow=obj.form.validation_workflow)
+                    .filter(Q(validationnode__isnull=True) | Q(validationnode__status=ValidationNodeStatus.UNKNOWN))
```
hugo093: spotted a bug : we never check ValidationNode upon the current instance : 

* active_nodes should be checked upon the instance 
* validationnode__isnull => that's wrong , it could be that some validation node related to another instance exist. 

will change it later to : 

```python 

if most_recent_new_version:
    # in case of resubmission, we need to filter out previous nodes
    qs = (
        ValidationNodeTemplate.objects.annotate(
            active_nodes=FilteredRelation(
                "validationnode",
                condition=Q(validationnode__created_at__gte=most_recent_new_version.created_at, validationnode__instance=obj),
            )
        )
        .prefetch_related(
            "validationnode_set", Prefetch("roles_required", UserRole.objects.select_related("group"))
        )
        .filter(can_skip_previous_nodes=True, workflow=obj.form.validation_workflow)
        .filter(Q(active_nodes__isnull=True) | Q(active_nodes__status=ValidationNodeStatus.UNKNOWN))
    )
else:
    qs = (
        ValidationNodeTemplate.objects
        .prefetch_related(
            "validationnode_set",
            Prefetch("roles_required", UserRole.objects.select_related("group"))
        )
        .filter(
            can_skip_previous_nodes=True,
            workflow=obj.form.validation_workflow,
        )
        .filter(
            Q(
                ~Exists(
                    ValidationNode.objects.filter(
                        instance=obj,
                        node=OuterRef("pk"),
                    )
                )
            )
            |
            Q(
                Exists(
                    ValidationNode.objects.filter(
                        instance=obj,
                        node=OuterRef("pk"),
                        status=ValidationNodeStatus.UNKNOWN,
                    )
                )
            )
        )
    )

```

iaso/api/validation_workflow_instances/serializers.py
```diff
@@ -74,26 +74,83 @@ class Meta:
...
 
     @extend_schema_field({"type": "string", "nullable": True})
     def get_rejection_comment(self, obj):
         if obj.general_validation_status == ValidationWorkflowArtefactStatus.REJECTED:
+            if getattr(obj, "all_validation_nodes", None):
+                rejected_validation_node = next(
+                    (n for n in obj.all_validation_nodes if n.status == ValidationNodeStatus.REJECTED), None
+                )
+                return getattr(rejected_validation_node, "comment", "")
             return obj.validationnode_set.filter(status=ValidationNodeStatus.REJECTED).first().comment
```
quang-le: This will throw if `.first()` returns None. 
I had the case with an instance of a form that had a rejected status from an old workflow. It's an edge case/misconfiguration for now though, but maybe stuff to keep in mind

quang-le: This will throw if `.first()` returns None. 
I had the case with an instance of a form that had a rejected status from an old workflow. It's an edge case/misconfiguration for now though, but maybe stuff to keep in mind

hugo093:
After discussing with @quang-le I'll rewrite the backend completly so it returns what the front expect

hat/assets/js/apps/Iaso/domains/instances/components/ValidationWorkflow/ValidationModal.tsx
```diff
@@ -1,63 +1,112 @@
...
+    instanceId,
+    nodeId,
+    closeDialog,
+    isOpen,
+}: ModalProps) => {
+    const { formatMessage } = useSafeIntl();
+    const { mutateAsync: save } = useCompleteNode({ instanceId, nodeId });
+    const validationSchema = useNodeValidationRejectSchema();
+    const formik = useFormik<{
+        comment?: string;
```
bmonjoie: Isn't a comment required for rejection?

hat/assets/js/apps/Iaso/domains/instances/components/ValidationWorkflow/useSaveNode.ts
```diff
@@ -4,37 +4,64 @@ import { postRequest } from 'Iaso/libs/Api';
...
+type CompleteNodeByPassBody =
+    | {
+          comment?: string;
+          approved: true;
+          node: string;
+      }
+    | { approved: false; comment: string; node: string };
+
+const completeNode =
+    ({ instanceId, nodeId }: { instanceId: number; nodeId: number }) =>
```
bmonjoie: I find this syntax confusing. I believe it would be better to use:

```suggestion
    (instanceId: number, nodeId: number) =>
```

hat/assets/js/apps/Iaso/domains/instances/components/ValidationWorkflow/useSaveNode.ts
```diff
@@ -4,37 +4,64 @@ import { postRequest } from 'Iaso/libs/Api';
...
+    (body: CompleteNodeBody) =>
+        postRequest(
+            `${API_URL}instance/${instanceId}/nodes/${nodeId}/complete/`,
+            body,
+        );
+const completeNodeByPass =
+    ({ instanceId }: { instanceId: number }) =>
```
bmonjoie: Same as above:

```suggestion
    (instanceId: number) =>
```

hat/assets/js/apps/Iaso/domains/instances/components/ValidationWorkflow/ValidationModal.tsx
```diff
@@ -1,63 +1,112 @@
...
+                        labelString={formatMessage(MESSAGES.comment)}
+                        onChange={onChange}
+                        withMarginTop
+                    />
+                </Box>
+            </ConfirmCancelModal>
+        </FormikProvider>
+    );
+};
+export const ValidationRejectModal = ({
```
bmonjoie, line 158: It seems this code is a duplicate of `ValidationApproveModal` with only the texts and the `approved` value that differ.
Maybe it would be better to have a base method and use it in both methods to not repeat all the code.

hat/assets/js/apps/Iaso/domains/instances/components/ValidationWorkflow/ValidationModal.tsx
```diff
@@ -69,27 +118,79 @@ export const ValidationModal = ({
...
-                        options={approvalOptions}
                         required
                     />
+                </Box>
+            </ConfirmCancelModal>
+        </FormikProvider>
+    );
+};
+
+export const ValidationApproveByPassModal = ({
```
bmonjoie, line 187: Same remark as before.

hat/assets/js/apps/Iaso/domains/instances/components/ValidationWorkflow/ValidationModal.tsx
```diff
@@ -102,5 +203,88 @@ export const ValidationModal = ({
         </FormikProvider>
     );
 };
-const modalWithButton = makeFullModal(ValidationModal, ValidateButton);
-export { modalWithButton as ValidateNodeModal };
+export const ValidationRejectByPassModal = ({
```
bmonjoie, line 216: Same remark as before

iaso/api/validation_workflow_instances/serializers/retrieve.py
```diff
@@ -1,92 +1,278 @@
...
 
-class NextTasksSerializer(ModelSerializer):
-    user_roles = UserRoleNameSerializer(read_only=True, many=True, source="node.roles_required")
-    name = serializers.CharField(read_only=True, source="node.name")
-    node_template_slug = serializers.CharField(read_only=True, source="node.slug")
+    def get_updated_by(self, obj):
+        return None
+
+    def get_comment(self, obj):
+        return None
```
bmonjoie, line 121: What's the point of adding them if they always return `None`?

iaso/api/validation_workflow_instances/serializers/retrieve.py
```diff
@@ -1,92 +1,278 @@
...
+
+    def get_type(self, obj):
```
bmonjoie, line 123: Shouldn't you annotate this with:
```
@extend_schema_field(serializers.CharField)
```
?

iaso/api/validation_workflow_instances/serializers/retrieve.py
```diff
@@ -1,92 +1,278 @@
...
+            user_role_ids = self.context["user_roles"]
+        else:
+            user_role_ids = user.iaso_profile.user_roles.values_list("pk", flat=True)
+
+        required_roles_id = [x.pk for x in obj.roles_required.all()]
+
+        if not set(required_roles_id).issubset(set(user_role_ids)):
+            return False
+
+        return True
```
bmonjoie:
```suggestion
       return set(required_roles_id).issubset(set(user_role_ids))
```

iaso/api/validation_workflow_instances/serializers/retrieve.py
```diff
@@ -1,92 +1,278 @@
...
+        return None
+
+    def get_comment(self, obj):
+        return None
+
+    def get_type(self, obj):
+        return "NEXT_BYPASS"
+
+    @extend_schema_field(serializers.BooleanField)
+    def get_user_can_do_actions(self, obj):
```
bmonjoie: Isn't that the same code as above?
It might be a good idea to make a utility method to share the code.

iaso/api/validation_workflow_instances/serializers/retrieve.py
```diff
@@ -1,92 +1,278 @@
...
+    @extend_schema_field(NestedTimelineSerializer(many=True))
+    def get_timeline(self, obj):
+        instance = self._get_instance(obj)
+
+        # get history
+        nodes = instance.prefetched_validation_nodes
+        nodes_timeline = [
+            n for n in nodes if n.status not in [ValidationNodeStatus.NEW_VERSION, ValidationNodeStatus.SUBMISSION]
+        ]
+        nodes_timeline = [n for n in nodes_timeline if n.created_at >= obj.created_at]
```
bmonjoie: Why did you split this? 

```suggestion
        nodes_timeline = [
            n for n in nodes if n.status not in [ValidationNodeStatus.NEW_VERSION, ValidationNodeStatus.SUBMISSION]
       and n.created_at >= obj.created_at]
```

iaso/api/validation_workflow_instances/serializers/retrieve.py
```diff
@@ -1,92 +1,278 @@
...
-        fields = ["slug", "name", "user_roles"]
+        nodes = [
+            n
+            for n in nodes
+            if n.status not in [ValidationNodeStatus.NEW_VERSION, ValidationNodeStatus.SUBMISSION]
+            and n.created_at >= obj.created_at
+        ]
+
+        if obj.next_created_at:
+            nodes = [n for n in nodes if n.created_at < obj.next_created_at]
```
bmonjoie: It seems this logic is duplicated quite a lot. This should be moved into a utility method.

iaso/api/validation_workflow_instances/permissions.py
```diff
@@ -14,3 +15,12 @@ def has_permission(self, request, view):
         return request.user.has_perm(CORE_VALIDATION_WORKFLOW_PERMISSION.full_name()) and request.user.has_perm(
             CORE_SUBMISSIONS_PERMISSION.full_name()
         )
+
+
+class HasAccountFeatureFlag(BasePermission):
```
bmonjoie: I would rename this into `HasSubmissionValidationWorkflowAccountFeatureFlag`, it's a bit long, but it won't be confusing if we have many of them at some point with the same name.

iaso/api/validation_workflow_instances/views.py
```diff
@@ -59,14 +63,52 @@ def get_queryset(self):
...
+                    .prefetch_related("node__roles_required")
+                    .order_by("-created_at"),
+                    to_attr="prefetched_validation_nodes",
+                ),
+                Prefetch(
+                    "form__validation_workflow__node_templates",
+                    queryset=ValidationNodeTemplate.objects.prefetch_related(
+                        Prefetch(
+                            "next_node_templates",
+                            queryset=ValidationNodeTemplate.objects.order_by("slug"),
```
bmonjoie: Why do you order them by `slug`?

iaso/api/validation_workflows_nodes/permissions.py
```diff
@@ -12,3 +13,12 @@ def has_permission(self, request, view):
         if request.user.is_superuser:
             return True
         return request.user.has_perm(CORE_VALIDATION_WORKFLOW_PERMISSION.full_name())
+
+
+class HasAccountFeatureFlag(BasePermission):
```
bmonjoie, line 18: That's exactly the same as the one in `iaso/api/validation_workflow_instances/permissions.py`

iaso/api/validation_workflows_nodes/views.py
```diff
@@ -26,6 +28,14 @@ def get_queryset(self):
                 Instance.objects.select_related("project__account", "form", "form__validation_workflow")
                 .filter_for_user(self.request.user)
                 .filter(form__deleted_at__isnull=True)
+                .annotate(
+                    annotate_last_submission=Subquery(
```
bmonjoie: I would call it `annotate_last_submission_created_at` because it was confusing to me when you compared it later.

iaso/models/validation_workflow/templates.py
```diff
@@ -229,27 +229,37 @@ def move_node_template(self, node, position=None, parent_node_templates=None):
         )
 
     def dump_nodes(self):
-        start = self.get_starting_node()
+        if "node_templates" in getattr(self, "_prefetched_objects_cache", {}):
+            node_templates = self._prefetched_objects_cache["node_templates"]
+        else:
+            node_templates = self.node_templates.all().order_by("slug")
```
bmonjoie: Same here, why order by `slug`?

iaso/tests/api/validation_workflow_instances/test_views/test_list.py
```diff
@@ -109,6 +109,16 @@ def setUp(self):
...
             uuid=str(uuid.uuid4()),
         )
+        self.enable_validation_workflow_feature_flag(self.account)
+
+    @staticmethod
+    def enable_validation_workflow_feature_flag(*accounts):
+        feature_flag, _ = AccountFeatureFlag.objects.get_or_create(
+            code="SUBMISSION_VALIDATION_WORKFLOW",
+            defaults={"name": "Web: Enable validation workflow"},
+        )
```
bmonjoie: Shouldn't that be in a migration, and therefore, just a get by code be enough?

iaso/tests/api/validation_workflow_nodes/test_views/common.py
```diff
@@ -51,7 +51,18 @@ def setUp(self):
...
 
     def setup_start(self):
         ValidationWorkflowEngine.start(self.validation_workflow, self.john_wick, self.instance)
+
+    @staticmethod
+    def enable_validation_workflow_feature_flag(*accounts):
+        feature_flag, _ = AccountFeatureFlag.objects.get_or_create(
+            code="SUBMISSION_VALIDATION_WORKFLOW",
+            defaults={"name": "Web: Enable validation workflow"},
+        )
```
bmonjoie: Same remark as before

iaso/api/validation_workflow_instances/serializers/retrieve.py
hugo093: Not really the same , fetching the required_roles_id is different

iaso/api/validation_workflows_nodes/views.py
```diff
@@ -7,16 +7,18 @@
 from rest_framework.response import Response
 from rest_framework.viewsets import GenericViewSet
 
+from iaso.api.validation_workflows_node_templates.permissions import HasAccountFeatureFlag
```
bmonjoie, line 10: You should use only one class with a better name

bmonjoie (approved):
As discussed, this is a small change that should probably be moved to its own ticket.

PR #2970, opened 2026-04-29
IA-5024: fix django-axes errors on failed login

Phil-V:

This fixes an issue introduced with PR https://github.com/BLSQ/iaso/pull/2860 (rate limiting on login), where a data race could trigger on failed login attempts. 

(See upstream issue https://github.com/jazzband/django-axes/issues/1144)

### Related JIRA tickets

IA-5024 

## Changes

Switched from the cache handler to the `AxesDatabaseHandler`, which shouldn't have that issue.

## How to test

- Try to login more than 10 times with a wrong password:
  - on the login page on the dashboard
  - or using the api (POST /api/token/, present in our Bruno config)
- Check that you can login again after a cooldown period (30 seconds)

## Print screen / video

/ 

## Notes

Used `AXES_DISABLE_ACCESS_LOG = True` so that we don't keep an ever-growing access log that we'd have to maintain. This feature didn't exist for the cache backend. Only the table `axes_accessattempt` should have entries (when failing to login), the two other axes tables should stay empty.

## Doc

/

### DIFF
--- a/hat/settings.py
+++ b/hat/settings.py
@@ -781,10 +781,9 @@ AUTHENTICATION_BACKENDS = [
 
 # The API will rate limit login attempts based on 2 parameters:
 # - AUTH_RATE_LIMIT_MAX_FAILURES:
-#  Number of failed login attempts allowed before the user is locked out temporarily
+#  Number of failed login attempts before the user is locked out temporarily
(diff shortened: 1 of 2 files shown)


hugo093 (approved):
LGTM 👍

PR #2973, opened 2026-04-29
IA-4749 assignments table

beygorghor:

We need a table of all the org units linked to a planning. We should be able to click on a row to assign an org unit to a team or a user.

Bulk update will be done in another PR

### Related JIRA tickets

IA-4749

## Changes

- dedicated paginated endpoint to retrieve org units linked to a planning
- python tests
- AssignmentsTable.tsx to handle list of org unit
- common search for map and list
- small css override to have small input
- frontend tests

## How to test

Open assignments for a specific planning. 
Make sure you can open the list tab
Select a team or a user, and click on a row to assign the team or the user to the org unit

## Print screen / video

Screenshot 2026-04-29 at 16 47 33
Screenshot 2026-04-29 at 16 47 26
Screenshot 2026-04-29 at 16 47 20
Screenshot 2026-04-29 at 16 47 10


iaso/api/microplanning/views.py
```diff
@@ -45,21 +48,68 @@
 )
 
 
+class PlanningOrgUnitChildrenPagination(Paginator):
+    page_size_query_param = LIMIT
+    page_query_param = PAGE
+    page_size = 50
+    max_page_size = 500
```
hugo093: put this in a separate pagination.py file

iaso/api/microplanning/views.py
```diff
@@ -45,21 +48,68 @@
...
-class PlanningOrgunitsViewSet(AuditMixin, ViewSet):
+class PlanningOrgunitsViewSet(AuditMixin, GenericViewSet):
     """List orgunits for a planning."""
 
+    queryset = OrgUnit.objects.none()
+    serializer_class = PlanningOrgUnitTableSerializer
     http_method_names = ["get", "head", "options"]
     permission_classes = [IsAuthenticated, ReadOnlyOrHasPermission(CORE_PLANNING_WRITE_PERMISSION)]
+    results_key = "results"
+    remove_results_key_if_paginated = True
```
hugo093: this is tied to the `CustomPaginationListModelMixin`. While i understand you don't use the native list there, you have two choices : 

* change the url definition in urls.py to be /microplannings/orgunits/root/ and use the native list method. 
* keep your code as is, inherit from custompaginationlistmodelmixin but hides the generated list endpoint like 
```python 
@extend_schema(exclude=True)
    def list(self, request, *args, **kwargs):
        return super().list(request, *args, **kwargs)
``` but this is shady

File: iaso/api/microplanning/views.py [hunk 3]
```diff
@@ -45,21 +48,68 @@
...
+    filter_backends = [filters.OrderingFilter, filters.SearchFilter]
+    search_fields = ["name"]
+    ordering_fields = ["id", "name"]
+    ordering = ["id"]
+
+    def pagination_class(self):
+        return PlanningOrgUnitChildrenPagination(self.results_key)
```
hugo093:
```suggestion
    pagination_class = PlanningOrgUnitChildrenPagination
```

iaso/api/microplanning/views.py
```diff
@@ -45,21 +48,68 @@
...
 
     @action(detail=False, methods=["get"])
     def children(self, request, *args, **kwargs):
+        queryset = self._children_org_units_queryset(request, with_geo=True)
+        search = request.query_params.get("search", None)
+        if search:
+            queryset = queryset.filter(name__icontains=search)
+        return Response(PlanningOrgUnitSerializer(queryset, many=True).data, status=status.HTTP_200_OK)
```
hugo093: this should be more inspired by the list method. (and you axctually could do a super().list(...) by doing multiple steps : 

* implement a proper get_queryset() method that returns the queryset by action
* implement a proper get_serializer_class method that returns the proper serializer_class by action
* implement a proper filter_backends method cause currently the search filter might be applied to other enpoints. 

This would allow you to move the logic to the right places , and clean the endpoints methods to be more generic 

a quick example : 

```python 

def get_queryset(self, request):
    qs = OrgUnit.objects
    if self.action == "root":
        # filter out here
    elif self.action == "children":
        # filter out here
    elif self.action == "children_paginated":
        # filter out here
    return qs


@property
def filter_backends(self):
    if self.action in ["children", "children_paginated"]:
        return [filters.OrderingFilter, filters.SearchFilter]
    return []

def get_serializer_class(self):
    if self.action == "children":
        return PlanningOrgUnitSerializer
    elif self.action == "children_paginated":
        return PlanningOrgUnitTableSerializer
    elif self.action == "root":
        return PlanningOrgUnitSerializer
    raise NotImplementedError(f"serializers not implemeneted for {self.action}")

```

iaso/api/microplanning/views.py
```diff
@@ -45,21 +48,68 @@
...
+
+    @action(detail=False, methods=["get"], url_path="children-paginated")
+    def children_paginated(self, request, *args, **kwargs):
         planning = self._get_planning(request)
```
hugo093: is there any reason we didn't do something like `/microplanning/{planning_pk}/orgunits/` ? cause every endpoints need a planning object ... we could use `drf-nested-routers` for that

Moreover, it'll be nearly impossible to document correctly with swagger

iaso/api/microplanning/views.py
```diff
@@ -45,21 +48,68 @@
...
+                ),
+                to_attr="_planning_assignments_prefetched",
+            )
+        )
+        queryset = self.filter_queryset(queryset)
+        page = self.paginate_queryset(queryset)
+        ctx = {**self.get_serializer_context(), "planning": planning}
+        serializer = PlanningOrgUnitTableSerializer(page if page is not None else queryset, many=True, context=ctx)
+        if page is not None:
+            return self.get_paginated_response(serializer.data)
```
hugo093: as said above, most of this can be moved to other places. 
P.S:  useful native function get_serializer_context ;)

iaso/api/microplanning/views.py
```diff
@@ -73,10 +123,7 @@ def children(self, request, *args, **kwargs):
         else:
             raise ValidationError({"planning": [_("Planning is missing sampling group or target org unit scope")]})
 
-        children_queryset = queryset.order_by("id")
-        serializer_children = PlanningOrgUnitSerializer(children_queryset, many=True)
-
-        return Response(serializer_children.data, status=status.HTTP_200_OK)
+        return queryset.filter(validation_status=OrgUnit.VALIDATION_VALID).order_by("id")
```
hugo093: some issues here : 

* if i'm not mistaking, your base queryset already filter by validation_status, no need to write it again. 
* after your refactoring, if your endpoints calls filter_queryset , they should call orderingfilter, which will apply ordering default id, so no need to add it there either

iaso/api/microplanning/serializers.py
```diff
@@ -475,3 +475,64 @@ def get_geo_json(self, org_unit: OrgUnit):
...
+
+class PlanningOrgUnitTableSerializer(serializers.ModelSerializer):
+    """Paginated planning org units for tables (minimal columns + assignment for this planning)."""
+
+    assignment = serializers.SerializerMethodField()
+
+    class Meta:
+        model = OrgUnit
+        fields = ["id", "name", "assignment"]
+        read_only_fields = fields
```
hugo093, line 512: if i'm not mistaking, by default read_only_fields = fields, so no need to write it

iaso/tests/api/microplanning/test_microplanning.py
```diff
@@ -921,6 +926,198 @@ def test_planning_orgunits_children_with_target_org_unit_type(self):
         self.assertEqual(ids, [child.id])
         self.assertTrue(r[0]["has_geo_json"])
 
+    def test_planning_orgunits_children_search_by_name(self):
```
hugo093, line 928: missing tests to assertNumQueries

iaso/api/microplanning/views.py
```diff
@@ -45,21 +48,68 @@
...
+class PlanningOrgUnitChildrenPagination(Paginator):
+    page_size_query_param = LIMIT
+    page_query_param = PAGE
+    page_size = 50
+    max_page_size = 500
+
+
 @extend_schema(tags=["Micro plannings", "Org units", "Plannings"])
-class PlanningOrgunitsViewSet(AuditMixin, ViewSet):
+class PlanningOrgunitsViewSet(AuditMixin, GenericViewSet):
```
hugo093: the AuditMixin is now useless ?

iaso/api/microplanning/serializers.py
```diff
@@ -475,3 +475,64 @@ def get_geo_json(self, org_unit: OrgUnit):
...
+            return {
+                "team": PlanningOrgUnitTableAssignmentTeamSerializer(assignment.team).data,
+                "user": None,
+            }
+        if assignment.user_id and assignment.user:
+            return {
+                "team": None,
+                "user": PlanningOrgUnitTableAssignmentUserSerializer(assignment.user).data,
+            }
+        return None
```
hugo093: I get the idea but this is weird in the sense that the response is inconsistent. 

My suggestion would be to return both things + a "type" field that tells the front which one to use. 

```python 

class PlanningOrgUnitTableAssignmentSerializer(ModelSerializer):
    team = PlanningOrgUnitTableAssignmentTeamSerializer(allow_null=True, read_only=True)
    user = PlanningOrgUnitTableAssignmentUserSerializer(allow_null=True, read_only=True)
    ouput = serializers.SerializerMethodFiedl()

    class Meta:
        model = Assignment
        fields = ("team", "user")
    
    @extend_schema_field(ChoiceField)
    def get_output(self):
          if ... : 
                return USER
          ....

```

Or you use one outputfield (merge team and user), but then to be drf-spectacular compliant, you'd have to use polymorpĥicserailizer and so on

iaso/api/microplanning/serializers.py
```diff
@@ -475,3 +475,64 @@ def get_geo_json(self, org_unit: OrgUnit):
...
+
+    assignment = serializers.SerializerMethodField()
+
+    class Meta:
+        model = OrgUnit
+        fields = ["id", "name", "assignment"]
+        read_only_fields = fields
+
+    def get_assignment(self, org_unit: OrgUnit):
+        return PlanningOrgUnitTableAssignmentSerializer().to_representation(self._assignment_for_org_unit(org_unit))
```
hugo093: PlanningOrgUnitTableAssignmentSerializer(instance=self._assignment_for_org_unit(org_unit)).data

iaso/api/microplanning/serializers.py
```diff
@@ -475,3 +475,64 @@ def get_geo_json(self, org_unit: OrgUnit):
...
         return hasattr(org_unit, "geo_json") and org_unit.geo_json is not None
+
+
+class PlanningOrgUnitTableAssignmentTeamSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Team
+        fields = ["id", "name", "color"]
+
+
+class PlanningOrgUnitTableAssignmentUserSerializer(serializers.ModelSerializer):
```
hugo093: inherit from iaso.common  ModelSerializer

iaso/api/microplanning/serializers.py
```diff
@@ -475,3 +475,64 @@ def get_geo_json(self, org_unit: OrgUnit):
 
     def get_has_geo_json(self, org_unit: OrgUnit) -> bool:
         return hasattr(org_unit, "geo_json") and org_unit.geo_json is not None
+
+
+class PlanningOrgUnitTableAssignmentTeamSerializer(serializers.ModelSerializer):
```
hugo093: inherit from iaso.common  ModelSerializer

hugo093 (changes requested):
Did not test the code yet, just read it. 

A lot of refactoring, moving around code and so on to do. 

While the suggestion to move the endpoint to `/microplanning/{planning_pk}/orgunits/` can be done in another ticket (lot of work). The rest should can be done now :)

iaso/api/microplanning/serializers.py
```diff
@@ -475,3 +466,67 @@ def get_geo_json(self, org_unit: OrgUnit):
...
+
+    team = PlanningOrgUnitTableAssignmentTeamSerializer(allow_null=True, read_only=True)
+    user = PlanningOrgUnitTableAssignmentUserSerializer(allow_null=True, read_only=True)
+    assignment_type = serializers.SerializerMethodField()
+
+    class Meta:
+        model = Assignment
+        fields = ("id", "team", "user", "assignment_type")
+
+    @extend_schema_field(OpenApiTypes.STR)
```
hugo093: if i'm not mistaking, could be serializers.ChoiceField(choices=()) => so it's basically an enum with potential null value

iaso/api/microplanning/serializers.py
```diff
@@ -475,3 +466,67 @@ def get_geo_json(self, org_unit: OrgUnit):
...
+    def get_assignment_type(self, obj: Assignment):
+        if obj.team_id and obj.team:
+            return "team"
+        if obj.user_id and obj.user:
+            return "user"
+        return None
+
+
+class PlanningOrgUnitTableSerializer(serializers.ModelSerializer):
```
hugo093: use ModelSerializer from iaso

iaso/api/microplanning/serializers.py
```diff
@@ -475,3 +466,67 @@ def get_geo_json(self, org_unit: OrgUnit):
...
+
+    class Meta:
+        model = OrgUnit
+        fields = ["id", "name", "assignment"]
+        read_only_fields = fields
+
+    def get_assignment(self, org_unit: OrgUnit):
+        assignment = self._assignment_for_org_unit(org_unit)
+        if assignment is None:
+            return None
```
hugo093: i think that line is unecessary now, as the serializer would return none if assignement is none. + add @extend_schema_field(PlanningOrgUnitTableAssignementSerializer)

iaso/api/microplanning/views.py
```diff
@@ -46,20 +48,58 @@
 
 
 @extend_schema(tags=["Micro plannings", "Org units", "Plannings"])
-class PlanningOrgunitsViewSet(AuditMixin, ViewSet):
-    """List orgunits for a planning."""
+class PlanningOrgunitsViewSet(GenericViewSet):
```
hugo093, line 51: If you plan to use it as a nested router , you must add NestedViewSetMixin => see doc https://chibisov.github.io/drf-extensions/docs/#nested-router-mixin

iaso/api/microplanning/views.py
```diff
@@ -46,20 +48,58 @@
...
+            return PlanningOrgUnitSerializer
+        return self.serializer_class
+
+    def get_planning(self) -> Planning:
+        cached = getattr(self, "_planning_for_orgunits", None)
+        if cached is not None:
+            return cached
+        pk = self.kwargs.get("parent_lookup_pk")
+        if pk is None:
+            raise ValidationError({"planning": [_("Planning id is required in the URL path.")]})
```
hugo093: this won't happen, or am i missing something?

iaso/api/microplanning/views.py
```diff
@@ -46,20 +48,58 @@
...
+            Planning.objects.filter_for_user(user)
+            .select_related("org_unit", "selected_sampling_result__group")
+            .prefetch_related("target_org_unit_types"),
+            pk=pk,
         )
+        return self._planning_for_orgunits
+
+    def get_queryset(self):
+        if self.kwargs.get("parent_lookup_pk") is None:
+            return OrgUnit.objects.none()
```
hugo093: again , this won't happen , or i am missing something

iaso/api/microplanning/views.py
```diff
@@ -73,26 +113,45 @@ def children(self, request, *args, **kwargs):
...
+                        "user__iaso_profile", "team"
+                    ),
+                    to_attr="_planning_assignments_prefetched",
+                )
+            )
+        return queryset
+
+    def get_serializer_context(self):
+        ctx = super().get_serializer_context()
+        if getattr(self, "action", None) == "children_paginated" and self.kwargs.get("parent_lookup_pk") is not None:
```
hugo093: self.kwargs.get("parent_lookup_pk") is not None won't happen

iaso/api/microplanning/views.py
```diff
@@ -73,26 +113,45 @@ def children(self, request, *args, **kwargs):
...
 
-        return Response(serializer_children.data, status=status.HTTP_200_OK)
+    @action(detail=False, methods=["get"], url_path="children-paginated")
+    def children_paginated(self, request, *args, **kwargs):
+        queryset = self.filter_queryset(self.get_queryset())
+        page = self.paginate_queryset(queryset)
+        serializer = self.get_serializer(page if page is not None else queryset, many=True)
+        if page is not None:
+            return self.get_paginated_response(serializer.data)
+        return Response({"results": serializer.data})
```
hugo093: return just Response(serializer.data) to be consistent with how the rest of drf works

hugo093 (changes requested):
some details here and there, otherwise starting to look good :)

iaso/api/microplanning/views.py
beygorghor: no indeed we had this discussion yesterday

iaso/api/microplanning/serializers.py
beygorghor: don't really get it here sorry ...

hugo093 (approved):
LGTM. Thanks for all the work. 

I pushed what I meant for the choicefield and drf-spectacular annotation :)

hat/assets/js/apps/Iaso/domains/assignments/components/AssignmentsMap.tsx
```diff
@@ -132,127 +127,140 @@ export const AssignmentsMap: FunctionComponent<Props> = ({
...
-                style={{ height: defaultHeight }}
-                center={defaultViewport.center}
-                zoom={defaultViewport.zoom}
-                scrollWheelZoom={false}
-                zoomControl={false}
-                boundsOptions={boundsOptions}
-            >
-                <CloseTooltipOnMoveStart />
-                <CustomZoomControl
+            {!isLoading && (
```
quang-le: I think this conditional may cause the flicker effect

hat/assets/js/apps/Iaso/domains/assignments/hooks/requests/useSaveAssignment.ts
```diff
@@ -46,6 +46,7 @@ export const useSaveAssignment = ({
         queryClient.invalidateQueries('assignmentsList');
         queryClient.invalidateQueries('planningDetails');
         queryClient.invalidateQueries('orgUnitsList');
+        queryClient.invalidateQueries('planningChildrenOrgUnitsPaginated');
```
quang-le, line 48: For some reason it doesn't seem to work as expected

hat/assets/js/apps/Iaso/domains/assignments/components/AssignmentsTable.tsx
```diff
@@ -0,0 +1,97 @@
...
+        <Paper sx={styles.paper}>
+            <Box sx={styles.tableContainer}>
+                <TableWithDeepLink
+                    baseUrl={baseUrls.assignments}
+                    params={params}
+                    marginBottom={false}
+                    marginTop={false}
+                    columns={columns}
+                    defaultSorted={[{ id: 'ended_at', desc: true }]}
+                    data={data?.results ?? []}
```
quang-le, line 80: Or maybe the table doesn't re-render because `data` is referentially stable?

hat/assets/js/apps/Iaso/domains/assignments/types/assigment.ts
```diff
@@ -2,6 +2,11 @@ import { Shape } from '../../orgUnits/types/shapes';
 
 export type AssignmentParams = {
     planningId: string;
+    tab: 'list' | 'map';
+    pageSize?: string;
+    page?: string;
+    order?: string;
+    search?: string;
```
quang-le, line 9: I think there's a type for these 4

hat/assets/js/apps/Iaso/domains/teams/hooks/requests/useGetPlanningOrgUnits.ts
```diff
@@ -1,17 +1,31 @@
...
+    params?: AssignmentParams,
 ): UseQueryResult<PlanningOrgUnits[], Error> => {
-    const url = makeUrlWithParams('/api/microplanning/orgunits/children/', {
-        planning: planningId,
-    });
+    const apiParams = {
+        search: params?.search,
+    };
+    const url = makeUrlWithParams(
+        `/api/microplanning/plannings/${planningId}/orgunits/children/`,
```
quang-le: I've started moving the endpoints baseUrls in constants like `const PLANNING_API_URL = '/api/microplanning/'`

hat/assets/js/apps/Iaso/domains/teams/hooks/requests/useGetPlanningOrgUnits.ts
```diff
@@ -1,17 +1,31 @@
...
+        apiParams,
+    );
     return useSnackQuery({
-        queryKey: ['planningChildrenOrgUnits', planningId],
+        queryKey: ['planningChildrenOrgUnits', planningId, apiParams],
```
quang-le, line 29: Not blocking or specific to this PR, but we should also move query keys to constants I think

quang-le (changes requested):
- There's flicker when assigning on the map
https://github.com/user-attachments/assets/3ae8a357-85ba-403f-b935-681f50167867

- When deleting assignments from the list tab, the list isn't emptied:

hat/assets/js/apps/Iaso/domains/assignments/types/assigment.ts
beygorghor: Not really, this should be gather too in a specific type

quang-le (approved):
fix the failing tests otherwise lgtm

PR #2974, opened 2026-04-30
IA-4982 feat: restyle form submissions timeline

62496:

This PR restyles the validation timeline so the comment/details area is visually distinct from the main timeline.

### Related JIRA tickets

IA-4982

## Changes

- This PR updates the validation timeline rendering in  
  `hat/assets/js/apps/Iaso/domains/instances/components/ValidationWorkflow/InstanceValidation.tsx`.

The main changes are:
- removed the default MUI `StepContent` left border, which was too close visually to the actual timeline 
- grouped each step’s details inside a dedicated content card so comment/description metadata is easier to distinguish from the main timeline
- added a simpler secondary vertical separator inside the card
- aligned the `VALIDATE` button inside the same card as the step content
- kept the existing timeline data logic unchanged: this PR only changes the presentation layer

Tests were kept in place and still pass:
- `InstanceValidation.test.tsx`
- `InstanceValidation.a11y.test.tsx`

## How to test

Create a validation workflows with at least a node template that can skip previous steps
Submit an instance that triggers that VF

You can use this python script to help you

- in your terminal you can use this commande :

```bash
docker compose exec -T iaso ./manage.py shell <<'PY'
```

```python 

import os
import django
os.environ.setdefault("DJANGO_SETTINGS_MODULE", "hat.local_settings")


django.setup()

from iaso.engine.validation_workflow import ValidationWorkflowEngine
from iaso.models import ValidationWorkflow, Instance
import time


workflow_template_slug = "test"
user = get_user_model().objects.get(username="axkoefm")
instance = Instance.objects.get(id="259")

# uncomment if you want a fresh new data, otherwise it will resubmit if possible
# instance.validationnode_set.all().delete()

workflow_template = ValidationWorkflow.objects.get(slug=workflow_template_slug)

ValidationWorkflowEngine.start(workflow_template, user, instance)

```

## Print screen / video
Capture d’écran du 2026-04-30 11-42-46
Capture d’écran du 2026-04-30 11-41-22

## Notes

- This PR only affects the visual rendering of the validation timeline.

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/instances/components/ValidationWorkflow/InstanceValidation.tsx
+++ b/hat/assets/js/apps/Iaso/domains/instances/components/ValidationWorkflow/InstanceValidation.tsx
@@ -33,14 +33,6 @@ const timelineContentSx = {
     backgroundColor: 'rgba(0, 0, 0, 0.04)',
 };
 
-const timelineLineSx = {
-    width: 0,
-    flexShrink: 0,
-    alignSelf: 'stretch',
-    my: 0.5,
-    borderLeft: '3px dashed rgba(0, 0, 0, 0.32)',
-};
-
 const timelineBodySx = {
     flex: 1,
     display: 'flex',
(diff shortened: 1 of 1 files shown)


quang-le (approved):
I think the vertical separator inside the card is not necessary, since the card itself already nicely delimit the comment, but it does the job.
I'll let you decide whether to keep it or not

62496:
Yes, thanks for the feedback, I agree with you, I don’t think it’s really useful. I mainly added it just to make sure I was following the instructions but I will remove it 👍

PR #2975, opened 2026-04-30
IA-4961: Assignments checkboxes should be a radio buttons

beygorghor:

On assignments page, on the right panel, while selecting a user or a team, checkboxes should be radio button
image

### Related JIRA tickets

IA-4961

## Changes

Just replace check boxes with radio button, adding radio group name.

## How to test

Open assignments for a specific planning, make sure you can still assign a user or a team to an org unit

## Print screen / video

Screenshot 2026-04-30 at 12 21 48

## Notes

No frontend tests here, this part of the code will change to allow nested teams and users


quang-le (approved):
lgtm

PR #2977, opened 2026-05-04
POLIO-2135 improve perf by avoid cross products and use exist, defer geometries related fields

mestachs:

Slow queries triggering a lot of tmp storage due to crossjoin generated between 1500 records and lot of orgunits and unicity on non trivial fields (geom, simplified_geom).

### Related JIRA tickets

POLIO-2135

## Changes

The previous sql was outer joining orgunits, reports,... 
The new one use exists

## How to test

The bad news to have enough data to trigger the problem you need to restore a production like copy.

## Print screen / video

before locally

image

after 

https://github.com/user-attachments/assets/abf8e231-e419-40f3-9a54-26cf141337a4

## Notes

I'm not a polio expert

## Doc

### DIFF
--- a/plugins/polio/api/vaccines/repository_reports.py
+++ b/plugins/polio/api/vaccines/repository_reports.py
@@ -1,6 +1,6 @@
 """API endpoints and serializers for vaccine repository reports."""
 
-from django.db.models import Q
+from django.db.models import Exists, OuterRef
 from drf_spectacular.types import OpenApiTypes
 from drf_spectacular.utils import OpenApiParameter, extend_schema
 from rest_framework import filters, permissions, serializers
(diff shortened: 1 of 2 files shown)


plugins/polio/api/vaccines/repository_reports.py
```diff
@@ -111,17 +112,29 @@ class VaccineRepositoryReportsViewSet(GenericViewSet, ListModelMixin):
...
         )
 
         if self.request.user and self.request.user.is_authenticated:
             base_qs = base_qs.filter(account=self.request.user.iaso_profile.account)
 
-        return base_qs.filter(Q(destructionreport__isnull=False) | Q(incidentreport__isnull=False))
+        return base_qs.filter(
+            Exists(DestructionReport.objects.filter(vaccine_stock_id=OuterRef("pk")))
+            | Exists(IncidentReport.objects.filter(vaccine_stock_id=OuterRef("pk")))
+        )
```
quang-le, line 137: TIL. We have a few `distinct` in the codebase, maybe some coud be refactored that way

quang-le (approved):
We should hotfix this

PR #2986, opened 2026-05-05
[IA-3839] adding team members count

Crebert08:

This ticket is about implementing a new column "Count" on the API table. 

### Related JIRA tickets

[IA-3889](https://bluesquare.atlassian.net/browse/IA-3839)

#### Backend
- In the serializer, I implemented the get_members_count function to annotate the queryset with the number of teams(users and teams)
- In the viewset, I implemented the check to see if the fields `members_count` is needed in the response
- Added some tests

#### Frontend
- Defined the types with the new fields
- updated the table configuration
- Added a select columns drawer
- Added the translations
- Added some tests for ColumnSelectDrawer Component

## How to test

> Go to users modules in the teams list
> You should see a select columns drawer with the members_count deactivated by default

## Print screen / video

[Screencast from 07-05-26 13:02:31.webm](https://github.com/user-attachments/assets/e8898e2d-19ff-48ae-952f-8dc714dc29bb)

## Notes

## Doc


iaso/api/teams/serializers.py
```diff
@@ -82,13 +85,39 @@ class Meta:
...
+            "sub_teams",
+            "sub_teams_details",
+        ]
+        read_only_fields = ["created_at", "members_count"]
 
     users_details = NestedUserSerializer(many=True, read_only=True, source="users")
     sub_teams_details = NestedTeamSerializer(many=True, read_only=True, source="sub_teams")
     project_details = NestedProjectSerializer(many=False, read_only=True, source="project")
 
+    def get_members_count(self, team):
```
hugo093, line 115: annotate with @extend_schema_field(serializers.IntegerField)

iaso/api/teams/views.py
```diff
@@ -59,12 +61,26 @@ def get_queryset(self):
...
+            requested_fields = self.request.query_params.get("fields", None)
+            order = self.request.query_params.get("order", "")
+
+            needs_members_count = is_field_referenced("members_count", requested_fields, order)
+
+            if needs_members_count:
+                queryset = queryset.annotate(
+                    annotated_users_count=Count("users", distinct=True),
+                    annotated_sub_teams_count=Count("sub_teams", distinct=True),
+                )
```
hugo093, line 82: side note if we have a perf problem on the count, this article was useful for me , especially point 2 :  https://pganalyze.com/blog/pagination-django-postgres#postgresql-pagination-in-django-option-2--approximating-the-count

iaso/api/teams/views.py
hugo093, line 47: the filters must now contain DynamicFieldsFilterBackendBackwardCompatible since you use it in the serializer :)

iaso/api/teams/views.py
```diff
@@ -59,12 +61,26 @@ def get_queryset(self):
...
             self.queryset.filter_for_user(user)
             .select_related("project")
             .prefetch_related("users", "users__iaso_profile", "sub_teams")
         )
 
+        if self.action != "dropdown":
+            requested_fields = self.request.query_params.get("fields", None)
+            order = self.request.query_params.get("order", "")
+
+            needs_members_count = is_field_referenced("members_count", requested_fields, order)
```
hugo093, line 76: i didn't even know about this function, good to know

hat/assets/js/apps/Iaso/domains/teams/types/team.ts
```diff
@@ -32,6 +32,7 @@ export type Team = {
     deleted_at?: string;
     parent?: number;
     color: string;
+    members_count: number;
```
hugo093: it's optional, so it should be members_count?: number;

hugo093 (changes requested):
Really good job, just a few details to change :) thanks for the tests 👍

hugo093 (approved):
Thanks for the changes, LGTM 👍

PR #2987, opened 2026-05-06
IA-5036: disable API query if no perm

quang-le:

Users without WF validation enabled would get API errors on submission details page

### Related JIRA tickets

IA-5036

## Changes

- Disable API hook if user doesn't have permission

## How to test
- Create a user without the validation workflow permission
- go to an instance details page
- No API error snack should pop up, no call to workflow validation API should be made 

## Print screen / video

https://github.com/user-attachments/assets/e69d5621-d719-4169-b16b-c64d9e73d5ca

Screenshot 2026-05-06 at 11 59 31

## Notes

Will be hotfixed


hugo093 (approved):
lgtm

PR #2990, opened 2026-05-06
IA-5023 centralize user role name seralization

62496:

Role names exposed by validation workflow APIs should not include the account ID prefix.

### Related JIRA tickets

IA-5023

## Changes

Centralized user role name serialization in a shared `UserRoleNameField` and `UserRoleNameSerializer`.

The shared field removes only the exact `{account_id}_` prefix from `UserRole.group.name`, so names like `data_manager` are preserved when they do not match the account prefix.

Updated validation workflow serializers to use the shared serializer for role names in:
- validation workflow retrieve responses
- validation workflow node template list/retrieve responses
- validation workflow instance 

Added focused tests for the shared field/serializer and updated impacted API tests to assert that prefixed group names are returned without the account prefix.

## How to test

1. Start the app locally.
2. Open `/dashboard/forms/submissions/validation/detail/`.
3. Check the `Roles required` column in the steps table.
4. Confirm that the role name is displayed without the account ID prefix.

## Print screen / video

image

### DIFF
--- a/iaso/api/common/serializer.py
+++ b/iaso/api/common/serializer.py
@@ -3,6 +3,10 @@ from django.contrib.auth.models import User
 from django.utils.module_loading import import_string
 from rest_framework import serializers
 
+from iaso.models import UserRole
+
+from .serializer_fields import UserRoleNameField
+
 
 class UserSerializer(serializers.ModelSerializer):
     full_name = serializers.CharField(source="get_full_name")
@@ -10,6 +14,14 @@ class UserSerializer(serializers.ModelSerializer):
     class Meta:
         model = User
         fields = ["id", "first_name", "last_name", "username", "full_name"]
+
+
+class UserRoleNameSerializer(serializers.ModelSerializer):
+    name = UserRoleNameField(source="*", read_only=True)
+
+    class Meta:
+        model = UserRole
+        fields = ["name", "id"]
 
 
 class ModelSerializerFieldMappingMixin:
--- a/iaso/api/common/serializer.py
+++ b/iaso/api/common/serializer.py
@@ -7,6 +7,7 @@ from iaso.models import UserRole
 
 from .serializer_fields import UserRoleNameField
 
+
 class UserSerializer(serializers.ModelSerializer):
     full_name = serializers.CharField(source="get_full_name")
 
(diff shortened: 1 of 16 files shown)


iaso/api/common/serializer.py
```diff
@@ -12,6 +16,14 @@ class Meta:
         fields = ["id", "first_name", "last_name", "username", "full_name"]
 
 
+class UserRoleNameSerializer(serializers.ModelSerializer):
```
hugo093: use built-in iaso ModelSerializer instead of drf one

iaso/api/common/serializer.py
```diff
@@ -12,6 +16,14 @@ class Meta:
...
+    name = UserRoleNameField(source="*", read_only=True)
```
hugo093: source should be "group.name"

iaso/api/common/serializer_fields.py
```diff
@@ -77,6 +77,12 @@ def to_internal_value(self, data):
         return super().to_internal_value(prefixed_name)
 
 
+@extend_schema_field(OpenApiTypes.STR)
+class UserRoleNameField(serializers.Field):
```
hugo093: inherit from serializers.CharField instead , as it is a charfield :) we could then use things like max_length and so on in the future

iaso/api/common/serializer_fields.py
```diff
@@ -77,6 +77,12 @@ def to_internal_value(self, data):
...
+    def to_representation(self, user_role):
+        return user_role.group.name.removeprefix(f"{user_role.account_id}_")
```
hugo093: instead , with the use of "group.name" in the source : 

```python 
def to_representation(self, obj):
       head, sep, tail = obj.partition("_")
        return tail if sep else obj
```

hugo093 (changes requested):
Good job overall, a few details to change. Also you should check all the other api and not just validation workflow :) (i know that profile as this issue as well for instance) 

Search for model = UserRole and you should find most of the endpoints.

iaso/api/common/serializer_fields.py
62496: It is annoying that this kind of group.name can expose some issues ? If yes could I use something like this instead?.
 ```python  
return re.sub(r"^\d+_", "", value)
```
image

hugo093: Indeed you're right, good catch :+1:

hugo093 (approved):
LGTM 👍 Good job

PR #2995, opened 2026-05-07
[SNT-451] add captchas with `django-simple-captcha`

tdethier:

add captchas with `django-simple-captcha`

### Related JIRA tickets

SNT-451

## Changes

- added the [django-simple-captcha](https://github.com/mbi/django-simple-captcha/) lib
- added Bruno files to illustrate how to generate captchas
- added a missing value in `.env.example`

## How to test

- Try to generate a captcha (see Bruno files or README for explanations)
  - there's no endpoint that consumes captchas yet (it's going to be used first [in SNT](https://github.com/BLSQ/snt-malaria/pull/283))
  - captcha generation should be public, no need to be logged in

## Print screen / video

image

image

## Notes

/

## Doc

See README

### DIFF
--- a/hat/settings.py
+++ b/hat/settings.py
@@ -234,6 +234,7 @@ INSTALLED_APPS += [
     "django_json_widget",
     "phonenumber_field",
     "axes",
+    "captcha",
 ]
 
 if USE_CELERY:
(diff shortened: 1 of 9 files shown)


hat/settings.py
```diff
@@ -911,6 +915,7 @@ def sentry_error_sampler(_, hint):
     }
     if not ENCRYPTED_TEXT_FIELD_KEY:
         ENCRYPTED_TEXT_FIELD_KEY = "71Eax4PGazWNj7vaXrucAD1bYUzjI-Fxubv8MZzcSyk="
+    CAPTCHA_TEST_MODE = True
```
Bewi, line 918: I guess this should be overrided before releasing ?

Bewi (approved):
LGTM

hat/settings.py
tdethier: No, because it's in the `if` block that is triggered when running tests (line 903)

PR #2996, opened 2026-05-07
IA-2576 date can be modified with the keyboard

62496:

Date can't be changed with the keyboard.

### Related JIRA tickets

IA-2576

## Changes

Updated `DatesRange` so temporary invalid dates emitted while typing are not cleared when  blockInvalidDates is True 

## How to test

1. Go to the entities list or any other page that has a date filters and use one.
2. Verify that the field is not cleared while editing the date with the keyboard.

## Notes

when editing in the middle of the date, digits can shift to the left, for example year digits moving into the month segment after deleting month digits.

Sould we  accept this behavior  ?


beygorghor (approved):

beygorghor (changes requested):
Ok this is working, can you try to add the missing typing please ? line 31 and keyValue, date line 105. Also i have eslint issue on the file

Image

I have merged develop in your branch, meanwhile we have added tests and it's failing now

beygorghor (approved):

beygorghor:
We can merge this ! @62496

PR #2999, opened 2026-05-08
IA-3848 add description field in Project

62496:

Projects need a description field to explain their purpose when multiple projects exist in the same instance.

### Related JIRA tickets

IA-3848

## Changes

Added a `description` field to projects and exposed it through the project/app APIs.

The project create/edit form now displays a Description textarea between App ID and Color, and saves the value.

## How to test

1. Go to the Projects page.
2. Create or edit a project.
3. Fill the Description field and save.
4. Reopen the project and verify that the description is still displayed.

## Print screen / video

image
image


Crebert08 (changes requested):
Good job ! 

I would add a column on the UI for the description field

62496:
Thanks for the feedback. Should I remove the current textarea for the description and add the column instead, or should I leave it as it is and just add the column?

Crebert08:
> Thanks for the feedback. Should I remove the current textarea for the description and add the column instead, or should I leave it as it is and just add the column?

You should keep the current textarea and just add the new Column

Crebert08:
LGTM !

@beygorghor any other comment on this PR ?

Crebert08 (approved):

hat/assets/js/apps/Iaso/domains/projects/config.tsx
```diff
@@ -33,6 +33,21 @@ export const useColumns = (
...
+                width: 300,
+                Cell: settings => {
+                    const description = settings.row.original.description;
+                    if (!description) return textPlaceholder;
+                    if (description.length <= 80) return description;
+                    return (
+                        <span title={description}>
+                            {`${description.slice(0, 80)}...`}
+                        </span>
+                    );
```
beygorghor: You can type your setting element like with stock item here:
` Cell: ({ row: { original: item } }: Setting<StockItem>) => (`

I prefer to destructure when declaring like this `const { description } = settings.row.original;`

Last thing you should use something like this to truncate text 
```
overflow: hidden;
text-overflow: ellipsis;
white-space: nowrap;
```

iaso/diffing/exporter.py
beygorghor, line 1: This should not be there i think

beygorghor (changes requested):
@tdethier can you have a look here on the python side please ?

iaso/migrations/0385_merge_20260508_0926.py
```diff
@@ -0,0 +1,12 @@
+# Generated by Django 4.2.30 on 2026-05-08 09:26
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("iaso", "0384_project_description"),
+        ("iaso", "0384_rename_bulkcreateusercsvfile_bulkcreateuserfile"),
+    ]
```
tdethier: You've got this extra migration because another migration was introduced while you worked on this feature. `makemigrations --merge` works, because there's no migration conflict, but it creates an extra unnecessary migration.

In this case, you should: 
- delete both migration files (since you don't have custom code in there, it's auto generated by django)
- run again `makemigrations`: you'll have a single file that's properly placed in the migration tree

iaso/diffing/exporter.py
tdethier: It looks like it's just formatting, it's strange 😅

@62496 do you know what happened? Was this done by `ruff format`? If so, maybe something is not right with your `ruff` config

iaso/api/apps/serializers.py
```diff
@@ -100,11 +104,14 @@ def update(self, instance, validated_data):
         forms = validated_data.pop(self.FORMS, None)
         app_id = validated_data.pop(self.APP_ID, None)
         name = validated_data.pop(self.NAME, None)
+        description = validated_data.pop(self.DESCRIPTION, None)
         color = validated_data.pop(self.COLOR, None)
         if app_id is not None:
             instance.app_id = app_id
         if name is not None:
             instance.name = name
+        if description is not None:
```
tdethier: What you wrote works, but since that field is not nullable, I'd rather use `""` instead of `None`, and check on `if description` so that everything is more coherent

62496:
Yes i just format the code with ruff or you talk about the migration ?

ps : it is the answer of iaso/diffing/exporter.py I don't now why my answers still in pending

iaso/tests/api/test_apps.py
```diff
@@ -19,7 +19,12 @@ class AppsAPITestCase(APITestCase):
     def setUpTestData(cls):
```
tdethier, line 19: I would probably add a test somewhere to show that you can still create apps and projects without passing the `description` field and that it's value at the model level is `""` when you do that

tdethier (changes requested):

62496:
>yes normally ruff is only for python maybe ? can you confirm that I'am not sure 

Personnally, I use the Ruff extension in VS code and it does the work

tdethier:
>yes normally ruff is only for python maybe ? can you confirm that I'am not sure

Yes, it will format and check your python files. It should be run using the config that's at the root level in `pyproject.toml`. It should be the default behavior but I'm not sure how you ran it

beygorghor:
@62496 you need to reformat you migration file with ruff please

iaso/diffing/exporter.py
```diff
@@ -346,8 +346,10 @@ def update_groups_without_groupsets(self, api, diffs: List[Diff], fields):
         support_by_update_fields = [field for field in fields if field.startswith("group:")]
         to_update_diffs = list(
             filter(
-                lambda x: (x.status == Differ.STATUS_MODIFIED or x.status == Differ.STATUS_NEW)
-                and x.are_fields_modified(support_by_update_fields),
+                lambda x: (
```
tdethier: I'm wondering why you still have changes there, something looks really weird with the `ruff` configuration. You're on ubuntu 24.04? which `ruff` version are you using?

iaso/api/apps/serializers.py
```diff
@@ -100,11 +104,15 @@ def update(self, instance, validated_data):
         forms = validated_data.pop(self.FORMS, None)
         app_id = validated_data.pop(self.APP_ID, None)
         name = validated_data.pop(self.NAME, None)
+        has_description = self.DESCRIPTION in validated_data
```
tdethier: Nitpick: you don't need this variable, you could check directly for that in your `if` below

iaso/permissions/core_permissions.py
```diff
@@ -365,7 +365,6 @@ def full_name(self) -> str:
     ui_group=PERMISSION_GROUP_ADMIN,
 )
 
-
```
tdethier, line 368: This is again a formatting issue, I guess it will show up again for everybody else if we include it

iaso/migrations/0386_project_description.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 4.2.30 on 2026-05-13 09:49
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("iaso", "0385_alter_corepermissionsupport_options"),
```
tdethier: Another migration was merged in the meantime, I'm afraid you'll need to rebase and redo the migration file 🙈

tdethier (dismissed):
Backend looks good 👍 I'll let @beygorghor validate the frontend latest changes

Before merging, please rebase and remake the migration file or it will cause some conflicts

tdethier (approved):
LGTM, but let's wait tomorrow to merge 👍

beygorghor (approved):

PR #3001, opened 2026-05-08
IA-5044 active list home page

quang-le:

Active List plugins needs to be able to override the homepage redirection

### Related JIRA tickets

IA-5044

## Changes

- Add `ROOT_REDIRECT_PATTERN_NAME` in settings.py to provide default root redirection
- Add `resolve_root_redirect_pattern_name` in `urls.py` to check if there's any override coming from the plugins
- Add tests
- Add testable example in `/plugins/test/`
- Port changes made to the iaso code base from the active list plugin branch
 
 
## How to test

Launch iaso with PLUGINS=test
login
see that redirection is to `test/homepage` i.o `dashboard/home`

## Print screen / video

Screenshot 2026-05-08 at 17 24 11
Screenshot 2026-05-08 at 17 24 11


cron.yaml
```diff
@@ -1,14 +1,31 @@
...
+ - name: "vaccine_stock_history_update_for_rounds"
+   url: "/tasks/launch_task/plugins.polio.tasks.archive_vaccine_stock_for_rounds.archive_vaccine_stock_for_rounds/polio_cron_task_user/"
+   schedule: "0 2 * * *"
+
+ - name: "export_entities_to_records"
+   url: "/tasks/launch_task/plugins.active_list.tasks.export_entities_to_records.export_entities_to_records/demo/"
+   schedule: "*/4 * * * *"
+ - name: "export_records_to_entities"
+   url: "/tasks/launch_task/plugins.active_list.tasks.export_records_to_entities.export_records_to_entities/demo/"
+   schedule: "*/4 * * * *"
```
beygorghor: We need to find a more `plugin` way for this

hat/settings.py
```diff
@@ -61,6 +61,7 @@ def get_env_var_or_default(env_var, default):
 TESTING = get_env_var_or_default("TESTING", "").lower() == "true"
 IN_TESTS = len(sys.argv) > 1 and sys.argv[1] == "test"
 PLUGINS = os.environ["PLUGINS"].split(",") if get_env_var_or_default("PLUGINS", "") else []
+ROOT_REDIRECT_PATTERN_NAME = get_env_var_or_default("ROOT_REDIRECT_PATTERN_NAME", "dashboard:home_iaso")
```
hugo093: I think the idea of a settings is good, however , i wouldn't dynamically resolve it with a plugin like it's done in the resolve_root_redirect_pattern_name. 

I would instead do it directly in the settings.py. 
Why ? The logic of this redirect belongs to iaso web app, not the plugin in itself. It's up to the webapp then to say "if that plugin is installed, then i should use this homepage instead" 

e.g 

```python 
if [plugin] in PLUGINS:
     ROOT_REDIRECT_PATTERN_NAME = "" 
```
This makes the behavior explicit and centralized.

Moreover , in the current implementation, what will happen if we have multiple plugins  that redefines a root?

iaso/templatetags/getattribute.py
```diff
@@ -0,0 +1,23 @@
...
+
+from django import template
+from django.conf import settings
+
+
+numeric_test = re.compile("^\d+$")
+register = template.Library()
+
+
+def getattribute(value, arg):
```
hugo093: I see it's used in active_list, so it should be in active_list repo, not iaso

requirements.txt
```diff
@@ -120,6 +119,8 @@ psutil==7.2.2 # https://pypi.org/project/psutil/
 # ------------------------------------------------------------------------------
 https://github.com/BLSQ/django-webpack-loader/archive/5fa2b6897b27dc6ff5bb51162e1375fd0aa8afea.tar.gz#egg=django_webpack_loader
 
+https://github.com/madewulf/spreadsheet-xlsform-validator/archive/refs/heads/main.zip
```
hugo093, line 124: can't we either move this to blsq own repo and/or publish it to pypi? 

It's not a good idea of using a library that belongs to an employee on its own repo and that is not published to pypi.

requirements.txt
```diff
@@ -162,6 +163,10 @@ gpxpy==1.4.0
 snaql==0.4.2
 tsp-solver2==0.4.1
 
+
+# active_list plugin
+openhexa.sdk==0.3.1
```
hugo093, line 170: side note : i'll tackle that once i redo the plugin architecture , but it shouldn't be there, it should be in pyproject.toml dependencies of active_list

hat/settings.py
```diff
@@ -619,6 +619,8 @@ def is_superuser(u):
     os.path.join(BASE_DIR, "iaso/static"),
     os.path.join(BASE_DIR, "hat/assets/webpack"),
 ]
+for plugin_name in PLUGINS:
+    STATICFILES_DIRS.append(os.path.join(BASE_DIR, "plugins/%s/static" % plugin_name))
```
hugo093: side note : shouldn't happen like that imho , but i'll tackle that when i redo plugin architecture , so out of scope

hugo093 (changes requested):
see my comments, while i find the idea good, i don't agree at all with architecture

hat/settings.py
beygorghor: For the second point, we already have the same issue in the front, i would take the last one like in the front

requirements.txt
beygorghor: I agree, i'll create a separate ticket for that if you don't mind !

beygorghor (approved):

hugo093 (approved):
as discussed, the plugin architecture thingy will be tackled in the plugin ticket, and the repo issue in another one. LGTM

PR #3002, opened 2026-05-08
[IA-5014] add `environs` + update settings to cast env vars

tdethier:

add `environs` + update settings to cast env vars

### Related JIRA tickets

IA-5014

## Changes

- installed `environs` for:
  - reading `.env` files
  - parsing env vars
  - checking for invalid variables (wrong type)

[more info here](https://github.com/sloria/environs)

## How to test

- if you have a local installation (backend without docker):
  - set values in your `.env` file, they should be read from the file directly, without preparing a specific configuration in your IDE (and setting env vars there)
  - you can now run any python command directly
  - everything should work as usual
- if you have a docker installation (your backend runs in a docker container):
  - everything should work as usual, no change for you, unless you have values in your `.env` that can't be parsed to their right type (or vars that are defined but don't have any value)

## Print screen / video

/

## Notes

- because of docker compose setting missing env vars as `""`, I had to override a method from `environs` - hopefully we'll manage to change this behavior and remove this override, but we'll work on that later

## Doc

/


hat/settings.py
```diff
@@ -808,27 +804,27 @@ def sentry_error_sampler(_, hint):
 
 SOCIALACCOUNT_PROVIDERS = {}
 
-WFP_AUTH_CLIENT_ID = get_env_var_or_default("WFP_AUTH_CLIENT_ID", False)
+WFP_AUTH_CLIENT_ID = env.str("WFP_AUTH_CLIENT_ID", default=None)
 # for now, only WFP uses social_accounts
-ACTIVATE_SOCIAL_ACCOUNT = WFP_AUTH_CLIENT_ID is not False
+ACTIVATE_SOCIAL_ACCOUNT = WFP_AUTH_CLIENT_ID is not None
```
tdethier, line 809: That's the variable I mentioned @butofleury: it's an ID so it should probably be `None` and not a boolean but I'm not sure how it's used in WFP.

hat/settings.py
```diff
@@ -808,27 +804,27 @@ def sentry_error_sampler(_, hint):
...
         "APP": {
-            "client_id": get_env_var_or_default("WFP_AUTH_CLIENT_ID", None),
+            "client_id": WFP_AUTH_CLIENT_ID,
             "secret": None,  # Secret is not accepted since we use PKCE
         },
         "OAUTH_PKCE_ENABLED": True,
         # To which tenant this is linked
         "IASO_ACCOUNT_NAME": iaso_account,
-        "EMAIL_RECIPIENTS_NEW_ACCOUNT": get_env_var_or_default("WFP_EMAIL_RECIPIENTS_NEW_ACCOUNT", "").split(","),
+        "EMAIL_RECIPIENTS_NEW_ACCOUNT": env.list("WFP_EMAIL_RECIPIENTS_NEW_ACCOUNT", default=[], delimiter=","),
```
tdethier, line 827: Also @butofleury this is now going to return a list if you don't provide a value

hat/settings.py
butofleury: @tdethier  `WFP AUTH CLIENT_ID` seems to be used for authentication via CIAM, but I don't know how!  
I think @madewulf  could now more about it!

tdethier: According to [allauth documentation](https://docs.allauth.org/en/dev/socialaccount/provider_configuration.html) it's indeed an ID, so the version in this PR should work fine

hat/settings.py
```diff
@@ -88,63 +91,63 @@ def get_env_var_or_default(env_var, default):
...
-DEV_SERVER = get_env_var_or_default("DEV_SERVER", "").lower() == "true"
-ENVIRONMENT = get_env_var_or_default("SENTRY_ENVIRONMENT", "development").lower()
-SENTRY_URL = get_env_var_or_default("SENTRY_URL", "")
-SENTRY_FRONT_ENABLED = get_env_var_or_default("SENTRY_FRONT_ENABLED", "false").lower() == "true"
-AVAILABLE_LANGUAGES = get_env_var_or_default("AVAILABLE_LANGUAGES", "en,fr")
+DEV_SERVER = env.bool("DEV_SERVER", default=False)
+ENVIRONMENT = env.str("SENTRY_ENVIRONMENT", default="development").lower()
+SENTRY_URL = env.str("SENTRY_URL", default="")
+SENTRY_FRONT_ENABLED = env.bool("SENTRY_FRONT_ENABLED", default=False)
+AVAILABLE_LANGUAGES = env.list("AVAILABLE_LANGUAGES", default=["en", "fr"], delimiter=",")
```
hugo093, line 104: side note : shouldn't we use this to filter out the LANGUAGES  settings ? Now we have two sources for the languages , it's weird

hat/settings.py
```diff
@@ -31,52 +32,54 @@
...
 # This should be the naked domain (no http or https prefix) that is
 # hosting Iaso, this is used when sending out emails that need a link
 # back to the Iaso application.
 #
 # This should be the same as the one set on: `/admin/sites/site/1/change/`
 
-DNS_DOMAIN = get_env_var_or_default("DNS_DOMAIN", "localhost:8081")
-TESTING = get_env_var_or_default("TESTING", "").lower() == "true"
+DNS_DOMAIN = env.str("DNS_DOMAIN", default="localhost:8081")
+TESTING = env.bool("TESTING", default=False)
```
hugo093, line 64: Side note again : that's my opinion and we already talked about it, but just writing it here for posterity. Those kind of settings shouldn't even exist in the settings.py and we should definitly split it like so : 

settings/base.py
settings/test.py
settings/dev.py

and use any setting accordingly with the DJANGO_SETTINGS_MODULE env variable

hat/settings.py
```diff
@@ -641,20 +645,12 @@ def is_superuser(u):
 try:
     from hat.__version__ import VERSION
 except Exception as e:
-    print("error importing hat.__version", e)
+    print("error importing hat.__version__", e)
```
hugo093, line 648: good catch 👍

hat/settings.py
```diff
@@ -629,7 +633,7 @@ def is_superuser(u):
             "assets/webpack",
             (
                 "webpack-stats.json"
-                if (DEBUG and not get_env_var_or_default("TEST_PROD", None) and not USE_S3)
+                if (DEBUG and not env.str("TEST_PROD", default=None) and not USE_S3)
```
hugo093: should be a boolean , no?

hat/settings.py
```diff
@@ -748,24 +744,24 @@ def sentry_error_sampler(_, hint):
...
-HIDE_BASIC_NAV_ITEMS = get_env_var_or_default("HIDE_BASIC_NAV_ITEMS", "no")
+APP_TITLE = env.str("APP_TITLE", default="IASO")
+FAVICON_PATH = env.str("FAVICON_PATH", default="images/iaso-favicon.png")
+LOGO_PATH = env.str("LOGO_PATH", default="images/logo.png")
+LOGIN_LOGO_PATH = env.str("LOGIN_LOGO_PATH", default=None)
+THEME_PRIMARY_COLOR = env.str("THEME_PRIMARY_COLOR", default="#006699")
+THEME_SECONDARY_COLOR = env.str("THEME_SECONDARY_COLOR", default="#ff7961")
+THEME_PRIMARY_BACKGROUND_COLOR = env.str("THEME_PRIMARY_BACKGROUND_COLOR", default="#F5F5F5")
+SHOW_NAME_WITH_LOGO = env.str("SHOW_NAME_WITH_LOGO", default="yes")
+HIDE_BASIC_NAV_ITEMS = env.str("HIDE_BASIC_NAV_ITEMS", default="no")
```
hugo093, line 764: should be a boolean?

hugo093 (changes requested):
few very minor details, few side notes where we could create follow up tickets. Otherwise LGTM 👍

hat/settings.py
tdethier: Yeah I'm not sure why we have both. There's another thing related to languages as well in polio 🤷‍♂️

hat/settings.py
tdethier: Yeah they should, definitely! But I didn't want to refactor stuff in this PR, I'd rather do this in a dedicated PR, with frontend changes

hat/settings.py
tdethier: Yeah it looks like a boolean. I wasn't sure because that's the only place in the whole codebase where it appears and I didn't know how it was used

hat/settings.py
tdethier: Yeah, we should definitely discuss this topic in a deep dive and see what the team wants to do

hugo093 (approved):
LGTM, maybe create follow up tickets for what was discussed

PR #3003, opened 2026-05-11
IA-4639: use `PaginationType` from bluesquare-components

beygorghor:

We need to remove duplicate Pagination type in IASO code and use the one from bluesquare-components

### Related JIRA tickets

IA-4639

## Changes

Correct usage and remove type

## How to test

Nothing to test here


hugo093 (approved):
lgtm

PR #3004, opened 2026-05-11
IA-5018 add field created at in submission csv

62496:

Submissions CSV/XLSX exports did not include the IASO creation/submission date
### Related JIRA tickets

IA-5018

## Changes

Added the `Création dans IASO` column to submissions CSV and XLSX exports.

This new column exports `Instance.created_at`, which corresponds to the date when the submission was created in IASO. It is distinct from the existing `Date de création` column, which uses `source_created_at`

Updated the submissions export tests to check that:
- the CSV export contains the new header
- the CSV row includes the IASO creation date
- the XLSX export includes the same column and value

## How to test

Go to the submissions list for a single form and export the results as CSV or XLSX.

The exported file should contain:
- `Date de création`: creation date from the source/device
- `Création dans IASO`: creation date in IASO, based on `Instance.created_at`

## Print screen / video

image


Phil-V (approved):
Looking good! 👍

PR #3006, opened 2026-05-11
IA-5017 pipeline permission

beygorghor:

Pipeline api and pages are accessible to all users if an openhexa workspace is configured, we should use a specific admin permission to make it inaccessible to users without this permission

### Related JIRA tickets

IA-5017

## Changes

- new chore permision, `permissions.py` and adapt `views.py`. + tests
- update the doc, 
- frontend integration (urls and menu)

## How to test

Follow [this](docs/pages/dev/how_to/openhexa-integration/openhexa-integration.en.md) readme to configure an Openhexa workspace on IASO 
Create a basic user without the pipeline permission, log in with this user, pipeline should not be visible in the menu and urls should not be accessible

## Print screen / video

Screenshot 2026-05-11 at 13 33 08
Screenshot 2026-05-11 at 13 28 58
Screenshot 2026-05-11 at 13 15 31


Bewi (approved):
LGTM and WaE

PR #3008, opened 2026-05-11
IA-4001 Add  test for data source read-only toggle

62496:

Add a regression test for updating a data source from read-only back to editable.

### Related JIRA tickets

IA-4001

## Changes

Added a backend API test covering the `read_only=True -> False` update case for data sources.

The test verifies that:
- the update request succeeds
- `read_only=False` is persisted in DB
- a later GET still returns `read_only: false`
 
## How to test

aso.tests.api.test_data_sources

### DIFF
--- a/iaso/tests/api/test_data_sources.py
+++ b/iaso/tests/api/test_data_sources.py
@@ -286,7 +286,7 @@ class DataSourcesAPITestCase(APITestCase):
         response = self.client.get("/api/datasources/dropdown/?order=name")
         self.assertJSONResponse(response, 403)
 
-     def test_datasource_update_can_unset_read_only(self):
+    def test_datasource_update_can_unset_read_only(self):
         self.client.force_authenticate(self.joe)
         self.data_source.read_only = True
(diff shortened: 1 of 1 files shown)


iaso/tests/api/test_data_sources.py
```diff
@@ -285,3 +285,33 @@ def test_dropdown_datasource_with_user_without_permission(self):
         self.client.force_authenticate(self.jim)
         response = self.client.get("/api/datasources/dropdown/?order=name")
         self.assertJSONResponse(response, 403)
+
+    def test_datasource_update_can_unset_read_only(self):
```
hugo093, line 289: good test, can you add another one where data source has the `read_only` set to `True`, and when you update without `read_only` field (json doesn't contain key `read_only`), the `read_only` isn't altered ?

hugo093 (changes requested):
Just asking for more test, otherwise GJ and LGTM

hugo093 (approved):
LGTM , good job

PR #3010, opened 2026-05-11
Feature SNT-374: public account creation additions

michiwend:

* Custom `?next=<path>` redirect after `/logout-iaso`
* Permission-free PATCH of `first_name`, `last_name`, `email` for logged-in users

### Related JIRA tickets

SNT-374

## Changes

This PR covers two changes which are prerequisites for the public account creation in SNT Malaria

### Custom `?next=<path>` redirect after `/logout-iaso`

The custom `?next=<path>` redirect after logout is for a use case where account creation was not successful. We log the user out and redirect them to the public account creation to start over if they want to. Without this, the user would see the login page instead.

Technically this is achieved by subclassing Django's default `LogoutView` as a backwards-compatible drop-in for the `/logout-iaso` endpoint.

Since this is a potential attack vector (see [CWE-601](https://cwe.mitre.org/data/definitions/601.html)), only allow-listed redirects (`settings.LOGOUT_NEXT_ALLOWED_PATHS`) are being followed.

### Permission-free PATCH of `first_name`, `last_name`, `email` for logged-in users

This  simply allows any logged-in user to `PATCH` their own `first_name`, `last_name` and `email` which is required by our account-creation wizard.

### DIFF
--- a/hat/urls.py
+++ b/hat/urls.py
@@ -13,7 +13,7 @@ from django.urls import include, path
 from django.views.generic import RedirectView, TemplateView
 from drf_spectacular.views import SpectacularAPIView, SpectacularRedocView, SpectacularSwaggerView
 
-from iaso.auth.views import IasoPasswordResetView
+from iaso.auth.views import IasoLogoutView, IasoPasswordResetView
 from iaso.views import ModelDataView, health, health_clamav, page, robots_txt
 
 
@@ -90,7 +90,7 @@ else:
         path("api/", include("iaso.urls")),
         path("pages/<page_slug>/", page, name="pages"),
         path("i18n/", include("django.conf.urls.i18n")),
-        path("logout-iaso", auth.views.LogoutView.as_view(next_page="/login/"), name="logout-iaso"),
+        path("logout-iaso", IasoLogoutView.as_view(next_page="/login/"), name="logout-iaso"),
         path(
             "forgot-password/",
             IasoPasswordResetView.as_view(
--- a/iaso/api/profiles/serializers/update.py
+++ b/iaso/api/profiles/serializers/update.py
@@ -15,7 +15,25 @@ class BaseProfileUpdateSerializer(ModelSerializer):
 
 
 class ProfileMeUpdateSerializer(BaseProfileUpdateSerializer):
-    pass
+    """
+    Serializer used by `PATCH /api/profiles/me/`.
+
+    Any logged-in user can update a small set of self-service fields without
+    holding any user-admin permission (see `HasProfilePermission`, which
+    short-circuits when `pk == PK_ME`).
+
+    `first_name`, `last_name` and `email` live on the `User` model rather than
+    on `Profile`. They are declared here as plain serializer fields (matching
+    the pattern in `ProfileUpdateSerializer`) and written back to the user by
+    `ProfilesViewSet._post_update_user` when they appear in `validated_data`.
+    """
+
+    email = serializers.EmailField(required=False, allow_blank=True, write_only=True)
+    first_name = serializers.CharField(required=False, allow_blank=True, write_only=True, max_length=150)
+    last_name = serializers.CharField(required=False, allow_blank=True, write_only=True, max_length=150)
+
+    class Meta(BaseProfileUpdateSerializer.Meta):
+        fields = BaseProfileUpdateSerializer.Meta.fields + ["first_name", "last_name", "email"]
 
 
 class ProfileUpdateSerializer(BaseProfileUpdateSerializer):
--- a/iaso/auth/views.py
+++ b/iaso/auth/views.py
@@ -1,5 +1,7 @@
-from django.contrib.auth.views import PasswordResetView
+from django.conf import settings
+from django.contrib.auth.views import LogoutView, PasswordResetView
 from django.contrib.sites.shortcuts import get_current_site
+from django.utils.http import url_has_allowed_host_and_scheme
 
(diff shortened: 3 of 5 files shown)


iaso/auth/views.py
```diff
@@ -12,3 +14,36 @@
...
+
+    Plugins extend the allow-list via their ``plugin_settings.CONSTANTS``.
+    """
+
+    def get_redirect_url(self):
+        candidate = (
+            self.request.POST.get(self.redirect_field_name) or self.request.GET.get(self.redirect_field_name) or ""
+        )
+        if candidate and self._is_allowed(candidate):
+            return candidate
```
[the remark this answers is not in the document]
michiwend: nice catch but... doesn't my implementation do exactly what is suggested here? 🫠

iaso/auth/views.py
```diff
@@ -12,3 +14,36 @@ def form_valid(self, form):
...
+
+    def _is_allowed(self, candidate: str) -> bool:
+        allowed = set(getattr(settings, "LOGOUT_NEXT_ALLOWED_PATHS", []))
+        if candidate not in allowed:
+            return False
+        return url_has_allowed_host_and_scheme(
+            candidate,
+            allowed_hosts={self.request.get_host()},
+            require_https=self.request.is_secure(),
+        )
```
hugo093, line 49: all of this can be done like : 

```python 
def get_success_url(self):
       redirect = super().get_success_url()
       if redirect in set(getattr(settings, "LOGOUT_NEXT_ALLOWED_PATHS", [])):
              return redirect
       return ""
```

  although be careful with that check and settings (an edge case for example:  /dashboard could be accepted but not /dashboard/  or /dashboard/?form=x) )

hat/urls.py
```diff
@@ -90,7 +90,7 @@ def custom_404_view(request, exception):
         path("api/", include("iaso.urls")),
         path("pages/<page_slug>/", page, name="pages"),
         path("i18n/", include("django.conf.urls.i18n")),
-        path("logout-iaso", auth.views.LogoutView.as_view(next_page="/login/"), name="logout-iaso"),
+        path("logout-iaso", IasoLogoutView.as_view(next_page="/login/"), name="logout-iaso"),
```
hugo093, line 93: next_page should be declared in settings as LOGOUT_REDIRECT_URL instead, more clean

iaso/api/profiles/serializers/update.py
```diff
@@ -15,7 +15,25 @@ class Meta:
...
+    the pattern in `ProfileUpdateSerializer`) and written back to the user by
+    `ProfilesViewSet._post_update_user` when they appear in `validated_data`.
+    """
+
+    email = serializers.EmailField(required=False, allow_blank=True, write_only=True)
+    first_name = serializers.CharField(required=False, allow_blank=True, write_only=True, max_length=150)
+    last_name = serializers.CharField(required=False, allow_blank=True, write_only=True, max_length=150)
+
+    class Meta(BaseProfileUpdateSerializer.Meta):
+        fields = BaseProfileUpdateSerializer.Meta.fields + ["first_name", "last_name", "email"]
```
hugo093, line 36: should have moved this to BaseProfileUpdateSerializer actually, and update the ProfileUpdateSerializer after

Also, worth to be check if there is any impact on IASO front end

iaso/tests/auth/test_logout_view.py
```diff
@@ -0,0 +1,68 @@
...
+from django.test import override_settings
+from django.urls import reverse
+
+from iaso.test import TestCase
+
+
+User = get_user_model()
+
+
+@override_settings(LOGOUT_NEXT_ALLOWED_PATHS=["/example/allowed-target"])
```
hugo093, line 11: see remark above about the edge cases (trailing slashes, get param, ..)

iaso/auth/views.py
```diff
@@ -12,3 +14,36 @@ def form_valid(self, form):
...
+class IasoLogoutView(LogoutView):
+    """LogoutView that honours ``?next=<path>`` only when it appears in
+    ``settings.LOGOUT_NEXT_ALLOWED_PATHS`` (and is a same-host relative URL).
+    Anything else falls back to ``next_page``.
+
+    Strict allow-listing (rather than just a same-host check) closes
+    CWE-601 / OWASP "Unvalidated Redirects and Forwards": an internal page
+    on the same host can still be weaponised for phishing.
+
+    Plugins extend the allow-list via their ``plugin_settings.CONSTANTS``.
```
hugo093, line 28: how do they extend this ? 

By adding an internal url to the plugin ? 
By adding any url ? 

I don't see anything about it in hat.settings, so a bit hard to understand how it'll be implemented.

hugo093 (changes requested):

iaso/auth/views.py
tdethier: There's another PR in SNT that implements this, @michiwend forgot to include it in this description: https://github.com/BLSQ/snt-malaria/pull/288

hugo093: Ok I see, thanks. Then, I have the same remarks I made in this [PR](https://github.com/BLSQ/iaso/pull/3001) 
I don't quite agree with the architecture , but that's a big wrapped "imho" and should maybe be discussed more

michiwend:
@hugo093 I addressed your feedback and pushed everything in a new branch #3014 to follow the convention of naming branches equally to the related plugin branch. Sorry for the mess but otherwise our PR in SNT is not building.

PR #3014, opened 2026-05-12
SNT-374: public account creation (prerequisites)

michiwend:

* Custom `?next=<path>` redirect after `/logout-iaso`
* Permission-free PATCH of `first_name`, `last_name`, `email` for logged-in users

### Related JIRA tickets

SNT-374

## Changes

This PR covers two changes which are prerequisites for the public account creation in SNT Malaria

### Custom `?next=<path>` redirect after `/logout-iaso`

The custom `?next=<path>` redirect after logout is for a use case where account creation was not successful. We log the user out and redirect them to the public account creation to start over if they want to. Without this, the user would see the login page instead.

Technically this is achieved by subclassing Django's default `LogoutView` as a backwards-compatible drop-in for the `/logout-iaso` endpoint.

Since this is a potential attack vector (see [CWE-601](https://cwe.mitre.org/data/definitions/601.html)), only allow-listed redirects (`settings.LOGOUT_NEXT_ALLOWED_PATHS`) are being followed.

### Permission-free PATCH of `first_name`, `last_name`, `email` for logged-in users

This  simply allows any logged-in user to `PATCH` their own `first_name`, `last_name` and `email` which is required by our account-creation wizard.

### DIFF
--- a/iaso/auth/views.py
+++ b/iaso/auth/views.py
@@ -1,5 +1,7 @@
-from django.contrib.auth.views import PasswordResetView
+from django.conf import settings
+from django.contrib.auth.views import LogoutView, PasswordResetView
 from django.contrib.sites.shortcuts import get_current_site
+from django.utils.http import url_has_allowed_host_and_scheme
 
 from iaso.mail.branding import core_email_branding_context
 
@@ -12,3 +14,36 @@ class IasoPasswordResetView(PasswordResetView):
         domain = get_current_site(self.request).domain
         self.extra_email_context = core_email_branding_context(protocol=protocol, domain=domain)
         return super().form_valid(form)
+
+
+class IasoLogoutView(LogoutView):
+    """LogoutView that honours ``?next=<path>`` only when it appears in
+    ``settings.LOGOUT_NEXT_ALLOWED_PATHS`` (and is a same-host relative URL).
+    Anything else falls back to ``next_page``.
+
+    Strict allow-listing (rather than just a same-host check) closes
+    CWE-601 / OWASP "Unvalidated Redirects and Forwards": an internal page
+    on the same host can still be weaponised for phishing.
+
+    Plugins extend the allow-list via their ``plugin_settings.CONSTANTS``.
+    """
+
+    def get_redirect_url(self):
+        candidate = (
+            self.request.POST.get(self.redirect_field_name) or self.request.GET.get(self.redirect_field_name) or ""
+        )
+        if candidate and self._is_allowed(candidate):
+            return candidate
+        # Returning "" makes ``get_success_url`` fall through to
+        # ``get_default_redirect_url`` (which uses ``next_page``).
+        return ""
+
+    def _is_allowed(self, candidate: str) -> bool:
+        allowed = set(getattr(settings, "LOGOUT_NEXT_ALLOWED_PATHS", []))
+        if candidate not in allowed:
+            return False
+        return url_has_allowed_host_and_scheme(
+            candidate,
+            allowed_hosts={self.request.get_host()},
+            require_https=self.request.is_secure(),
+        )
--- a/iaso/auth/views.py
+++ b/iaso/auth/views.py
@@ -1,7 +1,6 @@
 from django.conf import settings
 from django.contrib.auth.views import LogoutView, PasswordResetView
 from django.contrib.sites.shortcuts import get_current_site
-from django.utils.http import url_has_allowed_host_and_scheme
 
 from iaso.mail.branding import core_email_branding_context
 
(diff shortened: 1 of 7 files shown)


iaso/auth/views.py
```diff
@@ -12,3 +14,37 @@ def form_valid(self, form):
...
+
+    def _is_allowed(self, candidate: str) -> bool:
+        allowed = set(getattr(settings, "LOGOUT_NEXT_ALLOWED_PATHS", []))
+        if candidate not in allowed:
+            return False
+        return url_has_allowed_host_and_scheme(
+            candidate,
+            allowed_hosts={self.request.get_host()},
+            require_https=self.request.is_secure(),
+        )
```
hugo093: all of this can be done like :

```python
def get_success_url(self):
       redirect = super().get_success_url()
       if redirect in set(getattr(settings, "LOGOUT_NEXT_ALLOWED_PATHS", [])):
              return redirect
       return ""
```

although be careful with that check and settings (an edge case for example: /dashboard could be accepted but not /dashboard/ or /dashboard/?form=x) )

iaso/tests/auth/test_logout_view.py
```diff
@@ -0,0 +1,68 @@
...
+from django.test import override_settings
+from django.urls import reverse
+
+from iaso.test import TestCase
+
+
+User = get_user_model()
+
+
+@override_settings(LOGOUT_NEXT_ALLOWED_PATHS=["/example/allowed-target"])
```
hugo093, line 11: see remark above about the edge cases (trailing slashes, get param, ..) , tests should be added for this

iaso/auth/views.py
```diff
@@ -12,3 +14,37 @@ def form_valid(self, form):
...
+            self.request.POST.get(self.redirect_field_name) or self.request.GET.get(self.redirect_field_name) or ""
+        )
+        if candidate and self._is_allowed(candidate):
+            return candidate
+        # Returning "" makes ``get_success_url`` fall through to
+        # ``get_default_redirect_url`` (which uses ``next_page``).
+        return ""
+
+    def _is_allowed(self, candidate: str) -> bool:
+        allowed = set(getattr(settings, "LOGOUT_NEXT_ALLOWED_PATHS", []))
```
hugo093: can you add in hat.settings this LOGOUT_NEXT_ALLOWED_PATHS ? (and default it to [])

hugo093 (changes requested):
same remarks as in the previous PR :)

iaso/auth/views.py
michiwend: overriding `get_success_url` might cause issues: at that point Django has already resolved the fallback, so when there's no `?next` parameter it returns `/login/`. Since that's not in the allowlist, we'd end up returning `""` and breaking normal logouts. We could add it to the allow list but we would still get `""` for non listed redirects. Current implementation would always fall back to  `LOGOUT_REDIRECT_URL`. Does that make sense?

Bewi (dismissed):
The flow works well. 
Got few remarks on the code.

iaso/auth/views.py
```diff
@@ -12,3 +14,45 @@ def form_valid(self, form):
...
+        # ``get_default_redirect_url`` (which uses ``next_page``).
+        return ""
+
+    def _get_allowed_path(self, candidate: str) -> str:
+        """Return the matching allowlisted path, or empty string if not allowed.
+
+        Strips trailing slashes before comparison; returns the configured path.
+        """
+        normalized = candidate.rstrip("/")
+        allowed_paths = getattr(settings, "LOGOUT_NEXT_ALLOWED_PATHS", [])
```
tdethier: Architecture-wise, I would:
- move `LOGOUT_NEXT_ALLOWED_PATHS` from SNT settings to core settings
- set the env var on SNT servers

This makes more sense since you're using the settings in IASO

iaso/auth/views.py
```diff
@@ -12,3 +14,45 @@ def form_valid(self, form):
         domain = get_current_site(self.request).domain
         self.extra_email_context = core_email_branding_context(protocol=protocol, domain=domain)
         return super().form_valid(form)
+
+
+class IasoLogoutView(LogoutView):
```
tdethier, line 18: I kinda agree with both of you (@hugo093 and @michiwend) on this:
- current version looks complicated because it copies the current implementation of `get_redirect_url` and does not call `super()`
- Hugo's suggestion of overriding `get_success_url` does break the normal logout

I suggest:
- switching `LOGOUT_NEXT_ALLOWED_PATHS` from SNT to core settings. It can be overridden with env vars
- using a slightly modified version of Hugo's solution:

```
    def get_redirect_url(self):
        redirect = super().get_redirect_url()
        normalized = redirect.rstrip("/")
        if normalized in set(getattr(settings, "LOGOUT_NEXT_ALLOWED_PATHS", [])):
            return normalized
        return ""
```

This allows us to have a simpler class which does not break normal logout.

tdethier (changes requested):
The feature works, but I'm suggesting a few changes. See below ⬇️

tdethier (approved):
LGTM, thanks for making the changes 👍

hugo093 (approved):
lgtm

PR #3017, opened 2026-05-13
IA-5044: make it easier to override main app url

beygorghor:

Remove ROOT_REDIRECT_PATTERN_NAME logic and only use env variable

### Related JIRA tickets

IA-5044

## Changes

- remove logic
- only use settings variable

## How to test

Make sure you have active list plugins in plugins folder
This should be in your env variable:

PLUGINS=active_list
ROOT_REDIRECT_PATTERN_NAME=active_list:homepage

Visit home url, you should see active list ui


tdethier (approved):
LGTM 👍

PR #3018, opened 2026-05-13
[IA-5061] fix flaky stock test

tdethier:

 fix flaky stock test

### Related JIRA tickets

IA-5061

## Changes

- return nested objects ordered by ID

## How to test

- run tests, the `iaso.tests.api.stocks.test_serializers.StockKeepingUnitSerializerTestCase.test_serialize_instance` test should no longer fail

## Print screen / video

/

## Notes

/

## Doc

/


iaso/api/stocks/serializers.py
```diff
@@ -103,6 +103,14 @@ class Meta:
         ]
         read_only_fields = ["created_at", "updated_at", "created_by", "updated_by", "deleted_at"]
 
+    def to_representation(self, instance):
+        data = super().to_representation(instance)
+        data["projects"] = sorted(data["projects"], key=lambda p: p["id"])
+        data["org_unit_types"] = sorted(data["org_unit_types"], key=lambda o: o["id"])
+        data["forms"] = sorted(data["forms"], key=lambda f: f["id"])
+        data["children"] = sorted(data["children"], key=lambda c: c["id"])
```
hugo093: I wonder if it wouldn't be better to do this sorting at the get_queryset level (through prefetch ) ?  or to implement a custom ListSerializer 
Or even puttin ordering in related models (meta)

hugo093 (changes requested):
just a quick question to be clarified

iaso/api/stocks/serializers.py
tdethier: Yeah, ideally I would do it in `get_queryset`, just like the previous PR. But here, we're not even calling the endpoint, so it wouldn't change anything (that's why I overwrote `to_representation` in a second commit in the previous PR) 🫤

I'm wondering if ordering in meta will change anything here. If it works, then maybe we should add this to all models indeed

hugo093: Yeah could you try doing it in models Meta ? Cause I prefer doing this in a centralized way rather than overriding every to_representation :/

hugo093 (approved):
LGTM, tx for doing the changes

PR #3022, opened 2026-05-13
fix: using other way to install from git

hugo093:

pip install git+ doesn't work on deploy CI cause git is not installed


tdethier (approved):
LGTM 👍

PR #3023, opened 2026-05-13
[IA-5063]: refactor to centralise permission for HasAccountFeatureFlag

hugo093:

This code was repeated in a lot of place so it had to be refactored and centralized : 
```python
class HasAccountFeatureFlag(BasePermission):
    message = _("This feature is disabled for your account.")

        if not request.user.iaso_profile:
            return False
        return request.user.iaso_profile.account.feature_flags.filter(code="SUBMISSION_VALIDATION_WORKFLOW").exists()
```

### Related JIRA tickets

IA-5063

## Changes

* Refactored into one single class that we can parameter 
* Added tests
* Adapted current viewsets implementations 

## How to test

Python tests but the affected endpoints should still behave accordingly (aka raise a "This feature is disabled for your account."  message )

### DIFF
--- a/iaso/api/common/permissions.py
+++ b/iaso/api/common/permissions.py
@@ -1,3 +1,4 @@
+from django.utils.translation import gettext_lazy as _
 from rest_framework import permissions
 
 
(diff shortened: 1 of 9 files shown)


Phil-V (approved):
Tested manually with one of the endpoint, works well! No notes on the code, well done 👍

PR #3026, opened 2026-05-18
[IA-5064]: no data message in validation box of instance + fix query keys

hugo093:

When completing (and bypassing) a node, the screen wouldn't always rerender automatically. Moreover, if no validation had been triggered, the box on instance UI was empty
### Related JIRA tickets

IA-5064

## Changes

* Fix query keys
* Added no data message
* Added tests

## How to test

Go to instance detail UI: 
* No data message should appear
Configure and start a validation workflow
* Completing a node (with or without bypass) should rerender the screen automatically

## Print screen / video

image


bmonjoie (approved):
LGTM

PR #3027, opened 2026-05-18
[IA-5067]: fix api call in case of no validation workflow configured + tests

hugo093:

Calling /api/validation-workflows/instance/{pk}/ for an instance with a form that has no validation workflow configured would raise a 500. 

### Related JIRA tickets

IA-5067

## Changes

* Fix serializer + tests

## How to test

Call /api/validation-workflows/instance/{pk}/ for an instance with a form that has no validation workflow , should not return a 500


bmonjoie (approved):
LGTM

PR #3029, opened 2026-05-18
[IA-5043]: Validation workflow: switch to module permission access

hugo093:

For validation workflow, we should use module instead of feature flags.

### Related JIRA tickets

IA-5043

## Changes

* Backend changes:
  * New BasePermission 
  * Adapted API
  * Tests
  * Special migration
* Front end
  * Adapted calls + tests
  * Adapted module UI 
  
## How to test

Create an account with the module "validation workflow", you should be able to see and call everything
Do the opposite, you shouldn't see anything

## Notes

worth to see : used django-test-migrations to test the migration, quite useful


Phil-V (approved):
Good job on integrating testing tools for the migration, very handy! Also tested the migration and permissions manually. No notes on the code 👍

I have a small UX nitpick, maybe for a future PR. I don't know if there was conversation about this but I wonder if the validation block on the Instance details page shouldn't just be completely hidden when the module is inactive?

Image

PR #3034, opened 2026-05-19
[IA-5072] n+1 queries on Form submissions

Crebert08:

This PR is about fixing the n+1 issues on the map tab on the form submissions list

### Related JIRA tickets

[IA-5072](https://bluesquare.atlassian.net/browse/IA-5072)

### Map tab

- For the map context, there is no need for pagination and no need to return the as_dict format, so I implemented a function to return id and location only in `/home/crebert/Desktop/iaso/iaso/api/instances/views.py`
- Added the new location type in `/home/crebert/Desktop/iaso/hat/assets/js/apps/Iaso/domains/instances/types/instance.ts`
- Defined a new request for the map tab in `/home/crebert/Desktop/iaso/hat/assets/js/apps/Iaso/domains/instances/requests.ts`
- Added a new endpoint
- Added some tests

### List Tab
- Added an annotation on the queryset in `/home/crebert/Desktop/iaso/iaso/api/instances/views.py`
- Added a fallback in the properties in `/home/crebert/Desktop/iaso/iaso/models/instances.py`
- Added a test and fixed one in `/home/crebert/Desktop/iaso/iaso/tests/api/instances/test_instances.py`

## How to test

Explain how to test your PR.

//

### Map tab
The queries go from this: 
image

To this: 
image

### List tab
The queries go from this:
image

To this:
image

## Notes

## Doc

[IA-5072]: https://bluesquare.atlassian.net/browse/IA-5072?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iaso/api/instances/views.py
```diff
@@ -416,6 +417,22 @@ def list(self, request):
...
+                return Response(
+                    [
+                        {
+                            "id": i.id,
+                            "latitude": i.location.y,
+                            "longitude": i.location.x,
+                        }
+                        for i in location_queryset
+                    ]
+                )
```
hugo093: few remarks here: 

1. I know that api code is already a mess, but try defining serializers to use as much as possible. 
here for example: 
```python 
 return Response(
                    [
                        {
                            "id": i.id,
                            "latitude": i.location.y,
                            "longitude": i.location.x,
                        }
                        for i in location_queryset
                    ]
                )
```

2. why do we automatically return an empty response if with_location=False ? I feel like it's not a filter anymore then, it should be a new endpoint instead. 

3. Instance.objects.filter(pk__in=queryset).filter().only() => so just queryset.filter().only() ? 

4. what if we want this paginated ? actually it's not possible. And if it's on purpose, then it regroups with what i'm saying at point 2: should be a new endpoint, not a filter

iaso/tests/api/instances/test_instances.py
```diff
@@ -3083,6 +3083,43 @@ def test_check_bulk_push_gps_unselected_ids_warning_both(self):
...
+        with self.assertNumQueries(5) as ctx:
+            response = self.client.get("/api/instances/", {"withLocation": "true"})
+
+        self.assertEqual(response.status_code, 200)
+        data = response.json()
+        self.assertIsInstance(data, list)
+
+        returned_ids = {item["id"] for item in data}
+        for instance in instances_with_location:
+            self.assertIn(instance.id, returned_ids)
```
hugo093: coulb be smth like : self.assertCountEqual([isntance.id for instance in instances_with_location], returned_ids)

hugo093 (changes requested):
good job , the query count drop is quite impressive. 
However, some changes are needed imho: coding and "structural"

hat/assets/js/apps/Iaso/domains/instances/index.tsx
```diff
@@ -89,18 +90,10 @@ const Instances = () => {
     // Data for the map
     const { data: instancesSmall, isLoading: loadingMap } = useSnackQuery(
         ['instances', 'small', apiParams],
-        // Ugly fix to limit results displayed on map, IA-904
-        () =>
-            fetchInstancesAsSmallDict(
-                `${getEndpointUrl(apiParams, false, '', true)}&limit=${
-                    params.mapResults || 3000
```
quang-le, line 96: The limit was there for a reason: too many points will break the map on the front-end side. Have you tested with over 3000 instances? 30000? 
Please don't change such behaviours without making sure it's safe to do so

iaso/api/instances/views.py
quang-le: agreed, also see my previous comment.
My take is that, if possible, it should be a new endpoint, that receives the same query params so the relevant filters can be applied

@beygorghor  WDYT?

quang-le (changes requested):

hat/assets/js/apps/Iaso/domains/instances/index.tsx
Crebert08: Should I remove the Location filter box on the map tab and just keep it on the other tabs then ?

The `Without` filter doesn't make much sense IMO because the map won't show anything anyway

@quang-le 
@beygorghor

iaso/api/instances/views.py
```diff
@@ -515,6 +516,26 @@ def anwser_with_parquet_file(self, request, filters, queryset):
...
+        limit = int(request.GET.get("limit", 3000))
+        queryset = (
+            self.get_queryset()
+            .exclude(file="")
+            .exclude(device__test_device=True)
+            .for_filters(**filters)
+            .filter(location__isnull=False)
+            .select_related(None)
+            .only("id", "location")[:limit]
+        )
```
hugo093: I understand why you do the select_related(None) , but imho this queryset construction should be done within the get_queryset() method of the viewsets, it increase readability because it's the purpose of this function :) 

So it would be something like : 

```python 
    def get_queryset(self):
        request = self.request
        queryset: InstanceQuerySet = Instance.objects.order_by("-id").filter_for_user(request.user).filter_on_user_projects(user=request.user)

        if self.action == "map":
            queryset = queryset.exclude(file="").exclude(device__test_device=True).filter(location__isnull=False).only("id", "location")
        else:
            queryset = queryset.select_related("form", "created_by", "last_modified_by")


        return queryset
```

And your in your action itself : 
```python
queryset = self.get_queryset().for_filters(**filters)[:limit]
```

Note that, using proper filter backends and pagination, you coudl even rewrite your endpoint like the standard DRF way (see their code for list): 
```python
    def list(self, request, *args, **kwargs):

        page = self.paginate_queryset(queryset)
        if page is not None:
            serializer = self.get_serializer(page, many=True)
            return self.get_paginated_response(serializer.data)

        serializer = self.get_serializer(queryset, many=True)
```

Also, if you're willing to , you can implement the get_serializer_class method on the viewsets and apply it to your action , so drf-spectacular understands you're using this schema (or you wait for me to go through this api and refactor it)

iaso/api/instances/serializers/misc.py
```diff
@@ -116,3 +116,18 @@ class Meta:
 class UnlockSerializer(serializers.Serializer):
     lock = serializers.PrimaryKeyRelatedField(queryset=InstanceLock.objects.all())
     # we will  check that the user can access from the directly in remove_lock()
+
+
+class InstanceLocationSerializer(serializers.ModelSerializer):
```
hugo093: use iaso builtin ModelSerializer , not rest_framework serializers.

iaso/api/instances/serializers/misc.py
```diff
@@ -116,3 +116,18 @@ class Meta:
...
+    longitude = serializers.SerializerMethodField()
+
+    class Meta:
+        model = Instance
+        fields = ["id", "latitude", "longitude"]
+
+    def get_latitude(self, obj):
+        return obj.location.y if obj.location else None
+
+    def get_longitude(self, obj):
```
hugo093, line 136: as usual, don't forget the @extend_schema_field on those serializermethod

hugo093 (changes requested):
few code changes, for cleanliness :) otherwise lgtm

hat/assets/js/apps/Iaso/domains/instances/index.tsx
beygorghor: we can keep it like thaht, it will add more complexity to something already quiet complex

beygorghor (commented):
i'm still having lot's of duplicate on the non map endpoint 

Image

hugo093 (approved):
for my part and the comments I had, LGTM. I'll let the others review :) Thanks for the changes

beygorghor (approved):

PR #3037, opened 2026-05-20
[IA-5105] fix user creation

tdethier:

 fix user creation when no password is set but password validation is enabled

### Related JIRA tickets

IA-5105

## Changes

- triggered password validation only if a password was received
- small cleanup of dead code

## How to test

- have an `Account` where `enforce_password_validation=True` (set it in the admin panel in case you need)
- login, go to the users page and create a new user:
  - leave the `password` field empty
  - fill out the `email` field
  - tick the `send invitation email` box
- you should not get an error about a missing password, the email should be sent and the user should be created

## Print screen / video

/

## Notes

- we'll probably need to hotfix this

## Doc

/

### DIFF
--- a/iaso/api/profiles/serializers/create.py
+++ b/iaso/api/profiles/serializers/create.py
@@ -2,7 +2,6 @@ import django.core.exceptions as django_exceptions
 
 from django.contrib.auth.models import Permission
 from django.contrib.auth.password_validation import validate_password
-from django.utils.crypto import get_random_string
 from django.utils.translation import gettext_lazy as _
 from rest_framework import serializers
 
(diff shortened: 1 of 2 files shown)


iaso/api/profiles/serializers/create.py
```diff
@@ -160,7 +150,7 @@ def create(self, validated_data):
         )
 
         try:
-            if account.enforce_password_validation:
+            if account.enforce_password_validation and "password" in validated_data:
```
hugo093: as password has allow_blank=True, I would do instead:  validated_data.get("password")

iaso/tests/api/profiles/test_views/test_create.py
```diff
@@ -284,6 +285,27 @@ def test_create_profile_with_no_password_and_not_send_email(self):
...
+        data = {
+            "user_name": "userTest",
+            "first_name": "unittest_first_name",
+            "last_name": "unittest_last_name",
+            "send_email_invitation": True,  # password validation should not cause any issue because none was set
+            "email": "test@test.com",
+        }
+
+        response = self.client.post(reverse("profiles-list"), data=data, format="json")
+        result = self.assertJSONResponse(response, status.HTTP_201_CREATED)
```
hugo093, line 303: you can also check that an email was sent : 

```python 
 with self.captureOnCommitCallbacks(execute=True) as callbacks:
            response = self.client.post(reverse("profiles-list"), data=data, format="json")

        response_data = self.assertJSONResponse(response, 201)
        self.assertEqual(len(callbacks), 1)
        self.assertEqual(len(mail.outbox), 1)

```

hugo093 (changes requested):
Works but minor detaisl to change

hugo093 (approved):
LGTM

PR #3039, opened 2026-05-21
[IA-5075] Add `display_name` to mobile Instance Validation's endpoint

bmonjoie:

The mobile needs to display the labels for the validation to help differentiate them to the user.

### Related JIRA tickets

IA-5075

## Changes

Added a new field called `display_name` (same as on mobile) that concatenates the `Instance`'s values based on the `Form`'s label keys, if any.

## How to test

1. Have an instance with a validation status.
1. Add label keys to the corresponding form.
1. Call the mobile endpoint (`/api/mobile/validation-workflows/ `) 
1. Check that the `display_name` field is correct.

### DIFF
--- a/iaso/api/validation_workflows/serializers/mobile.py
+++ b/iaso/api/validation_workflows/serializers/mobile.py
@@ -1,3 +1,5 @@
+from typing import Optional
+
 from drf_spectacular.utils import extend_schema_field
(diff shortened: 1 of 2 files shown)


hugo093 (approved):
lgtm

PR #3043, opened 2026-05-21
[IA-5106]: new fields for /api/etl/instances

hugo093:

Following the validation workflow feature and discussion, data scientist need more fields for the instance endpoint:
* one legacy field that represents the validation status related to OU changes
* adapting the history field so it's the history of all the validation nodes 

### Related JIRA tickets

IA-5106

## Changes

Adapted backend + tests

## How to test

Call /api/etl/instances

* a new field org_unit_validation_status should be present, representing the OU change request status 
* history field has been adapted to provide a timeline of all the validation nodes for the validation workflow feature


dynamic_fields/filter_backends.py
```diff
@@ -39,7 +39,7 @@ def filter_queryset(self, request, queryset, view):
         if serializer_class:
             valid = serializer_class.get_valid_options()
             invalid = set(values) - set(valid)
-            if invalid:
+            if invalid and "__all__" not in valid:
```
quang-le: why not just filter out `"__all__" from `get_valid_options`?

iaso/api/common/views.py
```diff
@@ -20,6 +20,10 @@ class ModelViewSet(
     pass
 
 
+class ReadOnlyModelViewSet(RetrieveModelMixin, CustomPaginationListModelMixin, GenericViewSet):
```
quang-le, line 23: Why not extend EtlModelViewSet?

quang-le (commented):
Juts 2 questions

iaso/api/common/views.py
hugo093: Cause ETLModelViewSet is a full ModelViewSet (it then has delete create update .. ) and imho ETLModelViewSet should be a mixin :/ 

But maybe in another ticket ? Cause I see some potential issues with those pagination mixin tbh

dynamic_fields/filter_backends.py
hugo093: you're right, I changed this :)

iaso/api/common/views.py
hugo093: created IA-5137 because indeed, something was buggy x)

quang-le (approved):

PR #3044, opened 2026-05-22
[IA-5112] fix: don't display bypass nodes if the user can't take action

hugo093:

In the validation workflow timeline:
* bypass validation nodes were displayed even though the user can't take any action. 
* User , for a pending submission, wanted to still see all the steps. 

### Related JIRA tickets

IA-5112

## Changes

Adapted front end and backend to change display of validation timeline: 

* If the submission is APPROVED or REJECTED: just display the history 
* If the submission is pending : 
  * We display all the steps 
  * Bypass steps where user cannot take any action or steps without any pending action are displayed greyed out without any text beside the title
  * Bypass step where user can take action and/or next pending action are displayed in orange with two buttons "approve" and "reject"

## How to test

Create a validation workflow with some bypass nodes and required roles
Trigger that workflow for an instance

User with the roles should see the bypass nodes 
User without the roles should not see them

## Notes

Will write fe tests once orval is implemented, will be easier with msw


quang-le (approved):
Nice!

PR #3045, opened 2026-05-22
[IA-4741] remove current user from user management list

62496:

Remove the current user from the user management list when listing manageable users.

### Related JIRA tickets

IA-4741

## Changes

Updated the managed users listing policy so users with admin permissions no longer see their own profile in `GET /api/profiles/?managedUsersOnly=true`.

Added backend tests for both superuser and user admin cases to ensure the current user is excluded while the rest of the manageable profiles are still returned.

## How to test

1. Log in with a user admin account.
2. Open the Users / User management page.
3. Confirm the logged-in user does not appear in the table.
4. Confirm other users from the account are still listed.

Run the tests : 
```bash
iaso.tests.api.profiles.test_views.test_list.ProfileListAPITestCase
```
## Print screen / video

image

### DIFF
--- a/iaso/api/profiles/policies.py
+++ b/iaso/api/profiles/policies.py
@@ -126,7 +126,7 @@ class ManagedUsersPolicy:
     @staticmethod
     def authorize_list(requester, queryset):
         if requester.has_perm(CORE_USERS_ADMIN_PERMISSION.full_name()):
-            return queryset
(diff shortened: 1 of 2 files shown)


quang-le (changes requested):
This will not do: no users have no path in the UI to see or update their own data. I think this ticket requires more spec than you were given or should be done after other related tickets

PR #3047, opened 2026-05-26
[IA-4884] nfc cards hidden when no external storage module

Crebert08:

This PR is about rendering the nfc_cards fields on the entity profile when the user has the right module permissions

### Related JIRA tickets

[IA-4884](https://bluesquare.atlassian.net/browse/IA-4884)

## Changes

- Overrided the serializer to check if the user has the core storage permissions and the right modules
- Changed the frontend to render this field dynamically after checking the permissions and modules
- Tested the backend behavior
- Added a function in the instances views to check if the user has the right permissions before returning the full entity object
- Tested this behavior

## How to test

- Connect to PostgreSQL in docker with `sudo docker compose exec db psql -U postgres -d postgres`
- Connect to your database
- Run this query in docker to see the users without the core storage permissions:
```
SELECT id, username, email 
FROM auth_user
WHERE is_superuser = FALSE 
  -- 1. Exclude users who have the permission directly
  AND id NOT IN (
      SELECT aup.user_id 
      FROM auth_user_user_permissions aup
      JOIN auth_permission ap ON aup.permission_id = ap.id
      WHERE ap.codename = 'iaso_storages'
  )
  -- 2. Exclude users who have the permission via a group
  AND id NOT IN (
      SELECT aug.user_id 
      FROM auth_user_groups aug
      JOIN auth_group_permissions agp ON aug.group_id = agp.group_id
      JOIN auth_permission ap ON agp.permission_id = ap.id
      WHERE ap.codename = 'iaso_storages'
  )
  -- 3. NEW: Only include users who have MORE THAN 1 unique permission in total
  AND id IN (
      SELECT user_id 
      FROM (
          -- Get all direct permissions
          SELECT user_id, permission_id 
          FROM auth_user_user_permissions
          
          UNION
          
          -- Get all group permissions
          SELECT aug.user_id, agp.permission_id 
          FROM auth_user_groups aug
          JOIN auth_group_permissions agp ON aug.group_id = agp.group_id
      ) AS all_user_permissions
      GROUP BY user_id
      HAVING COUNT(DISTINCT permission_id) > 1
  );
```
- Connect to iaso with one of the users without the permissions
- The NFC Cards fields shouldn't display

## Print screen / video

https://github.com/user-attachments/assets/3662e82d-8a19-4e10-bf1a-1cd32e1adcbe

## Notes

//

## Doc

[IA-4884]: https://bluesquare.atlassian.net/browse/IA-4884?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hat/assets/js/apps/Iaso/domains/entities/hooks/useGetEntityFields.ts
Phil-V, line 1: Here, instead of inspecting the payload for the presence of the NFC key, I think a cleaner approach would be to check on the frontend side if the user has access to the external storages module.
See `userHasPermission` and `userHasAccessToModule` for examples of how to do that.

iaso/api/entities/views.py
Phil-V, line 1: Instead of modifying the serializer dynamically in the view, it might be better to implement this on the serializer. See `get_nfc_cards` on the `EntitySerializer`.

iaso/api/instances/views.py
Phil-V, line 1: This one is tricky. It's best to avoid burdening the view with too much logic, as you can see it gets pretty messy over time.

Currently, the response for the detail view is built with `instance.as_full_model(with_entity=True)`, which calls `as_small_dict_with_nfc_cards` on the `Entity` model, which itself calls the `get_nfc_cards` method. That might be a better place for the logic you have implemented.

In general, I think it's best not to change the API surface dynamically based on permissions. My preference would be to return `None` for `nfc_cards` but retain the key in the payload when the user lacks the permission.

Phil-V (changes requested):
Hi @Crebert08! Thanks for the work! Here are a couple of recommendations.

iaso/api/entities/views.py
Crebert08: My initial thought was to implement the permissions at the view level but you're right, this case requires handling the permission check directly in the serializer

Phil-V:
Hi @Crebert08 Thanks for the changes. I've feel I've led you astray with my comments 😅

There's one more little fix to make :
- On the frontend, in addition to what you have done, can you add a check for userHasAccessToModule as well, so that the field will be hidden for admins as well when the external storage module is disabled? You can see an example of how it's used in hat/assets/js/apps/Iaso/components/validationWorkflows/ValidationWorkflowDropdown.tsx

Phil-V (approved):
Well done! 👍 Appreciate the extra tests.

PR #3049, opened 2026-05-27
[IA-5123] [IA-5138]: Account management feature

hugo093:

Implemented account management feature

### Related JIRA tickets

IA-5123 IA-5138

### Backend

Here's a summary of the API related changes (note that everything is available under `/swagger-ui/`):  

1. new endpoints for account feature flags

| Method | URL                              | Old URL | Notes |
|--------|----------------------------------|---------|-------|
| GET    | /account-feature-flags/          | -       | New   |
| GET    | /account-feature-flags/dropdown/ | -       | New   |

2. new endpoints for account   

| Method    | URL                                     | Old URL             | Notes                                                       |
|-----------|-----------------------------------------|---------------------|-------------------------------------------------------------|
| GET       | /accounts/                              | -                   | New, does not display anthropic_api_key                     |
| GET       | /accounts/{id}/                         | -                   | New, does not display anthropic_api_key                     |
| PUT/PATCH | /accounts/{id}/                         | -                   | New, does not update anthropic_api_key                      |
| GET       | /accounts/{id}/ai-api-key/              | -                   | New, output truncated anthorpic_api_key                     |
| PUT       | /accounts/{id}/ai-api-key/              | -                   | New                                                         |
| DELETE    | /accounts/{id}/ai-api-key/              | -                   |                                                             |
| GET       | /api/accounts/{id}/custom-translations/ | custom_translations | New URL (was GET /api/custom_translations/?account_id={id}) |
| PUT/PATCH | /api/accounts/{id}/set-default-version/ | set_default_version | New URL (was PUT /api/accounts/{id}/)                       |
| GET       | /api/accounts/me/                       |                     | New, does not display anthropic_api_key                     |
| POST      | /api/accounts/switch/                   | switch              | Changed method (was PATCH)                                  |

3. Changed permission on /api/modules/

4. Fixed a bug spotted in profiles serializers

5. Added tests 

### Frontend

1. Created and/or adapted form components that better integrates with formik (with unit and a11y tests):
    1. `BooleanInput`
    2. `PasswordInput`
    3. `ArrayCheckboxInput`
2.  Added menu entry `Accounts` under `Settings`
3. Added the needed pages and components for this account managements, all under `domains/accounts`
   1. An account list page 
   2. An account detail page
   3. An account edit page
   4. A modal that allows to edit the AI api key
4. Following some backend changes, adapted the related query calls / hooks
5. Integrated zod locale change within `LocaleContext `
6. Created two methods under utils/forms that allows us to display Drf errors directly in formik. Thus allowing us also to display errors translated from the backend without having to translate it again in the front end :tada: 
7. Orval related configuration for the new endpoints
8. Tests, tests, tests

### Other

* Removed a CI step from ci.yml in plugins that didn't have any purpose
* Updated doc in `api_synchronization.en.md`
* Updated eslint config to avoid conflicts between prettier and eslint/react
* Updated vitest.config.ts to add another project : `a11y`
* Updated vitest.config.ts to restore integration tests that were left out in the filter

## How to test

* Python and front end tests
* Try to switch account in the front end, should still work
* Try to get custom translations in the front end should still work
* Play around the new account UI 

## Notes

* All files under `hat/assets/js/apps/Iaso/api` are orval generated files and thus don't need any review
* Some follow up tickets have been created because this PR is already way too big : IA-5169 IA-5168 IA-5178
* Should we implement api tests ? (cfr `validationWorkflow.api.test.tsx`)


iaso/api/accounts/serializers/retrieve.py
```diff
@@ -0,0 +1,18 @@
...
+        model = Account
+        fields = [
+            "id",
+            "name",
+            "created_at",
+            "user_manual_path",
+            "forum_path",
+            "modules",
+            "enforce_password_validation",
+            "anthropic_api_key",
```
mestachs: 1. I would not put it in the defaults... 
2. is this really intented to be used by the console or the page ? 
3. if it's the real key we want to return I would put that in permission (source admin looks like a weak permission) 

IMO it should be considered as a write only field
and perhaps returning a "diagnosable" version  sk-ant-api....xxx showing last numbers but never the unencrypted key in clear text

iaso/api/accounts/views.py
```diff
@@ -0,0 +1,150 @@
...
+    http_method_names = ["get", "options", "patch", "put", "head", "trace", "post"]
+    pagination_class = AccountPagination
+    filter_backends = [OrderingFilter]
+    ordering = ["id"]
+
+    @property
+    def permission_classes(self):
+        if self.action == "set_default_version":
+            return [IsAuthenticated, HasPermission(CORE_SOURCE_PERMISSION)]
+        return [IsAuthenticated]
```
mestachs: Am I wrong to say that any user authenticated will get access to the anthropic key ?
and Any one authenticated can patch/update ?

iaso/models/account.py
```diff
@@ -22,12 +24,28 @@ def __str__(self):
...
+                profile__in=tenant_user.main_user.tenant_users.values_list(
+                    "account_user__iaso_profile__id",
+                    flat=True,
+                )
+            )
+
+        return self.filter(profile=user.iaso_profile)
+
+
+class Account(CreatedAndUpdatedModel):
```
mestachs, line 43: apparently we already have created_at, updated_at on iaso_account
don't we need in a migration to officialize the inheritance change ?

iaso/api/accounts/views.py
```diff
@@ -0,0 +1,150 @@
...
+            raise NotFound
+
+        qs = list(self.get_queryset())
+
+        other_account_qs = [a for a in qs if a.id != request.user.iaso_profile.account_id]
+        instance = next(iter([a for a in qs if a.id == request.user.iaso_profile.account_id]), None)
+        if not instance:
+            raise NotFound
+
+        # other_accounts_qs = Account.objects.filter_for_user(self.request.user).distinct("id").exclude(id=request.user.iaso_profile.account_id)
```
mestachs: dead code ?

iaso/api/accounts/views.py
```diff
@@ -0,0 +1,150 @@
...
+        instance = self.get_object()
+        serializer = self.get_serializer(instance, data=request.data)
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+        return Response(status=status.HTTP_204_NO_CONTENT)
+
+    @extend_schema(responses={200: AccountRetrieveCurrentSerializer})
+    @action(detail=False, methods=["GET"], url_path="me")
+    def me(self, request):
+        if not request.user and not request.user.iaso_profile and not request.user.iaso_profile.account:
```
mestachs: does it happen that it can be true it's not a OR ?

mestachs:
While the structure looks better... it's hard to "detect/understand" all the api changes.

so from my understanding in API endpoint changes we have (sometime would dream of committing the openapi schema and see the changes in the PR, but probably not super easy to handle for diffs)

# New/modified

| Method | URL | Action | Change |
|--------|-----|--------|--------|
| `GET` | `/api/accounts/` | `list` | **New** (was 405) |
| `GET` | `/api/accounts/{id}/` | `retrieve` | **New** (was 405) |
| `PUT`/`PATCH` | `/api/accounts/{id}/` | `update` | **Changed** — now updates many fields, was only `default_version` |
| `GET` | `/api/accounts/{id}/custom-translations/` | `custom_translations` | **New URL** (was `GET /api/custom_translations/?account_id={id}`) |
| `PUT`/`PATCH` | `/api/accounts/{id}/set-default-version/` | `set_default_version` | **New URL** (was `PUT /api/accounts/{id}/`) |
| `POST` | `/api/accounts/switch/` | `switch` | **Changed method** (was `PATCH`) |
| `GET` | `/api/accounts/me/` | `me` | **New** |

# Deleted

| Method | URL | Reason |
|--------|-----|--------|
| `GET` | `/api/custom_translations/` | Moved under accounts |

looks safe since it's a iaso front end concern and you adapted the code


My main point is exposing the anthropic key in clear text... I would handle that differently

iaso/models/account.py
hugo093: No need for migration :)

hugo093:
@mestachs thanks for the review; while working on the front end , I noticed the same issues and made some changes  => this [PR](https://github.com/BLSQ/iaso/pull/3067) ; I will merge it when I'm finished and ask for a complete review (maybe with a front end dev)

iaso/api/accounts/views.py
hugo093: Yes , modified that in the other PR (see my general comment)

iaso/api/accounts/views.py
hugo093: don't get it, sorry

hat/assets/js/apps/Iaso/components/forms/PasswordInput.tsx
beygorghor, line 1: we already have a PasswordInput in bluesquare-component...

hat/assets/js/apps/Iaso/components/papers/WidgetPaperRow.tsx
```diff
@@ -13,10 +13,13 @@ const useStyles = makeStyles(theme => ({
...
     field: { label: string; value: any };
-    placeholder?: string
+    placeholder?: string;
 };
 
-export const WidgetPaperRow = ({ field, placeholder=textPlaceholder }: RowProps) => {
+export const WidgetPaperRow = ({
+    field,
+    placeholder = textPlaceholder,
+}: RowProps) => {
```
beygorghor: let's fit to FunctionComponent for now, has the rest of the code base is using it

hat/assets/js/apps/Iaso/domains/accounts/components/modals/EditAIApiKeyModal.tsx
```diff
@@ -0,0 +1,87 @@
...
+            anthropic_api_key: '',
+        },
+        validationSchema: toFormikValidationSchema(
+            AccountUpdateAIApiKeyRequest,
+        ),
+        onSubmit: withFormikSubmitAsync(values =>
+            saveAIApiKey({
+                id: accountId,
+                data: values,
+            }).then(() => closeDialog()),
```
beygorghor: i prefer using `onSuccess` on `useApiAccountsAiApiKeyUpdate`

hat/assets/js/apps/Iaso/domains/accounts/config.tsx
```diff
@@ -0,0 +1,49 @@
...
+                Cell: DateCell,
+            },
+            {
+                Header: formatMessage(MESSAGES.actions),
+                id: 'actions',
+                accessor: 'actions',
+                sortable: false,
+                Cell: (settings: Setting<any>) => {
+                    return (
+                        <>
```
beygorghor: This fragment is useless

hat/assets/js/apps/Iaso/domains/accounts/details.tsx
```diff
@@ -0,0 +1,173 @@
...
+    return {
+        id: `account-tab-${value}`,
+        'aria-controls': `account-tabpanel-${value}`,
+    };
+}
+
+const useStyles = makeStyles((theme: any) => {
+    return { ...commonStyles(theme) };
+});
+const AccountsDetails = () => {
```
beygorghor: Again use FunctionComponent here please

hat/assets/js/apps/Iaso/domains/accounts/details.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+    const { data: modules, isLoading: isLoadingModules } =
+        useGetModulesDropDown();
+
+    const {
+        data: accountFeatureFlags,
+        isLoading: isLoadingAccountFeatureFlags,
+    } = useApiAccountFeatureFlagsDropdownList();
+
+    const [tab, setTab] = useState('general');
```
beygorghor: we have a useTabs in bluesqaure-components

hat/assets/js/apps/Iaso/domains/accounts/details.tsx
```diff
@@ -0,0 +1,173 @@
...
+        isLoading ||
+        isLoadingModules ||
+        isLoadingAIApiKey ||
+        isLoadingAccountFeatureFlags;
+
+    const handleChange = (_: React.SyntheticEvent, newValue: string) => {
+        setTab(newValue);
+    };
+
+    const baseRedirectUrl = `${baseUrls.accounts}`;
```
beygorghor: We can put this one outside of the component

hat/assets/js/apps/Iaso/domains/accounts/details.tsx
```diff
@@ -0,0 +1,173 @@
...
+                    goBack={() => redirectTo(baseRedirectUrl)}
+                />
+                <LoadingSpinner />
+            </>
+        );
+    }
+
+    if (!account) {
+        return <Page404 displayTopBar={true} />;
+    }
```
beygorghor, line 104: this is useless, here those pages are only visible if the user has an account

hat/assets/js/apps/Iaso/domains/accounts/details.tsx
```diff
@@ -0,0 +1,173 @@
...
+import { CustomTabPanel } from 'Iaso/domains/accounts/components/details/CustomTabPanel';
+import { GeneralInfoPanel } from 'Iaso/domains/accounts/components/details/GeneralInfoPanel';
+import { ModulePanel } from 'Iaso/domains/accounts/components/details/ModulePanel';
+import { useGetModulesDropDown } from 'Iaso/domains/setup/hooks/useGetModulesDropDown';
+import { userHasAccessToModule } from 'Iaso/domains/users/utils';
+import { useParamsObject } from 'Iaso/routing/hooks/useParamsObject';
+import { useCurrentUser } from 'Iaso/utils/usersUtils';
+import MESSAGES from './messages';
+
+function a11yProps(value: string) {
```
beygorghor: arrow function ?

hat/assets/js/apps/Iaso/domains/accounts/edit.tsx
```diff
@@ -0,0 +1,158 @@
...
+import { ModulesEditPanel } from 'Iaso/domains/accounts/components/edit/ModulesEditPanel';
+import { useGetModulesDropDown } from 'Iaso/domains/setup/hooks/useGetModulesDropDown';
+import { useParamsObject } from 'Iaso/routing/hooks/useParamsObject';
+import { withFormikSubmitAsync } from 'Iaso/utils/forms';
+import MESSAGES from './messages';
+
+const useStyles = makeStyles((theme: any) => {
+    return { ...commonStyles(theme) };
+});
+export const AccountsEdit = () => {
```
beygorghor: FunctionComponent please

eslint.config.mjs
```diff
@@ -168,8 +168,8 @@ export default defineConfig([
                 { extensions: ['.js', '.jsx'] },
             ],
 
-            'react/jsx-indent': ['error', 4],
-            'react/jsx-indent-props': ['error', 4],
+            'react/jsx-indent': 'off',
+            'react/jsx-indent-props': 'off',
```
beygorghor, line 203: Why ?

beygorghor (changes requested):
- If can try to make everything fit in a non scrollable window it can be nice. Modules an feature flags should be scrollable with a fix header.
- Tables should use small rows (size="small"), those are way too big, and maybe on row one to a bit darker.
- eslint is now complaining in the CI 👼

hat/assets/js/apps/Iaso/domains/accounts/details.tsx
hugo093: I disagree there: 
* In the case of multi account, you could access another account
* It's still worth it in case you manipulate the  URL directly (e.g /dashboard/settings/accounts/detail/accountId/1/id/wron_thing_there/)

eslint.config.mjs
hugo093: was clashing with prettier and imho this kind of stuff belong to prettier not eslint

hat/assets/js/apps/Iaso/components/forms/PasswordInput.tsx
hugo093: I don't mind using the one in bluesquare-component , but it seems I can't do something as easy as : 
```typescript
<Field
                        label={formatMessage(MESSAGES.aiApiKeyLabel)}
                        name={'anthropic_api_key'}
                        component={PasswordInput}
                        margin={'normal'}
                        required
                    />
```

quang-le: We have a few cases like that in the polio codebase. I think the convoluted but best way is to do a wrapper around `InputComponent` and make it formik compatible. That way you get the flexibility of `Field` and we still get the benefit of having (mostly) one `InputComponent` to rule them all

hat/assets/js/apps/Iaso/domains/accounts/details.tsx
```diff
@@ -0,0 +1,181 @@
...
+const baseRedirectUrl = `${baseUrls.accounts}`;
+
+const AccountsDetails: FunctionComponent = () => {
+    const { formatMessage } = useSafeIntl();
+    const params = useParamsObject(baseUrls.accountsDetail);
+
+    const accountId = parseInt(params.id);
+
+    const { data: account, isLoading } = useApiAccountsRetrieve(
+        parseInt(params.id),
```
quang-le: you're parsing `params.id` twice

hat/assets/js/apps/Iaso/domains/accounts/details.tsx
```diff
@@ -0,0 +1,181 @@
...
+        undefined,
+        { query: { enabled: !!accountId } },
```
quang-le: I personally prefer `Boolean(accountId)`

hat/assets/js/apps/Iaso/domains/accounts/components/edit/ModulesEditPanel.tsx
```diff
@@ -0,0 +1,74 @@
...
+                                    </TableRow>
+                                );
+                            })}
+                        </TableBody>
+                    </Table>
+                </>
+            ) : (
+                <Alert severity={'info'} sx={{ mb: 2 }}>
+                    {formatMessage(MESSAGES.noResultsFound)}
+                </Alert>
```
quang-le: I don't get this error handling: 
- If `modules?.length` is falsy, that means something went wrong with the API call, and in that case the Alert/Snackbar should already be handled by the hook, shouldn't it?
- No results found is misleading: you get the active modules from the account retrieve call, here it's the modules list that couldn't be fetched

hat/assets/js/apps/Iaso/domains/accounts/components/edit/ModulesEditPanel.tsx
quang-le, line 1: I think it would make sense to display the active module as a fallback in case the modules dropdown cannot be fetched (and display the keys). It's an edge case, but since you handle it with alert, it would complement that behaviour

hat/assets/js/apps/Iaso/domains/accounts/details.tsx
quang-le: @hugo093 not sure which case you're referring to, but:
- user with multiple accounts needs to select the active account in the interface. I'm not sure this protects them from updating another account they have access to, but worth checking
- IIRC we do have a check preventing users accessing data from other accounts everywhere in the web app, so changing the accountId in the url should not lead to updating the wrong account. Worth double checking, but I'm reasonably confident on this one

hat/assets/js/apps/Iaso/domains/accounts/edit.tsx
```diff
@@ -0,0 +1,167 @@
...
+                    goBack={() => redirectTo(redirectBackUrl)}
+                    displayBackButton
+                />
+                <LoadingSpinner />
+            </>
+        );
+    }
+    if (!generalLoading && !data) {
+        return <Page404 displayTopBar={true} />;
+    }
```
quang-le, line 95: In general when we display a table and no results are found we show something like:

Image

hat/assets/js/apps/Iaso/domains/accounts/edit.tsx
```diff
@@ -0,0 +1,167 @@
...
+                            />
+                        </Grid>
+                        <Grid item xs={12} md={6}>
+                            <ModulesEditPanel modules={modulesData} />
+                        </Grid>
+                        <Grid
+                            item
+                            xs={12}
+                            sx={{
+                                justifyContent: 'space-between',
```
quang-le: I know the UI isn't super consistent, but we tend to keep cancel and save buttons closer together

Image

eslint.config.mjs
quang-le: Funny I was advocating dropping prettier for the same reason. We should discuss that together before changing though

iaso/api/account_feature_flags/views.py
```diff
@@ -0,0 +1,44 @@
...
+from iaso.api.account_feature_flags.serializers.dropdown import AccountFeatureFlagDropdownSerializer
+from iaso.api.account_feature_flags.serializers.list import AccountFeatureFlagListSerializer
+from iaso.api.common import HasPermission
+from iaso.api.common.mixin import CustomPaginationListModelMixin
+from iaso.models import AccountFeatureFlag
+from iaso.permissions.core_permissions import CORE_ACCOUNT_MANAGEMENT_PERMISSION
+
+
+@extend_schema(tags=["Account feature flags"])
+class AccountFeatureFlagViewSet(CustomPaginationListModelMixin, GenericViewSet):
```
quang-le, line 18: Why not ModelViewSet? Also I'd have expected an `http_method_names` property

iaso/api/accounts/views.py
```diff
@@ -0,0 +1,187 @@
...
+        if self.action == "me":
+            qs = qs.select_related(
+                "default_version", "default_version__data_source", "default_version__data_source__credentials"
+            ).prefetch_related("feature_flags")
+        return qs
+
+    @extend_schema(responses={204: None})
+    def update(self, request, *args, **kwargs):
+        super().update(request, *args, **kwargs)
+        return Response(status=status.HTTP_204_NO_CONTENT)
```
quang-le, line 103: why 204 no content? We sometimes use the API response from an update to avoid refetching: we just update the react-query cache with the response. Not saying it's what we should do here, but curious about the override

quang-le (changes requested):
Huge job, congrats!

Nice to see Orval in action, though I must say I find the generated code hard to read.

Some Ui comments on top of what I left in the code:

- Use the same icons as the existing  Modules table (and for account FF as well):
Image

- I know you followed the mock up on this , bu t I think the tabs should be in the top bar like similar pages in iaso (There might even be some reusable code, but no promises):

Image


That's pretty nitpicky considering the size of the work

iaso/api/account_feature_flags/views.py
hugo093: ModelViewSet implies that you can do those actions : 
* destroy
* update
* list
* get
* patch
* create

Where here, doing it so makes it that you can only do list (and other custom actions defined). Hence no need for http_method_names

iaso/api/accounts/views.py
hugo093: My serializer actually returns nothing (all fields are write_only) , so 204 is more logic here

hat/assets/js/apps/Iaso/domains/accounts/edit.tsx
hugo093: But there it's about the edit; it's just a safeguard when you try to access the url directly and modify the id to something that returns a 404

hat/assets/js/apps/Iaso/domains/accounts/components/edit/ModulesEditPanel.tsx
hugo093: Yeah it's honestly an edge case that shouldn't happen. 

If modules.length is falsy, it just means that there's no module output from the backend (not an error, just no results) ; which shouldn't happen but better be prudent. 

I could change that though to display only the active modules retrieved from account, but it goes against the mock up that was provided.

hat/assets/js/apps/Iaso/components/forms/PasswordInput.tsx
hugo093: Can we maybe do it in another ticket then ?

hat/assets/js/apps/Iaso/domains/accounts/details.tsx
hugo093: screencapture-localhost-8000-dashboard-forms-submission-accountId-1-instanceId-259hhh-referenceFormId-3-2026-06-10-07_28_49
It's just to avoid this kind of things :) (where i put some wrong id in the url , sure the api returns a 404, but page displayed is not UI friendly ? http://localhost:8000/dashboard/forms/submission/accountId/1/instanceId/259hhh/referenceFormId/3 )

eslint.config.mjs
hugo093: Issue there is that i had some code that wouldn't pass both eslint and prettier (going back and forth between two) :sweat_smile: so that's why i turned it off

beygorghor:
Screenshot 2026-06-10 at 11 55 40
I did some re layout on the frontend part to fit on small screens

beygorghor (approved):

iaso/api/accounts/views.py
Phil-V: I'll echo the comment made by Stéphan, there's something off about this condition. You probably meant to chain ORs rather than ANDs.

hat/assets/js/apps/Iaso/components/forms/PasswordInput.tsx
quang-le: ok, but pls open it then

hat/assets/js/apps/Iaso/domains/accounts/details.tsx
quang-le: fine by me then

hat/assets/js/apps/Iaso/domains/accounts/edit.tsx
quang-le: ok then

iaso/api/accounts/views.py
hugo093: ha yes, well spotted indeed (now i get it 😅 )

Phil-V (approved):
Very good 👍

Everything LGTM on the backend side.

Not related to your work per se but if there's a future UX pass:

- activating the form AI module doesn’t reveal the api key setting unless a page refresh is forced
- naming: I think “address” is more common than  “chemin” or “path”  to refer to a full URL

quang-le (approved):

PR #3060, opened 2026-06-01
[IA-5143] increase max lenght formattachment file

Crebert08:

This PR is about increasing the max_length on the file field on FormAttachment

### Related JIRA tickets

[IA-5143](https://bluesquare.atlassian.net/browse/IA-5143)

## Changes

- Increased the max_length to 512 characters
- Added a better error handling in case the file name is longer than 512
- Tested the behavior
- Migration

## How to test

//

## Print screen / video

//

## Notes

## Doc

[IA-5143]: https://bluesquare.atlassian.net/browse/IA-5143?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iaso/api/form_attachments.py
```diff
@@ -40,6 +41,7 @@ def get_scan_timestamp(self, obj: FormAttachment):
 
     def validate(self, data: typing.MutableMapping):
         form: Form = data["form"]
+        file: InMemoryUploadedFile = data.get("file", None)
```
mestachs: are you it's the correct type ? I think dependening on the size of the file it can one of the two types
- InMemoryUploadedFile
- TemporaryUploadedFile

UploadedFile is the base class for both of them I think

mestachs: note that get("key", None) == get("key")
get just prevent a KeyError if the key isn't in the dict

and technically the type should be 
 file: UploadedFile | None = data.get("file")

mestachs (approved):

PR #3062, opened 2026-06-01
[IA-5073] fix model imports

tdethier:

Fix model imports in `0334` for localhosting

### Related JIRA tickets

IA-5073

## Changes

Change how models are imported in `0334` migration file

## How to test

- can't reproduce the same issue unless you have an old IASO environment
- just make sure that the migration still works (see CI)

## Print screen / video

/

## Notes

/

## Doc

/


hugo093 (approved):
LGTM (can't refuse my own suggestion :P )

PR #3068, opened 2026-06-02
POLIO-2148: add stricter exclusion rule

quang-le:

Sheet names with whitespace cannot be excluded using json config

### Related JIRA tickets

POLIO-2148

## Changes

- Added additional hardcoded string to to exclusion list for preparedness worksheet names

## How to test

- reproduce campaign in ticket on polio DB
- start worker
- launch preparedness refresh--> no 500

## Print screen / video
N/A

### DIFF
--- a/plugins/polio/preparedness/parser.py
+++ b/plugins/polio/preparedness/parser.py
@@ -26,6 +26,7 @@ def is_worksheet_exluded(worksheet, exclusions_list):
         worksheet.is_hidden
         or worksheet.title.lower() in WORKSHEET_EXCLUSION_LIST
         or worksheet.title.lower().startswith("sheet")
+        or worksheet.title.lower().startswith("synt_")
         or (worksheet.title in exclusions)
     )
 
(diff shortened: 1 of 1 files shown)


plugins/polio/preparedness/parser.py
```diff
@@ -26,6 +26,7 @@ def is_worksheet_exluded(worksheet, exclusions_list):
         worksheet.is_hidden
         or worksheet.title.lower() in WORKSHEET_EXCLUSION_LIST
         or worksheet.title.lower().startswith("sheet")
+        or worksheet.title.lower().startswith("synt_")
```
hugo093: combine the two startswith (see https://stackoverflow.com/questions/26040346/startswith-more-arguments )

hugo093 (changes requested):
small fix to do, otherwise lgtm

plugins/polio/preparedness/parser.py
hugo093:
```suggestion
or worksheet.title.lower().startswith(("sheet", "synt_"))
```

hugo093 (approved):
LGTM

PR #3072, opened 2026-06-02
POLIO-2149: reorder permission check

quang-le:
Users with right permission were getting a 403

### Related JIRA tickets

POLIO-2149

## Changes

- Re-order permission check so users with multiple perms have the stronger one checked first

## How to test

Explain how to test your PR.
On a polio DB
Create a user with Vaccine stcok management non-admin permission
Log in as this user
Create a temporary form A
Save
Edit the temp form A to add a file
Save


hugo093 (approved):
LGTM

PR #3076, opened 2026-06-04
POLIO-2023: handle dateless VRF in vaccine repository

quang-le:

Some documents can be uploaded without specifying the date field, so the API returns a `null`date. The front-end was not able to handle this case and treated the file as non-existent, though the API was returning it

### Related JIRA tickets

POLIO-2023

## Changes

- Change the conditional rendering of the table cells
- Add a custom message
- Fix some imports and typings

## How to test

Go to polio > supply chain
create a VRF without reception date, but with a file.
save
Go to vaccine repository
Look for the VRF 
It should appear with "No date found message"

## Print screen / video

https://github.com/user-attachments/assets/70b1379d-ebb0-499b-8304-26acf925ba3b


plugins/polio/js/src/domains/VaccineModule/Repository/forms/hooks/useVaccineRepositoryColumns.tsx
```diff
@@ -32,7 +32,7 @@ export const useVaccineRepositoryColumns = (
                 id: 'number',
                 accessor: 'number',
                 width: 20,
-                Cell: settings => (
+                Cell: (settings: Setting<{ number: number }>) => (
```
beygorghor, line 35: ❤️

beygorghor (approved):

PR #3080, opened 2026-06-04
POLIO-1199 improve polio admin

quang-le:

Explain here in one sentence.

### Related JIRA tickets

POLIO-1199

## Changes

- Improved list layout for polio models in admin:
    - Campaign
    - CampaignGroup
    - Round
    - RoundDateHistoryEntry
    - Subactivity
    - VaccineRequestForm
    -  OutgoingStockMovement (form a)
    - IncidentReport
    - DestructionReport
    - EarmarkedStock
    - VaccineStcokHistory
    - Chronogram
- Added some utils and Mixins for the polio admin 
- Marked vaccine (stock) related fields on Round model as deprecated in both F-E and B-E
- 

## How to test

On a polio DB with plugin activated, go to the page of listed models

### DIFF
--- a/iaso/locale/fr/LC_MESSAGES/django.po
+++ b/iaso/locale/fr/LC_MESSAGES/django.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: \n"
 "Report-Msgid-Bugs-To: \n"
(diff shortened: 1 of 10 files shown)


hugo093 (approved):
LGTM. 

The only "complaint" I've here is that admin.py should be splitted and reorganized into multiple files, one file per model :) (it's already 800+ lines for a django admin )

PR #3083, opened 2026-06-05
[IA-5148]: New modules API  + frontend

hugo093:

There was a bug that would allow modules api to display modules related to plugins that were not activated. 

### Related JIRA tickets

IA-5148

## Changes

Ended up cleaning the modules API and adapting front end :upside_down_face: 

### Backend

* Redo the module system : 
    * Remove fr_name , make name translatable. translations have been moved to django.po
    * Added related_plugin to be able to filter out 
 
* API: 
   * reorganize code 
   * Change permission system: the code was a bit useless and so anybody could call this API , while the front end specifies you cannot access module UI if you don't have the "modules" permission. So changed it to be like the front : you must have the module perm to list them
   * Changed the list output fields to better fit the front end needs + remove permission field that wasn't used
   * Added a new dropdown endpoint  
   * Filter out the modules related to plugins that are not activated
   * Test

### Frontend

* Removed the previous hooks 
* Added orval configuration + generated code
* Created a dropdown component for modules that checks upon the permissions
* Removed translations as it now comes directly from the backend
* Tests

## How to test

* The affected UIs are modules and the setupAccount, the module data should still work. 
* The list should auto refresh when changing locale
* The API should not return modules that are related to a plugin you didn't activate

## Notes

In `ModuleList.integration.test.tsx`, there's an interesting test that shows how to detect backend changes from the front end when filtering : `'uses the correct parameters for searching'`


iaso/api/modules/views.py
```diff
@@ -0,0 +1,71 @@
...
+from django_filters.rest_framework import DjangoFilterBackend
+from drf_spectacular.utils import OpenApiParameter, extend_schema
+from rest_framework import permissions
+from rest_framework.decorators import action
+from rest_framework.mixins import ListModelMixin
+from rest_framework.viewsets import GenericViewSet
+
+from iaso.modules import MODULES
+
+from ...permissions.core_permissions import CORE_MODULES_PERMISSION
```
tdethier: Nitpick: if it's more than `..`, I just import directly `iaso.something` but it's just my preference 😅

iaso/api/modules/views.py
```diff
@@ -0,0 +1,71 @@
...
+        if self.action == "dropdown":
+            return ModuleDropdownSerializer
+        raise NotImplementedError(f"Serializer not implemented for {self.action}")
+
+    def get_queryset(self):
+        # todo : this filtering out could technically be done in the MODULES itself, but it has such a huge impact we'll start here :)
+        return [
+            module
+            for module in MODULES
+            if not module.related_plugin or module.related_plugin in (settings.PLUGINS or [])
```
tdethier, line 65: Nitpick: `settings.PLUGINS` should always be a list with the default value and parsing

iaso/api/modules/filters.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def filter_search(self, queryset, name, value):
+        if value:
+            queryset = [module for module in queryset if value.lower() in module.name.lower()]
+        return queryset
+
+    def filter_exclude(self, queryset, name, value):
+        if value:
+            queryset = [module for module in queryset if value.lower() not in module.codename.lower()]
+        return queryset
```
tdethier, line 57: You're overwriting all of this because modules are not a django ORM queryset?

iaso/tests/api/modules/test_dropdown.py
```diff
@@ -0,0 +1,114 @@
...
+
+        res = self.client.get(reverse("modules-dropdown"), data={"search": self.related_modules[0].name.lower()})
+        res_data = self.assertJSONResponse(res, status.HTTP_200_OK)
+        self.assertValidData(res_data, 1)
+        self.assertEqual(res_data[0]["label"], self.related_modules[0].name)
+
+        res = self.client.get(reverse("modules-dropdown"), data={"search": self.related_modules[0].name[:2]})
+        res_data = self.assertJSONResponse(res, status.HTTP_200_OK)
+        self.assertValidData(res_data, 3)
+        self.assertEqual(res_data[0]["label"], self.related_modules[0].name)
```
tdethier: This test is a bit weird to read 😅 

It's checking if the first two letters of the first module (so, `Da` because data collection is the first one I guess), then checking how many modules contain this string?

I would either document this more, or hardcode a search string and make each expected module explicit, it would be a bit easier to understand

iaso/tests/api/modules/test_dropdown.py
```diff
@@ -0,0 +1,114 @@
...
+
+        res = self.client.get(reverse("modules-dropdown"), data={"search": "d" * 22})
+        res_data = self.assertJSONResponse(res, status.HTTP_200_OK)
+        self.assertValidData(res_data, 0)
+
+        res = self.client.get(reverse("modules-dropdown"), data={"exclude": self.related_modules[0].codename})
```
tdethier: I didn't know we could `exclude`, that's very interesting!

iaso/tests/api/modules/test_list.py
```diff
@@ -0,0 +1,116 @@
...
+from rest_framework import status
+
+from iaso.models import Account
+from iaso.modules import MODULE_FORM_AI, MODULE_STOCK, MODULES
+from iaso.permissions.core_permissions import CORE_MODULES_PERMISSION
+from iaso.test import APITestCase, SwaggerTestCaseMixin
+
+
+@override_settings(PLUGINS=None)
+class ModuleDropdownAPITestCase(SwaggerTestCaseMixin, APITestCase):
```
tdethier:
```suggestion
class ModuleListAPITestCase(SwaggerTestCaseMixin, APITestCase):
```

iaso/tests/api/modules/test_list.py
```diff
@@ -0,0 +1,116 @@
...
+
+        res = self.client.get(reverse("modules-list"), data={"search": self.related_modules[0].name.lower()})
+        res_data = self.assertJSONResponse(res, status.HTTP_200_OK)
+        self.assertValidData(res_data, 1)
+        self.assertEqual(res_data[0]["name"], self.related_modules[0].name)
+
+        res = self.client.get(reverse("modules-list"), data={"search": self.related_modules[0].name[:2]})
+        res_data = self.assertJSONResponse(res, status.HTTP_200_OK)
+        self.assertValidData(res_data, 3)
+        self.assertEqual(res_data[0]["name"], self.related_modules[0].name)
```
tdethier: Same comment since it's basically the same test

tdethier (approved):
Nice job refactoring this API, I've been wanting to do it for a long time 👍 

I tested the endpoint and new view with different plugins, it works fine
SNT does not appear on a vanilla server:

Image

but it's there on an SNT server:

Image

--------------

I made a few comments and suggestions, but nothing major

iaso/api/modules/filters.py
hugo093: yes :)

iaso/tests/api/modules/test_dropdown.py
hugo093: it's juts a parameter I added to match the front end actually :P

PR #3086, opened 2026-06-05
IA-5183 allow associating a form to org unit groups

madewulf:

Forms could only be associated with org unit types; this PR makes it possible to also associate a form with groups of org units, while keeping the org unit type association available (both can be combined), and exposes this to the mobile application. It also warns users about the impacts of changing the default source version of an account.

### Related JIRA tickets

IA-5183

## Changes

**Backend**
- New `Form.org_unit_groups` M2M to `Group` + migration `0392_form_org_unit_groups`.
- `FormSerializer`: read field `org_unit_groups` (list of group dicts, without org unit counts) and write field `org_unit_group_ids`. The write field is **optional**, so existing API/mobile clients that don't send it are unaffected.
- Validation: the writable `org_unit_group_ids` queryset is scoped to the user's account (field level), and `validate()` further restricts the groups to the ones of the account's **default source version** (the version currently in use), otherwise 400 `Invalid org unit group ids`.
- `/api/forms/` **and** `/api/mobile/forms/` (`views.py` / `views_mobile.py`): new `orgUnitGroupIds` query param mirroring `orgUnitTypeIds`; `org_unit_groups` is prefetched only when the field is requested (`is_field_referenced`), so requests with a `fields` param excluding it don't pay an extra query.

**Frontend**
- `FormFormComponent.tsx`: new searchable “Org unit groups” multi-select below the org unit types field, using the existing `useGetGroupDropdown` hook with `projectIds`, disabled until at least one project is selected (same UX as org unit types).
- `detail.tsx` / `types/forms.ts`: `org_unit_group_ids` wired into the form state and save payload, `org_unit_groups` added to the fields requested by the detail page.
- `config/index.tsx`: new optional “Org unit groups” column in the forms list (available through the column select drawer, not visible by default).
- `DataSourceDialogComponent.tsx`: when editing a data source in a way that **changes the account's default source version** (the source is/becomes the default source and the selected version differs from the current default), a warning is shown listing the impacts (past data not appearing under the new version's org units, forms linked to groups of the previous version, users restricted to old-version org units, planning/microplanning assignments, reference submissions, saved maps/dashboards/registry/filters, and other version-scoped features). **Saving is blocked until the user ticks a confirmation checkbox** acknowledging these impacts. The confirmation is UI-only and is stripped from the payload.
- `messages.ts` + en/fr/es translations.

## How to test

1. Run the migration.
2. In the web app, create or edit a form: select a project, then pick one or more org unit groups in the new “Org unit groups” field, save, reload and check the values are kept.
3. Check `GET /api/forms/<id>/` returns `org_unit_groups` and `GET /api/forms/?orgUnitGroupIds=<group_id>` filters correctly.
4. Same on mobile endpoints: `GET /api/mobile/forms/?app_id=...` exposes `org_unit_groups` and supports `orgUnitGroupIds`.
5. Try saving a form with a group that is not in the account's default source version via the API: it should be rejected with a 400.
6. In the Data sources screen, edit a source and change its default version (or make another source the default with a different version): the impacts warning appears and Save stays disabled until the acknowledgement checkbox is ticked.

Automated tests: `iaso.tests.api.test_forms`, `iaso.tests.api.test_mobileforms`, `iaso.tests.api.test_forms_serializers`, `FormFormComponent.test.tsx`.

## Print screen / video
CleanShot 2026-06-09 at 10 47 03@2x

image

## Notes

- The group dropdown options are filtered by the selected projects (same behaviour as the org unit types dropdown). As for other group dropdowns, the default version logic of `/api/groups/dropdown/` applies — which is consistent with the backend now validating groups against the account's default version.
- Because groups are pinned to a source version, a form's group associations become stale when the account switches its default version (e.g. after a pyramid re-import). This is the rationale for the new warning on default version change.
- The mobile app will need a corresponding change to consume `org_unit_groups` / `orgUnitGroupIds`.

## Doc

API change is self-documented through the serializer (`/api/forms/` & `/api/mobile/forms/` swagger).


iaso/models/forms.py
```diff
@@ -146,6 +146,7 @@ class Form(SoftDeletableModel):
     )
 
     org_unit_types = models.ManyToManyField("OrgUnitType", blank=True)
+    org_unit_groups = models.ManyToManyField("Group", blank=True, related_name="forms")
```
hugo093, line 149: Not related to this PR, but this is so confusing and clashing with the django built in model : Group. We should rename it sometime

iaso/tests/api/test_forms.py
```diff
@@ -186,6 +198,23 @@ def test_forms_list_filtered_by_org_unit_type(self):
...
 
+    def test_forms_list_filtered_by_org_unit_group(self):
+        self.client.force_authenticate(self.form_manager)
+        # Filter by org unit group `health_facilities_group`: only form_2 is linked to it.
+        response = self.client.get(
+            f"/api/forms/?orgUnitGroupIds={self.health_facilities_group.pk}",
+            headers={"Content-Type": "application/json"},
+        )
+        self.assertJSONResponse(response, 200)
+        self.assertValidFormListData(response.json(), 1)
```
hugo093:
```suggestion
        res_data = self.assertJSONResponse(response, 200)
        self.assertValidFormListData(res_data, 1)
```

iaso/tests/api/test_forms.py
```diff
@@ -186,6 +198,23 @@ def test_forms_list_filtered_by_org_unit_type(self):
...
+        # Filter by org unit group `empty_group`: no form is linked to it.
+        response = self.client.get(
+            f"/api/forms/?orgUnitGroupIds={self.empty_group.pk}",
+            headers={"Content-Type": "application/json"},
+        )
+        self.assertJSONResponse(response, 200)
+        self.assertValidFormListData(response.json(), 0)
```
hugo093:
```suggestion
        res_data = self.assertJSONResponse(response, 200)
        self.assertValidFormListData(res_data, 0)
```

iaso/tests/api/test_forms.py
```diff
@@ -186,6 +198,23 @@ def test_forms_list_filtered_by_org_unit_type(self):
         self.assertJSONResponse(response, 200)
         self.assertValidFormListData(response.json(), 0)
 
+    def test_forms_list_filtered_by_org_unit_group(self):
+        self.client.force_authenticate(self.form_manager)
+        # Filter by org unit group `health_facilities_group`: only form_2 is linked to it.
+        response = self.client.get(
+            f"/api/forms/?orgUnitGroupIds={self.health_facilities_group.pk}",
+            headers={"Content-Type": "application/json"},
+        )
```
hugo093:
```suggestion
        response = self.client.get(
            f"/api/forms/?orgUnitGroupIds={self.health_facilities_group.pk}"
        )
```
application/json is by default with our test case (see in settings.py     "TEST_REQUEST_DEFAULT_FORMAT": "json",  # The default format that should be used when making test requests.
 )

iaso/tests/api/test_forms.py
```diff
@@ -186,6 +198,23 @@ def test_forms_list_filtered_by_org_unit_type(self):
...
+        self.assertJSONResponse(response, 200)
+        self.assertValidFormListData(response.json(), 1)
+        # Filter by org unit group `empty_group`: no form is linked to it.
+        response = self.client.get(
+            f"/api/forms/?orgUnitGroupIds={self.empty_group.pk}",
+            headers={"Content-Type": "application/json"},
+        )
```
hugo093:
```suggestion
        response = self.client.get(
            f"/api/forms/?orgUnitGroupIds={self.empty_group.pk}"
        )
```

iaso/tests/api/test_forms.py
```diff
@@ -420,6 +449,163 @@ def test_forms_create_ok(self):
...
         self.assertEqual(2, form.org_unit_types.count())
 
+    def test_forms_retrieve_with_org_unit_groups(self):
+        """GET /forms/<form_id> returns the org unit groups of the form"""
+
+        self.client.force_authenticate(self.form_manager)
+        response = self.client.get(f"/api/forms/{self.form_2.id}/")
+        self.assertJSONResponse(response, 200)
+
+        form_data = response.json()
```
hugo093:
```suggestion
        form_data = self.assertJSONResponse(response, 200)
```

iaso/tests/api/test_forms.py
```diff
@@ -420,6 +449,163 @@ def test_forms_create_ok(self):
...
+        self.client.force_authenticate(self.form_manager)
+        fields = (
+            "id,name,org_unit_types,org_unit_groups,projects,period_type,derived,single_per_period,"
+            "periods_before_allowed,periods_after_allowed,device_field,location_field,label_keys,"
+            "possible_fields,legend_threshold,change_request_mode,validation_workflow"
+        )
+        response = self.client.get(f"/api/forms/{self.form_2.id}/?fields={fields}")
+        self.assertJSONResponse(response, 200)
+
+        form_data = response.json()
```
hugo093:
```suggestion
        form_data = self.assertJSONResponse(response, 200)
```

iaso/tests/api/test_forms.py
```diff
@@ -420,6 +449,163 @@ def test_forms_create_ok(self):
...
+                "periods_after_allowed": 0,
+                "project_ids": [self.project_1.id],
+                "org_unit_type_ids": [self.jedi_council.id],
+                "org_unit_group_ids": [self.health_facilities_group.id],
+            },
+            format="json",
+        )
+        self.assertJSONResponse(response, 201)
+
+        response_data = response.json()
```
hugo093:
```suggestion
        response_data = self.assertJSONResponse(response, 201)
```

iaso/tests/api/test_forms.py
```diff
@@ -420,6 +449,163 @@ def test_forms_create_ok(self):
...
+                "periods_before_allowed": 1,
+                "periods_after_allowed": 0,
+                "project_ids": [self.project_1.id],
+                "org_unit_type_ids": [self.jedi_council.id],
+            },
+            format="json",
+        )
+        self.assertJSONResponse(response, 201)
+
+        response_data = response.json()
```
hugo093:
```suggestion
        response_data = self.assertJSONResponse(response, 201)
```

iaso/tests/api/test_forms.py
```diff
@@ -420,6 +449,163 @@ def test_forms_create_ok(self):
...
+                "periods_before_allowed": 1,
+                "periods_after_allowed": 0,
+                "project_ids": [self.project_1.id],
+                "org_unit_type_ids": [self.jedi_council.id],
+                "org_unit_group_ids": [self.unrelated_group.id],
+            },
+            format="json",
+        )
+        self.assertJSONResponse(response, 400)
+        self.assertHasError(response.json(), "org_unit_group_ids")
```
hugo093:
```suggestion
        res_data = self.assertJSONResponse(response, 400)
        self.assertHasError(res_data, "org_unit_group_ids", f'Invalid pk "{self.unrelated_group.id}" - object does not exist.')
```

And maybe specify the validation error message ?

iaso/tests/api/test_forms.py
```diff
@@ -420,6 +449,163 @@ def test_forms_create_ok(self):
...
+                "periods_before_allowed": 1,
+                "periods_after_allowed": 0,
+                "project_ids": [self.project_1.id],
+                "org_unit_type_ids": [self.jedi_council.id],
+                "org_unit_group_ids": [group_on_other_version.id],
+            },
+            format="json",
+        )
+        self.assertJSONResponse(response, 400)
+        self.assertHasError(response.json(), "org_unit_group_ids", "Invalid org unit group ids")
```
hugo093:
```suggestion
        res_data = self.assertJSONResponse(response, 400)
        self.assertHasError(res_data, "org_unit_group_ids", "Invalid org unit group ids")
```

iaso/tests/api/test_forms.py
```diff
@@ -420,6 +449,163 @@ def test_forms_create_ok(self):
...
+        """PATCH /forms/<form_id> with only org_unit_group_ids cannot attach a group from outside the account"""
+
+        self.client.force_authenticate(self.form_manager)
+        response = self.client.patch(
+            f"/api/forms/{self.form_1.id}/",
+            data={"org_unit_group_ids": [self.unrelated_group.id]},
+            format="json",
+        )
+        self.assertJSONResponse(response, 400)
+        self.assertHasError(response.json(), "org_unit_group_ids")
```
hugo093:
```suggestion
        res_data = self.assertJSONResponse(response, 400)
        self.assertHasError(res_data, "org_unit_group_ids")
```

iaso/tests/api/test_forms.py
```diff
@@ -420,6 +449,163 @@ def test_forms_create_ok(self):
...
+        self.assertEqual(0, self.form_1.org_unit_groups.count())
```
hugo093, line 574:
```suggestion
        self.form_1.refresh_from_db()
        self.assertEqual(0, self.form_1.org_unit_groups.count())
```
Just to be sure even though it's not necessary

iaso/api/forms/serializers.py
```diff
@@ -125,6 +144,10 @@ class Meta:
     def get_org_unit_types(obj: Form):
         return [t.as_dict() for t in obj.org_unit_types.all()]
 
+    @staticmethod
+    def get_org_unit_groups(obj: Form):
+        return [g.as_dict(with_counts=False) for g in obj.org_unit_groups.all()]
```
hugo093, line 153: write a dedicated nested serializer instead of using the as_dict method that we shouldn't use anymore. 
extend also the method using drf-spectacular to be swagger compliant. E.g : 
```python
@extend_schema_field(NestedOrgUnitGroupSerializer(many=True)
```

iaso/api/forms/serializers.py
```diff
@@ -163,6 +186,17 @@ def validate(self, data: typing.Mapping):
...
+            # validate org_unit_groups: a group must belong to the account's default source version
+            # (the data source version currently in use)
+            org_unit_groups = data.get("org_unit_groups", [])
+            if org_unit_groups:
+                default_version = self.context["request"].user.iaso_profile.account.default_version
+                allowed_group_ids = set(
+                    Group.objects.filter(source_version=default_version).values_list("id", flat=True)
+                )
+                if any(group.id not in allowed_group_ids for group in org_unit_groups):
+                    raise serializers.ValidationError({"org_unit_group_ids": "Invalid org unit group ids"})
```
hugo093: this should be part of the child_relation.queryset , not the validate. 
e.g : 
 self.fields["org_unit_group_ids"].child_relation.queryset = Group.objects.filter_for_user(user).filter(source_version=default_version)

hat/assets/js/apps/Iaso/domains/forms/components/FormFormComponent.tsx
```diff
@@ -113,6 +118,13 @@ const FormForm: FunctionComponent<FormFormProps> = ({
                 currentForm.project_ids.value &&
                 currentForm.project_ids.value.length > 0,
         });
+    const { data: allOrgUnitGroups, isFetching: isOuGroupLoading } =
+        useGetGroupDropdown(
+            { projectIds: currentForm.project_ids.value.join(',') },
+            // we only want to fetch the groups if the project ids are set, project ids is a required field
+            currentForm.project_ids.value &&
+                currentForm.project_ids.value.length > 0,
```
hugo093:
```suggestion
            Boolean(currentForm.project_ids?.value?.length)
```

hat/assets/js/apps/Iaso/domains/forms/components/FormFormComponent.tsx
```diff
@@ -299,6 +311,31 @@ const FormForm: FunctionComponent<FormFormProps> = ({
...
+                            }
+                            value={orgUnitGroups}
+                            errors={currentForm.org_unit_group_ids.errors}
+                            type="select"
+                            options={allOrgUnitGroups || []}
+                            label={MESSAGES.orgUnitsGroups}
+                            loading={isOuGroupLoading}
+                            disabled={
+                                !currentForm.project_ids.value ||
+                                currentForm.project_ids.value.length === 0
```
hugo093:
```suggestion
                                !currentForm.project_ids?.value?.length
```

hat/assets/js/apps/Iaso/domains/dataSources/components/DataSourceDialogComponent.tsx
```diff
@@ -269,6 +290,69 @@ export const DataSourceDialogComponent: FunctionComponent<Props> = ({
...
                     )}
+                    {willChangeDefaultVersion && (
+                        <Box mt={1}>
+                            <Alert severity="warning">
+                                <AlertTitle>
+                                    {formatMessage(
+                                        MESSAGES.changeDefaultVersionWarningTitle,
+                                    )}
+                                </AlertTitle>
+                                <ul style={{ margin: 0, paddingLeft: '1.2em' }}>
```
hugo093: can you use built in mui List and ListItem component for this , and use sx prop instead of style then ? 
Also I would put this alert at the top of the modal and make it take the whole width , right now it feels weird on the screen considering the amount of text 

Image

hugo093 (changes requested):
It works, some details here and there to change though

iaso/models/forms.py
madewulf: The terminology "group" is coming from DHIS2 originally. I don't think we should change it in the UI and I think it's better when the UI matches the code (we already have Instance in the code that is called Submission in the UI and this is a little painful). So, I don't think another name could be used, really.

hat/assets/js/apps/Iaso/domains/dataSources/components/DataSourceDialogComponent.tsx
madewulf: I think it makes sense to have the popup just under the select box that triggers its apparition. For the rest, let me sse.

iaso/api/forms/serializers.py
madewulf: This matches the code styles in the same file, and I don't want to go into a large refactor to something used almost everywhere in IASO for now. Changing these style would make sense in a v2 of the API, but here, it's too much risk to my taste.

hugo093 (approved):
LGTM

PR #3088, opened 2026-06-08
[IA-5170]: Only return permissions from active modules in profile retrieve

hugo093:

Profile update would return a 400 when not modifying anything. It was caused by permission related to a module that was deactivated in the mean time. 

### Related JIRA tickets

IA-5170

## Changes

* In profile retrieve endpoint, filter out the permissions returned to only be related to active modules 
* Added test that mimick the UI workflow

## How to test

1. Activate a module for the user
2. Assign a permission related to the module to the user 
3. Deactivate the module
4. Edit the user and submits without changing anything
5. Should not crash


tdethier (approved):
LGTM, the permissions list is indeed filtered 👍

PR #3089, opened 2026-06-08
POLIO-2159: exclude copieds sheets from prep parser

quang-le:

Another way clients add sheets that shouldn't be parsed needs exclusion

### Related JIRA tickets

POLIO-2159

## Changes

- Add `"copy"` to the list of prefixes to exclude

## How to test

reproduce campaign in ticket on polio DB
start worker
launch preparedness refresh--> no 500

## Note
to hotfix


bmonjoie (approved):
LGTM

PR #3090, opened 2026-06-08
[IA-5169]: provide a way to add custom zod error messages

hugo093:

It was impossibl to override a global zod message + translation

### Related JIRA tickets

IA-5169

## Changes

See code in zodUtils.ts and why it was coded this way. 
Some refs : 
* https://orval.dev/docs/reference/configuration/output#params 
* https://react.i18next.com/

## How to test

Go to account management page, try to edit an account and set the name to empty. You should see the custom error message. Switch locale, it should be translated


quang-le (approved):
Let's revisit this once we've upgraded necessary deps

PR #3093, opened 2026-06-08
POLIO-2160: Eslint on polio plugin

beygorghor:

 679 problems (586 errors, 93 warnings)
 553 errors and 46 warnings potentially fixable with the `--fix` option.

### Related JIRA tickets

POLIO-2160

## Changes

- auto fix =>   ✖ 80 problems (33 errors, 47 warnings)
-  unsused vars; exhaustive deps; no array index key => ✖ 17 problems (10 errors, 7 warnings)
- manual fixes: 
      - Component definition is missing display name
      - wrong use of require
      - label props missing
      -  Do not pass children as props. Instead, nest children between the opening and closing tags
      - loop key issues 
-  include polio plugin in eslint tests
- force eslint in CI

## How to test

run `npm run export-lint-problems`it and check the report


plugins/polio/js/src/domains/Campaigns/RiskAssessment/RiskAssessmentForm.tsx
```diff
@@ -198,7 +198,7 @@ export const RiskAssessmentForm: FunctionComponent = () => {
                 <Grid item md={6}>
                     {rounds.map((round, i) => {
                         return (
-                            <Box mt={i === 0 ? 0 : 2}>
+                            <Box mt={i === 0 ? 0 : 2} key={round.id}>
```
quang-le: I believe this can be null if you're creating the campaign. IN this case `round.number` is probably safer.

quang-le (changes requested):

quang-le (approved):

PR #3097, opened 2026-06-09
[IA-5116] Submission diff API

hugo093:

First draft at Submission diff API

### Related JIRA tickets

IA-5116

## Changes

Using jsonpatch, made an api that allows to see the Modifications for an instance 

## How to test

Try and call the endpoint /api/diff/instances/{instance_id}/  after resubmitting the instance and making changes

### DIFF
--- a/iaso/api/instance_diff/serializers.py
+++ b/iaso/api/instance_diff/serializers.py
@@ -5,26 +5,7 @@ from rest_framework import serializers
 
 from hat.audit.models import Modification
 from iaso.api.common import ModelSerializer
-
-
-class JsonPatchOperationSerializer(serializers.Serializer):
-    """
-    Serializer just there for swagger compliance: see https://jsonpatch.com/
-    """
-
-    op = serializers.ChoiceField(
-        choices=[
-            "add",
-            "remove",
-            "replace",
-            "move",
-            "copy",
-            "test",
-        ]
-    )
-    path = serializers.CharField()
-    value = serializers.JSONField(required=False)
-    vars()["from"] = serializers.CharField(required=False)
+from iaso.api.common.serializer import JsonPatchOperationSerializer
 
 
 class ModificationSerializer(ModelSerializer):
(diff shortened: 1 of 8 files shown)


iaso/api/instance_diff/serializers.py
```diff
@@ -0,0 +1,40 @@
...
+            "remove",
+            "replace",
+            "move",
+            "copy",
+            "test",
+        ]
+    )
+    path = serializers.CharField()
+    value = serializers.JSONField(required=False)
+    vars()["from"] = serializers.CharField(required=False)
```
tdethier: That's funny, I had not thought of `from` being an invalid attribute name

iaso/api/instance_diff/serializers.py
```diff
@@ -0,0 +1,40 @@
+import jsonpatch
+
+from drf_spectacular.utils import extend_schema_field
+from rest_framework import serializers
+
+from hat.audit.models import Modification
+from iaso.api.common import ModelSerializer
+
+
+class JsonPatchOperationSerializer(serializers.Serializer):
```
tdethier: Shouldn't we move this to some `utils` file? I guess it could be used somewhere else to compare not just instances

iaso/api/instance_diff/serializers.py
tdethier: Unfortunately it looks like this is not working, I don't see a `form` entry. Unless something was wrong with the edit (made with Enketo)

Image

tdethier (commented):
API works, I left a suggestion to maybe move the generic serializer. I'm just not sure about the `from` field in that serializer, it looks like `vars()["from"]` does not work

iaso/api/instance_diff/serializers.py
hugo093: from https://jsonpatch.com => the from only appears with the move method :/  i added some test to cover it though

tdethier (approved):
LGTM 👍

PR #3100, opened 2026-06-10
IA-5202: Ignore some rules on Orval generated files

beygorghor:

Orval is generating ts files with some eslint warning that cannot be fixed with the auto-fix. We should ignore those warning only on Orval files.

### Related JIRA tickets

IA-5202

## Changes

- auto fix an error
- adapt config for orval
- allow vitest skip
- remove unused inline eslint disable

## How to test

run`npm run show-lint-problems` it should pass 

## Print screen / video

--


hugo093 (approved):
LGTM :) (took the liberty to rebase)

PR #3103, opened 2026-06-10
[IA-5151] change embedded links entry to check module

Crebert08:

This PR is about removing the check of the feature flag on the embedded links entry in the menu

### Related JIRA tickets

[IA-5151](https://bluesquare.atlassian.net/browse/IA-5151)

### Frontend
- Added the embedded links module in the module file
- implemented the check in the menu file
- removed the now unused show_page feature flags
- added some tests

## How to test

> Connect to iaso with a user  with the embedded links module and you should see the entry in the menu
> Connect to iaso with a user without the embedded links module and you shouldn't see the entry in the menu

### user with module
image

### User without module
image

## Notes

## Doc

[IA-5151]: https://bluesquare.atlassian.net/browse/IA-5151?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


quang-le (changes requested):
Nice work, functions as expected.

Nothing to say about what was done, but a few steps are missing:
- Most important: add a migration to remove the feature flag on the servers where it exists and make sure that the accounts that had it enabled have the module enabled
- There's still a reference to SHOW_PAGES in the setuper. It should be removed
- There's still a reference to SHOW_PAGES in a test, it should be cleaned up

quang-le (approved):

PR #3104, opened 2026-06-10
fix/IA-5205 dynamic fields validation error response)

62496:

Invalid `fields` query parameters were returning a server error instead of a clear API validation error.

### Related JIRA tickets

IA-5205

## Changes

Updated the dynamic fields filter backend to raise a DRF `serializers.ValidationError` attached to the `fields` parameter.

The error response now returns `400 Bad Request` with a clear message, for example:

```json
{
  "fields": ["Invalid dynamic fields: fake"]
}
```

## How to test

Call an endpoint using dynamic fields with an invalid field:

```bash
curl -i "http://localhost:8081/api/forms/?fields=fake"
```

Expected result: 400 Bad Request with an error message under fields.

Also check that valid fields still work:

```bash
curl -i "http://localhost:8081/api/forms/?fields=id,name"
```

Expected result: 200 OK.

## Print screen / video

image
image

### DIFF
--- a/dynamic_fields/filter_backends.py
+++ b/dynamic_fields/filter_backends.py
@@ -1,5 +1,6 @@
 from django.conf import settings
-from django.core.exceptions import ValidationError
+from django.utils.translation import gettext_lazy as _
+from rest_framework import serializers
 from rest_framework.filters import BaseFilterBackend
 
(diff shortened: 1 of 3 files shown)


hugo093 (approved):
Nothing to say, GJ and LGTM :)

PR #3105, opened 2026-06-11
fix: resolve remaining SSO CI failures (translations + auth enforcement)

madewulf:

Fixes the remaining `develop` CI failures from the SSO plugin merge (#3099): the backend translation check and the auth-enforcement test.

### Related JIRA tickets

POLIO-2115

## Changes

Follow-up to the SSO fixes already on `develop` (commit `6df02eb7`, which made the SSO tests self-contained and moved `login_subtemplate.html` into core `hat/templates/`). Two remaining CI jobs were still red:

**1. `Check Backend (Django) translations` job** — `hat/locale` + `plugins/sso/locale`
- Moving `login_subtemplate.html` into `hat/templates/` caused its `Login via %(name)s` string to be extracted into the **core** `hat` locale instead of the `sso` plugin locale (where it was already translated). `make_translations` then reported it as a missing translation.
- Added the French translation `Se connecter via %(name)s` to `hat/locale`, and let the now-unused entry be dropped from the `sso` plugin locale.

**2. `Django tests` — `test_auth_enforcement`** — `iaso/tests/test_auth_enforcement.py`
- When the `sso` plugin is enabled it registers login / callback / token endpoints (e.g. `/polio/token/`) that are legitimately public, like the existing `/wfp_auth/...` endpoints. The test flagged them as endpoints that should require auth.
- Added `_sso_paths()` to derive those endpoint paths from `settings.SSO_PROVIDERS` (same config keys the URLs are built from) and skip them in `should_skip()`, so it stays correct regardless of the per-provider path configuration.

## How to test

```
# Translations (fails before, passes after)
manage make_translations && manage compile_translations

# Auth enforcement, both with SSO disabled (CI default) and enabled (set SSO_WHO_* env)
manage test iaso.tests.test_auth_enforcement
```

## Print screen / video

N/A — test/locale-only change.

## Notes

- The SSO endpoints are intentionally public (login/token must be reachable by unauthenticated users); skipping them mirrors the existing treatment of `/wfp_auth/` and polio endpoints.
- The auth-enforcement change only affects environments where SSO is enabled. CI runs with SSO disabled, so those endpoints don't exist there.

## Doc

N/A.


quang-le:
changes cherry picked in https://github.com/BLSQ/iaso/pull/3106

PR #3106, opened 2026-06-11
Polio 2115 implement sso for https poliogo test afro who int

beygorghor:

The polio project needs SSO login via Azure (Microsoft Entra ID) for `https://poliogo-test.afro.who.int`, and the existing `wfp_auth` plugin was too WFP-specific to reuse.

### Related JIRA tickets

POLIO-2115

## Changes

**New generic SSO module under `hat/`** — a generic OAuth2/OIDC SSO mechanism. The runtime is provider-agnostic (any OAuth2/OIDC provider with authorize/token/userinfo URLs); the current configuration wires a single WHO/Entra provider via `SSO_WHO_*` env vars:

- `hat/views.py`: `SSOBaseAdapter` (extends allauth's `OAuth2Adapter`), `SSOLoginView`, `SSOCallbackView`, and a `make_token_view` factory for the mobile-app token exchange. Adapters and views are created dynamically at runtime from `settings.SSO_PROVIDERS`.
- `hat/provider.py`: dynamically creates and registers allauth `OAuth2Provider` subclasses so the module integrates with allauth's provider registry like any first-class provider.
- `hat/urls.py`: `get_sso_urlpatterns()` generates URL patterns for each configured provider (`login`, `login/callback`, `token`).
- `hat/templates/sso/login_subtemplate.html`: generic login button rendered for each provider.
- New-account email notification support (same pattern as `wfp_auth`).

**Login page integration** (`hat/`):

- `hat/urls.py`: `LoginView` / `TemplateLoginView` inject a `sso_providers_for_login` context variable (built from allauth's provider registry) so the login templates render a button per configured provider — covering both the `hat` SSO providers and the existing `wfp_auth` provider with no provider-specific references.
- `hat/templates/iaso/login.html` and `disabled_password_login.html`: replaced the hardcoded `{% if provider.id == 'wfp' %}` check with a generic loop over `sso_providers_for_login`.
- `hat/settings.py`: added WHO/Azure provider config (`SSO_WHO_*` env vars); `hat` is added to `INSTALLED_APPS` so the shared template is always reachable when SSO is active.

**Security hardening (mobile token exchange)** — the unauthenticated `/polio/token/` endpoint accepts an access token minted by the mobile app and previously trusted it after only a Graph `userinfo` call. This is hardened:

- The presented token is now pinned to our own app registration and tenant: its `appid`/`azp` claim must equal `SSO_WHO_CLIENT_ID` and its `tid` claim must equal `SSO_WHO_TENANT_ID`, otherwise it is rejected with `401`. This blocks token-replay / nOAuth-style impersonation, where a token minted by another Azure app or another tenant (potentially carrying a forged `email`) could be exchanged for an Iaso JWT. Authenticity is still established by the Graph `userinfo` call; the claim checks pin the genuine token to our app + tenant.
- Login now **fails with `409`** when several users share the same email in the account, instead of silently logging in as the first match.
- Error responses return static, generic messages (no exception content) to avoid leaking server state; the detail is logged server-side.
- Removed debug `print()` statements that logged the raw access token.
- `SSO_WHO_TENANT_ID` is now **required** — it is used both to build the Entra authorize/token URLs and to validate incoming tokens.

**WFP backward compatibility**: existing deployments using `WFP_AUTH_CLIENT_ID` require no env var changes — the `wfp_auth` plugin continues to handle its own OAuth flow, and the shared login template lives under `hat/`.

## How to test

Set the following env vars (WHO/Azure example):

```
SSO_WHO_CLIENT_ID=<azure-app-client-id>
SSO_WHO_TENANT_ID=<azure-tenant-id>
SSO_WHO_ACCOUNT=<iaso-account-id>
SSO_WHO_CLIENT_SECRET=<client-secret>
```

1. Start the server — a "Login via WHO" button should appear on `/login/`.
2. Clicking it redirects to Microsoft login; after authenticating you are redirected back and logged in.
3. If no `SocialAccount` exists for the user, a new Iaso user is created on first login.
4. With `DISABLE_PASSWORD_LOGINS=true`, the SSO button is the only option shown.
5. Existing WFP deployments (`WFP_AUTH_CLIENT_ID` set, no `SSO_WHO_*`) should show the WFP button unchanged.
6. Mobile token exchange: `POST /polio/token/?app_id=<project-app-id>` with body `{"token": "<azure access token>"}` returns an Iaso `{refresh, access}` JWT pair. A token whose `appid` or `tid` does not match the configured app/tenant is rejected with `401`; an app_id whose project is not in the configured account returns `400`; multiple users sharing the email return `409`.

The Azure app registration must have `http://localhost:8081/polio/login/callback` (or the equivalent production URL) registered as a redirect URI.

## Print screen / video

_To be added._

## Notes

- The SSO code (under `hat/`) has no models or migrations — it is safe to install alongside `wfp_auth` with no DB changes.
- The `wfp_auth` plugin is intentionally kept as-is for existing deployments. A future migration to the generic SSO config is possible by moving WFP's config into `SSO_PROVIDERS`.
- The login views discover providers via allauth's registry, so any future allauth-based or `SSO_PROVIDERS`-based provider will appear on the login page automatically.
- Token validation applies to the shared `complete_login`, so it also covers the web callback flow. Azure Graph access tokens are readable JWTs in this setup; if an environment ever issues opaque tokens, login fails closed.

## Doc

Configuration is documented in `hat/settings.py` alongside the existing `WFP_AUTH_CLIENT_ID` block. `SSO_WHO_TENANT_ID` is used both to build the Entra endpoints and to validate the `tid` of incoming tokens.


quang-le (dismissed):
We need to move the SSO in iaso-core, i.o plugin

hat/views.py
```diff
@@ -0,0 +1,294 @@
...
+            return JsonResponse(
+                {
+                    "result": "error",
+                    "message": _("Error login to auth server"),
+                    "details": _("Error login to auth server"),
+                },
+                status=500,
+            )
+        except (InvalidAppIdException, InvalidAccountConfiguration) as e:
+            return JsonResponse({"result": "error", "message": str(e), "details": str(e)}, status=400)
```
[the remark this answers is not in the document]
hugo093: Seeing the InvalidAppIdException and InvalidAccountConfiguration messages, there's no risk

hat/tests/test_sso_auth.py
```diff
@@ -108,3 +108,76 @@ def test_token_missing(self, mock_get):
...
+                "account_id": 100000,
+                "email_recipients_new_account": [],
+            },
+        }
+    )
+    @patch("requests.get")
+    def test_complete_login_wrong_account_in_settings(self, mock_get):
+        self.assertEqual(m.User.objects.count(), 0)
+        self.assertEqual(m.Profile.objects.count(), 0)
+        self.assertEqual(SocialAccount.objects.count(), 0)
```
madewulf: Shouldn't this be either NOT done or done at the end of the test?

hat/views.py
```diff
@@ -119,15 +131,22 @@ def complete_login(self, request, app, token: str, response) -> SocialAccount:
 
         uid = extra_data["sub"].lower().strip()
 
-        app_id = request.GET.get(APP_ID, None)
-        account_name = self.sso_config["account_id"]
+        app_id: str = request.GET.get(APP_ID, None)
+        account_name: int = self.sso_config["account_id"]
```
madewulf: This variable should be renamed account_id

hat/views.py
```diff
@@ -119,15 +131,22 @@ def complete_login(self, request, app, token: str, response) -> SocialAccount:
...
 
         if app_id:
-            account = get_object_or_404(Project, app_id=app_id).account
-            if app_id != account_name:
+            try:
+                account = Project.objects.filter(account_id=account_name).get(app_id=app_id).account
+            except Project.DoesNotExist:
+                raise InvalidAppIdException(app_id)
+
+            if app_id != str(account_name):
```
madewulf: I believe this test does not make sense and will always be false (You didn't add it, I'm just realizing that now)

plugins/wfp_auth/views.py
```diff
@@ -132,7 +132,9 @@ def complete_login(self, request, app, token: str, response) -> SocialAccount:
         app_id = request.GET.get(APP_ID, None)
 
         if app_id:
-            account = get_object_or_404(Project, app_id=app_id).account
+            account = get_object_or_404(
+                Project.objects.filter(account__name=self.settings["IASO_ACCOUNT_NAME"]), app_id=app_id
+            ).account
             if app_id != self.settings["IASO_ACCOUNT_NAME"]:
```
madewulf, line 137: I believe this might have been added to cope with the existing situation at WFP, where the first users have been created without the prefix, and we did not want to break their authentication

plugins/wfp_auth/tests/test_auth.py
```diff
@@ -76,14 +76,15 @@ def test_complete_login_ok_with_existing_username(self, mock_get):
...
-        with patch("plugins.wfp_auth.views.WFP2Adapter.settings", new={"IASO_ACCOUNT_NAME": "foo"}):
+        with patch("plugins.wfp_auth.views.WFP2Adapter.settings", new={"IASO_ACCOUNT_NAME": "Account"}):
             response = self.client.post(
                 f"/wfp_auth/wfp/token/?app_id={self.project.app_id}&app_version=2501",
                 format="json",
                 data={
                     "token": "f4k3-t0k3n",
                 },
             )
+        print(response.json())
```
madewulf: Forgotten print?

hat/views.py
```diff
@@ -0,0 +1,294 @@
...
+        account_name: int = self.sso_config["account_id"]
+
+        if app_id:
+            try:
+                account = Project.objects.filter(account_id=account_name).get(app_id=app_id).account
+            except Project.DoesNotExist:
+                raise InvalidAppIdException(app_id)
+
+            if app_id != str(account_name):
+                uid = f"{app_id}_{uid}"
```
madewulf: I think this worked for WFP but does not for us

hat/sso_views.py
```diff
@@ -0,0 +1,294 @@
...
+                account = Account.objects.get(id=account_name)
+            except Account.DoesNotExist:
+                raise InvalidAccountConfiguration
+
+        try:
+            social_account = SocialAccount.objects.get(uid=uid, provider=self.provider_id)
+            social_account.extra_data = extra_data
+        except SocialAccount.DoesNotExist:
+            user = User.objects.filter(iaso_profile__account=account, email=email).first()
+
```
madewulf, line 218: We should fail if there are multiple users on multiple accounts with that email

tdethier:
why is the code in `hat` instead of `iaso`?

madewulf (approved):

plugins/wfp_auth/views.py
bmonjoie: No, it was added so that the social account created for this `app_id` links to the correct user account.
The default one (that's the legacy part) didn't have a prefix so that's why this settings exists.

This should not be removed.
Also, the change above is entirely wrong, and it will be reverted.

Explanation: 
By doing 
```
account = (
                Project.objects.filter(account__name=self.settings["IASO_ACCOUNT_NAME"]).get(app_id=app_id).account
            )
``` 
You are filtering projects that only belong in the default account. Therefore, if you are trying to connect to a Project that is in a different Account, it will throw a `Project.DoesNotExists` and result in a 500.

PR #3112, opened 2026-06-11
IA-5141: refactor apps and projects api

quang-le:

while hotfixing the issue, it appeared that the write logic of the `apps` endpoint should be moved to `projects`

### Related JIRA tickets

IA-5141

## Changes

- Remove all write operations from `apps` endpoint
- Move all validation to the `projects` endpoint
- Make `forms` field write-only on `projects`endpoint to preserve feature parity with the legacy code, though it doesn't seem to be used anywhere
- Update front-end to use projects API i.o apps API. This entails passing the project id and not the app id as previously
- Add tests

## How to test

- Try accessing `api/apps` unauthenticated: read operations should still work, but write operations are now impossible
- In the UI try creating and editing a project, there should be no behaviour change

## Print screen / video

https://github.com/user-attachments/assets/9618bbf0-238c-4fe3-beff-bd88476e8a88

Screenshot 2026-06-11 at 15 32 56

## Notes
DS team has been polled about this potential breaking change, no impact expected.


iaso/api/query_params.py
```diff
@@ -41,3 +46,17 @@
 WITH_LOCATION = "withLocation"
 ONLY_REFERENCE = "onlyReference"
 REFERENCE_INSTANCES = "referenceInstances"
+
+
+def parse_strict_boolean_param(value: Optional[str], field_name: str = "query_param") -> bool:
```
tdethier, line 51: Ideally we'd use a serializer for that 😅

iaso/tests/api/test_projects.py
```diff
@@ -78,16 +121,55 @@ def test_projects_list_ok(self):
...
         with self.assertNumQueries(3):
             response = self.client.get("/api/projects/", headers={"Content-Type": "application/json"})
         self.assertJSONResponse(response, 200)
         self.assertValidProjectListData(response.json(), 2)
 
+    def test_projects_list_filter_by_app_id(self):
+        """GET /projects/?app_id= should return only the matching project."""
+        self.client.force_authenticate(self.jane)
+        response = self.client.get(f"/api/projects/?app_id={self.project_1.app_id}")
+        self.assertJSONResponse(response, 200)
```
tdethier: Nitpick: I prefer using a const from `rest_framwork.status` than numbers for HTTP codes

iaso/api/projects/viewsets.py
```diff
@@ -28,20 +31,23 @@ class ProjectsViewSet(ModelViewSet):
...
-    permission_classes = [permissions.IsAuthenticated]
-    filter_backends = [filters.OrderingFilter]
+    filter_backends = [filters.OrderingFilter, DjangoFilterBackend]
+    filterset_class = ProjectsFilter
     ordering_fields = ["app_id", "name"]
     ordering = ["id"]
     serializer_class = ProjectSerializer
     results_key = "projects"
-    http_method_names = ["get", "head", "options", "trace"]
+    http_method_names = ["get", "head", "options", "trace", "put", "post", "patch"]
```
tdethier, line 40: Nitpick: I would probably keep `patch` and remove `put`, but it does not change much

iaso/api/apps/viewsets.py
tdethier, line 46: I guess you can simplify this now

tdethier (approved):
LGTM 👍 

I made a couple of suggestions below

iaso/api/query_params.py
quang-le: It used to be in one, but frankly, I thought it was overkill, both in terms of lines of code and overhead

iaso/api/projects/viewsets.py
quang-le: quite the opposite in fact, the front updates with PUT 😅

tdethier: That's weird, unless you update all the fields

PR #3113, opened 2026-06-12
[IA-5198]: fix n+1 query on api orgunittypes v2

hugo093:

/api/v2/orgunittypes/ was causing a n+1 query. 

### Related JIRA tickets

IA-5198

### Analysis 

It could have been a quick and dirty fix like what was done in ticket IA-5189. But after investigation it turned out that : 

* Most fields returned by the list aren't used in the front end , especially the one causing N+1. 
* Some UI were using list endpoint instead of dropdown
* The with_unit_counts is actually mandatory for the list UI as the delete button is disabled if unit_counts > 0.
* Retrieve doesn't need with_unit_counts
* The edit modals were not calling the retrieve to load initial data, and there is no retrieve UI. So the retrieve was unused. 
* Not really swagger compliant and a single serializer to rule them all, very hard to read. 
* It was possible to access the API while being unauthenticated. While the results returned were an empty array, it's still weird. 
* It's possible to have a cycle in OUT with sub_unit_types, the patch would then crash automatically

The related fixes

### Backend 

1. Reorganize the code : serializers, viewsets, filters
2. Adapt the list/dropdown/retrieve payloads to match the front end needs
3. Testing
4. fix the get_parents method to avoid crash when there's a cycle
5. remove the validate_reference_forms method => move it to the queryset selection in the child_relation
6. change the permission so an unauthenticated user will receive a 401 instead of a 200 with empty list
7. For the list, added proper pagination, removed the custom results key unfortunately
8. Adapt the setuper

Perf changes : 

BEFORE

http://localhost:8000/api/v2/orgunittypes/?with_units_count=true
```
| RESP |  default  |    46    |    0     |    46    |     33     |
```

http://localhost:8000/api/v2/orgunittypes/dropdown/?with_units_count=true
```
| RESP |  default  |    11    |    0     |    11    |     0      |
```

http://localhost:8000/api/v2/orgunittypes/1/
```
| RESP |  default  |    18    |    0     |    18    |     3      |
```

AFTER 

http://localhost:8000/api/v2/orgunittypes/
```
| RESP |  default  |    13    |    0     |    13    |     2      |
```

http://localhost:8000/api/v2/orgunittypes/dropdown/?fields=:all
```
| RESP |  default  |    6     |    0     |    6     |     0      |
```

http://localhost:8000/api/v2/orgunittypes/1/
```
| RESP |  default  |    14    |    0     |    14    |     3      |
```

### Front-end

1. Generate orval files and adapt the current hooks
2. Change the org unit type edit/create modal to dynamically call the retrieve endpoint
3. Adapt some calls to use the dropdown endpoint instead of the list

## How to test

Test whole UI (OU , OUT , datasource, ... )  + api

## Notes

* It would be worth refactoring the dialog for org unit type 
* Front end : Should we move the whole orgUnitTypes folder under domains ? 
* Swagger/drf-spectacular has a hard time with recursive field (hierarchy endpoint), it could be worth making a drf-spectacular extension for that
* Maybe we should create separate endpoints/api just for the setuper. Right now some fields are there just becasue the setuper needs them.


iaso/api/org_unit_types/views/v2.py
```diff
@@ -0,0 +1,154 @@
...
+
+    @property
+    def filterset_class(self):
+        if self.action == "dropdown":
+            return OrgUnitTypeDropdownFilter
+
+        return OrgUnitTypeFilter
+
+    def perform_destroy(self, instance):
+        if instance.org_units.count():
```
mestachs, line 59: not sure it really matters but exists() is perhaps better for that check
even I would expect the table to be "relatively" small (a bit more than 4 millions) compared to instances/submissions

iaso/api/org_unit_types/views/v2.py
```diff
@@ -0,0 +1,154 @@
...
+                Prefetch(
+                    "allow_creating_sub_unit_types",
+                    queryset=OrgUnitType.objects.filter(projects__app_id=app_id).all().order_by("id")
+                    if app_id
+                    else OrgUnitType.objects.all().order_by("id"),
+                ),
+                Prefetch(
+                    "org_units",
+                    queryset=OrgUnit.objects.filter_for_user_and_app_id(self.request.user, app_id),
+                    to_attr="prefetched_org_units",
```
mestachs, line 107: this looks to return a whole orgunit list where I would expected the only the "count", especially when the list isn't in the returned payload I think
```sql
SELECT "iaso_orgunit"."id", "iaso_orgunit"."name", "iaso_orgunit"."uuid", "iaso_orgunit"."custom", "iaso_orgunit"."validation_status", "iaso_orgunit"."version_id", "iaso_orgunit"."parent_id", "iaso_orgunit"."path", "iaso_orgunit"."aliases", "iaso_orgunit"."org_unit_type_id", "iaso_orgunit"."sub_source", "iaso_orgunit"."source_ref", "iaso_orgunit"."simplified_geom"::bytea, "iaso_orgunit"."catchment"::bytea, "iaso_orgunit"."geom_ref", "iaso_orgunit"."gps_source", "iaso_orgunit"."location"::bytea, "iaso_orgunit"."created_at", "iaso_orgunit"."updated_at", "iaso_orgunit"."source_created_at", "iaso_orgunit"."creator_id", "iaso_orgunit"."extra_fields", "iaso_orgunit"."opening_date", "iaso_orgunit"."closed_date", "iaso_orgunit"."default_image_id", "iaso_orgunit"."code" FROM "iaso_orgunit" WHERE ("iaso_orgunit"."version_id" IN (SELECT DISTINCT U0."id" FROM "iaso_sourceversion" U0 INNER JOIN "iaso_datasource" U1 ON (U0."data_source_id" = U1."id") INNER JOIN "iaso_datasource_projects" U2 ON (U1."id" = U2."datasource_id") INNER JOIN "iaso_project" U3 ON (U2."project_id" = U3."id") WHERE U3."account_id" = 2) AND "iaso_orgunit"."org_unit_type_id" IN (8, 9, 10, 1, 11))
```

mestachs (commented):
something looks weird with with_units_count and units_count field
- the query materialized the list and do 'len(...)'
- whatever we do units_count is returned 
   - `with_units_count=true` or fields='id,name,units_count'
   - I thought it was there to have better perf and skip the count when not asked
   - ex `/api/v2/orgunittypes/?order=name&page=1&limit=20`
   - I think it was main pain point when mobile was asking for them
   - on prod : https://iaso.bluesquare.org/api/v2/orgunittypes/?order=name&page=1 doesn't return units_count by default

hugo093:
@mestachs about the `whatever we do units_count is returned`and the points related : 
* mobile  is not using this API
* like said in the PR description, the front end needs the with_counts , so not returning and making it optional is weird : in tableColumns.tsx  

```typescript 
<DeleteDialog
    keyName={settings.row.original.id.toString()}
    disabled={
        parseInt(settings.row.original.units_count, 10) > 0
    }
    titleMessage={MESSAGES.delete}
    message={MESSAGES.deleteWarning}
    onConfirm={() => {
        deleteType(settings.row.original.id);
    }}
/>```  
                    => the disabled part

mestachs:
> so not returning and making it optional is weird : in tableColumns.tsx

we have that in multiple places where performance has "bitten" us.
 - forms
 - orgunits

and clearly I'm not fan of api doing heavy queries when all you ask is fields=id,name

image

image

hugo093:
Then I need to adapt the front-end, and remove the delete button if count is not asked (I don't mind tbh, just pointing out the current state of things now, + the current UI list always ask for count as there's no dynamic columns)

hugo093:
I'll close this PR as it seems it involves too many changes (maybe a v3 in the future ?). I'll do a quick dirty fix instead

PR #3117, opened 2026-06-12
[IA-5034]: Disable account creation when 400 is raised from setup account

hugo093:

in setupaccount form/api, if a 400 is raised, the account might still be created. 

### Related JIRA tickets

IA-5034

## Changes

Made the transaction atomic, added tests

## How to test

Reproduce the video in the jira ticket , the account should not be created (i atatched that video here just in case)

## Print screen / video

https://github.com/user-attachments/assets/75cb9d3c-08b2-4cc8-94da-1c28a15b7227


iaso/tests/api/setup_account/test_password_validation.py
```diff
@@ -38,6 +40,7 @@ def test_setup_account_password_validation_error_too_similar_username(self):
         response = self.client.post(self.BASE_URL, data=data, format="json")
         result = self.assertJSONResponse(response, status.HTTP_400_BAD_REQUEST)
         self.assertHasError(result, "password", self.ERROR_PASSWORD_TOO_SIMILAR_USERNAME)
+        self.assertFalse(Account.objects.all().count())
```
quang-le: `self.assertFalse(Account.objects.exists())`?

quang-le (approved):

PR #3119, opened 2026-06-12
POLIO-2158: update scores endpoint to return more data

quang-le:

DS need more data to provide dashboard to client

### Related JIRA tickets
POLIO-2158

## Changes

- Pass spread_id i.o url to avoid parsing errors
- return full preparedness dict

## How to test

- On a polio DB
- go to `/api/polio/preparedness_dashboard/score/?date=<some date>&spread_id=<SpreadSheetImport.spread_id>`. Date should be greater than the SpreadSheetImport created_at. spread_id can be found in the django admin.
- Check that the payload contains the "critical activities" (See Jira ticket for details)

### DIFF
--- a/plugins/polio/tests/api/preparedness_dashboard/test_filters.py
+++ b/plugins/polio/tests/api/preparedness_dashboard/test_filters.py
@@ -20,7 +20,7 @@ class PreparednessScoreFilterAPITestCase(PreparednessDashboardAPIBase):
         self.client.force_authenticate(self.user_polio)
 
     def test_filter_url_returns_empty_for_nonexistent(self):
-        response = self.client.get(self.SCORE_URL, {"url": 999999, "date": "2030-01-01"})
+        response = self.client.get(self.SCORE_URL, {"spread_id": 999999, "date": "2030-01-01"})
         data = self.assertJSONResponse(response, status.HTTP_200_OK)
         self.assertEqual(data, {})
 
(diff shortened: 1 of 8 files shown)


plugins/polio/tests/api/preparedness_dashboard/test_filters.py
```diff
@@ -47,13 +47,22 @@ def test_filter_date_returns_entry_before_or_on_date(self, mock_get_preparedness
...
-        date_str = (now - timedelta(days=1)).strftime("%Y-%m-%d")
-        response = self.client.get(self.SCORE_URL, {"url": older_ssi.url, "date": date_str})
-        self.assertJSONResponse(response, status.HTTP_200_OK)
+        exact_date_ssi = SpreadSheetImport.objects.create(
+            url="https://docs.google.com/spreadsheets/d/exact-date",
+            content={"title": "Exact Date Sheet", "sheets": []},
+            spread_id="exact-date",
+        )
+        query_date = now - timedelta(days=1)
+        SpreadSheetImport.objects.filter(pk=exact_date_ssi.pk).update(created_at=query_date)
```
tdethier: This is the same date as `newer_ssi`

plugins/polio/tests/api/preparedness_dashboard/test_filters.py
```diff
@@ -47,13 +47,22 @@ def test_filter_date_returns_entry_before_or_on_date(self, mock_get_preparedness
...
+
+        date_str = query_date.strftime("%Y-%m-%d")
+        response = self.client.get(self.SCORE_URL, {"spread_id": exact_date_ssi.spread_id, "date": date_str})
+        data = self.assertJSONResponse(response, status.HTTP_200_OK)
+        self.assertIn("scores", data)
```
tdethier, line 61: We should definitely get the `SpreadSheetImport` ID in the returned payload to make sure that we're talking about the right one. This would be useful for tests, but probably also for debugging

tdethier (approved):
It worked, I'm already approving because it needs to be included in the release, but I think that there's a test to fix + a field to add

plugins/polio/tests/api/preparedness_dashboard/test_filters.py
quang-le: caught vibe coding

PR #3122, opened 2026-06-15
[IA-5137]: fix incorrect behaviors in CustomPaginationListMixin and EtlModelViewSet

hugo093:

Our customPaginationListMixin and EtlModelViewSet were not  behaving as expected (mismatch between doc/code and real behavior ) 

### Related JIRA tickets

IA-5137

## Changes

* Change the parameter `remove_results_key_if_paginated` to  `remove_results_key_if_not_paginated`
* Adapt the custom Paginator class to raise an error if results_key is not defined 
* Adapt EtlModelViewSet to behave as expected (aka force an EtlPaginator)
* Make it possible for both to use a custom results_key and a custom paginator
* Adapted current API code
 
## How to test

Cfr python tests


iaso/api/common/mixin.py
```diff
@@ -78,7 +75,7 @@ def list(self, request: Request, *args, **kwargs):
             return self.get_paginated_response(serializer.data)
 
         serializer = self.get_serializer(queryset, many=True)
-        if not self.remove_results_key_if_paginated:
+        if not self.remove_results_key_if_not_paginated:
```
tdethier: This is hard to read when you're not familiar with this mixin. Can we call it `include_results_key_if_not_paginated` and avoid the double negative with `if include_results_key_if_not_paginated:`?

I understand it's annoying because it means switching the boolean flag everywhere, but it would be much simpler to read and understand 😅

iaso/tests/api/common/test_mixin/test_custom_pagination_list_model_mixin.py
```diff
@@ -0,0 +1,90 @@
...
+    class Meta:
+        model = User
+        fields = ["id"]
+
+
+class RandomPagination(Paginator):
+    page_size = 5
+
+
+class RandomOtherPagination(PageNumberPagination):
```
tdethier: This is not used, I guess there should be another type of viewset in these tests that uses it

iaso/tests/api/common/test_mixin/test_custom_pagination_list_model_mixin.py
```diff
@@ -0,0 +1,90 @@
...
+        for i in range(10):
+            User.objects.create(username=f"user-{i}", password=f"-{i}")
+
+    def test_pagination_class(self):
+        paginator_class = ViewSet().paginator
+        self.assertIsInstance(paginator_class, RandomPagination)
+        self.assertIsInstance(paginator_class, Paginator)
+
+    def test_default_applied(self):
+        paginator_class = ViewSet().paginator
```
tdethier: I guess this should be `ViewSetWithoutPagination` instead of `ViewSet`?

iaso/tests/api/common/test_mixin/test_custom_pagination_list_model_mixin.py
```diff
@@ -0,0 +1,90 @@
...
+router.register("test", ViewSet, basename="test")
+router.register("test-remove-results", ViewSetRemoveResultsIfNotPaginated, basename="test-remove-results")
+router.register("test-default", ViewSetWithoutPagination, basename="test-default")
+urlpatterns = [
+    path("", include(router.urls)),
+]
+
+
+@override_settings(ROOT_URLCONF=__name__)
+class TestCustomListModelPaginationMixin(APITestCase):
```
tdethier, line 80: I would tests to:
- trigger the error in `get_results_key()`
- check what happens when a viewset does not override `results_key`

iaso/api/instances/views_mobile.py
```diff
@@ -40,6 +40,12 @@ class InstancesMobileViewSet(GenericViewSet, RetrieveModelMixin):
     serializer_class = MobileInstancesSerializer
     pagination_class = Paginator
 
+    @property
+    def results_key(self):
+        if self.action == "attachments":
+            return "attachments"
+        return "results"
```
tdethier, line 47: I would add a test for `results`, I didn't see it in the mobile instance api tests (`tests.api.instances.test_views_mobile`)

iaso/api/common/pagination.py
```diff
@@ -4,15 +4,23 @@
...
-    def __init__(self, results_key="results"):
-        self.results_key = results_key
+    def get_results_key(self):
+        if not getattr(self, "request", None):
+            return self.results_key
+
+        view = self.request.parser_context.get("view", None)
+        results_key = getattr(view, "results_key", None) or self.results_key
+        if not results_key:
+            raise AttributeError(f"{view.__class__.__name__} should define a results_key attribute")
```
tdethier, line 16: I would also add a test for this

tdethier (changes requested):
Main code looks good, I think the fix will work 👍 

I'm mostly requesting changes because I would add more tests in a couple of places (especially when editing the mobile API).

iaso/api/instances/views_mobile.py
hugo093: It's a safeguard and kind of dead code, this viewset only allows a retrieve and attachments action :sweat_smile:

tdethier (approved):
LGTM, kudos for spotting this issue 👍

PR #3123, opened 2026-06-15
[IA-5207]: Account management UI improvements

hugo093:

* activating the form AI module doesn’t reveal the api key setting unless a page refresh is forced

* naming: I think “address” is more common than “chemin” or “path” to refer to a full URL

### Related JIRA tickets

IA-5207

## Changes

Changed naming + invalidateQueryKeys if form_ai has been activated. Invalidatequerykey for currentUser as well since it can have some impacts.

## How to test

Go to account management, try to update your own account to enable/disable form_ai; the field key should accordingly disappear and reappear without hard refresh.


hat/assets/js/apps/Iaso/domains/accounts/edit.tsx
```diff
@@ -55,12 +58,25 @@ export const AccountsEdit: FunctionComponent = () => {
...
+                }
+                if (
+                    accountId === currentUser?.account?.id &&
+                    variables.data.modules?.includes('FORM_AI') &&
+                    !data?.modules?.includes('FORM_AI')
+                ) {
+                    queryClient.invalidateQueries(
+                        getApiAccountsAiApiKeyRetrieveQueryKey(accountId),
+                    );
+                }
```
quang-le, line 73: shouldn't this be wrapped in a useCallback since it depends on `redirectTo` and `queryClient`?

hugo093: Not sure tbh, as onSuccess is only executed (once) at the successful save and hence doesn't include rerendering the component once changed ?

quang-le: same not super sure either. works fine as is so unless we get degraded perf or weird bugs at some point, let's leave it as is

quang-le (changes requested):
translations for `iaso.accounts.label.userManualPath` and `iaso.accounts.label.forumPath`need their translations update in en.json and fr.json

Image

PR #3126, opened 2026-06-16
[IA-5227] switch to `uv` + new prod docker image + call CI actions from `iaso-ci`

oschvr:

Switch to `uv` + new prod docker image + call CI actions from iaso-ci

### Related JIRA tickets

IA-4812, IA-5227

## Changes

- replaced `pip` with `uv`:
  - in the local docker container
  - in the new prod docker container
  - in the old deployment CI
  - in the azure CI :arrow_right: not for now
  - added dependency groups
  - bumped deps:
    - django-celery-beat to [2.6.0](https://github.com/celery/django-celery-beat/tags) for compatibility
    - `boto3` to `1.42.44` to match what's required for deployments

- prepared a new CI:
  - prepared pull requests/develop/main/manual_deploy worklows that call [`iaso-ci`](https://github.com/BLSQ/iaso-ci) 
    - pull request in iaso-ci here: https://github.com/BLSQ/iaso-ci/pull/1
  - deleted old `main` CI file
  - added caching (shared as of now) that users the registry to cache docker build layers

- Updated `eb_deploy.py` python script to receive the type of environment to deploy

- Documented how to build locally

## How to test
- to test `uv`:
  - run the local docker container: all non-pip commands should run as before
  - test the prod image on https://iaso-upgrade.bluesquare.org/ - SNT is currently deployed there

- to test the CI:
  - check the CI to make sure that everything runs properly

## Notes

/
## Doc

README.md


scripts/eb_deploy.py
```diff
@@ -21,9 +21,9 @@ def eb_deploy(env_name, version_name):
         exit("Mission VERSION environment variable")
     version = os.environ["VERSION_NAME"]
 
-    # If the argument correspond to one of the env, deploy to it
-    if sys.argv[1].lower() in [x.lower() for x in eb_envs.keys()]:
-        exit(eb_deploy(sys.argv[1], version_name=version))
+    # # If the argument correspond to one of the env, deploy to it
+    # if sys.argv[1].lower() in [x.lower() for x in eb_envs.keys()]:
+    #     exit(eb_deploy(sys.argv[1], version_name=version))
```
tdethier: I guess we can uncomment those now that we know the base script works. Or even better: we add the platform checks here as well

.github/workflows/deploy_docker_manual.yml
```diff
@@ -0,0 +1,35 @@
...
+        required: false
+        default: iasops
+
+permissions:
+  contents: read
+  packages: read
+
+jobs:
+  deploy:
+    uses: BLSQ/iaso-ci/.github/workflows/docker_deploy.yml@IA-5227-decouple-ci-actions-into-dedicated-repo
```
tdethier: TODO: switch it back to `main` once the iaso-ci PR is merged

.github/workflows/develop.yml
```diff
@@ -0,0 +1,47 @@
...
+    secrets: inherit
+
+  tests:
+    uses: BLSQ/iaso-ci/.github/workflows/tests.yml@IA-5227-decouple-ci-actions-into-dedicated-repo
+    with:
+      iaso_branch: develop
+
+  deploy_on_staging:
+    needs: [ build_and_push_image, tests ]
+    uses: BLSQ/iaso-ci/.github/workflows/docker_deploy.yml@IA-5227-decouple-ci-actions-into-dedicated-repo
```
tdethier: TODO: switch it back to `main` once the iaso-ci PR is merged

docker/prod/Dockerfile
```diff
@@ -10,7 +10,7 @@ WORKDIR /opt/app
 
 # Copy package files
 COPY ../../package.json package-lock.json ./
-RUN npm install -g npm@11.13.0 && \
+RUN npm install -g npm@10.9.3 && \
```
hugo093: why deprecating it ?

docker/prod/Dockerfile
```diff
@@ -38,21 +38,22 @@ RUN apt-get update && apt-get install --yes --no-install-recommends \
...
-# Create virtual environment
-RUN python -m venv /venv
-ENV PATH="/venv/bin:$PATH"
-
 # Install python dependencies + gunicorn
-COPY ../../requirements.txt ./
-RUN pip install --no-cache-dir -r requirements.txt && \
-    pip install --no-cache-dir gunicorn
+COPY ../../pyproject.toml ./
+RUN pip install uv
```
hugo093, line 44: shouldn't we use instead one of the images that contains uv already ? like ghcr.io/astral-sh/uv:python3.9-alpine3.22

docker/prod/Dockerfile
```diff
@@ -38,21 +38,22 @@ RUN apt-get update && apt-get install --yes --no-install-recommends \
...
+# Use python from the system, not from uv
+# https://docs.astral.sh/uv/reference/environment/#uv_python_preference
+ENV UV_PYTHON_PREFERENCE=only-system
+ENV UV_NO_DEV=1
+RUN uv sync --python "$(which python3)" --group prod
```
hugo093: add --lock option?

hugo093 (changes requested):
Just a few questions that need clarifying, otherwise LGTM

docker/prod/Dockerfile
tdethier: That's probably a mistake during one of the many rebases, good catch

docker/prod/Dockerfile
tdethier: Yeah I was wondering about that too. I'm not sure if we would lose some features/commands that are used by blsqbot. I think we can keep our current base image for now because we've already introduced a lot of changes, but we should definitely run some tests on such an image later on

What do you think @mestachs?

tdethier: @oschvr FYI as well

docker/prod/docker-compose.prod.yml
Phil-V, line 1: There are some recently added env vars from the settings.py file that are missing from the list (it's required that they are listed here for them to work, correct?).

tdethier: Everything should appear in there, yeah 🙂

Which ones did you spot?

Phil-V: FORM_AI_MODEL, SETUPER_SANDBOX_PASSSWORD, AUTH_RATE_LIMIT_MAX_FAILURES, AUTH_RATE_LIMIT_COOLOFF_SECONDS

tdethier: Awesome, I'll add them, thank you very much

scripts/eb_deploy.py
```diff
@@ -21,8 +22,17 @@ def eb_deploy(env_name, version_name):
         exit("Mission VERSION environment variable")
     version = os.environ["VERSION_NAME"]
 
-    # If the argument correspond to one of the env, deploy to it
+    # # If the argument correspond to one of the env, deploy to it
     if sys.argv[1].lower() in [x.lower() for x in eb_envs.keys()]:
+        env_details = eb_envs[sys.argv[1]]
```
madewulf: The guard matches case-insensitively, but eb_envs[sys.argv[1]] uses  exact-case argument. If the env name is passed with different casing than the AWS EnvironmentName key (e.g. iaso-staging vs Iaso-staging), the guard passes but the lookup raises KeyError and the deploy crashes.

docker/prod/docker-compose.prod.yml
```diff
@@ -13,101 +13,109 @@ services:
...
-        USE_AZURE_STORAGE: ${USE_AZURE_STORAGE}
-        USE_CELERY: ${USE_CELERY}
-        USE_S3: ${USE_S3}
-        WFP_AUTH_ACCOUNT: ${WFP_AUTH_ACCOUNT}
-        WFP_AUTH_CLIENT_ID: ${WFP_AUTH_CLIENT_ID}
-        WFP_EMAIL_RECIPIENTS_NEW_ACCOUNT: ${WFP_EMAIL_RECIPIENTS_NEW_ACCOUNT}
-        WORKER: ${WORKER}
-
+        - APP_TITLE
+        - AUTH_RATE_LIMIT_COOLOFF_SECONDS
```
madewulf, line 17: switching to list-form env passing drops the `:-default` fallbacks.**
Beyond the *missing* vars already noted by Philippe, the change from `KEY: ${KEY:-default}` mappings to bare `- KEY` entries means an unset var is no longer passed at all, so Django falls back to its `settings.py` default rather than the compose default. This is a behavior change for vars that had a non-empty fallback — notably `ENABLE_GZIP` (compose default was `true`, but `settings.py` defaults it to **False**), so gzip would be off unless explicitly set. Same applies to the previously hardcoded `DEBUG: 'false'`, which is now `- DEBUG` (trusts the host env instead of forcing it off). Worth double-checking each var that previously had a `:-...` default still gets the intended value from the EB environment.

deployment/.elasticbeanstalk/config.yml
```diff
@@ -0,0 +1,15 @@
+branch-defaults:
+  default:
+    environment: iaso-upgrade
+global:
+  application_name: Iaso
+  default_ec2_keyname: devops
+  default_platform: Docker running on 64bit Amazon Linux 2023
+  default_region: eu-central-1
+  include_git_submodules: true
+  inoption_settingsnce_profile: null
```
madewulf: Typo!

madewulf (changes requested):

docker/prod/docker-compose.prod.yml
tdethier: - there's no missing env vars anymore, they were added in the meantime
- we can't provide defaults anymore because if we go back to the previous formatting, we keep on getting blank strings for all env vars that are not set, which messes with the typing in the settings. I don't mind changing the default value in settings though.... and if we don't want to change them and yet we do want another default when deploying, maybe it's time to introduce another settings file for production like @hugo093 already suggested. In the meantime I can check for `DEBUG` and `ENABLE_GZIP` on all environments

deployment/.elasticbeanstalk/config.yml
tdethier: fixed in [ce15d55](https://github.com/BLSQ/iaso/pull/3126/commits/ce15d55a8c5a236c588afecb8074bb6d850d8ebf)

scripts/eb_deploy.py
tdethier: fixed in [7744ef1](https://github.com/BLSQ/iaso/pull/3126/commits/7744ef1aa785390ddf17e4af985555e3ed2fe922)

docker/prod/docker-compose.prod.yml
tdethier: also I forgot to mention, but we don't have a docker compose right now since we deploy python directly, so we already don't have any default that are not in `settings.py`

hugo093 (approved):
LGTM

docker/prod/Dockerfile
tdethier: I made a ticket to check that later: https://bluesquare.atlassian.net/browse/IA-5282

madewulf (approved):

PR #3133, opened 2026-06-19
[CONSOLE-281]: Account usage feature

hugo093:

Implement account usage feature 

### Related JIRA tickets

CONSOLE-281

## Changes

* Created a folder `saas` within iaso that defines how iaso should behave considering the quota system in `saas`: 
  * `quotas` register the AccountUsage model in the saas registry
  * `signals` defines how iaso will increment/refresh the usage : here through signals but we could have modified the API endpoints instead.
 
## How to test
Create a project for example, go to the admin in account usage, you should see the corresponding metric. µ
Same goes for instance, profile, .. 

## Notes

Chose signals cause : 
* We don't really where we're going
* Easier to git revert if needed
* Impact the whole saves/delete , even through admin

But we should avoid signals if possible :) 

Linked to this [PR](https://github.com/BLSQ/saas/pull/12)

I don't really know how to make the CI and tests work considering the code and CI architecture 

If possible , we could use something like django-appconf to define the quotas system through a simple settings


Phil-V:
Hi! Great work with the architecture overall. Before doing a full line-by-line review I have a couple of high-level concerns with the file storage aspect:
- As there is no incremental logic yet for disk size, the full quota for file sizes is re-computed every time an instance is saved (if I understand correctly), on transaction commit. This could be a problem for the bulk import task, as we save instances in "batches" of 2000 (but each is created individually), this would make the signal fire off 2k times. If the computation of the quota is heavy, this could cause performance issues (for an already slow task). With that limitation in mind, I'm not sure the signal on Instance post-save is sustainable without refactoring the import task, unless the computation done by the receiver is very cheap, or we suppress the signal for that task somehow, or if we find another way to calculate the quota only once per transaction. 
- That brings me to the second point: the `sum_size` calculation. As we use s3 for storage, my understanding of the underlying storage backend is that accessing `.size` fires off an API call to s3 to get the size (it's not stored locally as far as I could tell, but I might be wrong). This could be manageable for an occasional one-time command, but if we have to recalculate the whole account-wide usage on each change I suspect it will be too slow/expensive. Even putting that aside, just iterating over all the objects of the account on the python side rather than aggregating from the database sounds like it would be pretty slow. 

Let me know if that sounds right and/or out of scope of the PR.

hugo093:
> Hi! Great work with the architecture overall. Before doing a full line-by-line review I have a couple of high-level concerns with the file storage aspect:
> 
> * As there is no incremental logic yet for disk size, the full quota for file sizes is re-computed every time an instance is saved (if I understand correctly), on transaction commit. This could be a problem for the bulk import task, as we save instances in "batches" of 2000 (but each is created individually), this would make the signal fire off 2k times. If the computation of the quota is heavy, this could cause performance issues (for an already slow task). With that limitation in mind, I'm not sure the signal on Instance post-save is sustainable without refactoring the import task, unless the computation done by the receiver is very cheap, or we suppress the signal for that task somehow, or if we find another way to calculate the quota only once per transaction.
> * That brings me to the second point: the `sum_size` calculation. As we use s3 for storage, my understanding of the underlying storage backend is that accessing `.size` fires off an API call to s3 to get the size (it's not stored locally as far as I could tell, but I might be wrong). This could be manageable for an occasional one-time command, but if we have to recalculate the whole account-wide usage on each change I suspect it will be too slow/expensive. Even putting that aside, just iterating over all the objects of the account on the python side rather than aggregating from the database sounds like it would be pretty slow.
> 
> Let me know if that sounds right and/or out of scope of the PR.

@Phil-V 
Indeed, you're right. 

Maybe we could switch that job to a cron job instead ? 
About the second point and the aggregation from DB, I don't see how we could do it as we only store the file location in DB?

Phil-V:
@hugo093 
A cron task might be a good solution (btw remember to use .iterator() on the querysets). 

In the long term we might have to collect and cache the file sizes (as we go or as part of the cron job) so that we don't recalculate everything from scratch every time.

Phil-V:
Very clean architecture, well done 👍

I'll pre-approve but found a minor bug: when running `manage sync_account_usage`, the submissions count for the current period keeps increasing by 1.

Phil-V (approved):

hugo093:
> Very clean architecture, well done 👍
> 
> I'll pre-approve but found a minor bug: when running `manage sync_account_usage`, the submissions count for the current period keeps increasing by 1.

Thanks i'll have a look at it :)

iaso/saas/quotas/disk_space.py
```diff
@@ -0,0 +1,103 @@
+from hat.api_import.models import APIImport
```
tdethier, line 1: i would not name anything `saas` in core, i would talk about usage in general

iaso/saas/quotas/disk_space.py
```diff
@@ -0,0 +1,103 @@
...
+    FormAttachment,
+    FormVersion,
+    ImportGPKG,
+    Instance,
+    InstanceFile,
+    Profile,
+    ReportVersion,
+)
+from plugins.saas.models.account_usage import PeriodTypeChoices, UsageTypeChoices
+from plugins.saas.registry import account_usage_registry
```
tdethier, line 13: when i see this, it makes me think that this whole directory should be in the plugin, not in here

iaso/saas/quotas/disk_space.py
```diff
@@ -0,0 +1,103 @@
...
+        return DiskSpaceAccountUsage.sum_size(map(lambda x: x.file, files))
+
+    @staticmethod
+    def calculate_import_gpkg_file_size(account) -> int:
+        """Calculate the ImportGPKG size for an account"""
+        imports = ImportGPKG.objects.filter(data_source__projects__account=account)
+        return DiskSpaceAccountUsage.sum_size(map(lambda x: x.file, imports))
+
+    @staticmethod
+    def calculate_api_import_file_size(account) -> int:
```
tdethier, line 64: all these `calculate_xxx` methods are going to go through all the existing files, which is going to make a lot of queries to S3, that's going to cost us a lot

tdethier (changes requested):
as discussed together, let's see with the team what we decide to go for

tdethier:
alright let's close this first draft and use the other PR, sorry about the mess :sweat_smile:

PR #3134, opened 2026-06-19
IA-5184 Eslint extend to all plugins

beygorghor:

ESLINT should be able to test all files in plugins repo 
CI is only testing IASO core and polio for now 
IA-5184 
## Changes

adapt config

## How to test

Eslint should pass on CI


tdethier (approved):
LGTM 👍

PR #3135, opened 2026-06-19
IA-5258: support SECRET_KEY rotation via SECRET_KEY_FALLBACKS

madewulf:
## Jira
[IA-5258](https://bluesquare.atlassian.net/browse/IA-5258)

## What

Adds support for rotating IASO's Django `SECRET_KEY` without logging everyone out, by leveraging Django's built-in [`SECRET_KEY_FALLBACKS`](https://docs.djangoproject.com/en/stable/ref/settings/#std-setting-SECRET_KEY_FALLBACKS) setting.

- `hat/settings.py`: read a new `SECRET_KEY_FALLBACKS` env variable (comma-separated) into Django's `SECRET_KEY_FALLBACKS` setting. Defaults to an empty list — no behaviour change for existing deployments.
- `.env.example` and `README.md`: document the variable and the rotation procedure.

## Why

When `SECRET_KEY` changes, Django can no longer validate signatures produced with the old key, so all sessions, password-reset tokens and signed cookies become invalid. `SECRET_KEY_FALLBACKS` lists old keys Django still uses to *verify* (but not *sign*) signatures, enabling a graceful rollover.

## Rotation procedure

1. Move the current `SECRET_KEY` into `SECRET_KEY_FALLBACKS`.
2. Set `SECRET_KEY` to the new value.
3. Deploy — existing signed data stays valid, new data is signed with the new key.
4. Remove the old key from `SECRET_KEY_FALLBACKS` once outstanding signed data has expired.

## Tests

None — this is a thin wrapper that reads an env var into a built-in Django setting.

[IA-5258]: https://bluesquare.atlassian.net/browse/IA-5258?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -3,6 +3,9 @@
 # Minimum variables needed to run the application
 # ============================================================
 SECRET_KEY="secret"
+# Optional: old secret key(s) kept valid for signature verification while rotating
+# SECRET_KEY. Comma separated, e.g. SECRET_KEY_FALLBACKS="previous_secret"
+# SECRET_KEY_FALLBACKS=""
 
 ## DATABASE
 RDS_DB_NAME="iaso"
--- a/README.md
+++ b/README.md
@@ -816,6 +816,9 @@ DB_READONLY_USERNAME=
 # Used for encryption and authorisation
 ENCRYPTED_TEXT_FIELD_KEY=
 SECRET_KEY=
+# Old secret key(s) kept valid for signature verification while rotating SECRET_KEY
+# (comma separated). See "Rotating SECRET_KEY" below.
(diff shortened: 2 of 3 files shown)


mestachs:
I think @tdethier will need to update it's docker compose file to have SECRET_KEY_FALLBACKS

mestachs (approved):

tdethier:
> I think @tdethier will need to update it's docker compose file to have SECRET_KEY_FALLBACKS

indeed, thanks for the heads up ❤️

PR #3136, opened 2026-06-22
IA-5246 feat: add task monitoring

62496:

Prevent deployments from interrupting queued or running background tasks.

### Related JIRA tickets

IA-5246

## Changes

Adds a `GET /api/tasks/deployment-status/` endpoint on the existing tasks API.

The endpoint returns whether deployment is currently safe based on global `QUEUED` and `RUNNING` tasks. It reuses the existing task filters where relevant and returns a compact response without sensitive task params.

Access is restricted to authenticated superusers for now.

## How to test

Run:

make test TARGET=iaso.tests.api.test_tasks.IasoTasksTestCase
```bash
docker compose exec iaso ./manage.py test iaso.tests.api.test_tasks.IasoTasksTestCase
```
Optional manual check:
```bash
GET /api/tasks/deployment-status/
```
You should have this result ⬇️ if there are not any task in RUNNING or QUEUED 
## Print screen / video

Capture d’écran du 2026-06-22 11-48-51

Capture d’écran du 2026-06-22 14-25-11

### DIFF
--- a/iaso/api/tasks/views.py
+++ b/iaso/api/tasks/views.py
@@ -26,8 +26,14 @@ from iaso.permissions.core_permissions import CORE_DATA_TASKS_PERMISSION, CORE_P
 from iaso.utils.s3_client import generate_presigned_url_from_s3
 
 from ...models import TaskLog
-from .serializers import ExternalTaskPostSerializer, ExternalTaskSerializer, TaskLogSerializer, TaskSerializer, \
-    DeploymentStatusSerializer
+from .serializers import (
+    DeploymentStatusSerializer,
+    ExternalTaskPostSerializer,
+    ExternalTaskSerializer,
+    TaskLogSerializer,
+    TaskSerializer,
+)
+
 
 task_service = LazyService("BACKGROUND_TASK_SERVICE")
 logger = logging.getLogger(__name__)
--- a/iaso/api/tasks/views.py
+++ b/iaso/api/tasks/views.py
@@ -2,7 +2,6 @@ import logging
 
 import django_filters
 
-from django.db.models import Count
 from django.shortcuts import get_object_or_404
 from drf_spectacular.utils import extend_schema
 from gql import Client, gql
@@ -37,6 +36,7 @@ from .serializers import (
 
 task_service = LazyService("BACKGROUND_TASK_SERVICE")
 logger = logging.getLogger(__name__)
+DEPLOYMENT_BLOCKING_STATUSES = [QUEUED, RUNNING]
 
 
 @extend_schema(tags=["Tasks"])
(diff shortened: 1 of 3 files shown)


iaso/api/tasks/views.py
```diff
@@ -152,6 +161,35 @@ def types(self, _request):
...
         return Response(available_types)
 
+    @extend_schema(responses=DeploymentStatusSerializer)
+    @action(
+        detail=False,
+        methods=["get"],
+        url_path="deployment-status",
+    )
+    def deployment_status(self, request):
+        if not request.user.is_superuser:
```
quang-le: I would add a check on `is_staff`, to make sure the user already has a status giving them access to cross account data

iaso/api/tasks/views.py
```diff
@@ -152,6 +161,35 @@ def types(self, _request):
...
+            return Response(status=status.HTTP_403_FORBIDDEN)
+
+        blocking_statuses = [QUEUED, RUNNING]
```
quang-le: Store that in a constant and use it both here and in `get_queryset`, it makes more sense

iaso/api/tasks/views.py
```diff
@@ -152,6 +161,35 @@ def types(self, _request):
...
+        blocking_tasks = self.filter_queryset(self.get_queryset())
+
+        statuses = {status: 0 for status in blocking_statuses}
```
quang-le: from ruff:
Unnecessary dict comprehension for iterable; use `dict.fromkeys` instead

iaso/api/tasks/views.py
```diff
@@ -152,6 +161,35 @@ def types(self, _request):
...
+        for status_count in blocking_tasks.values("status").annotate(count=Count("id")):
+            statuses[status_count["status"]] = status_count["count"]
```
quang-le: This is not super legible, please split it to extract at least : `blocking_tasks.values("status").annotate(count=Count("id"))`
in a variable

iaso/api/tasks/views.py
```diff
@@ -152,6 +161,35 @@ def types(self, _request):
...
+
+        blocking_tasks_count = blocking_tasks.count()
```
quang-le: unnecessary DB hit, you can compute it from `statuses`

quang-le (changes requested):
I think it's a good first go at it. I would force users to be superuser and staff to access the endpoint though. 

The current response shape is a bit hard to parse when you have many tasks running, or queued, which could happen in prod.

I would suggest to have add a level to `blocking_tasks`:
```json
{
blocking_tasks:
    {
        name: "org_unit_import",
        QUEUED: 10,
        RUNNING: 2,
        tasks: { // existing payoad}
    }
}
```

I wonder if a breakdown by account wouldn't be helpful, in case we need to contact someone (hotfix scenario)

@beygorghor  WDYT?

Image

quang-le (approved):

PR #3137, opened 2026-06-23
POLIO-2154: fix on hold filter and styling

quang-le:
The logic deciding whether to consider a campaign on hold based on its rounds status needed refinement, because the way it was displayed in the calendar was confusing

### Related JIRA tickets

POLIO-2154

## Changes

- BEFORE: if a campaign was on hold, or had at least 1 round on hold, it was considered on hold. This was creating confusion in the calendar, as all campaign rounds were marked on hold, even though only one round was.

- CHANGES:
**BACKEND**
- Remove legacy `filter_queryset`method from `campaigns` API and replace it with `CampaignFilters` that was already used by the calendar
- Remove overlapping parameter `on_hold`and keep only `campaign_category` to filter on that status
- Simplify filter logic: showing test campaigns is now entirely handled by passing the `show_test` query param. It defaults to `true` on the backend (previous behaviour) but the logic around test campaigns has been removed from campaign category filtering.
    - In effect, it means a small change in behaviour: the "All" filter will return all campaigns including tests. If that behaviour needs to be reverted, it should be changed front-end side by setting `show_test` to `false` by default and ticking the corresponding box in the UI.
- Refine logic for `on_hold` campaign category; returned campaigns are now:
    - on hold campaigns
    - campaigns that have the ongoing (if applicable) or next round on hold
    - past campaigns that have their last round on hold 
- Note that it means the 'on_hold`status takes precedence over e.g `preventive` status, i.e: a preventive campaign on hold will show in the `on_hold`category, but not the `preventive` category. This is consistent with existing behaviour. 
- Other category filters remain unchanged and mutually exclusive

**FRONT-END**
- Remove references to `on_hold` query param
- Fix some TS typing
- Use round status directly to set cell styling

## How to test
 On a polio DB, create:
 - a campaign with at least 1 future round. Put this round on hold
 - a campaign on hold
 - a finished campaign with it's last round on hold
 - a finished campaign with it penultimate round on hold
 
 Go to the calendar and tweak the campaign category filter:
 - All: should show all 4 campaigns
 - on hold: should show the first 3
     - check the display: campaign on hold should have all rounds styled as on hold, others should only have relevant rounds 
 - Combine with other filters and campaign categories (eg: preventive), see what breaks, hopefully nothing

## Print screen / video

Screenshot 2026-06-23 at 10 57 43

### DIFF
--- a/plugins/polio/js/src/domains/Calendar/campaignCalendar/types.ts
+++ b/plugins/polio/js/src/domains/Calendar/campaignCalendar/types.ts
@@ -37,7 +37,7 @@ export type CalendarRound = {
     hasSubActivities: boolean;
     subActivities: SubActivity[];
     is_planned: boolean;
-    on_holmd: boolean;
+    on_hold: boolean;
 };
 
 export type SubActivity = {
--- a/plugins/polio/js/src/domains/Calendar/campaignCalendar/types.ts
+++ b/plugins/polio/js/src/domains/Calendar/campaignCalendar/types.ts
@@ -37,6 +37,7 @@ export type CalendarRound = {
     hasSubActivities: boolean;
     subActivities: SubActivity[];
     is_planned: boolean;
+    on_holmd: boolean;
 };
 
 export type SubActivity = {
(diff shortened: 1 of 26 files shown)


plugins/polio/js/src/domains/Calendar/campaignCalendar/types.ts
```diff
@@ -37,6 +37,7 @@ export type CalendarRound = {
     hasSubActivities: boolean;
     subActivities: SubActivity[];
     is_planned: boolean;
+    on_holmd: boolean;
```
beygorghor:
```suggestion
    on_hold: boolean;
```

beygorghor (approved):
LGTM except the typo 😆

PR #3138, opened 2026-06-23
IA-5122 add geometry and code for change request sync

mestachs:

Explain here in one sentence.

### Related JIRA tickets

IA-5122

### Backend

New fields on OrgUnitChangeRequest (migrations/0395_oucr_geom_code.py)
  - Added new_geom / old_geom (MultiPolygon) and new_code / old_code fields to the change request model, so geometry and code changes can be proposed and reviewed like any other field.

Diffing (comparisons.py, synchronizer.py, synchronizer_serializers.py)
  - CodeFieldType and GeometryFieldType added to comparisons — the differ can now detect code and geometry changes between source versions.
  - synchronizer.py populates new_geom/old_geom/new_code/old_code when creating change requests, and logs a review URL at the end of a sync run.
  - dumper.py — CSV export now includes a "modified fields" column.

API (serializers/misc.py, views.py)
  - new_code, old_code, new_geom, old_geom exposed in the change request detail serializer.
  - Approve/reject endpoint handles new_geom (writes to org_unit.geom and recomputes simplified_geom) and new_code.

Model (org_unit.py)
  - __apply_changes extended for new_geom and new_code.

OrgUnitChangeRequestConfiguration
  - geometry and code added to the list of syncable fields.

  Seed data (seed_test_data.py)

  - New seed_geom_code_demo_version function: creates a demo datasource version with ~47 org unit mutations across 10 scenarios (polygon shift, point→polygon, code-only, geometry+code, etc.), runs a full sync, and prints direct URLs to the review list,
  a sample geometry CR, and the two swapped-code conflict CRs.

### Frontend 

New component: CompareGeoJsonMapComponent.tsx
  - Renders two GeoJSON layers (old = red, new = green) on one map with a legend. Used when a change request has both old_geom and new_geom.

Change request detail screen (useNewFields.tsx, types.ts, messages.ts)
  - new_geom field: shows CompareGeoJsonMap when both shapes exist, plain GeoJsonMap for new-geom-only.
  - new_code field: displayed as plain text.
  - Types updated with new_geom, old_geom, new_code, old_code.
  - New messages: geom ("Shape"), code ("Code").

Review list screen — filter and sync dropdown updated to support filtering by data_source_synchronization_id and source_version_id.

### lazy mode : 
   - `docker compose run --rm iaso manage seed_test_data --mode=seed --dhis2version=2.40.11.1`
   - adapt the dhis2 version (https://play.dhis2.org/)
   - visit the links listed in the seed command output
   - click approve, check the orgunit has its new code/
   
### use the UI 
   - go in datasource
   - pick "sync/compare" datasource
      - pick the "_geom_demo" datasource
      - create 

### test as a client

get geopackages or stuff like that and build your own test, but I guess it will be difficult
 

## Print screen / video

image

image

## Notes

Since I don't know well this part, I might have missed other features (more on the change request part)

## Doc

Technically I only lifted restrictions, don't really know the existing documentation or why we limited the sync to same datasource.


iaso/diffing/synchronizer.py
```diff
@@ -100,6 +103,21 @@ def synchronize(self) -> None:
         self._bulk_create_change_request_groups()
         self._report_progress("Bulk created change request groups")
 
+        if self.task:
+            domain = Site.objects.get_current().domain
```
tdethier: we usually use `settings.DNS_DOMAIN`, I'm not sure if this hold the correct value on every server

iaso/models/data_source.py
```diff
@@ -182,15 +182,12 @@ class DataSourceVersionsSynchronization(models.Model):
 
     Fields that can be synchronized:
 
-        ["name", "parent", "opening_date", "closed_date", "groups"]
+        ["name", "parent", "opening_date", "closed_date", "geometry", "code"]
```
tdethier: why did you remove `groups`?

iaso/models/data_source.py
```diff
@@ -182,15 +182,12 @@ class DataSourceVersionsSynchronization(models.Model):
...
 
-    Basic business use case:
-
-    - often, a pyramid of org units is created in DHIS2 and imported in IASO
-    - IASO is then used to update the pyramid
-    - but meanwhile, people may continue to make changes in DHIS2
-    - as a consequence, the two pyramids diverge
-    - so we need to synchronize the changes in the two pyramids
+    The two versions may belong to the same data source (e.g. comparing two
+    successive imports of the same DHIS2 instance) or to different data sources
```
tdethier:
```suggestion
    successive imports of the same DHIS2 instance) or two different data sources
```

iaso/models/data_source.py
```diff
@@ -267,11 +264,20 @@ def clean(self, *args, **kwargs):
...
+        # Guard against cross-account data leakage: if a data source has projects,
+        # at least one must belong to this sync's account.  Data sources with no
+        # projects are considered shared/admin resources and are exempt.
+        if self.account_id:
+            for version in (self.source_version_to_update, self.source_version_to_compare_with):
+                ds = version.data_source
+                if ds.projects.exists() and not ds.projects.filter(account_id=self.account_id).exists():
+                    raise ValidationError(
+                        f"Data source '{ds.name}' is not linked to the account of this synchronization."
+                    )
```
tdethier, line 279: I would go even further: if you're asking to sync 2 versions, they must be exclusively in your account. if one of them is linked to a project that belongs to another account, or to no project at all, then we should reject the request

In the web interface, it's currently not possible to create a data source without a project anyway

iaso/tests/models/test_data_source_versions_synchronization.py
```diff
@@ -207,23 +209,59 @@ def test_clean_data_source_versions(self):
...
+        own_project = m.Project.objects.create(name="Own project", account=self.account, app_id="own.app")
+        foreign_datasource.projects.add(own_project)
+        data_source_sync = m.DataSourceVersionsSynchronization(
+            name="Ok sync",
+            source_version_to_update=self.source_version_to_update,
+            source_version_to_compare_with=foreign_version,
+            account=self.account,
+            created_by=self.user,
+        )
+        data_source_sync.clean_data_source_versions()  # must not raise
```
tdethier, line 264: It's good to test this scenario, but I'm really not convinced about the outcome. Allowing shared stuff across accounts is going to create some crazy issues

iaso/models/org_unit.py
```diff
@@ -760,6 +760,8 @@ class Statuses(models.TextChoices):
     # `new_location_accuracy` is only used to help decision-making during validation: is the accuracy
     # good enough to change the location? The field doesn't exist on `OrgUnit`.
     new_location_accuracy = models.DecimalField(decimal_places=2, max_digits=7, blank=True, null=True)
+    new_geom = MultiPolygonField(null=True, blank=True, srid=4326, geography=True)
+    new_code = models.TextField(blank=True, default="")
```
tdethier, line 765: We should have used `CharField` on `OrgUnit.code` instead (my bad).

I'm not sure what better here:
- using `CharFields` here and having a mismatch with `OrgUnit`, but a better field
- or keeping `TextFields` everywhere but wasting DB space

iaso/models/org_unit.py
```diff
@@ -887,6 +893,17 @@ def __apply_changes(self, user: User, approved_fields: typing.List[str]) -> None
             if field_name == "new_groups":
                 self.org_unit.groups.clear()
                 self.org_unit.groups.add(*self.new_groups.all())
+            elif field_name == "new_geom":
+                from django.contrib.gis.geos import MultiPolygon as GEOSMultiPolygon
```
tdethier: You can import that globally

iaso/models/org_unit.py
```diff
@@ -887,6 +893,17 @@ def __apply_changes(self, user: User, approved_fields: typing.List[str]) -> None
...
+
+                self.org_unit.geom = self.new_geom
+                if self.new_geom is not None:
+                    simplified = self.new_geom.simplify(tolerance=0.001, preserve_topology=True)
+                    if simplified.geom_type == "Polygon":
+                        simplified = GEOSMultiPolygon(simplified)
+                    self.org_unit.simplified_geom = simplified
+                else:
+                    self.org_unit.simplified_geom = None
```
tdethier: In most places, we use `GEOSGeometry` for `geom` and `simplify_geom()` for `simplified_geom`, like [here](https://github.com/BLSQ/iaso/blob/10e72f733fb8541ac8fac455158477e87aa98422/iaso/api/org_units.py#L911), but it's not always the case (sometimes, we use `MultiPolygon` for `geom` like [here](https://github.com/BLSQ/iaso/blob/10e72f733fb8541ac8fac455158477e87aa98422/iaso/gpkg/import_gpkg.py#L174)).

Shouldn't we do something similar here, especially for simplifying the shape?

iaso/models/org_unit.py
```diff
@@ -786,6 +788,8 @@ class Statuses(models.TextChoices):
     )
     old_groups = models.ManyToManyField("Group", blank=True, related_name="+")
     old_location = PointField(null=True, blank=True, geography=True, dim=3, srid=4326)
+    old_geom = MultiPolygonField(null=True, blank=True, srid=4326, geography=True)
+    old_code = models.TextField(blank=True, default="")
```
tdethier, line 793: I'm a bit scared about the consequences of having these new two fields on `OUCR` on the mobile app

iaso/models/org_unit_change_request_configuration.py
```diff
@@ -94,6 +94,8 @@ class Type(models.TextChoices):
         "opening_date",
         "closing_date",
         "location",
+        "geom",
+        "code",
```
tdethier: Same comment about the mobile app, I'm a bit worried it might break something

iaso/tests/api/data_source_versions_synchronization/test_views.py
```diff
@@ -201,13 +201,13 @@ def test_create_json_diff_error_code_in_field_names(self):
             "source_version_to_compare_with_org_unit_group": self.group_2.id,
             "ignore_groups": True,
             "show_deleted_org_units": False,
-            "field_names": ["name", "code"],
+            "field_names": ["name", "not_a_valid_field"],
```
tdethier, line 206: I would probably rename the test and maybe make a new one where we can see that `code` is now valid

iaso/api/data_source_versions_synchronization/serializers.py
```diff
@@ -96,11 +96,19 @@ def validate(self, validated_data):
...
+        request = self.context.get("request")
+        if request:
+            account = request.user.iaso_profile.account
+            for version in (source_version_to_update, source_version_to_compare_with):
+                ds = version.data_source
+                if ds.projects.exists() and not ds.projects.filter(account_id=account.id).exists():
+                    raise serializers.ValidationError(
+                        f"Data source '{ds.name}' is not linked to the account of this synchronization."
+                    )
+
```
tdethier, line 122: Same comment about being more strict

iaso/tests/api/data_source_versions_synchronization/test_serializers.py
```diff
@@ -277,15 +309,16 @@ def test_deserialize_error_code_in_field_names(self):
                 "parent",
                 "opening_date",
                 "closed_date",
-                "code",  # code is not allowed in pyramid syncs
+                "code",
```
tdethier, line 313: Same comment about changing the test name. I would add an assert here to show that `code` no longer triggers an error

iaso/api/org_unit_change_requests/views.py
```diff
@@ -468,12 +476,32 @@ def get_reference_instance_ids(instances):
...
 
+            # Geometry (polygon) changes
+            geom_before_full = change_request.old_geom.wkt if change_request.old_geom else ""
+            geom_after_full = (
+                (change_request.new_geom.wkt if change_request.new_geom else "")
+                if "new_geom" in change_request.requested_fields
+                else (change_request.org_unit.geom.wkt if change_request.org_unit.geom else "")
+            )
+            geom_conclusion = get_conclusion(change_request, "geom", geom_before_full, geom_after_full)
+            row.extend([geom_before_full[:80], geom_after_full[:80], geom_conclusion])
```
tdethier: Why do you use `[:80]` only?

iaso/tests/api/org_unit_change_requests/test_org_unit_change_requests.py
```diff
@@ -893,6 +921,24 @@ def get_reference_instance_ids(instances):
 
         expected_row_data.extend([location_before, location_after, location_conclusion])
 
+        # Geometry changes
```
tdethier: These tests are weird, it's strange to copy the export logic inside the tests (I know it's not your fault)

iaso/tests/api/org_unit_change_requests/test_org_unit_change_requests.py
```diff
@@ -430,6 +432,32 @@ def test_partial_update_approve_fail_wrong_status(self):
         self.assertEqual(response.status_code, 400)
         self.assertIn("Status must be `new` but current status is `approved`.", response.content.decode())
 
+    @time_machine.travel(DT, tick=False)
+    def test_partial_update_approve_new_geom(self):
```
tdethier, line 444: We should probably have another test where we also approve a new `code` (or add it to this one?)

iaso/management/commands/seed_test_data.py
```diff
@@ -143,6 +145,12 @@ def handle(self, *args, **options):
         for proj in (project, project2):
             proj.feature_flags.add(flag_auth, flag_auto_upload)
 
+        site = Site.objects.get_current()
```
tdethier: Here as well, I would use `settings.DNS_DOMAIN` instead

iaso/management/commands/seed_test_data.py
```diff
@@ -489,6 +499,344 @@ def make_category_options_public(self, credentials):
...
+          I        1    null → new polygon            –                      –                               geometry
+          J        2    –                             swapped with each other  –  (non-unique code conflict)   code
+
+        This simulates merging from multiple authoritative sources (DHIS2, EF, INS)
+        across datasource boundaries.  Configure the sync with:
+          source_version_to_update       = source_version   (the original DHIS2 import)
+          source_version_to_compare_with = demo_version     (this source)
+          field_names = [name, geometry, code, opening_date, closing_date]
+        """
+        import json
```
tdethier: You can make this a global import

iaso/management/commands/seed_test_data.py
```diff
@@ -489,6 +499,344 @@ def make_category_options_public(self, credentials):
...
+        sync = DataSourceVersionsSynchronization.objects.create(
+            name="geom_code_demo_sync",
+            source_version_to_update=source_version,
+            source_version_to_compare_with=demo_version,
+            account=account,
+            created_by=self.user,
+        )
+
+        field_names = ["name", "geometry", "code", "opening_date", "closed_date"]
+        from iaso.management.commands.command_logger import CommandLogger
```
tdethier: Same comment for all local imports, please ask your agent to not do that unless there's a circular import issue 😅

iaso/management/commands/seed_test_data.py
```diff
@@ -489,6 +499,344 @@ def make_category_options_public(self, credentials):
                     except Exception as e:
                         print("Failed to fix ", category_option["name"], e)
 
+    @transaction.atomic
+    def seed_geom_code_demo_version(self, datasource, source_version):
```
tdethier: Honestly, I'm not going review this command further, it's too complicated to read and clearly looks AI-generated

iaso/diffing/synchronizer.py
```diff
@@ -394,6 +414,33 @@ def _prepare_modified_change_requests(self, diff: dict) -> tuple[OrgUnitChangeRe
             new_closed_date = self.parse_date_str(new_closed_date_value) if new_closed_date_value else None
             requested_fields.append("new_closed_date")
 
+        if "geometry" in changes:
+            after_wkt = changes["geometry"]
+            if after_wkt:
+                from django.contrib.gis.geos import GEOSGeometry
```
tdethier: Same comment about local imports

iaso/diffing/comparisons.py
```diff
@@ -1,6 +1,20 @@
...
 
+_COORD_RE = re.compile(r"(-?\d+(?:\.\d+)?) (-?\d+(?:\.\d+)?)")
+# 6 decimal places ≈ 0.11 m — tight enough to catch real edits, loose enough to absorb:
+#   - floating-point drift introduced by WKT serialization round-trips, and
+#   - precision differences between source systems (DHIS2, EF, INS) that store the
+#     same boundary with a different number of decimal places.
+_COORD_DECIMALS = 6
+
+
+def _round_coord(m):
```
tdethier, line 14: I would definitely add unit tests for this thing

iaso/diffing/comparisons.py
```diff
@@ -56,12 +70,32 @@ def access(self, org_unit):
             return org_unit.simplified_geom
         return None
 
+    def is_same(self, value, other_value):
```
tdethier, line 73: I would add tests for this

iaso/diffing/comparisons.py
```diff
@@ -56,12 +70,32 @@ def access(self, org_unit):
...
+        if value is None or other_value is None:
+            return False
+        if value.geom_type != other_value.geom_type:
+            return False
+        if isinstance(value, Point):
+            return round(value.x, _COORD_DECIMALS) == round(other_value.x, _COORD_DECIMALS) and round(
+                value.y, _COORD_DECIMALS
+            ) == round(other_value.y, _COORD_DECIMALS)
+        # Polygons: normalise ring orientation then round coordinates before comparing.
+        # normalize() mutates in place and returns None in Django 3.x, so call it
```
tdethier, line 85: what about django 4.2 that we are using? do we still need to do this?

we don't really care about django 3.x

iaso/tests/diffing/test_differ.py
```diff
@@ -185,15 +185,15 @@ def test_full_diff(self):
...
-        self.assertDictEqual(
-            region_diff_comparisons[2],
-            {
-                "field": "geometry",
-                "before": self.multi_polygon,
-                "after": new_multi_polygon,
-                "status": Differ.STATUS_MODIFIED,
-                "distance": None,
-            },
+        region_geom_diff = region_diff_comparisons[2]
```
tdethier, line 188: Good idea using a var and asserting each field individually 👍

iaso/tests/models/test_org_unit_change_request.py
```diff
@@ -487,3 +489,92 @@ def test_filter_on_user_projects(self):
...
+        change_request.approve(user=self.user, approved_fields=["new_geom"])
+
+        org_unit.refresh_from_db()
+        self.assertEqual(org_unit.geom, new_geom)
+        self.assertIsNotNone(org_unit.simplified_geom)
+        # simplified_geom is derived from geom — must be a (Multi)Polygon, not None.
+        self.assertIn(org_unit.simplified_geom.geom_type, ("Polygon", "MultiPolygon"))
+
+    @time_machine.travel(DT, tick=False)
+    def test_approve_new_geom_cleared(self):
```
tdethier, line 517: Good idea checking if fields are cleared 👍

hat/assets/js/apps/Iaso/components/maps/GeoJsonMapComponent.tsx
```diff
@@ -66,13 +64,7 @@ export const GeoJsonMap: FunctionComponent<Props> = ({ geoJson }) => {
...
                     setCurrentTile={setCurrentTile}
                 />
-
-                <GeoJSON
-                    style={{
-                        color: theme.palette.secondary.main,
-                    }}
-                    data={geoJson}
-                />
+                <GeoJSON data={geoJson} />
```
tdethier: why did you remove styling?

hat/assets/js/apps/Iaso/components/maps/CompareGeoJsonMapComponent.tsx
```diff
@@ -0,0 +1,128 @@
...
+                <Box display="flex" gap={2} mb={0.5}>
+                    {oldLabel && (
+                        <Box display="flex" alignItems="center" gap={0.5}>
+                            <Box
+                                sx={{
+                                    width: 16,
+                                    height: 4,
+                                    bgcolor: theme.palette.error.main,
+                                    borderRadius: 1,
+                                }}
```
tdethier: I think the frontend guidelines say that you need to define a const style outside of your component and call it here instead of doing it directly like that, but I'll let the frontend experts weigh in @beygorghor @quang-le

iaso/diffing/comparisons.py
```diff
@@ -1,6 +1,20 @@
...
+
 from django.contrib.gis.geos import Point
 
 
+_COORD_RE = re.compile(r"(-?\d+(?:\.\d+)?) (-?\d+(?:\.\d+)?)")
+# 6 decimal places ≈ 0.11 m — tight enough to catch real edits, loose enough to absorb:
+#   - floating-point drift introduced by WKT serialization round-trips, and
+#   - precision differences between source systems (DHIS2, EF, INS) that store the
+#     same boundary with a different number of decimal places.
+_COORD_DECIMALS = 6
```
tdethier, line 11: should we look at how many decimals we get from the mobile app and use this number instead? unless we get like 15 decimal, I guess it would make sense to use the precision of the mobile app, right?

tdethier (changes requested):
**Testing**:
- I managed to update `code` and `geom` of an org unit after a pyramid sync 👍 
- there's no option on OUCRC modal to configure an OUCRC for `code` and `geom` (not sure if we actually want that or not, but fields were added in the backend, so the feature is halfway there. Currently, they are simply ignored anyway, since the mobile app doesn't let users change these fields)
Image

- I managed to make a new CR from the mobile app and update the `location` of an org unit: everything seems to be working fine there, but I'd like @bmonjoie to confirm that

Image

------------------

**Code**:
- I left a few comments here and there, mostly about code readability and adding more tests
- I would be more strict when checking tenancy of requested versions, but that probably needs to be discussed
- it definitely reads like AI in some places

------------------
Overall I'm requesting changes to:
- improve tests
- determine what we want to do with tenancy checks
- wait for a green light from our mobile expert

but this was some nice work 👍

iaso/diffing/comparisons.py
tdethier: @bmonjoie what do you think?

iaso/models/org_unit.py
bmonjoie: It's not a problem to have them in OUCR for mobile

iaso/models/org_unit_change_request_configuration.py
bmonjoie: This should not be added here. When those fields are added to a configuration, they will be ignored by the mobile, which will be confusing for the users and hard to debug.

iaso/diffing/comparisons.py
mestachs: see /iaso/iaso/tests/diffing/test_comparisons.py

iaso/diffing/comparisons.py
mestachs: see /iaso/iaso/tests/diffing/test_comparisons.py

iaso/diffing/comparisons.py
mestachs: same see PolygonNormalizeBehaviorTests

hat/assets/js/apps/Iaso/components/maps/CompareGeoJsonMapComponent.tsx
mestachs: switched to useStyles

iaso/diffing/comparisons.py
mestachs: IMO for comparison 6 decimals is enough (note we store what we receive if they were empty/null)

Added a test to show that "too much decimals doesn't matter much" since ST_AsGeoJSON drop 9 decimals. 

https://github.com/BLSQ/iaso/pull/3138/changes#diff-bbaf0642a83b8b79b6357da572fa8a7f92a770c11fdf2bf4faba4958fe2c29c7R196

iaso/api/org_unit_change_requests/views.py
mestachs: we don't put the whole geometry in the xls (it can adds up MB depending if it has been simplified or not)

note it's now on ChangeField(...,max_len=80)
I changed this code to "handle an array of fields"
also had big perf issues with the export to csv, so I tweaked the query.
the bigger bottleneck was `.defer("data_source_synchronization__json_diff")`
that can end up in Gb when multiplied by the number of rows (and not used by the csv or list)

iaso/tests/api/org_unit_change_requests/test_org_unit_change_requests.py
mestachs: modified them to test more the output vs recalculating the formatting

image

iaso/tests/api/org_unit_change_requests/test_org_unit_change_requests.py
mestachs: added test_partial_update_approve_new_code

iaso/management/commands/seed_test_data.py
mestachs: moved it to seed_geom_code_demo.py and it's a bit better

image

mestachs:
So I think I've addressed most comments.
While doing this exercise I found that the export to csv of the change request was just horribly slow (minutes to get 1300 lines), so this adds more content to re-review.

iaso/management/commands/seed_geom_code_demo.py
```diff
@@ -0,0 +1,321 @@
...
+
+def _make_square_polygon(cx, cy):
+    size = 0.1  # ~10 km side — large enough to be unmistakably distinct from the original
+    poly = Polygon(
+        [
+            (cx - size, cy - size),
+            (cx + size, cy - size),
+            (cx + size, cy + size),
+            (cx - size, cy + size),
+            (cx - size, cy - size),
```
tdethier, line 69: i guess you don't need this line because it's a copy of line `65`, right?

iaso/api/org_unit_change_requests/views.py
```diff
@@ -284,43 +295,28 @@ def export_to_csv(self, request):
                 "created_by",
                 "updated_by",
             )
+            # old_parent/new_parent/org_unit__parent are only ever read for their `.name` below.
+            # Without this, select_related() pulls their full geometry columns for every row,
+            # which blows up the query result size (observed: "out of memory for query result"
+            # on ~1300 rows) even though that geometry is never used.
```
tdethier, line 301: nice catch 👍

iaso/api/org_unit_change_requests/views.py
```diff
@@ -387,101 +362,8 @@ def get_reference_instance_ids(instances):
...
-            # Reference instances changes
-            reference_before = get_reference_instance_ids(change_request.old_reference_instances)
-            reference_after = (
-                get_reference_instance_ids(change_request.new_reference_instances)
-                if change_request.new_reference_instances.exists()
-                else get_reference_instance_ids(change_request.org_unit.reference_instances)
-            )
-            row.extend([reference_before, reference_after])
+            for field in change_fields:
+                field.append_to(row, change_request)
```
tdethier, line 366: nice 👍

iaso/api/data_source_versions_synchronization/serializers.py
```diff
@@ -92,15 +92,34 @@ def to_representation(self, instance):
             representation["created_by"] = UserNestedSerializer(instance.created_by, read_only=True).data
         return representation
 
+    def _get_account_id(self, source_version):
+        ds = source_version.data_source
+        account_ids = list(ds.projects.values_list("account_id", flat=True).distinct())
+        if len(account_ids) > 1:
+            raise serializers.ValidationError(f"Data source '{ds.name}' is linked to more than one account.")
+        return account_ids[0] if account_ids else None
```
tdethier: this doesn't work when your account has multiple projects linked to the same data source:
Image


```suggestion
        account_ids = set(ds.projects.values_list("account_id", flat=True))
        if len(account_ids) > 1:
            raise serializers.ValidationError(f"Data source '{ds.name}' is linked to more than one account.")
        return list(account_ids)[0] if account_ids else None

```

tdethier (approved):
The feature works mostly fine (see suggestion for a bug below), I managed to create change requests for code and geometry.

Checking for other accounts when comparing sources works fine, it spotted something I wasn't even aware of on my local setup 👍 

Image

Please fix the `account_ids` below before merging, but good job 👍

PR #3139, opened 2026-06-23
[IA-5198]: Fix org unit types n+1 query

hugo093:

/api/v2/orgunittypes/ was causing a n+1 query.

### Related JIRA tickets

IA-5198

## Changes

Did the same fix as in IA-5189 + added some tests

## How to test

Make sure that the projects have some feature flags; call /api/v2/orgunittypes/ ; no n+1 should be triggered due to projects feature flags


beygorghor (approved):
Thx

PR #3140, opened 2026-06-23
POLIO-1816: deprecate vaccine authorisation

quang-le:
Client doesn't need the feature but asked that we keep it under wraps just in case

### Related JIRA tickets

POLIO-1816
## Changes

- deprecate API: return 410 on all requests
- remove from front-end menu
- deprecate front-end components

## How to test

On a polio DB:
- Open menu > Vaccine module: NOPV authorizations should be gone
- go to swagger UI: same
- try reaching `/api/polio/vaccineauthorizations/`: you should get a 410

## Print screen / video
Screenshot 2026-06-23 at 16 44 15

Screenshot 2026-06-23 at 16 41 28

### DIFF
--- a/plugins/polio/api/vaccines/stock_management/incidents/views.py
+++ b/plugins/polio/api/vaccines/stock_management/incidents/views.py
@@ -11,7 +11,7 @@ from plugins.polio.permissions import (
 )
 
 
-@extend_schema(tags=["Polio - Inicdent reports"])
+@extend_schema(tags=["Polio - Incident reports"])
 class IncidentReportViewSet(VaccineStockSubitemBase):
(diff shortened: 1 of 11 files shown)


plugins/polio/api/vaccines/stock_management/incidents/views.py
```diff
@@ -11,7 +11,7 @@
 )
 
 
-@extend_schema(tags=["Polio - Inicdent reports"])
+@extend_schema(tags=["Polio - Incident reports"])
```
tdethier, line 14: 👍

tdethier (approved):
LGTM 👍

PR #3143, opened 2026-06-24
[IA-5252]: handle soft deleted validation workflows

hugo093:

API was still returning some results related to soft deleted validation workflows

### Related JIRA tickets

IA-5252

## Changes

Checked all the endpoints and filter out soft deleted workflows
Adapted ETL endpoint by adding a new field : workflow_deprecated.

## How to test

Create a validation workflow, attach some form and start some process with instances. 
Delete the validation workflow
Everything should disappear from the UI

## Notes

Couldn't fix the form API

* Tried to implement a custom PK related field serializer that would filter out soft deleted models; but as the api code is not optimized, it would trigger n+1 queries on validation_workflow field for list endpoint. (solution would be to split serializer per operation)
* Cascading soft delete is not a good solution either because:
  * Form is attached to a validation workflow
  * Validation workflow is soft deleted, form FK is set to null
  * We decide to restore the validation workflow => form FK is still broken ? (even though if the user resaves the form from the UI , it'll be set to null)

### DIFF
--- a/hat/assets/js/apps/Iaso/constants/menu.tsx
+++ b/hat/assets/js/apps/Iaso/constants/menu.tsx
@@ -509,7 +509,8 @@ export const useMenuItems = (): MenuItems => {
                 {
                     label: formatMessage(MESSAGES.submissionsTitle),
                     key: 'submissions',
-                    permissions: paths.instancesPath.permissions,
+                    permissions:
+                        paths.validationWorkflowInstancesPath.permissions,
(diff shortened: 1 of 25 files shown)


quang-le (changes requested):
Please manage impact on the front-end and make sure nothing breaks on the mobile side.
Instance details get a 404 error when deleting workflow:

Image

quang-le (approved):

PR #3149, opened 2026-06-26
IA-5268 delete account command is no more working

mestachs:

new models are added, this is never re-tested, cycle between models make it sometimes hard to cleanup

### Related JIRA tickets

IA-5268

## Changes

tried to use to use dependencies between models to produce a delete order
some code comes from the previous command version, to handle these volumes

note that we have several cycle in our models that make toposort not working at his best
- Planning -> PlanningSamplingResult -> Planning.selected_sampling_result
- DataSource -> SourceVersion -> default_version
- OrgUnit -> OrgUnit
that's why if falls back on Bread First Search

## How to test

assuming you have a fresh db (adapt version against https://play.dhis2.org/)

```
docker compose run --rm iaso manage seed_test_data --mode=seed --dhis2version=2.40.12
docker compose run --rm iaso manage seed_test_data --mode=seed --dhis2version=2.41.8.2
docker compose run --rm iaso manage seed_test_data --mode=seed --dhis2version=2.42.5.1
docker compose run --rm iaso manage seed_test_data --mode=seed --dhis2version=2.43.0.1
```
add perhaps setuper created account

```
docker compose run --rm iaso manage delete_accounts --account-to-keep 1
```

the real test is to have a copy of the production db and test against it

## Doc

https://github.com/BLSQ/ops-runbooks/blob/f14506f3cef2b40d4e983a64fae96912b7acdc06/local-hosting/iaso-produce-tenant-dump.md

### DIFF
--- a/iaso/management/commands/delete_accounts.py
+++ b/iaso/management/commands/delete_accounts.py
@@ -63,7 +63,7 @@ from django.contrib.sessions.models import Session
 from django.core.management.base import BaseCommand
 from django.db import connection
 from django.db.models import ForeignKey, ManyToManyRel, OneToOneField, Q, QuerySet, TextField
-from django.db.models.deletion import SET_NULL, Collector, ProtectedError
+from django.db.models.deletion import PROTECT, SET_NULL, Collector, ProtectedError
 from django.db.models.functions import Cast
 
 from hat.audit.models import Modification
@@ -465,6 +465,55 @@ class Command(BaseCommand):
 
         return total
 
+    def _delete_with_protect_retries(self, queryset, label):
+        """
+        Deletes `queryset`, auto-resolving ProtectedError by deleting the blocking rows
+        first — recursively, so a PROTECT chain deeper than one level (e.g. a multi-level
+        Team.parent hierarchy, where unblocking a parent uncovers a grandparent still
+        blocked by it) doesn't escape as an uncaught exception. Each level reuses this
+        same retry logic instead of a bare, non-retried delete.
+        """
+        deleted_count = 0
+        for attempt in range(_MAX_PROTECT_UNBLOCK_ATTEMPTS):
+            try:
+                deleted_count += self._delete_qs_in_chunks(queryset, label=label)
+                return deleted_count
+            except ProtectedError as exc:
+                blocking_pks_by_model = defaultdict(list)
+                for obj in exc.protected_objects:
+                    blocking_pks_by_model[type(obj)].append(obj.pk)
+                for blocker_model, pks in blocking_pks_by_model.items():
+                    self._log(f"  [auto-unblock] {blocker_model.__name__} ×{len(pks)} blocking {label}")
+                    self._delete_with_protect_retries(
+                        blocker_model.objects.filter(pk__in=pks),
+                        label=f"{blocker_model.__name__}[unblock]",
+                    )
+        raise RuntimeError(
+            f"Could not delete {label} after {_MAX_PROTECT_UNBLOCK_ATTEMPTS} attempts — PROTECT cycle not resolved"
+        )
+
+    def _null_self_referential_fks(self, model, qs, label):
+        """
+        Nulls any nullable, self-referential PROTECT FK on `qs` (e.g. Team.parent,
+        Entity.merged_to) before it gets deleted. Mirrors the OrgUnit.parent handling
+        above: on_delete=PROTECT raises as soon as ANY row references the one being
+        deleted, whether or not that row is also part of the same delete — so a tree
+        deleted in one batch always trips over itself unless the self-reference is
+        broken first.
+
+        Scoped to on_delete=PROTECT specifically (not e.g. CommentIaso's self-ref
+        CASCADE) — CASCADE already deletes the whole subtree without our help, so
+        nulling it first would just be unnecessary writes on rows about to disappear.
+        """
+        for field in model._meta.local_fields:
+            if (
+                isinstance(field, ForeignKey)
+                and field.related_model is model
+                and field.null
+                and field.remote_field.on_delete is PROTECT
+            ):
+                self._update_qs(qs, label=f"{label}.{field.name}=NULL[self-ref]", **{field.name: None})
+
     def _update_qs(self, queryset, label=None, **fields):
         """Dry-run-aware wrapper around queryset.update(**fields)."""
         label = label or queryset.model.__name__
@@ -512,7 +561,7 @@ class Command(BaseCommand):
                     blocking_pks_by_model[type(obj)].append(obj.pk)
                 for blocker_model, pks in blocking_pks_by_model.items():
                     self._log(f"  [auto-unblock] {blocker_model.__name__} ×{len(pks)} blocking {label} (collect phase)")
-                    self._delete_qs_in_chunks(
+                    self._delete_with_protect_retries(
                         blocker_model.objects.filter(pk__in=pks), label=f"{blocker_model.__name__}[unblock]"
                     )
         else:
@@ -1003,6 +1052,21 @@ class Command(BaseCommand):
             devices = Device.objects.filter(pk__in=device_ids)
             self._delete_qs(devices, label="Device")
 
+        # ---- Step 4g: Out-of-graph — Dashboard.owned_by (third-party, no FK to Account) ----
+        # django_sql_dashboard is only installed in some environments (see the _Dashboard
+        # import guard above) and isn't part of the auto-discovered graph (no FK to
+        # Account). Its FK to User must be cleared explicitly before Users are deleted
+        # below, or Postgres raises an IntegrityError — Django's on_delete is enforced by
+        # the ORM Collector, not a DB-level constraint, so a plain User.delete() only
+        # cascades relations the auto-discovery step actually walked.
+        if _Dashboard is not None:
+            with self._doing(f"account={account.id} Dashboard.owned_by=NULL"):
+                self._update_qs(
+                    _Dashboard.objects.filter(owned_by_id__in=user_ids),
+                    label="Dashboard.owned_by=NULL",
+                    owned_by=None,
+                )
+
         # ---- Step 5: Out-of-graph — User (upstream from Profile, not in reverse FK graph) ----
         # Profile was deleted in the auto step; use the user_ids collected at the top.
         with self._doing(f"account={account.id} User"):
@@ -1077,30 +1141,21 @@ class Command(BaseCommand):
             if info.is_partial_coverage:
                 label += " ⚠ partial"
 
+            # Self-referential PROTECT FKs (e.g. Team.parent, same idea as OrgUnit.parent
+            # above) always raise ProtectedError on delete, even when both ends of the
+            # reference are in the same batch — Collector's PROTECT check fires on any
+            # existing referencing row regardless of it also being staged for deletion.
+            # Null it upfront so the whole tree deletes in one shot, correct order.
+            self._null_self_referential_fks(model, qs, label)
+
             if self.dry_run:
                 self._log(f"  [DRY RUN] would delete {label}")
                 continue
 
-            deleted_count = 0
-            for attempt in range(_MAX_PROTECT_UNBLOCK_ATTEMPTS):
-                try:
-                    deleted_count += self._delete_qs_in_chunks(qs, label=label)
-                    break
-                except ProtectedError as exc:
-                    # Auto-clear blocking objects (e.g. nullable PROTECT FKs from partial-coverage models)
-                    blocking_pks_by_model = defaultdict(list)
-                    for obj in exc.protected_objects:
-                        blocking_pks_by_model[type(obj)].append(obj.pk)
-                    for blocker_model, pks in blocking_pks_by_model.items():
-                        self._log(f"  [auto-unblock] {blocker_model.__name__} ×{len(pks)} blocking {label}")
-                        self._delete_qs_in_chunks(
-                            blocker_model.objects.filter(pk__in=pks),
-                            label=f"{blocker_model.__name__}[unblock]",
-                        )
-            else:
-                raise RuntimeError(
-                    f"Could not delete {label} after {_MAX_PROTECT_UNBLOCK_ATTEMPTS} attempts — PROTECT cycle not resolved"
-                )
+            # Belt-and-braces: auto-clears any remaining blocking objects (e.g. nullable
+            # PROTECT FKs from partial-coverage models), recursively resolving PROTECT
+            # chains deeper than one level.
+            deleted_count = self._delete_with_protect_retries(qs, label=label)
 
             if deleted_count:
                 self._log(f"  {label}: {deleted_count:,} deleted")
@@ -1235,10 +1290,12 @@ class Command(BaseCommand):
         if mode == MODE_SHOW_GRAPH:
             return self._mode_show_graph(options, discovered_models, deletion_order)
         if mode == MODE_DELETE_ACCOUNTS:
-            return self._mode_delete_accounts(options, discovered_models, deletion_order)
-        if mode == MODE_KEEP_SINGLE_ACCOUNT:
-            return self._mode_keep_single_account(options, discovered_models, deletion_order)
+            result = self._mode_delete_accounts(options, discovered_models, deletion_order)
+        elif mode == MODE_KEEP_SINGLE_ACCOUNT:
+            result = self._mode_keep_single_account(options, discovered_models, deletion_order)
+        else:
+            raise ValueError(f"unknown mode, please fix parameters: {mode!r}")
 
         self._log("Done!")
         self._print_model_stats()
-        return 0
+        return result
--- a/iaso/management/commands/delete_accounts.py
+++ b/iaso/management/commands/delete_accounts.py
@@ -165,9 +165,21 @@ _OUT_OF_GRAPH_CLEANUP_NOTES = [
     ),
     OutOfGraphCleanupNote(label="iaso.Form", reason="linked via projects M2M"),
     OutOfGraphCleanupNote(label="iaso.OrgUnitType", reason="linked via projects M2M"),
+    OutOfGraphCleanupNote(label="iaso.MatchingAlgorithm", reason="linked via projects M2M"),
+    OutOfGraphCleanupNote(label="iaso.RecordType", reason="linked via projects M2M"),
+    OutOfGraphCleanupNote(label="iaso.Device", reason="linked via projects M2M"),
+    OutOfGraphCleanupNote(label="iaso.AccountFeatureFlag", reason="linked via Account.feature_flags M2M, not a FK"),
     OutOfGraphCleanupNote(
         label="iaso.CommentIaso",
         reason="GenericForeignKey (content_type + object_pk) to OrgUnit, not a real FK",
+    ),
+    OutOfGraphCleanupNote(
+        label="auth.User",
+        reason="upstream of Profile — Profile.user is a FK held BY the discovered model, pointing outward; BFS never follows that direction",
+    ),
+    OutOfGraphCleanupNote(
+        label="iaso.OpenHEXAInstance",
+        reason="upstream of OpenHEXAWorkspace, same shape as auth.User/Profile — OpenHEXAWorkspace.openhexa_instance points outward",
     ),
     OutOfGraphCleanupNote(
         label="iaso.ReportVersion",
@@ -180,6 +192,7 @@ _OUT_OF_GRAPH_CLEANUP_NOTES = [
     ),
     OutOfGraphCleanupNote(label="django_sql_dashboard.Dashboard", reason="no FK to Account"),
     OutOfGraphCleanupNote(label="django.contrib.sessions.Session", reason="no FK to Account"),
+    OutOfGraphCleanupNote(label="iaso.Config", reason="no relation to Account at all — only M2M to auth.User"),
     OutOfGraphCleanupNote(
         label="hat_api_import.APIImport [vector_control_apiimport]",
         reason="not a iaso/plugins app and no FK to Account",
--- a/iaso/management/commands/delete_accounts.py
+++ b/iaso/management/commands/delete_accounts.py
@@ -91,13 +91,6 @@ except (ImportError, RuntimeError):
 
 
 # ---------------------------------------------------------------------------
-# Logging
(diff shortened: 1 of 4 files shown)


iaso/management/commands/delete_accounts.py
```diff
@@ -1,9 +1,55 @@
...
+
+What still needs manual code (M2M gap):
+  - DataSource / SourceVersion / OrgUnit — linked to Account via
+    Project.data_sources M2M, not a direct FK; the BFS path via credentials
+    is SET_NULL and therefore partial
+  - Form — linked via projects M2M
+  - Orphan audit/export log cleanup (content-type based, not FK based)
+
+Usage:
+  docker compose run --rm iaso manage delete_accounts_v3 --account-to-keep 1
```
tdethier: Why didn't you use the setuper?

iaso/management/commands/delete_accounts.py
```diff
@@ -12,575 +58,879 @@
...
-    # Iterate over primary keys in chunks
-    for i in range(0, total_to_delete, chunk_size):
-        chunk_pks = pk_list[i : i + chunk_size]
 
+def _chunked_delete(queryset, chunk_size=5000, label=None):
+    """Delete in streaming chunks — never loads more than chunk_size PKs at once."""
+    model = queryset.model
+    label = label or model.__name__
+    total = 0
+    t0 = time.monotonic()
```
tdethier: this variable is not used

iaso/management/commands/delete_accounts.py
```diff
@@ -12,575 +58,879 @@
...
-                print(f"Deleted {total_deleted} / {total_to_delete} records so far in {elapsed_time:.2f}s...")
-            except Exception as err:
-                print(
-                    f"Failed to delete chunk :  {total_deleted} / {total_to_delete} | {elapsed_time:.2f}s...",
-                    err,
-                    repr(err),
-                )
+            deleted = chunk_qs._raw_delete(using=queryset.db)
+        except Exception as exc:
+            _log(f"  {label}: _raw_delete failed ({exc!r}), falling back to .delete()")
```
tdethier: TIL `something!r`

iaso/management/commands/delete_accounts.py
```diff
@@ -12,575 +58,879 @@
...
+            _, counts = chunk_qs.delete()
+            deleted = counts.get(model._meta.label, 0)
```
tdethier: why don't you reuse the `label` variable?

iaso/management/commands/delete_accounts.py
```diff
@@ -12,575 +58,879 @@
...
+        Directed edges from FK holder → FK target (used for topo sort)
+    """
+    discovered = {}  # model → (filter_kwarg, on_delete_name, is_partial)
+    graph_edges = []
+
+    visited = {root_model: (None, False)}  # model → (filter_kwarg, is_partial)
+    queue = deque([(root_model, None, False)])
+
+    while queue:
+        current, current_filter, current_partial = queue.popleft()
```
tdethier: `current` is the current model? if so, I would rename it explicitly `current_model`

iaso/management/commands/delete_accounts.py
```diff
@@ -12,575 +58,879 @@
...
+
+        for rel in current._meta.related_objects:
+            if isinstance(rel, ManyToManyRel):
+                continue
+
+            child = rel.related_model
+            if child in visited:
+                continue
+            if getattr(child._meta, "abstract", False) or getattr(child._meta, "swapped", None):
```
tdethier: What about `proxy`? I guess we should also skip them, right?

iaso/management/commands/delete_accounts.py
```diff
@@ -12,575 +58,879 @@
...
+                continue
+            if not _is_project_model(child):
+                continue
+
+            field_name = rel.field.name
+            new_filter = field_name if current_filter is None else f"{field_name}__{current_filter}"
+            od_name = getattr(rel.on_delete, "__name__", str(rel.on_delete))
```
tdethier: i would call that `on_delete_value`

iaso/management/commands/delete_accounts.py
```diff
@@ -12,575 +58,879 @@
...
+
+    # Collect cross-edges between discovered models (for topo sort)
+    for model in discovered:
+        for field in model._meta.get_fields():
+            if not isinstance(field, (ForeignKey, OneToOneField)):
+                continue
+            target = field.related_model
+            if target not in discovered or target is model:
+                continue
+            od_name = getattr(field.remote_field.on_delete, "__name__", "?")
```
tdethier: same comment about naming

iaso/management/commands/delete_accounts.py
```diff
@@ -12,575 +58,879 @@
...
+            queue.append((child, new_filter, is_partial))
+
+    # Collect cross-edges between discovered models (for topo sort)
+    for model in discovered:
+        for field in model._meta.get_fields():
+            if not isinstance(field, (ForeignKey, OneToOneField)):
+                continue
+            target = field.related_model
+            if target not in discovered or target is model:
+                continue
```
tdethier, line 278: why do you skip it if it's not already in the discovered list?

iaso/management/commands/delete_accounts.py
```diff
@@ -12,575 +58,879 @@
...
-        )
+    for model in discovered:
+        ts.add(model)
 
-        print(
-            "OrgUnit remove reference_instance",
-            OrgUnitReferenceInstance.objects.filter(instance__in=Instance.objects.filter(project=project)).delete(),
-        )
-        print("Instance update", Instance.objects.filter(project=project).update(entity=None))
+    for child, parent, od_name in graph_edges:
```
tdethier: same comment about naming

iaso/management/commands/delete_accounts.py
```diff
@@ -12,575 +58,879 @@
...
+    """
+    skip_models = skip_models or set()
+    models_processed = 0
+    models_with_rows = 0
+    total_deleted = 0
 
-        chunked_delete_records(Instance.objects.filter(project=project))
+    for model in deletion_order:
+        if model not in discovered or model in skip_models:
+            continue
```
tdethier: maybe we should log some kind of warning? I guess it should not happen, but just in case

iaso/management/commands/delete_accounts.py
```diff
@@ -12,575 +58,879 @@
...
+        label = f"{model.__name__}[{filter_kwarg}]"
+        if is_partial:
+            label += " ⚠partial"
+
+        if dry_run:
+            _log(f"  [DRY RUN] would delete {label}")
+            continue
+
+        n = 0
+        for attempt in range(5):
```
tdethier: why 5 attempts?

iaso/management/commands/delete_accounts.py
```diff
@@ -12,575 +58,879 @@
...
+        models_processed += 1
+        label = f"{model.__name__}[{filter_kwarg}]"
+        if is_partial:
+            label += " ⚠partial"
+
+        if dry_run:
+            _log(f"  [DRY RUN] would delete {label}")
+            continue
+
+        n = 0
```
tdethier: could you use an explicit variable name?

iaso/management/commands/delete_accounts.py
```diff
@@ -12,575 +58,879 @@
...
 
-        print("deleting datasources might take awhile (all versions, orgunits, groups,...)")
-        print("datasources", datasources.delete())
-        print("StoragePassword", StoragePassword.objects.filter(project=project).delete())
-        print("project", project.delete())
+        if n:
+            _log(f"  {label}: {n:,} deleted")
+        if n:
+            models_with_rows += 1
+            total_deleted += n
```
tdethier: why two `if n:`?

iaso/management/commands/delete_accounts.py
```diff
@@ -12,575 +58,879 @@
...
+                count_str = "  [?]"
+        _log(f"  {flag} {model._meta.label:53s} {od_name:12s}  {filter_kwarg}{count_str}")
+
+    _log("")
+    _log("⚠ = path contains a SET_NULL FK — rows where that FK is NULL are NOT found by this filter")
+    _log("    These models need manual handling or supplementary filters.")
+    _log("")
+
+    # Models NOT discovered (manual handling required)
+    manual = [
```
tdethier: this should be a const list outside of this function

iaso/management/commands/delete_accounts.py
```diff
@@ -12,575 +58,879 @@
...
         )
+        parser.add_argument("--for-account", type=int, metavar="ID", help="Account to count rows for with --show-graph")
+        parser.add_argument("--chunk-size", type=int, default=5000)
+        parser.add_argument("--dry-run", action="store_true")
+
+    # -----------------------------------------------------------------------
+    # Helpers
+    # -----------------------------------------------------------------------
+
+    def _sql(self, sql, params=None, label="SQL"):
```
tdethier: i would probably move all external functions (like `build_fk_graph()`) here, or move all internal functions outside

iaso/management/commands/delete_accounts.py
```diff
@@ -12,575 +58,879 @@
...
+        if self.dry_run:
+            _log(f"  [DRY RUN] {label}: {sql[:100]}")
+            return 0
+        if params:
+            self.cursor.execute(sql, params)
+        else:
+            self.cursor.execute(sql)
+        n = self.cursor.rowcount
```
tdethier: why not call it `row_count`?

iaso/management/commands/delete_accounts.py
```diff
@@ -12,575 +58,879 @@
...
+        """
+        Drop-in replacement for queryset.delete() that:
+        - Uses Django's Collector to discover the full cascade (respects model evolution)
+        - Logs each model being deleted with its count
+        - Deletes in chunks instead of one giant transaction
+        For models without signals (most iaso models), Collector stores querysets in
+        fast_deletes — no instances are loaded into memory.
+        """
+
+        for attempt in range(10):
```
tdethier: why 10 attempts here?

iaso/management/commands/delete_accounts.py
```diff
@@ -12,575 +58,879 @@
...
+        collector.sort()
+
+        # Build a unified deletion plan from both fast_deletes (querysets) and
+        # data (instance dicts). Topo-sort them together so FK holders are always
+        # deleted before the models they reference, regardless of which list they
+        # came from.
+        plan_items = []  # list of (model, queryset_or_None, pk_list_or_None)
+        for qs in collector.fast_deletes:
+            plan_items.append((qs.model, qs, None))
+        for model, instances in collector.data.items():
```
tdethier: since this structure comes back a lot (model, queryset, other stuff), why not use a dataclass to hold them and access its members by using their names? no need to go back to the comments to find out what each variable means, you can simply use an explicit attribute name

iaso/management/commands/delete_accounts.py
```diff
@@ -12,575 +58,879 @@
...
+        # Topo sort via Kahn's algorithm: if model A has a FK to model B, delete A first.
+        # Use _meta.local_fields (forward FK/O2O only) — _meta.get_fields() also returns
+        # reverse relations which would flip the dependency direction and break the sort.
+        # Kahn's algorithm handles cycles gracefully: cycle participants are appended at the
+        # end in plan_items insertion order (fast_deletes before data, so SourceVersion before
+        # DataSource). DataSource.default_version is already nulled, so either order is safe.
+        # plan_models as dict preserves insertion order for cycle-node fallback.
+        plan_models = dict.fromkeys(m for m, _, _ in plan_items)
+        successors = {m: set() for m in plan_models}
+        in_degree = {m: 0 for m in plan_models}
```
tdethier: you could populate these variables directly when creating `plan_items`

tdethier: i don't understand what `in_degree` means and does

iaso/management/commands/delete_accounts.py
```diff
@@ -12,575 +58,879 @@
...
+                            in_degree[target] += 1
+        queue = [m for m in plan_models if in_degree[m] == 0]
+        ordered_models = []
+        while queue:
+            m = queue.pop(0)
+            ordered_models.append(m)
+            for s in successors[m]:
+                in_degree[s] -= 1
+                if in_degree[s] == 0:
+                    queue.append(s)
```
tdethier: please give your variables proper names, i'm sure they're sad :/

iaso/management/commands/delete_accounts.py
```diff
@@ -12,575 +58,879 @@
...
+        # Append cycle participants at the end in their original insertion order
+        cycle_nodes = [m for m in plan_models if m not in set(ordered_models)]
+        if cycle_nodes:
+            _log(f"  {label}: topo-sort cycle (ignored): {[m.__name__ for m in cycle_nodes]}")
+        ordered_models.extend(cycle_nodes)
+
+        model_to_item = {m: (qs, pks) for m, qs, pks in plan_items}
+        ordered_items = [(m, *model_to_item[m]) for m in ordered_models if m in model_to_item]
```
tdethier: what happens if a model is not in `model_to_item`? should we at least log a warning here?

iaso/management/commands/delete_accounts.py
```diff
@@ -12,575 +58,879 @@
...
+            t0 = time.monotonic()
+            if qs is not None:
+                n = qs._raw_delete(using=qs.db)
+            else:
+                n = 0
+                for i in range(0, len(pks), self.chunk_size):
+                    chunk = pks[i : i + self.chunk_size]
+                    n += model.objects.filter(pk__in=chunk)._raw_delete(using=queryset.db)
+            if n:
+                _log(f"  {lbl}: {n:,} deleted ({time.monotonic() - t0:.1f}s)")
```
tdethier: same comment about variables naming

iaso/management/commands/delete_accounts.py
```diff
@@ -12,575 +58,879 @@
...
+
+        model_to_item = {m: (qs, pks) for m, qs, pks in plan_items}
+        ordered_items = [(m, *model_to_item[m]) for m in ordered_models if m in model_to_item]
+
+        if plan_items:
+            _log(f"  {label}: cascade plan — {', '.join(sorted(m._meta.label for m, _, _ in plan_items))}")
+
+        total_steps = len(ordered_items)
+        for step, (model, qs, pks) in enumerate(ordered_items, 1):
+            lbl = f"{label}→{model._meta.label}"
```
tdethier: why don't you use `label` that you receive as parameter?

iaso/management/commands/delete_accounts.py
```diff
@@ -12,575 +58,879 @@
...
 
-        print("users_profile (tripl old table)")
-        User.objects.raw("delete from users_profile")
-
-        print(
-            "Instance created_by",
-            Instance.objects.filter(created_by__in=User.objects.filter(iaso_profile__account=account)).delete(),
+            with self._doing(f"Instance {lbl}"):
+                if not self.dry_run:
+                    _chunked_delete(instances_in_ds, self.chunk_size, f"Instance[{lbl}]")
```
tdethier: why do you use `_chunked_delete` here and not `_delete_qs` like just above? it feels like both do more or less the same, maybe we should refactor both into one, or one could be deleted

iaso/management/commands/delete_accounts.py
```diff
@@ -12,575 +58,879 @@
...
+                    Instance.objects.filter(form_version__in=f.form_versions.all()).delete()
+                    f.delete_hard()
+            except Exception:
+                _log(traceback.format_exc())
 
-    def delete_modification_logs_and_comments(self, cursor):
-        dump_modification_stats()
-        # this part is a bit brittle, may should become a loop on ContentType.objects.all() ?
+        try:
+            from django_sql_dashboard.models import Dashboard
```
tdethier: please make all imports global, unless there's a circular imports issue, which should not happen here

iaso/management/commands/delete_accounts.py
```diff
@@ -12,575 +58,879 @@
...
-        cursor.execute(
-            "delete from iaso_exportrequest where id not in ( select distinct export_request_id from iaso_exportstatus )"
-        )
-
-        print("******* Deleting Modification (might take a while too)")
+        # ---- Step 6: Account itself ----
+        if not self.dry_run:
+            account_id, account_name = account.id, account.name
+            account.delete()
+            _log(f"  Account {account_id} ({account_name!r}) deleted")
```
tdethier: not sure why `!r` is used here

iaso/management/commands/delete_accounts.py
```diff
@@ -12,575 +58,879 @@
...
 
-        self.delete_modification_logs_and_comments(cursor)
+        self._delete_qs(Device.objects.filter(projects=None), label="Device[orphan]")
```
tdethier: this is already done in `_post_flight()`

iaso/management/commands/delete_accounts.py
```diff
@@ -12,575 +58,879 @@
...
+        self._sql("DELETE FROM vector_control_apiimport", label="vector_control_apiimport")
```
tdethier: and this is also already done, in `_pre_flight()`

iaso/management/commands/delete_accounts.py
```diff
@@ -12,575 +58,879 @@
...
+        if full_cleanup:
+            self._post_flight(account_to_keep)
 
-        dump_diffs(counts_before, counts_after)
-        print("******* Review external credentials left")
-        for ext_cred in ExternalCredentials.objects.all():
-            print(ext_cred.id, ext_cred.url, ext_cred.login, ext_cred.name)
+        _log("Done!")
+        _log("Row counts after deletion:")
+        from django.apps import apps as django_apps
```
tdethier: same comment for all local imports

iaso/management/commands/delete_accounts.py
```diff
@@ -12,575 +58,879 @@
...
+
+        account_to_keep = None
+        full_cleanup = False
+
+        if options.get("account_to_keep") is not None:
+            account_id_to_keep = options["account_to_keep"]
+            account_to_keep = Account.objects.get(pk=account_id_to_keep)
+            _log(f"Keeping: {account_id_to_keep} — {account_to_keep.name!r}")
+            accounts_to_delete = list(Account.objects.exclude(pk=account_id_to_keep).order_by("-id"))
+            random.shuffle(accounts_to_delete)
```
tdethier: why do you shuffle?

iaso/management/commands/delete_accounts.py
```diff
@@ -12,575 +58,879 @@
...
+            self.cursor.execute(sql)
+        n = self.cursor.rowcount
+        if n:
+            _log(f"  {label}: {n:,} deleted")
+        return n
+
+    def _delete_qs(self, queryset, label=None):
+        label = label or queryset.model.__name__
+        if self.dry_run:
+            _log(f"  [DRY RUN] {label}: ~{queryset.count()}")
```
tdethier: something's wrong with the logging here when running the command in dry run

Image

iaso/management/commands/delete_accounts.py
```diff
@@ -12,575 +58,879 @@
...
+        try:
+            qs = manager.filter(**{filter_kwarg: account})
+        except Exception as exc:
+            _log(f"  [skip] {model.__name__}: cannot build queryset ({exc})")
+            continue
+
+        models_processed += 1
+        label = f"{model.__name__}[{filter_kwarg}]"
+        if is_partial:
+            label += " ⚠partial"
```
tdethier:
```suggestion
            label += " ⚠ partial"
```

iaso/management/commands/delete_accounts.py
tdethier: well actually it's not just in dry run, something's off in that section of the command

Image

iaso/management/commands/delete_accounts.py
```diff
@@ -12,575 +58,879 @@
...
+    # -----------------------------------------------------------------------
+    # Pre/post-flight cleanup
+    # -----------------------------------------------------------------------
+
+    def _pre_flight(self):
+        try:
+            self._sql("DELETE FROM users_profile", label="users_profile")
+        except django.db.utils.ProgrammingError:
+            pass
+        self._sql("DELETE FROM vector_control_apiimport", label="vector_control_apiimport")
```
tdethier: this deleted all API imports of all accounts when i wanted to delete a single account

Image

iaso/management/commands/delete_accounts.py
```diff
@@ -12,575 +58,879 @@
...
+        with self._doing(f"User account={account.id}"):
+            self._delete_qs(users, label="User")
+
+    # -----------------------------------------------------------------------
+    # Pre/post-flight cleanup
+    # -----------------------------------------------------------------------
+
+    def _pre_flight(self):
+        try:
+            self._sql("DELETE FROM users_profile", label="users_profile")
```
tdethier: what is this table, i don't have it? something from trypelim?

iaso/management/commands/delete_accounts.py
tdethier: and this will also delete data from all accounts, even the one you want to keep

tdethier (changes requested):
**Testing**:
- the new options are great (listing accounts, showing graph, choosing what to delete or what to keep)
- when testing the `account-to-keep <account_id>` mode, the command became unresponsive after a while (I had around 60 accounts, nearly all of them are setupers)

Image

- deleting a single account worked fine...
- ...but it deleted `APIImports` for all accounts :arrow_right: that's a major bug to fix

-------------

**Code**:
- I left plenty of commends, most of them are related to improving code readability (in general, code was hard to read, please use proper names for variables)
- going for a graph-based flow is a great idea, but what if we properly used cascade constraints everywhere? wouldn't it make everything much simpler?

iaso/management/commands/delete_accounts.py
mestachs: it exists in prods... I suspect a trypelim artefacts

image

tdethier:
update: the command to delete all accounts except one went through, it was just slow
image

mestachs:
la prochaine je vous laisse corriger

hugo093 (approved):
LGTM 

Just afraid of the static lists of M2M that is not dynamic and thus we might forget to update it. But it can be an improvement considering the size of the PR.

PR #3159, opened 2026-07-01
IA-5117: submission diff UI

quang-le:

Users need an interface to see difference between original submission and re-submission when validation via workflow

### Related JIRA tickets

IA-5117

### Back-end
- **Change audit logging in instance bulk upload to avoid having multiple audit entries on instance creation**. This is necessary because the whole feature assumes that 1 audit entry = 1 save. **Review this carefully** because there were tests making sure the targeted logging was happening. However, as mentioned, this feature needs to have only one `Modification` object per `Instance` at creation
-  Add `InstanceModificationSerializer`(extends `ModificationSerializer`) to return fields necessary for front-end component (form descriptor, possible fields)

### Front-end
- Deduplicate `useGetInstance` hook
- Run Orval for submission-diff endpoint
- Add new route fro diff view
- Add `ValidateInstance`component + related components and utils
- Update `CompareInstanceLogs` and related components (fix TS, share some code).
- Add front-end tests

### Feature behaviour

- From instance details view:
    - If the instance to validate is a first submission: flow is not changed
    - If it's a resubmission: approve/reject buttons are replace with a "review" button that redirects to the diff comparison page
- When comparing:
    - Only fields with diff are shown by default
    - All fields can be shown using a toggle switch
    - User can choose among available levels which one they want to validate (based on user roles)
    - If user bypasses some steps, the steps to be bypassed are shown in the UI
    - Reject button is disabled unless user leaves a comment. A helper text under the comment's textarea and a tooltip on the disabled button make that behaviour explicit to the user
    - On save, user is redirected to the submission details page

## How to test

- required: running task worker, mobile device/emulator, form with workflow, user with user roles appropriate for the workflow (caution: super users can't be configured via the interface for this)

- With a mobile device/emulator:
    - Make a submission of a form that is associated with a workflow
    - go to that submission's details page (Web)
    - reject it (Web)
    - back on mobile: refresh data and resubmit the instance
- On the web:
    - Go to the submission's details
    - click one of the "review buttons"
    - See "feature behaviour" above for behaviour to test 

## Print screen / video

https://github.com/user-attachments/assets/91913167-bc98-4c71-a135-707007c86f3c

## Notes

There is a known bug with steps that don't have the `can_skip_previous_step` active. I will open a separate ticket for that: If you try to validate a `NEXT_STEP` without the flag, the bypass API will refuse and the regular API will throw a 404.


iaso/api/instance_diff/serializers.py
```diff
@@ -18,4 +19,61 @@ class Meta:
...
+    files = serializers.SerializerMethodField()
+    possible_fields = serializers.SerializerMethodField()
+    form_descriptor = serializers.SerializerMethodField()
+
+    class Meta(ModificationSerializer.Meta):
+        fields = [*ModificationSerializer.Meta.fields, "files", "possible_fields", "form_descriptor"]
+
+    @classmethod
+    def instance_queryset(cls, queryset):
+        return queryset.select_related(*cls.INSTANCE_SELECT_RELATED).prefetch_related(*cls.INSTANCE_PREFETCH_RELATED)
```
hugo093: it doesn't belong here but in get_queryset() method of the viewset

iaso/api/instance_diff/serializers.py
```diff
@@ -18,4 +19,61 @@ class Meta:
 
     @extend_schema_field(JsonPatchOperationSerializer(many=True, allow_empty=True))
     def get_diff(self, obj):
-        return jsonpatch.JsonPatch.from_diff(obj.past_value[0]["fields"], obj.new_value[0]["fields"]).patch
+        past_value = obj.past_value[0].get("fields", None) if obj.past_value else None
+        new_value = obj.new_value[0].get("fields", None) if obj.new_value else None
+        return jsonpatch.JsonPatch.from_diff(past_value, new_value).patch
+
+
+class InstanceModificationSerializer(ModificationSerializer):
```
hugo093: ModificationSerializer on itself has no purpose, just merge the two serializers

iaso/tests/api/instance_diff/test_list.py
hugo093, line 13: I notice that i didn't actually checked upon swagger compliance .. can you do it ?

iaso/tasks/process_mobile_bulk_upload.py
```diff
@@ -87,9 +87,12 @@ def process_mobile_bulk_upload(api_import_id, project_id, task=None):
 
                     for instance_data in instances_data:
                         uuid = instance_data["id"]
+                        instance = Instance.objects.get(uuid=uuid)
```
hugo093, line 90: could it happen that it might raise a 404 ?

hugo093 (changes requested):
A first pass on the backend , a few minor comments. Will look at the front tomorrow

iaso/api/instance_diff/serializers.py
quang-le: The `get_queryset` method of the viewset returns a `Modification` queryset.

iaso/tasks/process_mobile_bulk_upload.py
quang-le: No, `import_instances` creates the objects in the DB based on the data in the zip file.
The instance is actually already accessed this way in `process_instance_xml`, I'll deduplicate that `get` and pass the instance directly

iaso/api/instance_diff/serializers.py
hugo093: but the only place you use this in actually in the viewsets , so it should belong there ? I just don't get why you made this a classmethod on a serializer ?

hat/assets/js/apps/Iaso/domains/instances/validate/index.tsx
```diff
@@ -0,0 +1,200 @@
...
+                                    isLogDetailError={false}
+                                    emptyPlaceholder={
+                                        MESSAGES.noPreviousVersion
+                                    }
+                                    headerA={MESSAGES.previous}
+                                    headerB={MESSAGES.current}
+                                />
+                            )}
+                        </Grid>
+                    )}
```
hugo093, line 195: This is a big source of CLS imho, while it works we should improve that later. Moreover it doesn't display a general loading state. 

E.g instance and instanceDiff are loading/fetching; you get kind of a blank empty page

hat/assets/js/apps/Iaso/domains/instances/validate/index.a11y.test.tsx
```diff
@@ -0,0 +1,115 @@
...
+    InstanceLogDetail: ({
+        emptyPlaceholder,
+    }: {
+        emptyPlaceholder?: { defaultMessage?: string };
+    }) => (
+        <div data-testid="instance-log-detail">
+            {emptyPlaceholder && <span>{emptyPlaceholder.defaultMessage}</span>}
+        </div>
+    ),
+}));
```
hugo093, line 51: while i appreciate the a11y test, mocking half of the components defeats its purpose. 

it should be done through msw and integration like testing, but in another ticket :)

hat/assets/js/apps/Iaso/domains/instances/validate/index.test.tsx
```diff
@@ -0,0 +1,253 @@
+import React from 'react';
```
hugo093, line 1: all of this should actually be moved to integration testing. but in another ticket again

hat/assets/js/__tests__/integration/instanceDiff/InstanceDiff.integration.test.tsx
```diff
@@ -0,0 +1,197 @@
...
+            content_type: 'iaso.instance',
+            object_id: instanceId,
+            diff: [{ op: 'add', path: '/field', value: 'new' }],
+            past_value: {},
+            new_value: { field: 'new' },
+        },
+    ],
+});
+
+describe('Instance diff integration', () => {
```
hugo093, line 60: this test suite is kind of wrong; you're not testing the page in itself, just the hooks. you should call renderWithThemeAndIntlProvider everytime

hugo093 (approved):
Some details to change, fixed some stuff myself on backend and frontend. I created a follow up ticket though. LGTM otherwise

PR #3167, opened 2026-07-03
POLIO-2162: fix campaign creation save bug

quang-le:

When creating a new polio campaign, saving wouldn't redirect, but subsequent saves would return a 400 because the front-end would attempt a POST i.o PUT

### Related JIRA tickets

POLIO-2162

## Changes

- When saving, onSuccess: check if id is saved in state, if not, redirect to details page

## How to test
 
On a polio Db, open creation modal
Fill in enough data to enable saving 
save
change data
save again
yay, it works

## Print screen / video

https://github.com/user-attachments/assets/7ecbc82d-d253-4abf-aad8-6a8d6e20e39a


tdethier (approved):
LGTM 👍

PR #3171, opened 2026-07-06
[IA-4872]: Missions CRUD feature : backend

hugo093:

Mission CRUD feature  : backend

### Related JIRA tickets

IA-4872

### Data modelling

Re-implemented the `Mission` data modelling to use [django-polymorphic](https://django-polymorphic.readthedocs.io/)

The base model is `Mission`
We have 3 different mission types models that are derived from `Mission`: 
    1. MissionForm
    2. MissionOrgUnitType
    3. MissionEntityType

Each one of those models has its own specificity and fields while still allowing to retrieve them all through `Mission.objects.all()` 

The `Mission` model cannot be abstract as it's linked through a M2M field in `Planning`
 
The get_assignements abstract method will be used later like a strategy pattern to retrieve all assignements (see 
`get_assignements` method in `MobilePlanningV2Serializer`)

All mission models are soft deletable.

Old planning have been migrated to new plannings through migrations + testing

### Backend

* Basic CRUD operations for mission with polymorphic support under `/api/microplanning/missions/`
* Basic endpoint for FE to retrieve mission type options
* Note that it won't be possible to update the mission_type directly, it will have too much consequences and it's better to delete and create a new one.
* Testing
* Swagger compliance

### Admin

* Added a quick admin integration for Mission, MissionForm, MissionOrgUnitType and MissionEntityType

### Misc

* Some ESLint fixes on the front end
* Moved `CurrentAccountDefault` to `iaso.api.common.serializer_fields` and adapted codebase
* Translations
* Reorganized `models.microplanning`
* Added new permissions
* New dependencies in uv.lock
* Skipped and commented some tests while awaiting for next backend feature (forms, assignments and planning)

## Notes

The nice to have feature would be, once we agree on this data modelling and finish the integration with planning: how to handle soft deletions through the m2m relationships and UI ? It will be a separate ticket


iaso/api/missions/serializers/update/mission_entity_type.py
```diff
@@ -0,0 +1,133 @@
...
+        through_data = validated_data.pop("forms")
+
+        instance = super().update(instance, validated_data)
+
+        # get the current m2m fields
+        existing = {obj.form_id: obj for obj in instance.missionentitytypethroughform_set.all()}
+        incoming = {item["form"].id: item for item in through_data}
+
+        # delete
+        MissionEntityTypeThroughForm.objects.filter(form_id__in=list(existing.keys() - incoming.keys())).delete()
```
Phil-V: All `MissionEntityTypeThroughForm` are deleted for the same form ids, not just for the same mission? (same question for the update views of the other mission types that have similar delete methods)

iaso/migrations/0395_mission_alter_corepermissionsupport_options_and_more.py
```diff
@@ -0,0 +1,403 @@
...
+    Mission = apps.get_model("iaso", "Mission")
+    MissionForm = apps.get_model("iaso", "MissionForm")
+    MissionFormThroughForm = apps.get_model("iaso", "MissionFormThroughForm")
+    PlanningFormThrough = apps.get_model("iaso", "Planning_forms")
+
+    # get content_type for mission_form
+    mission_form_ct = ContentType.objects.get_for_model(MissionForm)
+
+    for planning in Planning.objects.prefetch_related("forms").select_related("project__account").all():
+        account = planning.project.account
```
Phil-V: Perhaps we should filter out projects with a null account (if they exist), add a check for null when accessing the account, or verify the prod data just to be safe.

iaso/models/missions.py
```diff
@@ -0,0 +1,190 @@
...
+    MISSION_TYPE = MissionType.ORG_UNIT_AND_FORM
+
+    org_unit_type = models.ForeignKey("OrgUnitType", on_delete=models.PROTECT, related_name="mission_org_unit_type")
+    forms = models.ManyToManyField("Form", related_name="mission_org_unit_types", through=MissionOrgUnitTypeThroughForm)
+    min_cardinality = models.PositiveIntegerField(
+        default=1, help_text="Minimum number of times this form should be filled"
+    )
+    max_cardinality = models.PositiveIntegerField(
+        null=True, blank=True, help_text="Maximum number of times this form can be filled (null = unlimited)"
+    )
```
Phil-V, line 152: Is it intended to define the cardinality both on the mission model and the the throughform model?

Phil-V (approved):
Good job with the polymorphic implementation! 👍

I'll pre-approve as this is part of a broader WIP PR, overall LGTM aside from a couple of notes.

iaso/api/missions/serializers/update/mission_entity_type.py
hugo093: good spot :) will fix

iaso/models/missions.py
hugo093: Yes, but only for entity type and org unit type

PR #3172, opened 2026-07-06
[IA-5298] implement form dropdown endpoint on form

Crebert08:

This PR is about implementing a form dropdown

### Related JIRA tickets

[IA-5298](https://bluesquare.atlassian.net/browse/IA-5298)

### Features
- Implemented the dropdown serializer 
- Implemented the dropdown's view 
- Refactored the queryset to check which serializer is being used
- Modified the hook that call the dropdown endpoint

### Refactor
- Moved the serializers into a single file
- Refactored the test file for clear structure

## How to test

//

## Print screen / video

//

## Notes

## Doc

[IA-5298]: https://bluesquare.atlassian.net/browse/IA-5298?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iaso/api/forms/views.py
```diff
@@ -202,6 +214,21 @@ def get_queryset(self):
...
+            return FormDropdownSerializer
+        return FormSerializer
+
+    @extend_schema(responses={200: FormDropdownSerializer(many=True)})
+    @action(detail=False, methods=["get"], pagination_class=None)
+    def dropdown(self, request: Request, *args, **kwargs):
+        """
+        Lightweight forms list for dropdowns.
+        """
+        queryset = self.get_queryset()
```
hugo093: Apply the filter as well
```suggestion
        queryset = self.filter_queryset(self.get_queryset())
```

iaso/api/forms/views.py
```diff
@@ -202,6 +214,21 @@ def get_queryset(self):
...
+        serializer = self.get_serializer(queryset, many=True)
+        return Response({self.results_key: serializer.data})
```
hugo093: No need for the result key here, especially since you tell swagger with your extend schema that we return only a list and not a {key: list} 


```suggestion
        return Response(serializer.data)
```

iaso/api/forms/views.py
```diff
@@ -202,6 +214,21 @@ def get_queryset(self):
 
         return queryset
 
+    def get_serializer_class(self):
```
hugo093, line 217: ❤️

iaso/api/forms/serializers.py
```diff
@@ -219,3 +219,12 @@ def create(self, validated_data):
         form = super(FormSerializer, self).create(validated_data)
         log_modification(None, form, FORM_API, user=self.context["request"].user)
         return form
+
+
+class FormDropdownSerializer(FormSerializer):
```
hugo093: Don't make it depend on the FormSerializer:
* it might have unwanted side effect in the future (like you modify something on FormSerializer and you didn't realise it affects this serializer)
* The formserializer is a dynamic field serializer, we don't need that in the dropdown serializer
* By doing so you inherit all the fields from FormSerializer .. and swagger is wrong then 

Image

Also try to use label and value, to match the FE needs :) 

suggestion : 
```python
class FormDropdownSerializer(ModelSerializer):
    value = serializers.IntegerField(source="id", read_only=True)
    label = serializers.CharField(source="name", read_only=True)
    class Meta:
         model = Form
         fields = ["value", "label"]
```

iaso/tests/api/test_forms.py
```diff
@@ -1141,3 +1143,41 @@ def test_form_details_full_details(self):
...
+        dropdown_forms = self.assertJSONResponse(self.client.get("/api/forms/dropdown/"), 200)["forms"]
+        self.assertTrue(dropdown_forms)
+        for form in dropdown_forms:
+            self.assertEqual(set(form.keys()), {"id", "name"})
+
+        list_form = self.assertJSONResponse(self.client.get("/api/forms/"), 200)["forms"][0]
+        for heavy_field in ["period_type", "latest_form_version", "instances_count"]:
+            self.assertIn(heavy_field, list_form)
+
+    def test_dropdown_returns_only_the_requested_fields(self):
```
hugo093: I don't agree with this, dropdown should only return id, name. It shouldn't be dynamic. If we want dynamic fields, better use the standard list endpoint

iaso/tests/api/test_forms.py
```diff
@@ -1141,3 +1143,41 @@ def test_form_details_full_details(self):
...
+        self.client.force_authenticate(user=self.yoda)
+        response = self.client.get("/api/forms/dropdown/", {"fields": "id,name,form_id,period_type"})
+        forms = self.assertJSONResponse(response, 200)["forms"]
+
+        survey = {f["name"]: f for f in forms}["Hydroponic public survey"]
+        self.assertEqual(set(survey.keys()), {"id", "name", "form_id", "period_type"})
+        self.assertEqual(survey["form_id"], "sample2")
+        self.assertEqual(survey["period_type"], "QUARTER")
+
+    def test_dropdown_default_avoids_the_list_prefetch_and_version_queries(self):
```
hugo093: I get why you tested that, but if we want this, we should maybe redo the get_queryset() method , and catch `if self.action == "dropdown"` just after the filtering and return the queryset

iaso/tests/api/test_forms.py
```diff
@@ -1141,3 +1143,41 @@ def test_form_details_full_details(self):
 
         self.assertTrue(at_least_one_query_with_max_and_count)
         self.assertEqual(len(ctx.captured_queries), 11)
+
+    def test_dropdown_uses_a_lighter_serializer_than_the_list(self):
```
hugo093: General note : 
Can you split this dropdown testing into another testing class and/or file ?

like : 

```
class FormsDropdownAPITestCase(SwaggerTestCaseMixin,  APITestCase):
...
```

And add standard tests and assert : 

```
def assertValidData(data, expected_length):
      self.assertEqual(len(data), expected_length)
      self.assertResponseCompliantToSwagger(data, "FormDropdown", True)

def test_filters(self):
     pass

def test_dropdown(self):
      pass

def test_num_queries(self):
       pass

def test_permissions(self):
     pass
```

As the test file is already 1000+ lines long , it would be great :)

hat/assets/js/apps/Iaso/domains/forms/hooks/useGetFormsDropdownOptions.ts
```diff
@@ -85,7 +85,7 @@ export const useGetFormsDropdownOptions = (
 
     const queryKey = useMemo(() => ['forms', queryParams], [queryParams]);
 
-    const url = makeUrlWithParams('/api/forms/', queryParams);
+    const url = makeUrlWithParams('/api/forms/dropdown/', queryParams);
```
hugo093: Seeing how it's used everywhere, I wouldn't modify this I'm afraid (I see in the codebase there are some extra fields and so on). 

I'd instead make a separate component FormDropdown (and FormAsyncDropdown ?) with the related react query hook . That component should also check if the user has the related permission :) 

That way we can gradually replace in the code those forms dropdown without breaking anything.

hat/assets/js/apps/Iaso/domains/forms/hooks/useGetFormsDropdownOptions.test.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+    it('appends extraFields to the requested fields', () => {
+        renderHook(() =>
+            useGetFormsDropdownOptions({
+                extraFields: ['period_type', 'label_keys'],
+            }),
+        );
+
+        expect(requestedFields()).toBe('id,name,period_type,label_keys');
+    });
```
hugo093: see backend remarks, most of those tests won't be needed
BUT you can keep them for the list endpoint and current component :D

hat/assets/js/apps/Iaso/domains/forms/hooks/useGetFormsDropdownOptions.test.tsx
```diff
@@ -0,0 +1,132 @@
...
+    }));
+};
+
+// The `fields` query param the hook built for the request.
+const requestedFields = (): string => {
+    const [, queryParams] = mockMakeUrlWithParams.mock.calls.at(-1) ?? [];
+    return queryParams?.fields;
+};
+
+describe('useGetFormsDropdownOptions', () => {
```
hugo093: thanks for the front end tests :) you can also have a look at msw and how we use it if you're interested

hugo093 (changes requested):
Changes to be done, both on front-end and back-end.

hat/assets/js/apps/Iaso/domains/forms/hooks/useformsDropdown.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+    return useSnackQuery({
+        queryKey,
+        queryFn: () => getRequest(url),
+        options: {
+            enabled,
+            staleTime: 1000 * 60 * 15,
+            cacheTime: 1000 * 60 * 5,
+            keepPreviousData: true,
+            select: (data: FormDropdown[]): FormDropdown[] => data ?? [],
```
hugo093: the select is useless there

hat/assets/js/apps/Iaso/domains/forms/hooks/useFormsDropdown.tsx
```diff
@@ -0,0 +1,44 @@
+import { useMemo } from 'react';
```
hugo093, line 1: rename file => useformsDropdown to useFormsDropdown

iaso/tests/api/test_forms_dropdown.py
```diff
@@ -0,0 +1,129 @@
...
+        return reverse("forms-dropdown")
+
+    def assertValidData(self, data, expected_length):
+        self.assertEqual(len(data), expected_length)
+        self.assertResponseCompliantToSwagger(data, "FormDropdown", as_array=True)
+
+    def test_permissions(self):
+        res = self.client.get(self.url)
+        data = self.assertJSONResponse(res, status.HTTP_200_OK)
+        self.assertValidData(data, 0)
```
hugo093, line 53: not your fault there, but it still bugs me that we return a 200 with empty results instead of a 401/403

hugo093 (approved):
a few details here and there to adjust.
I'd also have expected a FormDropdown React component, but it can be done in another ticket.

I'll pre approve this as changes are minor, thanks for the work

PR #3174, opened 2026-07-07
IA-5302 prefetch user when issuing a dropdown query

mestachs:

When searching/pausing on a short non selective you have a lot of profiles and for each of them we do query to get the user.

### Related JIRA tickets

IA-5302

## Changes

select_related if dropdown call

## How to test

see the test
but you can create plenty of user (via ./manage.py shell or xlsx import)
and then use the entity screen (filter on submitter)

## Doc

### DIFF
--- a/iaso/api/profiles/views.py
+++ b/iaso/api/profiles/views.py
@@ -124,7 +124,14 @@ class ProfilesViewSet(ModelViewSet):
 
     def get_queryset(self):
         account = self.request.user.iaso_profile.account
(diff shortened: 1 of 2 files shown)


hugo093 (approved):
LGTM went from : 

```
http://localhost:8000/api/profiles/dropdown/
| RESP |  default  |    9     |    0     |    9     |     3      |
```

To : 

```
http://localhost:8000/api/profiles/dropdown/
| RESP |  default  |    7     |    0     |    7     |     0      |
```

PR #3176, opened 2026-07-07
POLIO-2174: limit profiles api

beygorghor:
Vulnerability Description

It allows a low-privilege account to view sensitive user details that should only be accessible to administrators.
For example, the poliogo-test.afro.who.int/api/ or poliogo-test.afro.who.int/api/profile/[userID] endpoints are accessible to all user roles. Attackers can reach this endpoint just by fuzzing the path.

### Related JIRA tickets

POLIO-2174

## Changes

- limit  retrieve, list and export  profiles if not admin or having correct permission 
- let only pass /me and dropdown search

## How to test
Create a user without user permission, make sure profile dropdown are working and that you cannot acces profiles api

## Print screen / video

Screenshot 2026-07-07 at 17 08 26
Screenshot 2026-07-07 at 17 06 00
Screenshot 2026-07-07 at 17 05 51
Screenshot 2026-07-07 at 17 05 43

### DIFF
--- a/iaso/api/profiles/permissions.py
+++ b/iaso/api/profiles/permissions.py
@@ -20,7 +20,7 @@ class HasProfilePermission(permissions.BasePermission):
         if request.user.has_perm(CORE_USERS_MANAGED_PERMISSION.full_name()):
             return self.has_permission_over_user(request, pk, view.action)
 
-        return view.action in ["retrieve", "list", "export_csv", "export_xlsx", "dropdown"]
+        return view.action == "dropdown"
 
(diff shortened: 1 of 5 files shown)


madewulf (approved):

beygorghor:
@madewulf i double check in mobile code , only profiles/me is used and this behavior hasn't change.

bmonjoie:
@beygorghor, I confirm the mobile only calls `/api/profiles/me/` which was untouched in this PR, so there should not be any problem with the mobile application.

PR #3180, opened 2026-07-08
[IA-5305]: upgrade orval

hugo093:

Upgrade orval to latest version

### Related JIRA tickets

IA-5305

## Changes

* Upgrade dependency
* Redo patch
* Remove the { query: enabled: ... } as it's handled by orval 
* Regenerate orval files
* adapt tests

## How to test

CI should be enough


beygorghor (approved):
We can probably override the template to avoid version changes on each file no ? 
Also there is this `Unexpected use of continue statement` a bit everywhere, not sure there is another way.
For the rest, it's running on my machine !

PR #3191, opened 2026-07-13
WC2-943: Display Storage IDs as HEX

bmonjoie:

Storage IDs are displayed as HEX on some cards; we should therefore be able to match what is displayed on the card.

### Related JIRA tickets

WC2-943

## Changes

Add the HEX value in front of the base64 value

## How to test

Have some Storage device with a BASE64 customer chosen ID and go in the External Storage screen.

Make sure to try to search for the HEX string as well.

## Print screen / video

image

## Notes

Example values are given in the ticket.

### DIFF
--- a/iaso/api/storage.py
+++ b/iaso/api/storage.py
@@ -1,10 +1,12 @@
 # TODO: need better type annotations in this file
+import base64
+import binascii
 import datetime
 
 from typing import Any, List, Tuple, Union
 
 from django.core.paginator import Paginator
-from django.db.models import Q, QuerySet
+from django.db.models import Case, Q, QuerySet, When
 from django.http import HttpResponse, StreamingHttpResponse
 from drf_spectacular.utils import extend_schema
 from rest_framework import exceptions, permissions, serializers, status, viewsets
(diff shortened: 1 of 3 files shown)


iaso/api/storage.py
```diff
@@ -133,6 +136,14 @@ class Meta:
             "entity",
         )
 
+    @staticmethod
+    def get_storage_id(obj: StorageDevice):
+        base64_id = obj.customer_chosen_id
+        try:
+            return f"{base64.b64decode(base64_id).hex(' ').upper()} ({base64_id})"
+        except (TypeError, binascii.Error):
+            return base64_id
```
hugo093, line 145: why don't you reuse the annotation in your filter (and pass it to the queryset in the viewset) ?  

You could even put that annotation as a queryset method. 

That way you could also easily test the fact that hex_id should return None in case of error (aka not a base64)

hugo093 (changes requested):

iaso/api/storage.py
bmonjoie: Because the annotation is not formatted the way we want (a space every 2 characters).

hugo093: then lgtm

hugo093 (approved):

PR #3193, opened 2026-07-13
[IA-5315] fix SSO tests with `--keepdb`

tdethier:

fix SSO tests with `--keepdb`

### Related JIRA tickets

IA-5315

## Changes

- updated the test class to store ID from the `Account` object that was created by `setUpTestData`

## How to test

- run SSO tests (`hat.tests.test_sso_auth`) multiple times with the `--keepdb` flag: they should not fail

## Print screen / video

/

## Notes

/

## Doc

/

### DIFF
--- a/hat/tests/test_sso_auth.py
+++ b/hat/tests/test_sso_auth.py
@@ -1,4 +1,5 @@
 import base64
+import copy
 import json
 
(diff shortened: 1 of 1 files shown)


Phil-V (approved):
Looking good! 👍

PR #3194, opened 2026-07-14
docs: guides to read and modify the org units of a data source via the API

madewulf:

External organisations that need to read and write the org units of an IASO data source from their own systems have no documentation to work from, and have to reverse-engineer the API.

### Related JIRA tickets

[CHAICMR-233](https://bluesquare.atlassian.net/browse/CHAICMR-233)

## Changes

Adds two developer how-tos, each in English and French, aimed at an external organisation integrating with IASO in Python. They cross-link: FHIR for reads, the org unit API for writes.

- `docs/pages/dev/how_to/modify_org_units_via_api/` — **writing** org units.
- `docs/pages/dev/how_to/read_org_units_via_fhir/` — **reading** them through the FHIR API.
- `mkdocs.yml` — one nav entry per page per language tree (en, fr, es), next to the existing "Use the APIs" guide. The es nav points at the same pages and falls back to English, which is what `use_iaso_apis` already does.

### Guide 1 — modifying org units

Organised around the two write paths and when to use each:

- **Direct modification** — `POST /api/orgunits/create_org_unit/` and `PATCH /api/orgunits/<id>/`, plus the async `orgunitsbulkupdate` task. Requires `iaso_org_units`.
- **Change requests** — `POST /api/orgunits/changes/`. Requires no permission beyond being authenticated and seeing the org unit; a reviewer holding `iaso_org_unit_change_request_review` decides.

It makes explicit something that is easy to get wrong: **a change request cannot create an org unit on its own.** `org_unit_id` is mandatory, so proposing a new org unit means first creating it directly with `validation_status: "NEW"` (which needs `iaso_org_units`) and then submitting a change request against it, which IASO records as kind `org_unit_creation`. A caller with no permissions at all can only propose changes to existing org units.

It also documents the sharp edges an integrator will otherwise hit at runtime: `create_org_unit` parses dates only as `%d-%m-%Y` (an ISO date gives a 500, not a clean 400) while `PATCH` accepts four formats and change requests want ISO; `GET /api/orgunits/` defaults to `validation_status=VALID`, so units you just created as `NEW` are invisible; the list envelope key flips between `orgunits` and `orgUnits` depending on whether you paginate; latitude/longitude/altitude are read as a trio; `groups` replaces rather than appends.

Closes with a complete, copy-pasteable CSV → IASO sync script that runs in either mode.

### Guide 2 — reading org units through FHIR

Documents the read-only FHIR R4 `Location` API: the OrgUnit → Location mapping field by field, the search parameters, `Bundle` pagination, the `children` operation, the `CapabilityStatement`, and `OperationOutcome` errors. States plainly what is *not* there — one resource type only, no writes, no `_include`/`_sort`/`_lastUpdated`, no delta sync.

Two findings worth a reviewer's attention:

- **`iaso/api/fhir/README.md` is wrong on three counts**, and the guide contradicts it deliberately. The base path is `/api/fhir/`, not `/fhir/` (the router is registered in `iaso/urls.py`, which `hat/urls.py:136` mounts under `/api/`). The identifier systems are emitted as `https://`, not `http://`. And the status mapping is `NEW → inactive`, `REJECTED → suspended` — the README states those two backwards, which would make an integrator misread every unvalidated org unit. `constants.py` and `test_validation_status_mapping` both agree with the guide.
- **List pagination is not stably ordered.** `OrgUnit` declares no default ordering and the FHIR viewset adds none, yet results are paged with limit/offset — so paging a large result set can repeat one org unit and skip another. There is no `_sort` to fix it. The guide documents this and the examples deduplicate by `id`.

Neither is fixed here; see Notes.

## How to test

No runtime code is touched — this is documentation only. To review the rendering:

```
mkdocs serve
```

then open *Developers → How to*, check both new pages, and confirm the language switcher shows the French versions.

Note that `docs/venv` in the repo is currently broken on Apple Silicon (its Python symlink points at `/usr/local/opt/python@3.13`, an Intel-homebrew path), so it needs a fresh env. I validated that `mkdocs.yml` parses, that all nav targets resolve to files on disk, and that the two guides' cross-links point at real pages — but I did not run a full `mkdocs build`.

The API behaviours documented here were each read off the current implementation (`iaso/api/org_units.py`, `iaso/api/org_unit_change_requests/`, `iaso/api/fhir/`) and its tests rather than assumed. Reviewers who know these endpoints well are the right people to sanity-check the quirks lists.

## Print screen / video

N/A — text documentation.

## Notes

- **The quirks are documented, not fixed.** Several look like genuine bugs worth their own tickets rather than warnings in prose: the `create_org_unit` 500 on an ISO date, the `orgunits`/`orgUnits` envelope switch, the unordered FHIR pagination, the `fullUrl` double slash (`.../Location//123`), and `Accept: application/fhir+json` being rejected with a 406 by a FHIR API. Happy to split these out if you agree they are real.
- **`iaso/api/fhir/README.md` is left untouched** even though this PR shows it to be wrong in three places. Correcting or deleting it felt out of scope for a docs PR, but it should not survive much longer in its current state — it actively misleads.
- The write guide deliberately stops at the submitter's side of change requests. The review API (`PATCH /api/orgunits/changes/<id>/`, `bulk_review`) is not documented, since an external organisation does not hold the review permission.
- No Spanish translations; the es nav entries fall back to English, consistent with the neighbouring guides.

## Doc

This PR *is* the doc: `docs/pages/dev/how_to/modify_org_units_via_api/` and `docs/pages/dev/how_to/read_org_units_via_fhir/`, published under Developers → How to.

[CHAICMR-233]: https://bluesquare.atlassian.net/browse/CHAICMR-233?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/pages/dev/how_to/modify_org_units_via_api/modify_org_units_via_api.en.md
+++ b/docs/pages/dev/how_to/modify_org_units_via_api/modify_org_units_via_api.en.md
@@ -1,0 +1,694 @@
+# Modify the org units of a data source via the API
+
+This guide is for an external organisation that needs to create and update org units in an IASO
+data source from its own systems, using Python. It covers the two write paths IASO offers —
+**direct modification** and **change requests** — how to choose between them, and the constraints
+and quirks you will run into.
+
+Everything below is plain JSON over HTTPS. The examples use
+[`requests`](https://requests.readthedocs.io/); nothing else is required.
+
+## The two write paths
+
+| | Direct modification | Change request |
+|---|---|---|
+| Endpoints | `POST /api/orgunits/create_org_unit/`, `PATCH /api/orgunits/<id>/` | `POST /api/orgunits/changes/` |
+| Effect | Applied immediately to the pyramid | Recorded as a proposal, applied only once a reviewer approves it |
+| Permission needed | `iaso_org_units` | None beyond being authenticated and able to see the org unit |
+| Reviewed by a human | No | Yes — by a user holding `iaso_org_unit_change_request_review` |
+| Audit trail | Modification log | Full before/after snapshot, per-field approval, rejection comment |
+| Can create an org unit | Yes | Not on its own — see below |
+| Bulk | Async bulk-update task available | None; one request per org unit |
+
+Use **change requests** when the IASO team wants to keep editorial control over the pyramid: your
+system proposes, they decide. Use **direct modification** when your organisation is the authority on
+this data and is trusted to write to it.
+
+!!! warning "A change request cannot create an org unit by itself"
+    Every change request points at an org unit that **already exists** — the `org_unit_id` field is
+    mandatory in practice. To get a change request of kind `org_unit_creation`, you first create the
+    org unit directly with `validation_status: "NEW"`, then submit a change request against it. IASO
+    detects that the target is still `NEW` and treats the request as a creation; approving it flips
+    the org unit to `VALID`, rejecting it flips it to `REJECTED`.
+
+    That first step is a direct write, so it needs the `iaso_org_units` permission. **A user with no
+    permission at all can only propose changes to existing org units, never add new ones.**
+
+## Prerequisites
+
+Ask the IASO administrator of the account for:
+
+- a dedicated **user account** (do not share a human's credentials — the audit trail records the author of every change);
+- the **permissions** matching the path you intend to use (see the table above);
+- the **id of the data source and of the source version** you are allowed to write into;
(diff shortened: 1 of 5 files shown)


beygorghor (changes requested):
@madewulf Can we force users to use [iaso-client](https://github.com/BLSQ/iaso-client) instead please, it has been done on purpose.

PR #3198, opened 2026-07-15
SLEEP-1817: soft deleted instances in entity filters

Phil-V:
refs: SLEEP-1817

Soft-deleted Instances are taken into account when performing various searches on Entities (filter by date, jsonlogic filters, etc.)

### Related JIRA tickets

SLEEP-1817 SLEEP-1793

## Changes

Exclude soft-deleted Instances from Entity filters and jsonlogic filters.  

## How to test

Go to the entity list `/dashboard/entities/list/` and verify that the search and search in submitted fields behaves correctly and doesn't return results for soft-deleted instances.

## Notes

/

### DIFF
--- a/iaso/api/entities/views.py
+++ b/iaso/api/entities/views.py
@@ -197,7 +197,9 @@ class EntityViewSet(ModelViewSet):
             queryset = queryset.prefetch_related(
                 Prefetch(
                     "instances",
-                    queryset=Instance.objects.only("id", "entity_id", "source_created_at", "created_at"),
+                    queryset=Instance.objects.filter(deleted=False).only(
+                        "id", "entity_id", "source_created_at", "created_at"
+                    ),
                 ),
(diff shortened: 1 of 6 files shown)


iaso/api/entities/views.py
beygorghor, line 1: @bmonjoie is this a problem for the mobile app ? i guess no, but just to be sure

beygorghor (approved):

bmonjoie (changes requested):
@Phil-V, I don't foresee any issues with those changes on mobile, but I think you should make the modification and add some tests for the mobile endpoints as well.

Those are: 
- /api/mobile/entities/
- /api/mobile/entitytypes/{id}/entities/ (the important part of this one is the `json_content` parameter)

bmonjoie (approved):
LGTM

PR #3202, opened 2026-07-16
[IA-5308]: missions mobile API

hugo093:

Mission mobile API implementation

### Related JIRA tickets

IA-5308

## Changes

Refactored and adapted the mobile viewsets related to missions:
* No changes on the payload of v1
* Two changes on payload of v2:
  *  removed the id in mission_forms
  * Pagination
* Adapted serializers + query optimization
* tests (split + reorganize)
* Quick fixes on planning api

## How to test

Python tests


iaso/api/instances/mission_resolution.py
```diff
@@ -0,0 +1,31 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def resolve_planning(planning_id, project: Project) -> Optional[Planning]:
+    if planning_id is None:
+        return None
+
+    # Planning.objects does not filter out soft-deleted plannings.
+    planning = Planning.objects.filter(pk=planning_id, project__account=project.account).first()
```
bmonjoie, line 16: I find this dangerous. Why is `Planning.objects` not filtering out `soft-deleted` objects? It should!

If one day we change the behavior of `Planning.objects` to match what other soft-deletable models are doing, this will break.

iaso/api/instances/mission_resolution.py
```diff
@@ -0,0 +1,31 @@
...
+        logger.error("Instance references planning %s, unknown for account %s", planning_id, project.account_id)
+
+    return planning
+
+
+def resolve_mission(mission_id, project: Project) -> Optional[Mission]:
+    if mission_id is None:
+        return None
+
+    mission = Mission._base_manager.filter(pk=mission_id, account=project.account).first()
```
bmonjoie, line 27: My understanding is that we should not be using `_base_manager` and especially not in this case.

The convention in this project is to use 3 `objects` for soft-deletable models like this:

```python
objects = DefaultSoftDeletableManager.from_queryset(MissionQuerySet)()

objects_only_deleted = OnlyDeletedSoftDeletableManager.from_queryset(MissionQuerySet)()

objects_include_deleted = IncludeDeletedSoftDeletableManager.from_queryset(MissionQuerySet)()
```

Then, you can use `Mission.objects_include_deleted` which will be much clearer.

iaso/api/microplanning/serializers/mobile/nested_mission.py
```diff
@@ -0,0 +1,172 @@
...
+        fields = [
+            "form",
+            "min_cardinality",
+            "max_cardinality",
+        ]
+
+        extra_kwargs = {
+            "min_cardinality": {"read_only": True},
+            "max_cardinality": {"read_only": True},
+        }
```
bmonjoie, line 41: Why use `extra_kwargs` here instead of `read_only_fields`?

iaso/api/microplanning/serializers/mobile/nested_mission.py
```diff
@@ -0,0 +1,172 @@
...
```
bmonjoie, line 58: Why use extra_kwargs here instead of read_only_fields?

iaso/api/microplanning/serializers/mobile/nested_mission.py
```diff
@@ -0,0 +1,172 @@
...
+            "form",
+            "min_cardinality",
+            "max_cardinality",
+        ]
+
+        extra_kwargs = {
+            "form": {"read_only": True},
+            "min_cardinality": {"read_only": True},
+            "max_cardinality": {"read_only": True},
+        }
```
bmonjoie, line 76: Why use extra_kwargs here instead of read_only_fields?

iaso/api/microplanning/serializers/mobile/v2.py
```diff
@@ -0,0 +1,78 @@
...
+                #     if (
+                #         m_out == a.org_unit.org_unit_type.id
+                #         or m_out in a.org_unit.org_unit_type.sub_unit_types.values_list("id", flat=True)
+                #     ):
+                #         missions.append(NestedMissionSerializer(m).data)
+                # elif m.mission_type == MissionType.ENTITY_AND_FORM:
+                #     # We always assign entities as there are no enforcement on entities and OrgUnit types.
+                #     missions.append(NestedMissionSerializer(m).data)
+                # else:
+                #     raise NotImplementedError("Unknown mission type")
```
bmonjoie, line 74: Why leave this here?

iaso/api/microplanning/serializers/mobile/v2.py
```diff
@@ -0,0 +1,78 @@
...
+
+    @extend_schema_field(AssignmentSerializer(many=True))
+    def get_assignments(self, planning: Planning):
+        assignments = []
+        for a in planning.assignment_set.all():
+            missions = []
+            for m in planning.missions.all():
+                form_assignments = m.get_form_assignments(a.org_unit)
+
+                if form_assignments:
```
bmonjoie, line 46: I'm not a fan of mixing `get_form_assignments` with the logic of "is this assignable?".

I was very confused in the beginning as to why `get_form_assigbnments` returns a `boolean` and I saw that it only does so in the OrgUnit and Entity missions.

I understand the logic is to filter the list only once, but I would argue that, having two methods:
- `is_assignable(a.org_unit)`
- `get_form_assignments(a.org_unit)`

(with the first one keeping a cache for the second one if needed)

would make more sense.

iaso/api/microplanning/views_mobile.py
```diff
@@ -45,8 +53,8 @@ def get_queryset(self):
                 # We have to filter on FORM_FILLING only because this was the only type of missions before
                 Prefetch(
                     lookup="missions",
-                    queryset=Mission.objects.filter(mission_type=MissionType.FORM_FILLING).prefetch_related(
-                        "mission_forms"
+                    queryset=Mission.objects.instance_of(MissionForm).prefetch_polymorphic_related(
+                        MissionForm, "missionformthroughform_set"
```
bmonjoie, line 57: It seems this is added quite a few time.

Maybe you should add 4 methods to the default `objects`:
- `prefetch_polymorphic_form_mission()`
- `prefetch_polymorphic_org_unit_mission()`
- `prefetch_polymorphic_entity_mission()`
- `prefetch_polymorphic_all()`

iaso/models/microplanning/plannings.py
```diff
@@ -98,3 +102,6 @@ def has_pipeline_uuid(self, pipeline_uuid: str) -> bool:
     def get_pipeline_uuids(self) -> list:
         """Get the list of pipeline UUIDs for this planning."""
         return self.pipeline_uuids or []
+
+    def get_possible_org_units(self):
```
bmonjoie, line 106: Is this method ever used?

bmonjoie (changes requested):

Phil-V (approved):

PR #3210, opened 2026-07-19
IA-5330: declutter the submission detail page

madewulf:
Jira: [IA-5330](https://bluesquare.atlassian.net/browse/IA-5330)

Implements the **Page decluttering strategy** design for the submission (instance) detail page. Adapted to MUI + the existing theme rather than ported as raw CSS, so it inherits the app's palette and stays consistent with surrounding pages.

The main goals were:
- making the submission itself less cluttered by allowing to toggle on/off the technical ids
- allow to see more data on small screens with an optional 2 columns layout
- clean up the design in general, hiding more technical identifiers and cleaning up the typography
- Make the panel with additional data (locks, validation, exports, location) less cluttered by making these options foldable (this also allow to see immediately that these section exists, without scrolling)

## Layout

- The **submission is now the primary column on the left** (wide), with the metadata as a sidebar rail on the right.
- Top bar title reads **"Submission: &lt;form&gt; · 📍 &lt;org unit&gt;"** (the shared `TopBar` takes a plain string, so a pin emoji is used rather than an icon/pill).

## Submission panel

- Questions grouped into **sections with headers** and field counts, replacing the flat table.
- **Search** over question labels *and* ids, with match highlighting, a result count and an empty state; searching reveals a question's id when the id is what matched.
- **Show question ids** toggle and a **one / two column** density switch (blue segmented control).
- Values rendered **by question type**: formatted date pills, numbers, tinted choice / multi-choice chips, photos and files (left-aligned, capped, flowing within the two-column grid), and geopoints on a real `MarkerMap` with a **"show / hide exact values"** disclosure (lat / lng / alt / accuracy). ODK `start`/`end` metadata timestamps are recognised as dates; the `${…}` interpolation placeholder is stripped from metadata labels.
- The controls toolbar is a **muted "chrome" strip**, visually distinct from the answers.

## Left rail

- **General card**: promotes the fields people read (form, period, status, last activity), folds identifiers/provenance behind a **Technical details** disclosure, uses the design's left-aligned rows, a tight type scale (20 / 14 / 13 / 12) and an **info icon** heading. Reference + status pills grouped right.
- **Entity card** (when present): same card style, titled with the **entity type name**, with a link through to the entity.
- **Location**: the design's right-aligned key/value list and a landscape `MarkerMap`.
- **Files, validation, change requests, export requests, locks**: collapsible rows (outlined icons), each showing a short state summary; all collapsed by default.

## Scope and blast radius

- `InstanceFileContent` is **untouched**, so its other callers (registry, org unit reference submissions) keep the current rendering; the new panel is used only by the detail page.
- Export requests, locks history and validation expose their body separately from their `WidgetPaper` wrapper so both layouts share them; the default exports render exactly as before. The compare view keeps the legacy `InstanceDetailsField` rows (via a `FieldComponent` prop on `InstanceDetailsLocation`).
- Falls back to `InstanceFileContentBasic` when a submission has no form descriptor.
- Drive-by fixes: the duplicate `id="export-requests"` on locks history, and `InstanceDetailsField.valueTitle` typed optional to match its default.
- **Follow-up (not in scope here):** the entity popup and compare view still use the old card layout and can be aligned separately.

## Verification

Checked in a browser against several real submissions (including a "Test all fields" form with images, geopoints and metadata). New unit tests cover date formatting and section flattening (groups, nested groups, repeats, `meta` skipping, choice resolution, metadata-label cleanup, search filtering). `domains/instances` suite: **26 files / 175 tests pass**. ESLint clean; `tsc` clean on the changed/new files.

[IA-5330]: https://bluesquare.atlassian.net/browse/IA-5330?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Screenshots
Video Tour: 

https://github.com/user-attachments/assets/59b53d5e-c883-4375-8c16-dcec85849ab0

### Screenshots
CleanShot 2026-07-21 at 13 14 57@2x
CleanShot 2026-07-21 at 13 16 07@2x
CleanShot 2026-07-21 at 13 16 42@2x
Screenshot with export request:
CleanShot 2026-07-22 at 11 48 33@2x

Language switch added in the latest commits:
CleanShot 2026-07-22 at 14 21 57@2x


hat/assets/js/apps/Iaso/domains/instances/components/SubmissionContent/SubmissionValue.tsx
```diff
@@ -0,0 +1,353 @@
...
+                <Box
+                    sx={{
+                        display: 'flex',
+                        flexWrap: 'wrap',
+                        gap: 0.75,
+                        justifyContent: 'inherit',
+                    }}
+                >
+                    {value
+                        .split(/,\s*/)
```
mestachs: Not sure this works everytime. I had forms with choices with blanks but was perhaps already there. might end up as "not found".

hat/assets/js/apps/Iaso/domains/instances/components/SubmissionContent/useSubmissionSections.test.ts
```diff
@@ -0,0 +1,183 @@
+import { Descriptor } from '../InstanceFileContentRich';
+import {
+    buildSubmissionSections,
+    filterSubmissionSections,
+    getFieldKind,
+} from './useSubmissionSections';
+
+const descriptor: Descriptor = {
```
mestachs, line 8: I would add a test with translated forms (we had swahili and english/fr in cameroon)

hat/assets/js/apps/Iaso/domains/instances/components/InstanceDetailsExportRequests.tsx
```diff
@@ -17,23 +17,16 @@ type Props = {
     currentInstance: any;
 };
 
-const InstanceDetailsExportRequests: FunctionComponent<Props> = ({
+/**
+ * Body of the export requests panel, without any surrounding paper, so it can
+ * be dropped either in a WidgetPaper or in an accordion row of the detail rail.
+ */
+export const InstanceDetailsExportRequestsContent: FunctionComponent<Props> = ({
```
mestachs, line 25: the screenshot never shows it, IMO worth trying a "seed command" and exporting one and checking the rendered output

mestachs:
Can you show how it looks like when the question is a file (xlsx, docx, pdf like "pieces justificatives")

Apparently it was in the video

image

hat/assets/js/apps/Iaso/domains/instances/components/InstanceDetailsExportRequests.tsx
mestachs: I think we lost the link to  `/forms/mappings/formId/${currentInstance.form_id}/order/.../pageSize/20/page/1`

when there were no mapping/export this really help the "call to action" to create/debug why it wasn't exported (no mapping at all, not for that version)

madewulf: image
Done

hat/assets/js/apps/Iaso/domains/instances/components/SubmissionContent/useSubmissionSections.test.ts
madewulf: Added the feature allowing you to switch among the languages available

beygorghor (approved):

butofleury (approved):

PR #3213, opened 2026-07-23
SNT-545 Create re-usable ChatPanel

Bewi:

Create a re-usable ChatPanel

### Related JIRA tickets

SNT-545

## Changes

This is a copy of the ChatPanel from FormAI but with all custom messages and styling stripped down so it is reusable.

- Create a ChatPanel independent from any business logic
- Add markdown support for answers of ChatAI

## How to test

Nothing to test here, it is only used in SNT_Malaria at the moment.
In the future, FormAI should use it too.

## Print screen / video

N/A

## Notes

SNT Malaria PR: https://github.com/BLSQ/snt-malaria/pull/338

## Doc

N/A


michiwend (approved):
LGTM

PR #3214, opened 2026-07-23
POLIO-1959: use md5 for polio files

quang-le:

Vaccine stock file upload was using a not so reliable method to check that an uploaded file had been modified

### Related JIRA tickets

POLIO-1959

## Changes

- use md5 hash to compare files for changes:
    - Update `ModelWithFile` abstract class to add `md5` field (cf existing pattern in `FormVersion`etc)
    - Update serializers
    - Add 2 step migration:
        - first add field
        - then populate md5 field with non atomic transaction
- Remove some dead code
- Update local clamav doc with command for mac silicon

## How to test

On a polio DB, with clamav active (see modified md file for procedure)

- Go to polio> notifications:
    - click submit and use template to submit a file: check terminal output for file scan
- Go to polio > Vaccine module:
   - Upload/update file for supply chain: VRF, Pre-Alert
   - Upload/update file for stock management: FormA, Incident Report, Destruction report
   All uploaded files should appeared as scanned

## Print screen / video
N/A no feature change

### DIFF
--- a/iaso/tests/utils/test_virus_scan_serializers.py
+++ b/iaso/tests/utils/test_virus_scan_serializers.py
@@ -25,36 +25,42 @@ class ModelWithFileSerializerScanTestCase(TestCase):
 
     def test_create_always_needs_scanning(self):
         file = SimpleUploadedFile("doc.pdf", b"content")
-        self.assertTrue(self.serializer._check_if_file_exists_and_needs_scanning({"file": file}, obj=None))
+        new_md5 = calculate_md5(file)
+        self.assertTrue(self.serializer._check_if_file_exists_and_needs_scanning({"file": file}, None, new_md5))
 
     def test_update_needs_scanning_when_md5_empty(self):
         file = SimpleUploadedFile("doc.pdf", b"content")
+        new_md5 = calculate_md5(file)
         obj = MagicMock(spec=DummyModel)
         obj.file = SimpleUploadedFile("doc.pdf", b"content")
         obj.md5 = ""
-        self.assertTrue(self.serializer._check_if_file_exists_and_needs_scanning({"file": file}, obj=obj))
+        self.assertTrue(self.serializer._check_if_file_exists_and_needs_scanning({"file": file}, obj, new_md5))
 
     def test_update_skips_scanning_when_md5_matches(self):
         content = b"same content"
         file = SimpleUploadedFile("doc.pdf", content)
+        new_md5 = calculate_md5(file)
         obj = MagicMock(spec=DummyModel)
         obj.file = SimpleUploadedFile("old.pdf", content)
-        obj.md5 = calculate_md5(SimpleUploadedFile("x.pdf", content))
-        self.assertFalse(self.serializer._check_if_file_exists_and_needs_scanning({"file": file}, obj=obj))
+        obj.md5 = new_md5
+        self.assertFalse(self.serializer._check_if_file_exists_and_needs_scanning({"file": file}, obj, new_md5))
 
     def test_update_needs_scanning_when_md5_differs(self):
-        file = SimpleUploadedFile("doc.pdf", b"new content")
+        new_file = SimpleUploadedFile("doc.pdf", b"new content")
+        new_md5 = calculate_md5(new_file)
         obj = MagicMock(spec=DummyModel)
-        obj.file = SimpleUploadedFile("doc.pdf", b"old content")
-        obj.md5 = calculate_md5(SimpleUploadedFile("x.pdf", b"old content"))
-        self.assertTrue(self.serializer._check_if_file_exists_and_needs_scanning({"file": file}, obj=obj))
+        old_file = SimpleUploadedFile("x.pdf", b"old content")
+        obj.file = old_file
+        obj.md5 = calculate_md5(old_file)
+        self.assertTrue(self.serializer._check_if_file_exists_and_needs_scanning({"file": new_file}, obj, new_md5))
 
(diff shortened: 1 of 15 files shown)


iaso/utils/virus_scan/serializers.py
```diff
@@ -117,7 +122,7 @@ def _check_if_file_exists_and_needs_scanning(self, validated_data, obj):
         if not old_file:
             return True  # no previous file, so we need to scan the new one
 
-        # Compare the new file with the existing one
-        needs_scanning = os.path.basename(old_file.name) != os.path.basename(file.name) or old_file.size != file.size
-        # we should compare file hashes instead - see POLIO-1959
-        return needs_scanning
+        if new_md5 is None:
```
tdethier: why do you set `new_md5` default value to `None` in params? if it's passed, it will never be  `None` and it's always passed 😅

iaso/tests/utils/test_virus_scan_serializers.py
```diff
@@ -0,0 +1,51 @@
...
+from django.test import SimpleTestCase
+
+from iaso.utils.encryption import calculate_md5
+from iaso.utils.virus_scan.model import ModelWithFile, VirusScanStatus
+from plugins.polio.api.vaccines.stock_management.destructions.serializers import DestructionReportSerializer
+
+
+class ModelWithFileSerializerScanTestCase(SimpleTestCase):
+    def setUp(self):
+        self.serializer = DestructionReportSerializer()
```
tdethier: I would not add a polio serializer in a vanilla iaso test. I would create new dummy model and serializer inside this test file just for this test

iaso/tests/utils/test_virus_scan_serializers.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def test_update_needs_scanning_when_md5_differs(self):
+        file = SimpleUploadedFile("doc.pdf", b"new content")
+        obj = MagicMock(spec=ModelWithFile)
+        obj.file = SimpleUploadedFile("doc.pdf", b"old content")
+        obj.md5 = calculate_md5(SimpleUploadedFile("x.pdf", b"old content"))
+        self.assertTrue(self.serializer._check_if_file_exists_and_needs_scanning({"file": file}, obj=obj))
+
+    @patch("iaso.utils.virus_scan.serializers.scan_uploaded_file_for_virus")
+    def test_scan_file_if_exists_sets_md5(self, mock_scan):
```
tdethier, line 66: I would test `scan_file_if_exists()` as well for all other scenarios:
- creating an object without a file
- updating an object with a file, passing a new file
- updating an object with a file, passing nothing (so this means removing the attached file on the object)
  - this should probably set all fields related to virus scanning back to their default values
- updating an object without a file, passing a file (so this means setting a file on the object)
- updating an object without a file, passing nothing (so updating all the other fields)

plugins/polio/migrations/0256_backfill_modelwithfile_md5.py
```diff
@@ -0,0 +1,82 @@
...
+        .exclude(file__isnull=True)
+        .only("pk", "file", "md5")
+        .iterator(chunk_size=CHUNK_SIZE)
+    )
+
+    for instance in queryset:
+        try:
+            md5 = calculate_md5(instance.file)
+        except (FileNotFoundError, OSError):
+            # Missing/unreadable file: skip this row, don't abort the whole run.
```
tdethier: I guess we could print some info about skipping here as well

plugins/polio/api/vaccines/supply_chain.py
```diff
@@ -290,24 +290,7 @@ def validate(self, attrs: Any) -> Any:
...
-                if not old_file:
-                    is_different = True
-                    current_obj.file = new_file
-                    continue
-
-                if os.path.basename(old_file.name) != os.path.basename(new_file.name) or old_file.size != new_file.size:
-                    is_different = True
-                    current_obj.file = new_file
-            elif hasattr(current_obj, key) and getattr(current_obj, key) != attrs[key]:
+            if hasattr(current_obj, key) and getattr(current_obj, key) != attrs[key]:
```
tdethier, line 293: I'm not sure about this change, we should have tests that thoroughly cover all scenarios to make sure it's alright

plugins/polio/models/base.py
```diff
@@ -2034,7 +2034,7 @@ def create_notifications(self, created_by: User) -> None:
 
 
 @task_decorator(task_name="create_polio_notifications_async")
-def create_polio_notifications_async(pk: int, task: Task = None) -> None:
+def create_polio_notifications_async(pk: int, user: User = None, task: Task = None) -> None:
```
tdethier, line 2036: was this done on purpose? 😅 it's going to be overridden by line `2039` anyway

plugins/polio/api/vaccines/supply_chain.py
tdethier: actually why is this in arrival report, since there is no file on that model?

tdethier (changes requested):
Overall it works fine, the same file with another name is not scanned again 👍 

While testing, I noticed that:
- on pre alerts, if you update a pre alert file with another one that has the same md5, the new file is not saved, it's ignored
- on VRF, when you remove a file, the other virus scan related fields should also be reset - here's the DB after removing the file:
Image

I'm requesting changes to add a few more tests here and there and to remove the new polio mention inside vanilla iaso

beygorghor (approved):

hugo093 (approved):

PR #3217, opened 2026-07-24
[IA-5332] change requests module

tdethier:

rename "data validation" module into "change requests" module

### Related JIRA tickets

IA-5332

## Changes

- updated FR/EN/ES translations of the data validation module
- updated variable name of the data validation module (not its `codename`)
- moved `dynamic_fields` translations into their own app
- fixed the `-f` flag of the `generate_openapi_schema` command
- refreshed `orval` files that require modules

## How to test

- go to admin/modules: you should see the "change requests" module in the list, "data validation" is not
- same thing in the permissions list (when creating a new data role or user)
- same thing in the module list in the account page
- use the  `generate_openapi_schema` command with the `-f` flag: there should be a new file located at the provided value

## Print screen / video

account page in FR:

image

modules page in EN:
image

## Notes

- i wanted to edit the module `codename`, but since it's in some kind of enum (`Account.modules`), updating the value would imply making a new migration to update all existing accounts... so i simply updated translations

## Doc

/


bmonjoie (approved):
LGTM

PR #3220, opened 2026-07-27
IA-5339: make the MappingVersion admin interface load fast

madewulf:
## Problem

The `MappingVersion` admin is unusable in prod:

- **Detail / add page**: `form_version` (and `mapping`) are rendered as plain `<select>` widgets, so *every* `FormVersion` row is fetched and rendered as an `<option>`.
- **Changelist**: `list_filter = ("form_version_id",)` builds one filter entry per `FormVersion`, and rendering each entry's label (`FormVersion.__str__`) hits `form.name`, causing a query per form version.

## Fix

Replace the exhaustive dropdown/filter with autocomplete widgets + a text search, and `select_related` the FKs displayed in the list.

- `autocomplete_fields = ["form_version", "mapping"]`
- `search_fields` on form name / form_id / version_id, so you can still narrow down by form version
- `list_display` + `list_select_related` to avoid the N+1 when rendering rows
- `ordering = ("-id",)` for stable pagination

`MappingAdmin` had exactly the same pathology on `list_filter = ("form_id",)`, and needed `search_fields` anyway for the `mapping` autocomplete above, so it gets the same treatment.

## Measurements

Rendered through the Django test client on a local DB holding ~1180 form versions:

| page | before | after |
|---|---|---|
| `/admin/iaso/mappingversion/` | 1187 queries | **6 queries** |
| `/admin/iaso/mappingversion/add/` | 1185 queries, 1180 `<option>` | **5 queries, 0 `<option>`** |
| `/admin/iaso/mapping/` | 8 queries | **5 queries** |

Prod holds far more form versions than this local DB, so the real gain is bigger.

## Note for reviewers

The `form version` **filter** on the changelist is gone (it is the thing that was slow). Searching by form name / form id / version id replaces it.

CleanShot 2026-07-27 at 17 52 07@2x
CleanShot 2026-07-27 at 17 52 34@2x

### DIFF
--- a/iaso/admin/base.py
+++ b/iaso/admin/base.py
@@ -475,14 +475,21 @@ class LinkAdmin(admin.GeoModelAdmin):
 @admin.register(Mapping)
 @admin_attr_decorator
 class MappingAdmin(admin.GeoModelAdmin):
-    list_filter = ("form_id",)
-    autocomplete_fields = ["data_source"]
+    list_filter = ("mapping_type",)
+    search_fields = ("name", "form__name", "form__form_id")
+    list_display = ("name", "form", "mapping_type", "data_source", "created_at")
+    list_select_related = ("form", "data_source")
+    autocomplete_fields = ["data_source", "form"]
 
 
 @admin.register(MappingVersion)
 @admin_attr_decorator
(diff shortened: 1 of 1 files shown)


tdethier (approved):
LGTM 👍

PR #3221, opened 2026-07-27
IA-5341: stop the admin from loading whole tables in filters and selects

madewulf:
Follow-up to #3220 (IA-5339), which fixed the `MappingVersion` admin. Same problem, everywhere else.

## Measurements

Every touched page (list, search **and** add) rendered through the Django test client, queries counted with a `connection.execute_wrapper` (the usual `CaptureQueriesContext` silently under-reports here: `queries_log` is a capped deque and the 22k-query page overflows it).

On my machine, I get: 

| page | queries | `<option>` |
|---|---|---|
| `instancelock/add/` | **22,259 → 3** | 25,749 → 1 |
| `tasklog/` | **1,424 → 5** | |
| `tasklog/add/` | **1,322 → 3** | 1,444 → 0 |
| `bulkcreateuserfile/add/` | **709 → 3** | 6,326 → 0 |
| `workflowversion/` | 213 → 13 | |
| `orgunitchangerequestconfiguration/add/` | 161 → 4 | 629 → 5 |
| `storagedevice/` | 158 → 6 | |
| `formversion/` | 105 → 5 | |
| `instance/?q=` | 105 → 12 | |
| **total across the 69 pages** | **28,799 → 427** | |

Prod tables are much larger than this local DB, so the real gain is bigger.

## Verification

Tested all the admin page: I don't think the reviewer needs to do the same.

`iaso.tests.test_auth_enforcement` has one failing assertion (`/accounts/login/` should require auth) — **pre-existing on develop**, verified by re-running it with these changes stashed.

## Trade-offs to review

Some exhaustive filters are replaced by search rather than reproduced, because the listing itself was the cost:

- **`Instance` changelist loses its `form` filter** — the busiest page in the admin. `form__name` / `form__form_id` were added to `search_fields` instead. This is the one I'd most like an opinion on; if filtering instances by form is a common workflow, we could keep it behind a cheaper widget.
- `StorageDevice` loses the entity filter (search on `entity__uuid` instead), `ExportRequest` loses the launcher filter, `Assignment` loses the planning filter, `StockLedgerItem` loses the rule filter, `TaskLog` loses the task filter.

### DIFF
--- a/hat/api_import/admin.py
+++ b/hat/api_import/admin.py
@@ -6,7 +6,8 @@ from .models import APIImport
 @admin.register(APIImport)
 class APIImportAdmin(admin.GeoModelAdmin):
     date_hierarchy = "created_at"
-    search_fields = ("json_body", "headers", "exception")
+    search_fields = ("app_id", "json_body", "headers", "exception")
+    autocomplete_fields = ("user",)
     list_display = (
         "id",
         "created_at",
@@ -18,7 +19,6 @@ class APIImportAdmin(admin.GeoModelAdmin):
     list_filter = (
         "has_problem",
         "import_type",
-        "app_id",
     )
 
     def get_queryset(self, request):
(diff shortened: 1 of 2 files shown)


tdethier (approved):
LGTM 👍

PR #3222, opened 2026-07-28
[IA-4872] Refactor `Mission` model

bmonjoie:

The current model duplicates the MissionFormThroughForm table and creates 7 tables, of which 3 are not needed.

### Related JIRA tickets

IA-4872

## Changes

Use `MissionForm` as a base, as they all extend from that. 
This allows removing duplicate through-tables for the forms and reduces the number of queries.

If we ever need to have missions that do not extend `MissionForm`, we can refactor to re-add a level or abstraction above it and keep MissionForm as a parent for the others.

## How to test

Check that the tests are still passing.

## Notes

I don't know why the number of queries during creations has increased.
But querying is better overall.


hugo093 (changes requested):
I understand and agree with the issue of the 3 extra tables that does the same things. However, I do not agree with parts in this PR and data modelling. 

----------------
Biggest concern I have is this sentence in the PR itself : 

> If we ever need to have missions that do not extend MissionForm, we can refactor to re-add a level or abstraction above it and keep MissionForm as a parent for the others.

Why not making it generic now before it's in prod and really used ? Now is the time to do it, not later where the migration will be really complex to handle. 

-----------------
> This allows removing duplicate through-tables for the forms and reduces the number of queries.
....
> I don't know why the number of queries during creations has increased.
But querying is better overall.

Isn't that a bit contradictory ? Also it's not entirely true : 
* in create, the queries increased by 1
* in retrieve it effectively went from 7 to 5
* in update : 
  * 16 to 18
  * 16 to 18
  * 15 to 14
* in delete : 9 to 8 
* in list, the queries went from 11 to 13
* in dropdown : 10 to 8

------------

Data modelling in itself : 

Like said before I agree with the fact that 3 tables are created (and repeated) but it would allow easy customization in the future for each mission type if we need to change it. 

If we really don't want this, I'd avoid making everything inherit from MissionForm as it says confusingly that other mission (org unit and entity types) are also mission forms .. defeats the polymorphism actually in the sense that you're using a concrete subtype as a base, where it should be as "abstract" and generic as possible. 

May I suggest this kind of structure : 

```
Mission
   │
   └── MissionWithForms
          ├── MissionForm
          ├── MissionOrgUnitType
          └── MissionEntityType
```

So we keep everything in MissionWithForms and "querying is better overall" (meaning the count is more simple actually) 

But keep in mind that by doing this : 
* Future update to MissionX in the M2M will be harder to do 
* It adds an extra table

What are your thoughts ?

bmonjoie:
As discussed, I'm not entirely sure the top-level `Mission` table is necessary at the moment, but I can live with that.
I think that the important part is to remove the through table duplication and make this part cleaner.

Regarding, the query count, I thought maybe you could chime in and tell me why they changed :sweat_smile:

hugo093 (approved):
approving my own code

iaso/api/microplanning/views_mobile.py
```diff
@@ -45,7 +45,7 @@ def get_queryset(self):
                 # We have to filter on FORM_FILLING only because this was the only type of missions before
                 Prefetch(
                     lookup="missions",
-                    queryset=MissionForm.objects.filter(mission_type=MissionType.FORM_FILLING).prefetch_related(
+                    queryset=MissionWithForms.objects.filter(mission_type=MissionType.FORM_FILLING).prefetch_related(
```
bmonjoie, line 48: Why not call `MissionForm.objects` directly here?

bmonjoie (commented):
I'm just wondering if we should not call `MissionWithForm.objects` everywhere  we use `Mission.objects` now. It's going to be easy to change later, and it will remove a layer of unneeded queries.

iaso/api/microplanning/views_mobile.py
hugo093: There's just one place where Mission.objects is called, in microplanning serialziers .. so I dunno tbh. Maybe we'll let @madewulf decide

For the views_mobile I'd not focus too much on it as it's heavily reworked in another PR.

hugo093:
As discussed with @bmonjoie and related to the other remarks I made in this PR, I "refactored the refactoring" to include this structure : 

```
Mission
   │
   └── MissionWithForms
          ├── MissionForm
          ├── MissionOrgUnitType
          └── MissionEntityType
```

Where the forms M2M lies within `MissionWithForms`. 

Pros : 

* It avoids the duplicate tables for the forms M2M
* It keeps the data modelling quite explicit, separating mission types (and imho it's quite clear to understand the business concepts from there)
* It allows an easy update later (adding a MissionWithX and/or MissionY will be easier)
* It stays quite generic for future update

Cons :

* If a mission type needs a particular behavior in the forms m2m realtionship, we'll need to get back on the original model (each mission type having its own forms m2m)
* It adds an extra table MissionWithForms which means extra queries


Considering the pros and cons, imho this refactor wasn't mandatory. It might have been nicer , but not mandatory. 

I'll let @madewulf do the final review before moving on

hugo093:
As discussed I'll merge this and move on

PR #3223, opened 2026-07-28
[IA-5340] improve tests speed and readability

tdethier:

improve tests speed and readability

### Related JIRA tickets

IA-5340

## Changes

- improved tests speed:
  - skipped migration `iaso 0389` tests since the migration was successfully applied on deployed servers
  - moved validation tests to serializer tests to avoid object creation and transactions in `SetupAccount`, `Profile` update
  - updated polio helper method for creating `Campaign` objects directly with the ORM instead of calling the API
  - refactored `test_auth_enforcement.py`
- improved tests readability:
  - removed `print()` calls in multiple tests
  - replaced `int` code status in API tests with const values from `rest_framework.status` (sorry about that, that makes a lot of noise, check [this link](https://github.com/BLSQ/iaso/compare/IA-5340_improve-tests-for-review) if you want to see a diff of this PR without that commit)
- removed old `TESTING` setting which was redundant with `IN_TESTS`
- fixed issues with some old loggers (`logger.ok()`)
- fixed some flaky tests

## How to test

- manually run tests or check the CI
  - tests should be faster
  - tests should no longer print values on the standard output
  - tests should not fail

## Print screen / video

before, locally:

image

after, locally:
image

before, in the CI:
image

after, in the CI:
image

## Notes
- timing improvements are not as big in the CI as locally - probably because we split tests on multiple workers and we don't usually do that locally

## Doc

/


iaso/api/source_versions_serializers.py
```diff
@@ -93,14 +91,13 @@ def validate(self, attrs):
 
     def generate_csv(self):
         data = self.validated_data
-        iaso_logger = CommandLogger(sys.stdout)
```
Phil-V, line 96: It looks like the custom IASO CommandLogger and the standard lib Logger are not compatible in their api, so it might break some stuff. For instance the Dumper relies on a method`.colorize()` that doesn't exist on the standard logger as far as I can tell. They also differ in how they accept multiple arguments when logging. Might affect the Differ, Exporter and Dumper.

iaso/tests/api/test_orgunits.py
```diff
@@ -1177,10 +1177,10 @@ def test_create_org_unit_fail_on_group_not_found(self):
             data={
                 "name": "Test ou",
                 "org_unit_type_id": self.jedi_council.pk,
-                "groups": [34],
+                "groups": [3446456767657675756],
```
Phil-V, line 1180: Was this a cause of flakiness? Good catch 😄

Phil-V (changes requested):

iaso/tests/api/test_orgunits.py
tdethier: yeah, it failed at some point when I ran tests here :sweat_smile:

iaso/api/source_versions_serializers.py
tdethier: good catch!

i don't know why we built this whole system of loggers instead of using the default one. i think i'll change all of this to default loggers

Phil-V (approved):
Much better with default loggers 👍

PR #3224, opened 2026-07-28
IA-5319: Save file sizes on upload

Phil-V:

Accessing the file size of existing files required making individual API calls to s3, which was not scalable. This PR stores the size of files as they are initially uploaded so they can be easily accessed/queried later.

### Related JIRA tickets

IA-5319

Relates to the account usage PR https://github.com/BLSQ/iaso/pull/3133.

## Changes

- Added a custom `SizedFileField` that stores the file sizes on upload.
- Migrated models to use this new field: Instance, InstanceFile, BulkCreateUserFile, FormVersion, FormAttachment, ImportGPKG, ReportVersion and APIImport.
- The field adds a `<field_name>_size` field to each affected model that can be queried independently (aggregate stats, etc.).
- The size is also accessible the usual way through the model instance `my_instance.file.size`. If the size was stored, it will return it without accessing the underlying Storage backend. 

## How to test

- Check that the uploading of files for the affected models is not broken.
- For the reviewer, use the django shell to check if the sizes are stored correctly, and are accessible through `.file.size` and `.file_size`. 

## Print screen / video

/

## Notes

- For the reviewer, useful background information is Django's [FileField and FieldFile implementation](https://github.com/django/django/blob/stable/4.2.x/django/db/models/fields/files.py). 
- The pre_save logic is overly defensive checking for (int, float) instances because many tests were broken when using mock objects for files as part of their setup.
- django-composite-field was mentioned when discussing this feature. After evaluation I decided not to use it in the end, instead drawing inspiration from it.

## Doc

/


iaso/utils/models/sized_file_field.py
```diff
@@ -0,0 +1,88 @@
...
+        """
+        super().contribute_to_class(cls, name, private_only=private_only, **kwargs)
+
+        if not self.size_field_name:
+            self.size_field_name = f"{name}_size"
+
+        try:
+            cls._meta.get_field(self.size_field_name)
+        except FieldDoesNotExist:
+            size_field = models.BigIntegerField(null=True, blank=True, editable=False)
```
tdethier: i would probably use a `PositiveBigIntergerField` since we're not going to use negative values

iaso/utils/models/sized_file_field.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def pre_save(self, model_instance, add):
+        """
+        Populate the database size column prior to saving the model instance
+        if the file is newly uploaded.
+        """
+        if file := getattr(model_instance, self.attname):
+            if not file._committed:
+                try:
+                    file_size = file.size
```
tdethier, line 70: this calls `SizedFieldFile.size()` on line 13, right?

iaso/utils/models/sized_file_field.py
```diff
@@ -0,0 +1,88 @@
...
+    """
+
+    attr_class = SizedFieldFile
+
+    def __init__(self, *args, size_field_name=None, **kwargs):
+        """
+        Initialize the field with an optional custom database size column name.
+        """
+        self._explicit_size_field_name = size_field_name
+        self.size_field_name = size_field_name
```
tdethier: why do we need both of them?

iaso/tests/utils/models/test_sized_file_field.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+class CustomSizedFileModel(models.Model):
+    file = SizedFileField(upload_to="test_uploads/", size_field_name="custom_size")
+
+    class Meta:
+        app_label = "iaso"
+
+
+class SizedFileFieldTest(TestCase):
```
tdethier, line 25: nice coverage!

tdethier (commented):
this looks good, nice job 👍 

what about the command that fills this new field, do we do this in another PR or in this one?

i'm just commenting for now in case we already include the command

iaso/utils/models/sized_file_field.py
```diff
@@ -0,0 +1,88 @@
...
+        """
+        Return the file size from the cached database column if available.
+        Otherwise, fall back to default behavior (querying the storage backend).
+        """
+        if self.field and self.instance:
+            size_field_name = getattr(self.field, "size_field_name", None)
+            if size_field_name:
+                val = getattr(self.instance, size_field_name, None)
+                if val is not None:
+                    return val
```
tdethier, line 22: this works for creating new objects with a file, but it doesn't work when changing an existing file from an object: the old size is still stored when it should be recomputed because the file is different

i spotted this while updating a `FileAttachment`

iaso/utils/models/sized_file_field.py
Phil-V: It was a bit messy, pushed some changes to make the logic clearer.

iaso/utils/models/sized_file_field.py
Phil-V: This is correct and points to a bug that I must have introduced as I refined the size logic on the SizedFieldFile. Added a fix and a test for that case.

Phil-V:
@tdethier 
The code from the other PR on Account usage will have to be adapted: https://github.com/BLSQ/iaso/pull/3133/changes#diff-e4ed6202fa035454990bbd426ab65fd7163b12ebf2207634ed04db8c27bcec3c

tdethier (approved):
thanks for bringing fixes so quickly!

LGTM 👍

PR #3226, opened 2026-07-29
[IA-5342] Fix "Cannot resolve keyword 'uuid' into field" in OUCRC.

bmonjoie:

When passing a string value in `project_id`, the server would return a 500 instead of a 400.

### Related JIRA tickets

IA-5342

## Changes

Changed `IdOrUuidRelatedField` into `serializers.PrimaryKeyRelatedField` since there are no `uuid` on `Project`

## How to test

Run tests or call the API yourself with invalid values.


tdethier (approved):
LGTM 👍

PR #3227, opened 2026-07-29
[SLEEP-1816]Modifying the orval config to pass the plugins

Crebert08:

This PR is about modifying the orval config so that it can generate the plugins endpoints

### Related JIRA tickets

[SLEEP-1816](https://bluesquare.atlassian.net/browse/SLEEP-1816)

## Changes

- Added a function to check the plugins endpoints schemas

## How to test

//

## Notes

Linked to this [PR](https://github.com/BLSQ/trypelim/pull/305)

## Doc

[SLEEP-1806]: https://bluesquare.atlassian.net/browse/SLEEP-1806?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[SLEEP-1816]: https://bluesquare.atlassian.net/browse/SLEEP-1816?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/pages/dev/reference/guidelines/api/api_synchronization.en.md
+++ b/docs/pages/dev/reference/guidelines/api/api_synchronization.en.md
@@ -16,11 +16,13 @@ We will primarily rely on the following tools:
 - [Backend](#backend)
   - [Setting up drf-spectacular](#setting-up-drf-spectacular)
   - [Documenting your endpoint](#documenting-your-endpoint)
+  - [Documenting an endpoint that lives in a plugin](#documenting-an-endpoint-that-lives-in-a-plugin)
   - [Testing the generated OpenAPI schema](#testing-the-generated-openapi-schema)
   - [Utilities](#utilities)
 - [Frontend](#frontend)
   - [Configuration](#configuration)
   - [Generating new files](#generating-new-files)
+  - [Generating a client for a plugin](#generating-a-client-for-a-plugin)
   - [Integrating React Query hooks](#integrating-react-query-hooks)
   - [Using zod schemas](#using-zod-schemas)
   - [Integrating with formik](#integrating-with-formik)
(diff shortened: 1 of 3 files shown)


Phil-V (commented):
This is a great improvement for Trypelim and other plugins, thanks for pushing for this idea and making it work 👍

For this PR, I recommend a couple of changes:
- The directory resolution logic: as Orval loads the swagger data from a live dev environment to generate its config, we might run into an issue: if the plugin is not loaded (`PLUGINS` env variable), but the plugin folder is present, Orval will delete the old config. I'm not sure what is the best way to resolve this. Perhaps add optional support for that variable if that's not too complicated? 
- Could you add documentation for this feature? A good place might be `docs/pages/dev/reference/guidelines/api/api_synchronization.en.md`

hugo093 (approved):
I trust the first pass of @Phil-V  on this. 

I'm just sharing some concerns, which comes along the "plugin" system we're using (so not your fault in this PR). 

The fact that we're not making plugins but augmented web apps everytime is starting to show its """weaknesses""". 

In this case for example, I'd have expect that we move the orval configuration files to a common repo that every plugin can use on its own and then generate the appropriate files without the need of `iaso`. Of course we'd have to be careful about names clashing, but as you showed in this PR, we can adjust the tags to avoid this. 

Here we're instead adding another layer in iaso about plugins, checking upon the .env etc. And in the future, if some plugin need some special treatment, it'll be a succession of if else blocks... 

Otherwise good job on this PR and the doc, appreciate the hard work and hope orval was nice for you to use

PR #3229, opened 2026-07-30
SNT-551 Exclude composite layer when getting csv template

Bewi:

Exclude composite layer when getting csv template

### Related JIRA tickets

SNT-551

## Changes

If snt malaria plugin is activated, filter metric types to exclude composite layers hen getting CSV template

## How to test

Testable in SNT Malaria. In Data layer view, add some composite layers, verify their key.
Then export CSV template, it should not contains composite layers.

## Print screen / video

N/A

## Notes

N/A

## Doc

N/A


michiwend (approved):
Not touched by your changes but related: Do we also want to guard against someone trying to import into a composite variable/code? This is quite theoretical as we generate these composite variables but also doesn't harm to guard against. Otherwise LGTM

PR #3230, opened 2026-08-03
IA-5343 improve webpack dev, remove npm from deps

beygorghor:

- Launching webpack dev without docker can take 187 secs on my computer
Screenshot 2026-08-03 at 12 34 50

- We are using npm as dependency, this is useless, also on [ bluesquare-component](https://github.com/BLSQ/bluesquare-components/pull/224) leading to critical update 

- browserlist is not up to date

### Related JIRA tickets

IA-5343

## Changes

- `usePollling` should only be set when using Docker to have live reload
- removed npm from deps
- updated browser list
- force npm and node version with engines and .npmrc file
- adding 'min-release-age=7' to .npmrc => Only install package versions _older_ than 7 days

### Locally
- run` npm ci`
- run  `npm run dev`, it should take less than a minute
- hot reload should still work

### On docker

- run `docker-compose build webpack`
- run `docker-compose up webpack`
- hot reload should still work

## Print screen / video
Screenshot 2026-08-03 at 12 30 44

## Notes
We need to merge this [PR](https://github.com/BLSQ/bluesquare-components/pull/224) too that also remove npm from deps on bluesqaure-components


tdethier (approved):
tested on Ubuntu 24.04:
- frontend raises an error when wrong version of `npm` is detected
- hot reload still works
- no major performance difference
- installing a recent package fails, as expected 


LGTM 👍

Bewi (approved):
LGTM

PR #3232, opened 2026-08-05
IA-5348: npm 11.19 and min release age

beygorghor:

⚠️ to make the CI working properly we need to merge this [pr](https://github.com/BLSQ/iaso-ci/pull/4) first

Only install package versions _older_ than 7 days to mitigate supply chain attacks

**npm 11.19**: min-release-age converts to an internal --before date. 
On npm 11.13, preparing a github: dependency (bluesquare-components) spawned a child install that still saw min-release-age in .npmrc, so npm crashed with “--min-release-age cannot be provided when using --before”. That conflict was fixed in npm 11.15+; we pin 11.19 so cooldown and git installs of BSC can coexist.

### Related JIRA tickets

IA-5348

## Changes

- bump to npm 11.19 everywhere
- add `min-release-age` to `.npmrc`

## How to test

- run `npm ci`
- run `npm run dev`

OR

 - run `docker-compose build webpack`
 -  un `docker-compose up webpack`

Make sure dashboard is loading

### DIFF
--- a/docker/bundle/Dockerfile
+++ b/docker/bundle/Dockerfile
@@ -25,7 +25,7 @@ COPY package.json .
 COPY package-lock.json .
 
 
-RUN npm install -g npm@11.13.0
+RUN npm install -g npm@11.19.0
 
 RUN rm -rf node_modules
 RUN npm cache clean --force
--- a/plugins/template/.github/workflows/ci.yml
+++ b/plugins/template/.github/workflows/ci.yml
@@ -36,7 +36,7 @@ jobs:
                   cache: npm
             - name: upgrade npm
               run: |
-                  npm install --global npm@11.13.0
+                  npm install --global npm@11.19.0
             - name: Environment info
               run: |
                   node --version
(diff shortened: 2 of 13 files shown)


plugins/template/.github/workflows/ci.yml
```diff
@@ -36,7 +36,7 @@ jobs:
                   cache: npm
             - name: upgrade npm
               run: |
-                  npm install --global npm@11.13.0
+                  npm install --global npm@11.19.0
```
tdethier, line 39: the plugin template CI files were not updated after the switch to the new CI system, we should do that

these changes are not needed because now the npm version is set only in `iaso-ci` but since it's already done, it can stay 🤷‍♂️

docker/bundle/Dockerfile
```diff
@@ -25,7 +25,7 @@ COPY package.json .
 COPY package-lock.json .
 
 
-RUN npm install -g npm@11.13.0
+RUN npm install -g npm@11.19.0
```
tdethier, line 28: and we should get rid of this file too now that i think about it

tdethier (approved):
LGTM 👍 

let's merge the `iaso-ci` PR first so that we can get this CI green here

PR #3235, opened 2026-08-06
IA-5349: Support more types on entity type dialog

beygorghor:

Form possible field is using entities duplicate field types, we should keep it while launching an entity duplicate analysis but we should be able to select other field types while editing a entity type
We should probably use more field types here https://xlsform.org/en/#question-types
Maybe also see what is really supported on mobile and on mobile when rendering those fields

### Related JIRA tickets

IA-5349

## Changes

- split types per usage, deduplication or entity type conifg
- adapt frontend to use correct usage
- backend test

## How to test

Use this form

[dev_test_form_all_field_types_2026080701.xlsx](https://github.com/user-attachments/files/30826662/dev_test_form_all_field_types_2026080701.xlsx)

Create a entity type, make sure you can select barcode field
try to launch a entity duplicate analysis, you should not be able to select barcode, time, date, select types

## Print screen / video

Screenshot 2026-08-07 at 12 51 51
Screenshot 2026-08-07 at 12 46 08

Screenshot 2026-08-07 at 12 07 50


Phil-V (approved):
Looking good!

PR #3236, opened 2026-08-06
IA-5350: HOTFIX: Fix error with PATCH Instance endpoint

Phil-V:

The full OrgUnit queryset was accidentally loaded and evaluated in-memory when checking permissions.

### Related JIRA tickets

IA-5350

## Changes

Remove the truthy check on the queryset in Instance PATCH endpoint.

## How to test

Go to a submission details page, in the speed dial options pick the "Edit attached OrgUnit or period" option, change the Instance org unit and submit. It should succeed without any errors.

## Print screen / video

/ 

## Notes

Added the atomic decorator but that is unrelated to the issue. 

## Doc

/


tdethier (approved):
LGTM 👍

PR #3243, opened 2026-08-12
SLEEP-1803: Optimize mobile org unit api performance

Phil-V:

The org unit mobile api can consume large amounts of memory and CPU time (~480MB / 26 seconds per API call for 25k org units) when syncing a large health pyramid. This PR attempts to alleviate the issue by reducing the amount of processing done in the view (reduced by about 50% to ~200MB / 13.5s per 25k).

### Related JIRA tickets

refs: SLEEP-1803

## Changes

Modified the mobile org unit API to load and initialize fewer related objects per org unit by relying on annotations instead. Specifically, a number of prefetches and selected related objects used to determine a valid parent_id for the org unit were replaced.

## How to test

There are no changes to the API so this is mostly about checking nothing was accidentally broken. 

- Ideally, test the api with the mobile app (sync health pyramid) to make sure the api works properly.
- You can also query the API directly:
`/api/mobile/orgunits/?limit=25000&page=1&shapes=false&app_id=your_project_app_id`

## Print screen / video
/

## Notes

The reviewer should be aware that the view has a custom cache mechanism (not related to this PR). When testing before/after the changes it might be necessary to clear the cache between calls to the api:
```
from django.core.cache import cache
cache.clear()
```

## Doc

/

### DIFF
--- a/iaso/api/mobile/org_units.py
+++ b/iaso/api/mobile/org_units.py
@@ -164,13 +164,7 @@ class MobileOrgUnitSerializer(serializers.ModelSerializer):
                 if parent_org_unit_type_id not in self.valid_org_unit_type_ids:
                     return None
             else:
-                # Fallback logic if the serializer is used in isolation (e.g. tests)
-                parent = org_unit.parent
-                parent_org_unit_type = parent.org_unit_type if parent is not None else None
-                if parent_org_unit_type is None or not any(
-                    p.app_id == self.app_id for p in parent_org_unit_type.projects.all()
-                ):
-                    return None
+                raise NotImplementedError("The serializer requires a valid project in its context.")
 
         return parent_id
 
--- a/iaso/api/mobile/org_units.py
+++ b/iaso/api/mobile/org_units.py
@@ -6,10 +6,12 @@ from django.contrib.gis.db.models import GeometryField
 from django.contrib.gis.db.models.aggregates import Extent
 from django.contrib.gis.db.models.functions import GeomOutputGeoFunc
 from django.core.cache import cache
+from django.db.models import F
 from django.db.models.expressions import RawSQL
 from django.db.models.functions import Cast
 from django.http import Http404, HttpResponseNotFound
 from django.shortcuts import get_object_or_404
+from django.utils.functional import cached_property
 from drf_spectacular.utils import extend_schema
 from rest_framework import serializers
 from rest_framework.decorators import action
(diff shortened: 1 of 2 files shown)


iaso/api/mobile/org_units.py
```diff
@@ -124,17 +126,53 @@ class Meta:
...
+            return name_annotated
         return org_unit.org_unit_type.name if org_unit.org_unit_type else None
 
+    @cached_property
+    def valid_org_unit_type_ids(self):
+        valid_ids = set()
+        project = self.context.get("project")
+        if project is not None:
+            valid_ids = set(project.unit_types.values_list("id", flat=True))
+        return valid_ids
```
bmonjoie, line 140: As discussed, I understand why it's better to pass a "simple" `Project` in the context, as it doesn't expose the serializer's inner logic, but I'm afraid it could lead to N+1 if we don't do `select_related` and `prefetch` in the `View`.  
I don't know if this should be changed or not.

iaso/api/mobile/org_units.py
```diff
@@ -124,17 +126,53 @@ class Meta:
...
+                if parent_org_unit_type_id not in self.valid_org_unit_type_ids:
+                    return None
+            else:
+                # Fallback logic if the serializer is used in isolation (e.g. tests)
+                parent = org_unit.parent
+                parent_org_unit_type = parent.org_unit_type if parent is not None else None
+                if parent_org_unit_type is None or not any(
+                    p.app_id == self.app_id for p in parent_org_unit_type.projects.all()
+                ):
+                    return None
```
bmonjoie: I understand why it was done like this, but I believe it's wrong.

Doing so, your isolation test will always test the wrong branch, and your real logic won't be tested.

I would either throw an exception here, ensuring that the project context is always passed, or ignore this path completely and return the parent_id.

iaso/api/mobile/org_units.py
```diff
@@ -225,17 +263,25 @@ def results_key(self):
...
+        if not hasattr(self, "request") or self.request is None:
+            return None
         user = self.request.user
         app_id = self.get_app_id()
-
-        limit_download_to_roots = False
-
         try:
-            project = Project.objects.get_for_user_and_app_id(user, app_id)
+            return Project.objects.get_for_user_and_app_id(user, app_id)
```
bmonjoie, line 267: I believe you should add a `prefetch_related` to `unit_types` here.

iaso/api/mobile/org_units.py
```diff
@@ -280,8 +331,12 @@ def list(self, request, *args, **kwargs):
...
                 OrgUnit.objects.filter_for_user_and_app_id(None, app_id)
                 .order_by("path")
-                .prefetch_related("parent__org_unit_type__projects", "groups")
-                .select_related("org_unit_type", "parent", "parent__org_unit_type")
+                .annotate(
+                    parent_validation_status=F("parent__validation_status"),
+                    parent_org_unit_type_id=F("parent__org_unit_type_id"),
+                    org_unit_type_name_annotated=F("org_unit_type__name"),
+                )
+                .prefetch_related("groups")
```
bmonjoie, line 333: This is repeated from `get_queryset`, why not use `self.get_queryset()` here?

bmonjoie (commented):
As discussed, I think the tests should be modified to add `with self.assertNumQueries(...)` and see how those changes have impacted the performance at the database level.

iaso/api/mobile/org_units.py
Phil-V: I agree and replaced it with a NotImplementedError

Phil-V:
It seems like the test suite already asserts the number of queries, but coincidentally these changes didn't affect the total number of queries. I updated the comment in the test suite that breaks down all the queries to reflect the changes.

iaso/api/mobile/org_units.py
Phil-V: It seems that the view initializes a different queryset altogether when passing ids. I'm not 100% about the intent of this and don't feel super confident refactoring it, but if you feel like this is a good opportunity to refactor this, perhaps we could have a call about it.

iaso/api/mobile/org_units.py
bmonjoie: It seems that using `prefetch_related` actually increases the number of queries. Therefore, this can be resolved.

bmonjoie (approved):
LGTM

PR #3248, opened 2026-08-13
IA-5362: Fix TS on Forms

beygorghor:

Starting TS fixes on the code base, we need to fix TS issues domain per domain, starting with `forms`

### Related JIRA tickets

IA-5362

## Changes

- listed all errors in form domain using `npx tsc --noEmit 2>&1 | grep "domains/forms/"` 
- fixed those one by one

## How to test

- run `npx tsc --noEmit 2>&1 | grep "domains/forms/"`, result should be empty
- make sure the whole app is working, especially in the forms

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/forms/components/FormFormComponent.tsx
+++ b/hat/assets/js/apps/Iaso/domains/forms/components/FormFormComponent.tsx
@@ -48,7 +48,7 @@ const useStyles = makeStyles((theme: Theme) => ({
     },
 }));
 
-const formatBooleanForRadio = value => {
+const formatBooleanForRadio = (value: boolean) => {
     if (value === true) return 'true';
     if (value === false) return 'false';
     return null;
@@ -72,7 +72,7 @@ const FormForm: FunctionComponent<FormFormProps> = ({
     const { data: allProjects, isFetching: isFetchingProjects } =
         useGetProjectsDropdownOptions();
 
-    const setPeriodType = value => {
+    const setPeriodType = (value: string | null) => {
         let periodTypeValue = value;
         if (value === null || value === NO_PERIOD) {
             periodTypeValue = null;
@@ -226,7 +226,8 @@ const FormForm: FunctionComponent<FormFormProps> = ({
                                             );
                                         }}
                                         value={formatBooleanForRadio(
-                                            currentForm.single_per_period.value,
+                                            currentForm.single_per_period
+                                                .value ?? false,
                                         )}
                                         errors={
                                             currentForm.single_per_period
--- a/hat/assets/js/apps/Iaso/domains/forms/components/FormFormComponent.tsx
+++ b/hat/assets/js/apps/Iaso/domains/forms/components/FormFormComponent.tsx
@@ -48,11 +48,7 @@ const useStyles = makeStyles((theme: Theme) => ({
     },
 }));
 
-const formatBooleanForRadio = (value: boolean) => {
-    if (value === true) return 'true';
-    if (value === false) return 'false';
-    return null;
-};
+const formatBooleanForRadio = (value: boolean) => (value ? 'true' : 'false');
 
 type FormFormProps = {
     currentForm: FormDataType;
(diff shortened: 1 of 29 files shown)


hat/assets/js/apps/Iaso/domains/forms/components/formStasts/Filters.tsx
```diff
@@ -35,7 +35,7 @@ export const Filters: FunctionComponent<Props> = ({ baseUrl, params }) => {
     });
 
     const handleChange = useCallback(
-        (key, value) => {
```
hugo093, line 38: since there's only projectIds used, shouldn't we set string instead of any?

hat/assets/js/apps/Iaso/domains/forms/components/FormFormComponent.tsx
```diff
@@ -48,7 +48,7 @@ const useStyles = makeStyles((theme: Theme) => ({
     },
 }));
 
-const formatBooleanForRadio = value => {
+const formatBooleanForRadio = (value: boolean) => {
```
hugo093: now that TS says it should be a boolean, maybe the code below can be "optimized" ? 🙃

hat/assets/js/apps/Iaso/domains/forms/components/FormsFilterComponent.tsx
```diff
@@ -21,7 +21,7 @@ type instance = {
     longitude?: Instance['longitude'];
 };
 
-type newForms = {
+export type newForms = {
```
hugo093: detail but I'd name it NewForms instead of newForms

hat/assets/js/apps/Iaso/domains/forms/components/FormVersionsDialogComponent.tsx
```diff
@@ -117,24 +118,27 @@ const FormVersionsDialogComponent: FunctionComponent<Props> = ({
...
-                        const entryValue: any = entry[1];
-                        if (entryKey === 'xls_file_validation_errors') {
-                            setXlsFileErrors(
-                                entryValue.map(err => err.message),
-                            );
-                        } else {
-                            setFieldErrors(entryKey, entryValue);
-                        }
-                    });
+                if ((error as DjangoError).status === 400) {
```
hugo093: to avoid casting everytime the error : 

at the top : 

if (!isDjango400Error(error)) {
        return;
    }

where something like: 

const isDjango400Error = (
    error: unknown,
): error is Django400Error => {
    return (
        error instanceof Error &&
        typeof error === 'object' &&
        error !== null &&
        'status' in error &&
        error.status === 400 &&
        'details' in error
    );
};

hat/assets/js/apps/Iaso/domains/forms/components/FormVersionsDialogComponent.tsx
```diff
@@ -177,15 +181,19 @@ const FormVersionsDialogComponent: FunctionComponent<Props> = ({
                         setIsLoading(false);
                         return;
                     }
-                } catch (error) {
+                } catch (error: unknown) {
```
hugo093, line 183: same remark

hat/assets/js/apps/Iaso/domains/forms/detail.tsx
```diff
@@ -190,9 +191,9 @@ const FormDetail: FunctionComponent = () => {
                     formId: savedFormData.id,
                 });
             }
-        } catch (error) {
-            if (error.status === 400) {
-                Object.entries(error.details).forEach(
+        } catch (error: unknown) {
```
hugo093, line 209: same remark

hat/assets/js/apps/Iaso/domains/forms/detail.tsx
```diff
@@ -169,14 +170,14 @@ const FormDetail: FunctionComponent = () => {
                 omit(currentForm, ['form_id', 'possible_fields']),
                 v => v.value,
             );
-            saveForm = createForm(formData);
+            saveForm = createForm(formData as Form);
```
hugo093: those casting are weird tbh, we should maybe fix the typing of create/updateForm instead of that wild casting. You are not really fixing TS there, you're bypassing it :/

hugo093 (changes requested):
Good work overall, already a first step. A few remarks here and there and also I'd try to use "any" whenever possible

hugo093 (approved):
LGTM

PR #3260, opened 2026-08-14
test(frontend): Add SNT to vitest config

michiwend:

Enable vitest for SNT

### Related JIRA tickets

SNT-548


vitest.config.ts
beygorghor, line 1: We can probably use something more generic for all the plugins like `plugins/*/js/**/*.test.{ts,tsx}`

beygorghor (changes requested):

vitest.config.ts
michiwend: Done ✅

beygorghor (approved):

PR #3267, opened 2026-08-17
[IA-5167]: Use new hooks from account API and adapt profile backend/hooks

hugo093:

Following the new account API, most of the data in the current profile API are redundant. Moreover, frontend wasn't using some of the new hooks

### Related JIRA tickets

IA-5167

### Backend

* Modified the endpoint payload for profiles/me and accounts/me to be more optimized and avoid data redundancy
* Modified the code to have dedicated DRF actions for delete/retrieve/update for "me" instead of relying too much on PK_ME
* Added and adapted tests

### Frontend

* Regenerate orval files
* Adding proper typing to current user hooks
* Added a new convenient hook for getting current account `useCurrentAccount`
* `useHasNoAccount` hook has been moved from `Iaso/utils/usersUtils.ts` to `Iaso/domains/accounts/hooks.ts`( :warning: breaking change)
* `hasFeatureFlag`function now takes the argument (featureFlag, account?) instead of (user, featureflag) ( :warning: breaking change)
* `useCurrentUserHasAccessToModule` function now takes the arguments (module, account?) instead of (module, user) ( :warning: breaking change)
* Some utility hooks have been created for convenience : `useCurrentUserHasFeatureFlag`, `useCurrentUserHasAccessToModule`, `useCurrentUserHasAllPermissions`, `useCurrentUserHasOneOfPermissions`
* Adapted tests and current FE code regarding the new hooks. 
* Replace `useGetCustomTranslation` and `updateDefaultDataSource` with orval generated hooks from account API

## How to test

Python tests
FE test + navigating the app

## Notes

Files under Iaso/apps/api are generated automatically and thus don't need reviewing


hat/assets/js/apps/Iaso/components/nav/TopBarComponent.tsx
```diff
@@ -177,8 +176,8 @@ const TopBar: FunctionComponent<Props> = ({
                                 <HomePageButton />
                             </Box>
 
-                            {currentUser.is_staff === true &&
-                                currentUser.is_superuser === true && (
```
beygorghor, line 181: 😄

hat/assets/js/apps/Iaso/components/DisplayIfUserHasPerm.tsx
beygorghor, line 1: ❤️

hat/assets/js/apps/Iaso/domains/users/components/ProtectedRoute.tsx
```diff
@@ -63,32 +65,29 @@ const ProtectedRoute: FunctionComponent<Props> = ({
...
     // this should kick in if the above effect didn't redirect the user to a better page
     const hasNoPermWarning =
         isRootUrl &&
-        (!currentUser.permissions ||
-            (currentUser.permissions.length === 0 && !isAuthorized));
+        (!currentUser?.permissions ||
+            (currentUser?.permissions.length === 0 && !isAuthorized));
     if (!currentUser) {
         return null;
     }
```
beygorghor, line 93: normally with this, everything under should have a `currentUser`, so checking a bit everywhere that we have a `currentUse`r is a bit overkill

beygorghor (changes requested):
- We still have useGetCurrentUser used at top level in `LocalizedAppComponent`. This is still using `User` and should probably now use `CurrentUser` instead.
- So you removed all those fields too, not only account infos ? Is this correct ? 
    - date_joined
    - user_permissions
    - user_roles
    - user_roles_permissions
    - dhis2_id
    - home_page
    - country_code
    - editable_org_unit_types
    - user_roles_editable_org_unit_type_ids
    - color
- I have asked to @mestachs , we will have probably to update [this](https://github.com/BLSQ/iaso-pipelines-superset/blob/main/civ-extract-entities-2026/pipeline.py#L534-L566) after 
- On a DS side, i asked to our POLIO DS and they never used this endpoint
- We need also to double check if `setuper `is not using this api
- Did you check the other python script too ?
- At some point Iaso-client should be updated too

PR #3270, opened 2026-08-18
IA-5364: query builder extra operators

beygorghor:

It can be very useful to perform a search on submitted text fields with more operators:

- Contain
- Does not contain
- is empty
- is not empty

### Related JIRA tickets

IA-5364

## Changes

- adapt `jsonlogic` to support those operators
- tests instances search and entities search with those operators
- adapt frontend to support those with text field only

## How to test

- try to search a submission, or an entities using query builder and a text field
- you should see those operators
Screenshot 2026-08-18 at 11 43 22

- make sure that results are relevant !

## Notes

- Follow up on workflows is also using query builder, but mobile doesn't support those operators for now. 
- Make sure those operators are not visible when creating or editing a follow-up in a workflow attached to a entity type


Phil-V (approved):
Good test coverage and seems to work well! 👍

PR #3275, opened 2026-08-19
Fix error when creating missions in multi-project environments

Phil-V:

Mission creation fails with a MultipleObjectsReturnedError if OrgUnitTypes or EntityTypes are shared across Projects.

### Related JIRA tickets

refs: IA-4872

## Changes

- Added distinct() calls to make sure the missions api accounts for the M2M relationship between EntityTypes/OrgUnitTypes and Projects.
- Updated test case setup to have multiple projects.

## How to test

- Make sure you have OrgUnitTypes/EntityTypes shared across multiple Projects (setuper should take care of that).
- Under /dashboard/planning/missions/, click Create
- Create a "Org Unit + Form" mission
- Create a "Entity + Form" mission.
- It should succeed without error.

## Print screen / video

/ 
## Notes

/ 
## Doc

/

### DIFF
--- a/iaso/api/missions/serializers/create/mission_entity_type.py
+++ b/iaso/api/missions/serializers/create/mission_entity_type.py
@@ -18,8 +18,10 @@ class EntityTypeScopedFormField(serializers.PrimaryKeyRelatedField):
         if not entity_type_pk:
             return Form.objects.none()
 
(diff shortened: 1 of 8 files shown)


bmonjoie (approved):
LGTM

PR #3277, opened 2026-08-20
IA-5367: Add api_import to the Instance model and API

Phil-V:

There was no easy way to identify or query which Instances belonged to which API Import, or to query Instances created by a specific app version. This PR establishes a relationship between the Instance and APIImport models. 

### Related JIRA tickets

refs: IA-5367

## Changes

- Modified the Instance model to add API Import FK
- Added the app version to the Instance page under technical details

## How to test

- Upload a form submission through the app.
  - Alternatively, edit an Instance through the django admin and edit its app version. 
- Check the Instance details page under technical details, you should see the app version.

## Print screen / video

ia-5367

## Notes

- Updated an old already-broken script (reimport_failed_imports.py), probably no need to spend much time reviewing it.

## Doc

/


Phil-V:
Hi @mestachs! Do these changes address what you had in mind for this ticket?

mestachs:
What I really like is the traceability it adds 
 - no more guessing, you'll find the source xml/zip it comes from (and which method was used)

My only fear is that at some point we might 
- want to archive API_IMPORT log request (big table)  
   - the fact that it's nullable make it possible but we will loose the app version id
- might leak information from API_IMPORT at some point (REQUEST_HEADER_INFO_KEYS, HTTP_AUTHORIZATION looks removed, but I see HTTP_COOKIE)
- select n+1 on api import or load by ids on huge table (and impact on other queries like entities, orgunit reference form)
   - I think it's already the case in your PR

So I'm wondering if keeping 2 fields on Instance
 - the version string => exposing this one in the serializer and parquet
 - the api import fk
but not navigating  api_import and not exposing details for it might not be better ?

what do you think @madewulf @beygorghor ?

mestachs:
I've added the app_version to ease the testing via setuper and 

my fear is that this join is not a good idea and we fetch all the vector control infos not only the app_version 

de-normalizing the app_version is perhaps safer. 

https://gist.github.com/mestachs/24e9ded7cefa7f439eee506f82016a87#file-sample-sql

```
       "vector_control_apiimport"."id",
       "vector_control_apiimport"."created_at",
       "vector_control_apiimport"."user_id",
       "vector_control_apiimport"."import_type",
       "vector_control_apiimport"."json_body",
       "vector_control_apiimport"."headers",
       "vector_control_apiimport"."has_problem",
       "vector_control_apiimport"."exception",
       "vector_control_apiimport"."file",
       "vector_control_apiimport"."app_id",
       "vector_control_apiimport"."app_version",
       "vector_control_apiimport"."file_size"
...

LEFT OUTER JOIN "vector_control_apiimport" ON ("iaso_instance"."api_import_id" = "vector_control_apiimport"."id")
```

Phil-V:
Thanks for the review and updating the setuper, that’s a great addition!

In light of your comments (especially about future-proofing) I’ve updated the PR with the following changes:

- store the app_version on the Instance model directly
- don’t expose the api import id and remove it from the Instance api responses or filters

I’ve also added the app_version to the parquet export. 

At the moment, the app_version is only exposed in the detail view for submissions.

You mentioned the exposed headers by the APIImport api. That’s probably beyond the scope of this PR (and a little concerning), we should probably strip the HTTP_COOKIE information from the response or switch to a whitelist of attributes?

mestachs (approved):

PR #3278, opened 2026-08-21
IA-5366: Task dialog overflow issue

butofleury:

Handle non breakable string Dhis2 URL in the UI.

### Related JIRA tickets

[IA-5366](https://bluesquare.atlassian.net/browse/IA-5368)

## Changes

Decoded dhis2 URL when import task fail in order to show better formatted URL in the UI.

## How to test

1. Try to run a background worker  `uv run ./manage.py tasks_worker` or  `docker compose run iaso manage tasks_worker`

2. Create a DHIS2 data source with correct user name and password, then create a new version from dhis2 and launch a dhis2 org unit import task.

3. Change the username and/or password for the DHIS2 data source in order to trigger an error then relaunch dhis2 ou import. From [the tasks list page](http://localhost:8081/dashboard/settings/tasks/), try to visualize the task information. You should see a better formatted **dhis2 url**  in **Result Message** on the Task section for the failed task.

Tasks-Action

For the dhis2 you can use the [Demo instance](https://play.im.dhis2.org/stable-2-40-12/) : 

- URL: https://play.im.dhis2.org/stable-2-40-12
- Username: admin
- Password: district

**More details on how to test in the below video.**

## Print screen / video

[AwesomeScreenshot-21_08_202610_15_09.webm](https://github.com/user-attachments/assets/0d24ba85-a1b3-4472-8e73-96b0a0167eba)

[IA-5366]: https://bluesquare.atlassian.net/browse/IA-5366?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/tasks/components/TaskBaseInfo.tsx
+++ b/hat/assets/js/apps/Iaso/domains/tasks/components/TaskBaseInfo.tsx
@@ -1,5 +1,12 @@
 import React, { FunctionComponent, ReactNode } from 'react';
-import { Table, TableBody, TableRow, TableCell, Button } from '@mui/material';
+import {
+    Table,
+    TableBody,
+    TableRow,
+    TableCell,
+    Button,
+    Box,
+} from '@mui/material';
 import { TablePropsSizeOverrides } from '@mui/material/Table/Table';
 import { makeStyles } from '@mui/styles';
 import { OverridableStringUnion } from '@mui/types/esm';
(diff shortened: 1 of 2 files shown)


Bewi (changes requested):
It works indeed but I would use the opportunity to fix the overflow on the ResultMessage table, it is adding an horizontal scroll, we could just make sure it does not get out of its own cell. (Not sure why we have two columns there neither).

Image

butofleury:
> It works indeed but I would use the opportunity to fix the overflow on the ResultMessage table, it is adding an horizontal scroll, we could just make sure it does not get out of its own cell. (Not sure why we have two columns there neither).
> Image

@Bewi , I've added the fix on the ResultMessage to avoid an horizontal scroll: https://github.com/BLSQ/iaso/pull/3278/changes/aa3af48b6081199d84780415f3f9232f6e06524f

Bewi (approved):
Thanks for the change, LGTM

PR #3283, opened 2026-08-25
POLIO-2192: SSO should not auto-provision unknown users

madewulf:

The WHO SSO login flow silently created a brand-new IASO account whenever the email from the identity provider didn't match an existing user, instead of rejecting the login — and could even crash with a 500 error when the identity provider didn't supply name claims.

### Related JIRA tickets

POLIO-2192

## Changes

- `hat/sso_views.py`: `complete_login()` no longer auto-provisions a `User`/`Profile` when no existing user matches the SSO email. It now raises `UnknownUserException`, caught by both the web callback (`SSOCallbackView.dispatch`) and the mobile token endpoint (`token_view`), returning a clear "No account found for this email address. Accounts must be created by an administrator before you can sign in." message (404 on the mobile JSON endpoint).
- Removed the now-dead "new account" notification email path (`send_new_account_email`, its templates under `hat/templates/sso/emails/`, and the `email_recipients_new_account` setting), since nothing creates accounts through SSO anymore.
- Added `hat/templates/socialaccount/authentication_error.html` overriding allauth's default error template. That default template ignores the specific exception message and only ever shows a generic "An error occurred" text — this override is needed so this error (and the other pre-existing SSO error cases: invalid token, multiple users with same email, etc.) actually surfaces its message to the user on the web login flow.
- Updated `hat/tests/test_sso_auth.py`: replaced the test asserting auto-creation with one asserting the new rejection behaviour.

## How to test

- Automated: `./manage.py test hat.tests.test_sso_auth` (all 11 tests pass).
- Manual, with `SSO_WHO_*` env vars configured for a test account:
  - Log in via WHO SSO with an email that has **no** corresponding IASO user for the configured account → should land on a page showing "No account found for this email address...", and no `User`/`Profile`/`SocialAccount` should be created.
  - Log in with an email that **does** match an existing user → should log in normally, same as before.

## Print screen / video

Backend-only auth-error page, no significant visual to capture. See "Notes" for how this was diagnosed on production instead.

## Notes

- This is the root cause of the `/polio/login/callback` 500 error seen in production on `poliogo.afro.who.int` on 2026-08-25, for accounts whose Entra ID profile had no first/last name set.
- The auto-provisioning logic being removed here was originally copied from `plugins/wfp_auth`, where it's intentional (that flow *is* meant to self-provision). It was carried over by mistake when the WHO SSO flow was scaffolded from it, and was never meant to apply here — WHO accounts must be provisioned by an admin ahead of time.

## Doc

None.


beygorghor (approved):
@madewulf i have fixed the translations

PR #3284, opened 2026-08-26
IA-5380 improve entity sync to read less

mestachs:

The query could be improved so postgres don't do a Seq Scan of most entities.

### Related JIRA tickets

IA-5380

## Changes

1. iaso/models/entity.py — the actual fix (commit 078254b)

Why: `_filter_entities_with_instances()` (used by every mobile entity sync request) filtered entities via `id__in=Subquery(instances.values("entity_id").distinct())`. The DISTINCT inside that subquery blocked Postgres from planning it as a cheap semi-join, forcing it instead to Parallel Seq Scan the entire iaso_instance table and run a per-row org-unit path check (8.5M loop iterations in the prod EXPLAIN) before it could even start matching entities. 

Fix: swapped it for a correlated `Exists(instances.filter(entity_id=OuterRef("pk")))`, which lets Postgres use the indexed iaso_instance.entity_id column to do a per-entity point lookup instead of scanning/sorting/deduping the whole table.

What it helped for (measured against a real prod-data copy, same account/profile/app_id as the incident):

```
┌────────────────────┬────────────────┬─────────────────────┐
│                    │ Execution time │ Disk blocks touched │
├────────────────────┼────────────────┼─────────────────────┤
│ Old (warm-storage) │ 4.94s          │ 727,014 (~5.5GB)    │
├────────────────────┼────────────────┼─────────────────────┤
│ New                │ 3.52s          │ 169 (~1.3MB)        │
└────────────────────┴────────────────┴─────────────────────┘
```

Identical result set (951 rows) both ways — no behavior change, 124 existing entity tests still pass. Wall time improved modestly (~1.4x), but the number that actually matters for the incident is the ~4,300x drop in disk blocks touched per execution — the old query dragged ~5.5GB through shared buffers on every single call, which is what turns "one device syncing" into "the DB falls over" once hundreds of devices do it at once.

Since the query get executed nearly twice (once as count(*) and once to get the actual page), the benefits are nearly doubled.

2. iaso/management/commands/explain_request.py — diagnostic tooling (commit 7daf74a)

Why: this is the command that produced the original slow-query trace pasted at the start of this session (EXPLAIN (ANALYZE, BUFFERS, FORMAT JSON)). The buffer/I-O data was already being captured but only existed buried inside a single-line JSON blob meant for pasting into pev — nobody scanning logs would spot "745,341 blocks read" by eye.

What it helped for: added a human-readable Explain summary: line per query (shared_hit/shared_read blocks in MB, I/O read time, execution time), plus an Explain totals across N explained queries: line at the end summing all of them. Makes this exact kind of investigation — "how much disk did this request actually touch" — visible directly in the command output next time, instead of requiring the manual EXPLAIN reconstruction I did in this session.

## How to test

if you really want, setup a copy of the database

Launch the command against it 
```
docker compose exec -e DEBUG=true -e DEBUG_SQL=true iaso python manage.py explain_request     --url-path="/api/mobile/entities/?limit_date=2026-08-24&app_id=rci.pnlp.cps.denombrement.administration.2026&limit=344&page=1"     --username="CSU_FOUMBOLO_VOL_07"     --threshold-duration=0 --explain=true
```

before the change to entity query
```
Explain totals across 12 explained queries: shared_hit=10,716,799 blocks (~83725.0 MB), shared_read=1,450,456 blocks (~11331.7 MB), io_read_time=3727.6ms, execution_time=11877.5ms
Headers: {'Content-Type': 'application/json', 'Vary': 'Accept', 'Allow': 'GET, POST, HEAD, OPTIONS'}
Status: 200
Response size: (0.96 MB)
Total request time: 11.574s with 12 queries
```
After
```
Explain totals across 12 explained queries: shared_hit=16,039,733 blocks (~125310.4 MB), shared_read=0 blocks (~0.0 MB), io_read_time=0.0ms, execution_time=5998.6ms
Headers: {'Content-Type': 'application/json', 'Vary': 'Accept', 'Allow': 'GET, POST, HEAD, OPTIONS'}
Status: 200
Response size: (0.96 MB)
Total request time: 6.148s with 12 queries
```

### DIFF
--- a/iaso/management/commands/explain_request.py
+++ b/iaso/management/commands/explain_request.py
@@ -49,6 +49,12 @@ def enable_capture_queries(executed_queries=[]):
 
 
 def dump_executed_queries(executed_queries, threshold_duration=0.4, auto_explain=False, full_stack=False):
+    explained_count = 0
+    total_hit_blocks = 0
+    total_read_blocks = 0
+    total_io_read_ms = 0.0
+    total_execution_ms = 0.0
+
     for i, q in enumerate(executed_queries):
         if q["duration"] > threshold_duration:
             print(f"\n--- SQL #{i + 1} ---")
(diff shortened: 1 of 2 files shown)


oschvr (approved):

Phil-V (approved):
Looking good! Also love the extra debug info 👍

Before:
`shared_hit=4,108,307 blocks (~32096.1 MB), shared_read=1,809,872 blocks (~14139.6 MB), io_read_time=0.0ms, execution_time=12777.8ms`

After:
`shared_hit=682,676 blocks (~5333.4 MB), shared_read=1,024,283 blocks (~8002.2 MB), io_read_time=0.0ms, execution_time=4246.7ms`

PR #3286, opened 2026-08-26
IA-5382 restrict embeddable pages with CSP frame-ancestors, gate PowerBI page type

madewulf:

Embeddable pages (polio dashboard embeds, generic `Page` model pages) had `X-Frame-Options` exempted with no replacement restriction, so any external site could iframe them; PowerBI pages also had no way to be configured from the Pages UI and were offered even when the PowerBI service-principal isn't set up.

### Related JIRA tickets

IA-5382

## Changes

- **CSP `frame-ancestors` for embeds**: instead of a blanket `X-Frame-Options` exemption with no replacement, embeddable views now send `Content-Security-Policy: frame-ancestors 'self' <allowed origins>`.
  - Polio dashboard embed routes (`embeddedCalendar`, `embeddedVaccineRepository`, `embeddedVaccineStock`, `embeddedLqasCountry`, `embeddedLqasMap`) use a new `EMBED_FRAME_ANCESTORS` env-configurable setting (`hat/settings.py`), defaulting to the WHO AFRO RRT ArcGIS hub domain. Applied via a small `frame_ancestors_csp` decorator (`iaso/utils/http.py`).
  - The generic `Page` model (`/pages/<slug>/`) gets a new per-page `embed_allowed_origins` array field, so each page can independently allowlist which origins may embed it, rather than sharing one global setting. Defaults to `'self'` only if unset.
- **PowerBI fields in the Pages UI**: the Pages create/edit dialog didn't expose `powerbi_group_id` / `powerbi_report_id` even though the `Page` model and API already supported the `POWERBI` type. Added the PowerBI radio option with its Group ID / Report ID fields (mirrors the existing Superset dashboard-ID pattern).
- **Gate PowerBI behind actual configuration**: PowerBI report rendering requires a `Config` row with slug `powerbi_sp` (service-principal credentials for the PowerBI API). The "PowerBI" type option is now hidden in the create/edit dialog unless that `Config` exists, via a new `GET /api/pages/powerbi_available/` endpoint. That endpoint deliberately returns only `{"available": bool}` — never the `Config.content`, which holds the service-principal secret — and stays visible when editing a page that's already `POWERBI`-typed even if the config is later removed, so existing pages don't lose their selected type in the UI.

## How to test

- `GET /pages/<slug>/` and the polio `embedded*` dashboard routes: check the response has `Content-Security-Policy: frame-ancestors ...` instead of no restriction.
- Pages > Create: without a `powerbi_sp` Config row, "PowerBI" is not offered as a type.
- Create a `Config` (Django admin) with slug `powerbi_sp` and any JSON content: "PowerBI" appears as a type option; Group ID / Report ID become required fields and are saved correctly on create/edit.
- `iaso/tests/test_xframe_options.py` and `iaso/tests/api/test_pages.py` cover the CSP header and the `powerbi_available` endpoint.

## Print screen / video

N/A (see linked ticket for verification screenshots taken during development).

## Notes

- The existing `ConfigViewSet` (`/api/configs/<slug>/`) was deliberately *not* reused for the PowerBI availability check: it filters by `Config.objects.filter(users=request.user)` and returns the raw `content`, which would leak the PowerBI service-principal secret to whichever users happen to be attached to that Config row.
- No changes to the React create/edit dialog for the new `embed_allowed_origins` field — it follows the same admin/API-only pattern already used for the PowerBI/Superset config fields prior to this PR.

## Doc

None yet — flagging in case the team wants this documented for admins setting up new embeddable pages/PowerBI.


madewulf:
Fixed a backward-compatibility issue: since \`embed_allowed_origins\` defaults to an empty list for every existing \`Page\`, the CSP header is now only sent once a page explicitly opts in by setting that field. Existing pages keep behaving exactly as before (embeddable anywhere, no restriction) until an admin configures allowed origins for them.

mestachs (approved):

hat/settings.py
```diff
@@ -300,6 +300,13 @@ def _get_from_environ(self, name, default, **kwargs):
         "iaso.middlewares.query_count.SafeQueryCountMiddleware",
     ]
 
+# Extra origins (besides 'self') allowed to embed IASO's embeddable views (Page model, polio
+# embedded dashboards) in an <iframe>, e.g. "https://afro-rrt-who.hub.arcgis.com". Used to build
+# the Content-Security-Policy: frame-ancestors header set by iaso.utils.http.frame_ancestors_csp.
+EMBED_FRAME_ANCESTORS = env.list(
+    "EMBED_FRAME_ANCESTORS", default=["https://afro-rrt-who.hub.arcgis.com"], delimiter=","
```
mestachs, line 307: IMO better for leave empty and not mention thirdparty urls (here specific to who)

hat/dashboard/views.py
```diff
@@ -63,6 +65,7 @@ def iaso(request: HttpRequest) -> HttpResponse:
 
 
 @xframe_options_exempt
+@frame_ancestors_csp
```
mestachs, line 68: I guess that it's ok since it but it's more stricter then before.
(which parent origins are allowed to put this response inside an <iframe>.)
which is probably not done for a lot of pages.

madewulf: In the end, I decided not to do it like this, sorry for the time you spent on this while it's not going forward. (configuration was made at gateway level, which we might reconsider later)

PR #3288, opened 2026-08-27
POLIO-2195: SSO error pages crash with 500 instead of rendering

madewulf:

Any SSO login error (invalid token, cancelled login, multiple accounts sharing an email, etc.) crashed with a raw 500 error instead of showing a friendly message.

### Related JIRA tickets

POLIO-2195

## Changes

- `hat/templates/account/base.html` (new): a project-owned override of allauth's default `account/base.html`. The stock allauth template's nav menu references URL names (`account_login`, `account_signup`, `account_email`, `account_logout`) that only exist if `allauth.account.urls` is registered — this project never registers them, since it has its own separate login system. Every SSO error page extends this template (via `socialaccount/base.html`), so rendering *any* SSO error crashed with `NoReverseMatch`. The override drops that menu block entirely (nothing in this project needs it).
- `hat/urls.py`: registers the missing `socialaccount_login_cancelled` URL name (redirecting to `settings.LOGIN_URL`). allauth's `render_authentication_error()` reverses this name specifically for the "user cancelled the login on the provider's side" case (`?error=access_denied`), which is a separate crash from the template one above — same root cause (unregistered allauth URLs), different code path.
- `hat/tests/test_sso_auth.py`: two new regression tests hitting the web (browser) callback directly — previously only the mobile/JSON token endpoint had test coverage, so this whole code path was untested. One test covers the cancelled-login redirect, the other covers the generic error-page render.

## How to test

- Automated: `./manage.py test hat.tests.test_sso_auth` (13 tests, all pass).
- Manual: with `SSO_WHO_*` configured, trigger any SSO error on the web login flow (e.g. click "Cancel" on the Microsoft login screen, or use an account that maps to more than one IASO user) — should land on a clean redirect/error page instead of a 500.

## Print screen / video

N/A — backend error-handling fix, no significant visual to capture.

## Notes

- Found while investigating a real production 500 for a WHO user who has two separate IASO accounts sharing the same email. SSO correctly detected the ambiguity and refused to guess which account to log into — but rendering that refusal as a page is what crashed. This PR fixes the crash; the underlying duplicate-account issue for that specific user is a separate, non-blocking data cleanup (tracked in POLIO-2195's description) that needs an admin decision on which account is canonical.
- Independent of (and does not depend on) the still-open POLIO-2192 PR (#3283, "SSO should not auto-provision unknown users") — that PR also touches this file but for a different reason (removing auto-account-creation). No functional overlap; only proximity in the same file.

## Doc

None.

### DIFF
--- a/hat/urls.py
+++ b/hat/urls.py
@@ -34,20 +34,21 @@ def _sso_providers():
 def get_sso_urlpatterns():
     """Generate URL patterns for all configured SSO providers."""
     patterns = []
-    sso_providers = getattr(settings, "SSO_PROVIDERS", {})
-    if sso_providers:
-        # allauth's own error-handling code (render_authentication_error) reverses this
-        # URL name when the user cancels the login on the provider's side. This project
-        # doesn't include allauth's own account/socialaccount urls (it has its own login
-        # system), so without this the cancelled-login case crashes with NoReverseMatch.
-        patterns.append(
-            path(
-                "accounts/login/cancelled/",
-                RedirectView.as_view(url=settings.LOGIN_URL),
-                name="socialaccount_login_cancelled",
-            )
+    # Always register this name: allauth's render_authentication_error() reverses it
+    # when the user cancels login on the provider's side. This project doesn't include
+    # allauth's own account/socialaccount urls (it has its own login system), so without
+    # this the cancelled-login case crashes with NoReverseMatch.
+    #
+    # Must not be gated on SSO_PROVIDERS: WFP is configured via WFP_AUTH_CLIENT_ID and
+    # is never added to SSO_PROVIDERS, but its callback uses the same allauth helper.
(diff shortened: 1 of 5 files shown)


hat/urls.py
```diff
@@ -34,7 +34,20 @@ def _sso_providers():
 def get_sso_urlpatterns():
     """Generate URL patterns for all configured SSO providers."""
     patterns = []
-    for provider_id, config in getattr(settings, "SSO_PROVIDERS", {}).items():
+    sso_providers = getattr(settings, "SSO_PROVIDERS", {})
```
mestachs: if WFP_AUTH_CLIENT_ID implies SSO_PROVIDERS not empty it's ok?

mestachs (approved):

mestachs (approved):

PR #3291, opened 2026-08-31
IA-5377: Add follow task option to org unit bulk edit dialog

Phil-V:

Add a button to confirm and follow the task when editing org units in bulk. 

### Related JIRA tickets

refs: IA-5377

## Changes

Similar to what we already have for the data sources dialog, add a confirm & follow task button. 

## How to test

- Test the bulk edit confirmation dialog:
  - Go the Org Unit list page
  - select multiple org units
  - use the speed dial actions to edit the org units
  - pick a random attribute to modify and click "validate"
- To be extra thorough, check that the dialog works properly when the user doesn't have the tasks permissions ("Batch monitoring" under user permissions), in which case the follow button should not appear. 

## Print screen / video

Screenshot 2026-08-31 at 11 25 47

## Notes

/ 

## Doc

/

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/orgUnits/components/OrgUnitsMultiActionsDialog.tsx
+++ b/hat/assets/js/apps/Iaso/domains/orgUnits/components/OrgUnitsMultiActionsDialog.tsx
@@ -68,16 +68,6 @@ const useStyles = makeStyles(theme => ({
         display: 'flex',
         alignItems: 'center',
     },
-    warningIcon: {
-        display: 'inline-block',
-        marginLeft: theme.spacing(1),
-        marginRight: theme.spacing(1),
-    },
-    warningMessage: {
-        display: 'flex',
-        justifyContent: 'center',
-        alignItems: 'center',
-    },
 }));
 
 const stringOfIdsToArrayofIds = stringValue =>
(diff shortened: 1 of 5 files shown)


hat/assets/js/apps/Iaso/domains/orgUnits/messages.ts
```diff
@@ -633,6 +642,18 @@ const MESSAGES = defineMessages({
         id: 'iaso.orgUnits.orgUnitRejected',
         defaultMessage: 'Org Unit rejected successfully',
     },
+    goToCurrentTask: {
```
mestachs: it's now modifyAndFollow no ?

hat/assets/js/apps/Iaso/domains/orgUnits/components/OrgUnitsMultiActionsDialog.tsx
```diff
@@ -72,6 +76,7 @@ const useStyles = makeStyles(theme => ({
     warningMessage: {
```
mestachs: warningMessage and warningIcon are now unused (since you use Alert)

mestachs (approved):
minor cleanup possible but works as expected.

PR #3292, opened 2026-08-31
POLIO-2189: Round scope update

beygorghor:
When you select separate scope the map resets to no scope selected, thus loosing the scope details and the sub-activities.

When separate scope per round is selected, to by default assume the current scope for all the round and the user updates scope where needed.

### Related JIRA tickets

POLIO-2189

## Changes

- display dialog to handle scope when changing separate per scope
- save accordingly
- small refactor of history button position

## How to test
Edit a a campaign scope, and play with the separate scope per round option, you should see a dialog asking you what to do with the scope.

## Print screen / video

Screenshot 2026-08-31 at 17 19 43
Screenshot 2026-08-31 at 17 19 23
Screenshot 2026-08-31 at 17 19 18
Screenshot 2026-08-31 at 11 46 36
Screenshot 2026-08-31 at 11 46 32
Screenshot 2026-08-31 at 11 46 27


plugins/polio/js/src/domains/Campaigns/hooks/api/useGetCampaignScope.ts
beygorghor, line 1: This is unused

plugins/polio/js/src/domains/Campaigns/Scope/ScopeChangeDialog.tsx
```diff
@@ -0,0 +1,334 @@
...
+    warning: {
+        mt: 1,
+        mb: 1,
+    },
+    disclaimer: {
+        mt: 2,
+    },
+};
+
+const formatRoundDate = (
```
Bewi: Feels like to date formats helpers could be in a util (same for date range bellow)

plugins/polio/js/src/domains/Campaigns/Scope/ScopeChangeDialog.tsx
```diff
@@ -0,0 +1,334 @@
...
+        onConfirm({ mode, selectedRoundNumbers: selected });
+    }, [mode, onConfirm, roundNumbers, selectedRoundNumbers]);
+
+    const toLabel = formatMessage(MESSAGES.dateRangeTo);
+    const dropsDistricts =
+        mode === 'empty' ||
+        (mode === 'selectedRounds' &&
+            selectedRoundNumbers.length < roundNumbers.length);
+
+    const getRoundLabel = (round: ScopeChangeRound): string => {
```
Bewi, line 187: Could be in a useCallback

plugins/polio/js/src/domains/Campaigns/Scope/scopeChangeUtils.ts
```diff
@@ -0,0 +1,81 @@
+import cloneDeep from 'lodash/cloneDeep';
+
+import { Scope, Vaccine } from '../../../constants/types';
+
+export const uniqueDistrictIds = (scopes: Scope[] = []): Set<number> => {
+    const ids = new Set<number>();
+    scopes.forEach(scope => {
```
Bewi, line 7:
```typescript
const ouIds = scopes.flatMap(scope => scope?.group?.org_units ?? []);
const uniqueIds = new Set<number>(ouIds);
```

plugins/polio/js/src/domains/Campaigns/Scope/scopeChangeUtils.ts
```diff
@@ -0,0 +1,81 @@
...
+        scope.group?.org_units?.forEach(id => ids.add(id));
+    });
+    return ids;
+};
+
+export const countDistricts = (scopes: Scope[] = []): number =>
+    uniqueDistrictIds(scopes).size;
+
+export const countMergedDistricts = (scopeLists: Scope[][]): number => {
+    const ids = new Set<number>();
```
Bewi: Same here:
```typescript
const ids = scopeLists.flatMap(scopes => uniqueDistrictIds(scopes));
return new Set<number>(ids);
```

plugins/polio/js/src/domains/Campaigns/Scope/scopeChangeUtils.ts
```diff
@@ -0,0 +1,81 @@
...
+    });
+    return ids.size;
+};
+
+type RoundLike = { number: number; scopes?: Scope[] };
+
+/** Union of district ids. If a district appears in several lists, the last vaccine wins. */
+export const mergeScopes = (scopeLists: Scope[][]): Scope[] => {
+    const districtToVaccine = new Map<number, Vaccine | undefined>();
+    scopeLists.forEach(scopes => {
```
Bewi: Tempted to use flatMap here as well or something else to simplify the nesting of foreach, but didn't find anything :D

plugins/polio/js/src/domains/Campaigns/Scope/scopeChangeUtils.ts
```diff
@@ -0,0 +1,81 @@
...
+                districtToVaccine.set(id, scope.vaccine);
+            });
+        });
+    });
+    const byVaccine = new Map<string, number[]>();
+    districtToVaccine.forEach((vaccine, id) => {
+        const key = vaccine ?? '';
+        const orgUnits = byVaccine.get(key) ?? [];
+        orgUnits.push(id);
+        byVaccine.set(key, orgUnits);
```
Bewi:
```typescript
byVaccine.set(key, [...orgUnits, id]);
```

plugins/polio/js/src/domains/Campaigns/Scope/scopeChangeUtils.ts
```diff
@@ -0,0 +1,81 @@
...
+    campaignScopes,
+    selectedRoundNumbers,
+}: {
+    toRounds: boolean;
+    rounds: T[];
+    campaignScopes: Scope[];
+    selectedRoundNumbers: number[];
+}): { rounds: T[]; scopes: Scope[] } => {
+    const selected = new Set(selectedRoundNumbers);
+    if (toRounds) {
```
Bewi: Not sure what this means, I had to read multiple time the logic to understand it, maybe a more descriptive name ?

Bewi (changes requested):
Few technical improvements.

Also I would sort per region by default.
And I had some strange behavior with the cancel (sometimes it wipes the screen, other time when selectin a region, icon doesn't change anymore, etc...)

plugins/polio/js/src/domains/Campaigns/Scope/ScopeChangeDialog.tsx
beygorghor: Not really in fact, it returns only a string and label receives a string

beygorghor:
Cancel behavior is already like that in prod, i'll open another ticket for that.
Regarding order region, i'll prefer to see with the client if this is what they want.
Other issues has been fixed

Bewi (approved):

PR #3293, opened 2026-09-01
[IA-5393] Fix Mobile Validation Workflow returns all instances

bmonjoie:

The endpoint should return only instances created by the user. 
This PR also fixes a couple of N+1 issues.

### Related JIRA tickets

IA-5393

## Changes

The query for mobile was missing a filter on the `created_by`.

I also moved queries that were made directly inside the `Serializer` into a `QuerySet` object that can be inherited.

## How to test

1. Have validation workflow configured
2. Create new instances with different users
3. Call the mobile endpoint and check that only your own instances are returned.


Phil-V (approved):
Looking good, no notes!

PR #3294, opened 2026-09-02
IA-5387 speed up incremental and full mobile entity sync

mestachs:

Explain here in one sentence.

### Related JIRA tickets

IA-5387

## Changes

Fixes `/api/mobile/entities/` being significantly slower than the equivalent web endpoint, for both incremental syncs (`limit_date` set) and full/fresh-install syncs.

## Root cause

`filter_for_user` (org-unit scope) and `filter_for_mobile_entity` (`limit_date`) each issued their own `EXISTS(...)` subquery against `iaso_instance`. Two stacked `EXISTS()` clauses on the same table — even though each is individually cheap — pushed Postgres off a cheap per-entity incremental plan into a full sequential scan + hash join of the whole account, **regardless of org-unit scope size**.

Fix: merge both conditions into a single correlated `EXISTS(...)` when both apply (`iaso/models/entity.py`, threaded through `filter_for_user`/`filter_for_user_and_app_id`, backward compatible via a `skip_limit_date_filter` flag so other callers are unaffected).

Another big change is issuing a single sql to get the results and the count (with a window function). So doing only once the heavy query (compared to previously with 2 similar sql, one doing the count, one fetching the data)

**Measured (real accounts, clean before/after A/B):**

| Scenario | User | Before | After |
|---|---|---|---|
| Incremental | narrow org-unit scope | 6.06s / 16.8M blocks | 2.38s / 4.6M blocks |
| Incremental | whole-country scope | 5.53s / 12.5M blocks | 3.45s / 5.9M blocks |
| Full sync | narrow org-unit scope | 10.86s / 24.9M blocks | 5.11s / 10.7M blocks |
| Full sync | whole-country scope | 8.50s / 11.7M blocks | 5.26s / 5.9M blocks |

Verified the response content is identical before/after (same `count`, same entity ids, same order — hashed and diffed), not just faster.

## Other changes

- **`filter_for_app_id`**: `entity_type__in=(subquery)` instead of a join through the `projects` m2m — safe regardless of how many projects share a form, unconditional low-risk win.
- **`.only()`** on the `attributes` join in the mobile queryset — it was pulling every `iaso_instance` column (JSON payload, geometry, file paths) when the serializer only reads `.uuid`. Kept the join (already mandatory for the `attributes__deleted` filter) but trimmed the `SELECT` list. No N+1 introduced.
- **`MobileEntitiesSetPagination`**: merges the pagination `COUNT()` and the data-slice query into one via `Window(Count("id"))`, instead of DRF's default two separate queries.
- **Bonus fix**: `EntityFilterSet.groups` (web `/api/entities/`) could duplicate entities when an org unit belonged to 2+ of the requested groups — same class of join-multiplication bug on a different relation, found while auditing for similar cases. Rewrote as `Exists(...)`, matching the pattern above.

## Tooling

Added `setuper/emulate_mobile_sync.py` — a standalone `uv` script that replicates the mobile app's actual sync request pattern (traced from `FetchEntities.kt` in `iaso-mobile-app`: `has_next`-driven pagination, `limit_date` omitted for a full sync, paired `/entities/deleted/` fetch) for local testing against a real server.

## Tests

12 new/updated tests: duplicate-row regressions (shared reference form across projects, `groups` filter), correctness of the merged org-unit+`limit_date` filter (including the full-sync case — an entity a `limit_date` cutoff excludes must still appear when `limit_date` is omitted), pagination edge cases (empty page 1, page 2 of an empty result → 404, true last partial page), and a regression guard that a non-empty page issues exactly one entity query. Each was verified individually against unmodified `develop` to confirm it checks behavior, not implementation — the correctness tests pass on both, the regression-guard tests fail on `develop` for the expected reason.

## How to test

restore a copy of production

 - CSU_FOUMBOLO_VOL_07 (~7 org units) is the representative/common case; 
 - MIAN (~19k org units, whole pyramid)
 

```
# Incremental sync (limit_date set) -- narrow org-unit scope, the common case
docker compose exec iaso python manage.py explain_request \
  --url-path="/api/mobile/entities/?limit_date=2026-08-24&app_id=rci.pnlp.cps.denombrement.administration.2026&limit=1000&page=1" \
  --username="CSU_FOUMBOLO_VOL_07" --threshold-duration=0 --explain=true

# Incremental sync (limit_date set) -- whole-country org-unit scope
docker compose exec iaso python manage.py explain_request \
  --url-path="/api/mobile/entities/?limit_date=2026-08-24&app_id=rci.pnlp.cps.denombrement.administration.2026&limit=1000&page=1" \
  --username="MIAN" --threshold-duration=0 --explain=true

# Full sync (limit_date omitted -- fresh install) -- narrow org-unit scope
docker compose exec iaso python manage.py explain_request \
  --url-path="/api/mobile/entities/?app_id=rci.pnlp.cps.denombrement.administration.2026&limit=1000&page=1" \
  --username="CSU_FOUMBOLO_VOL_07" --threshold-duration=0 --explain=true

# Full sync (limit_date omitted -- fresh install) -- whole-country org-unit scope
docker compose exec iaso python manage.py explain_request \
  --url-path="/api/mobile/entities/?app_id=rci.pnlp.cps.denombrement.administration.2026&limit=1000&page=1" \
  --username="MIAN" --threshold-duration=0 --explain=true
```

you can also emulate what the mobile does

```
# Narrow org-unit scope (~2,022 entities / 3 pages) -- fine to let it run to completion
uv run setuper/emulate_mobile_sync.py \
  --app-id rci.pnlp.cps.denombrement.administration.2026 \
  --username CSU_FOUMBOLO_VOL_07 --password CSU_FOUMBOLO_VOL_07_PWD
```

or point it at the staging/production

```
uv run setuper/emulate_mobile_sync.py \
  --server https://your-staging-instance.com \
  --app-id someapp --username someuser --password somepass
```

### DIFF
--- a/iaso/api/entities/filters.py
+++ b/iaso/api/entities/filters.py
@@ -22,7 +22,7 @@ class EntityFilterSet(FilterSet):
     created_by_team_id = CharFilter(field_name="attributes__created_by__teams__id")
 
     entity_type_ids = CharInFilter(field_name="entity_type_id", lookup_expr="in")
-    groups = CharInFilter(field_name="attributes__org_unit__groups", lookup_expr="in")
+    groups = CharInFilter(method="filter_groups")
 
     show_deleted = BooleanFilter(field_name="deleted_at", lookup_expr="isnull", exclude=True)
     orgUnitId = CharFilter(method="filter_org_unit")
@@ -32,6 +32,16 @@ class EntityFilterSet(FilterSet):
     class Meta:
         model = Entity
         fields = []
+
+    def filter_groups(self, queryset, name, value):
+        # `attributes__org_unit__groups__in=value` (a direct field_name+lookup_expr filter, as this
+        # used to be) joins through Group.org_units, a genuine m2m: an org unit routinely belongs to
+        # several groups, so filtering by 2+ group ids that a single org unit belongs to would join
+        # once per matching membership and duplicate the Entity row. Exists(...) checks membership
+        # without joining, so it can't duplicate -- same pattern as EntityDateFilterBackend below.
+        return queryset.filter(
+            Exists(Instance.objects.filter(pk=OuterRef("attributes_id"), org_unit__groups__in=value))
+        )
 
     def filter_org_unit(self, queryset, name, value):
         try:
--- a/iaso/api/mobile/entity.py
+++ b/iaso/api/mobile/entity.py
@@ -1,8 +1,12 @@
+import math
+
+from django.db.models import Count, Window
 from django_filters.rest_framework import DjangoFilterBackend  # type: ignore
 from drf_spectacular.utils import extend_schema
 from rest_framework import filters, permissions, serializers
 from rest_framework.exceptions import AuthenticationFailed, NotFound, ParseError
 from rest_framework.pagination import PageNumberPagination
+from rest_framework.response import Response
 
 from iaso.api.common import DeletionFilterBackend, HasPermission, ModelViewSet, Paginator, TimestampField
 from iaso.api.query_params import LIMIT, PAGE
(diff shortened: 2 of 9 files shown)


Phil-V (approved):

Phil-V:
By the way, found one unnecessary prefetch here (not introduced by this PR): #3311

PR #3296, opened 2026-09-02
WC2-949: failing tests in Azure pipelines

beygorghor:
Due to latest change  in azure config, now we can see that python tests are failing since a while, but it wasn’t stopping the pipeline (due to order of commands in yaml config).

Those tests should pass in azure pipeline, we should not ignore them.
### Related JIRA tickets

WC2-949

## Changes

Azure was failing tests that are green on GitHub because the two CIs don’t share the same env: Azure sets DISABLE_PASSWORD_LOGINS, doesn’t compile .mo files, and (after #3233) actually fails the job when manage.py test is last.

Compile translations in Azure (same as GitHub)
Keep password logins off in YAML; force them on only during manage.py test
Don’t crash auth when token_obtain_pair is missing
Align the GPKG missing-column assertion with the current error message

Also fixed /health endpoint returning empty values on Azure

## How to test

- this should run on azure pipeline
Screenshot 2026-09-02 at 13 16 35

### DIFF
--- a/iaso/auth/backends.py
+++ b/iaso/auth/backends.py
@@ -1,9 +1,22 @@
 from django.contrib.auth import get_user_model
 from django.contrib.auth.backends import ModelBackend
-from django.urls import reverse
+from django.urls import NoReverseMatch, reverse
 
 
 UserModel = get_user_model()
+
+
+def _is_token_obtain_request(request) -> bool:
+    """True when this authenticate() call is for JWT token issuance.
+
+    `token_obtain_pair` is not registered when DISABLE_PASSWORD_LOGINS is on.
+    """
+    if not request or not hasattr(request, "path"):
+        return False
+    try:
+        return request.path == reverse("token_obtain_pair")
+    except NoReverseMatch:
+        return False
 
 
 class MultiTenantAuthBackend(ModelBackend):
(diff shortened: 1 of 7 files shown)


mestachs (approved):
I guess the real test will be to check the status from azure build

iaso/auth/backends.py
```diff
@@ -1,11 +1,24 @@
...
 
+def _is_token_obtain_request(request) -> bool:
+    """True when this authenticate() call is for JWT token issuance.
+
+    `token_obtain_pair` is not registered when DISABLE_PASSWORD_LOGINS is on.
+    """
+    if not request or not hasattr(request, "path"):
+        return False
+    try:
+        return request.path == reverse("token_obtain_pair")
```
mestachs, line 17: ce n'est pas plus facile de checker ? settings.DISABLE_PASSWORD_LOGINS

mestachs (approved):

iaso/auth/backends.py
beygorghor: That’s weaker than catching NoReverseMatch. URLs are wired at import time, and IN_TESTS forces DISABLE_PASSWORD_LOGINS = False in hat/settings.py, so override_settings can disagree with the live URLconf. The existing test mocks reverse for that reason.

PR #3299, opened 2026-09-02
[IA-5391] Add "OU Code" and "OU Status" in the submission export

bmonjoie:

Add "OU Code" and "OU Status" to the Submissions export.
And allow to filter on OU Status, in the Submissions list.

### Related JIRA tickets

IA-5391

## Changes

1. Add 2 columns to the CSV and XLS Submissions export.
1. Add a new filter to the Submissions list. 
1. Display the OU status in the list.

## How to test

1. Go to the Submissions list page.
1. Click on the XLS/CSV buttons to see if the OU code and status have been correctly added. 
1. Filter on the OU status and see if it works.

## Print screen / video

image
image

### DIFF
--- a/hat/assets/js/apps/Iaso/domains/instances/constants.tsx
+++ b/hat/assets/js/apps/Iaso/domains/instances/constants.tsx
@@ -221,6 +221,17 @@ export const INSTANCE_METAS_FIELDS = [
         type: 'location',
     },
     {
+        key: 'org_unit_status',
+        Cell: settings => {
+            const data = settings.row.original;
+            return data.org_unit.validation_status;
+        },
+        sortable: false,
+        active: false,
+        type: 'info',
+        tableOrder: 7,
+    },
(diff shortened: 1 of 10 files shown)


hat/assets/js/apps/Iaso/domains/instances/constants.tsx
```diff
@@ -220,6 +220,17 @@ export const INSTANCE_METAS_FIELDS = [
         tableOrder: 7,
         type: 'location',
     },
+    {
+        key: 'org_unit_status',
+        Cell: settings => {
```
beygorghor, line 225: Best way to proceed here with a type is to make like this:

`Cell: ({ row: { original: orgUnit } }: Setting<OrgUnit>) => (`

So you can use orgUnit typed as an Org unit

Bewi (approved):
WaE & LGTM

PR #3300, opened 2026-09-02
POLIO-2198: Cancel Campaign edition is emptying the form

beygorghor:

When editing a campaign, pressing on cancel is removing values in all the fields.

It should reset the form to the original state.

Cancel button should not be enabled if not having changes.

### Related JIRA tickets

POLIO-2198

## Changes

Changes made on this PR are fixing initial issue but creating this one, since the campaign edition and creation in on a dedicated page, `setSelectedCampaignId` is not used here anymore.

## How to test

Edit a campaign.
Cancel should be disable.
Change something.
Press on Cancel, your changes should be removed.
change something and save

## Print screen / video

https://github.com/user-attachments/assets/24ec7708-fef8-41be-bb86-4bd18b7e4634


Bewi (approved):
Was & LGtM

PR #3303, opened 2026-09-04
IA-5390: Add an extra field "CODE" in the "Select visible column" for orgunit

butofleury:

Allow to visualize the CODE of the orgunit in the list of visible columns

### Related JIRA tickets

[IA-5390](https://bluesquare.atlassian.net/browse/IA-5390) 

## Changes

Added the org unit code in the Visible columns to allow enabling if needed!

## How to test

From org unit list page, click on the `Select Visible collums` button and check if you see the code in the visible collumns and enable it if needed and see if the code is visible on each row from the org unit list!

## Print screen / video

image

image

[IA-5390]: https://bluesquare.atlassian.net/browse/IA-5390?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bmonjoie (changes requested):
Everything seems to work except that the request was not to make it visible by default and allow enabling it by going into the "Visible columns".

bmonjoie (approved):
LGTM!

PR #3306, opened 2026-09-07
POC-  orgunit v3

mestachs:

proposition for a new orgunit api

### Related JIRA tickets

IA-XXX, WC2-XXX, POLIO-XXX, POLIOPM-XXX SLEEP-XXX, SNT-XXX, CONSOLE-XXX

## Changes

mainly addition (new v3 name space)

main changes 
* fields= DSL + schema endpoint, 
* fixed-shape vs sub-selectable relation fields, 
* ancestor_id ltree descendants, 
* bbox/within/outside spatial filters, 
  * find descendants outside the shape of a country/region : `/api/v3/orgunits/?ancestor_id=<country_id>&location__outside_org_unit=<country_id>&fields=id,name,latitude,longitude,ancestors(id,name)`
  * find point outside a bbox : `/api/v3/orgunits?location__outside_bbox=-8.60,4.19,-2.49,10.74&fields=id,name,latitude,longitude,ancestors(id,name)`
* strict param validation with suggestions, 
* explicit django-filter FilterSet 
   - id/id__in, name/source_ref/code variants, 
   - date range filters (ISO 8601 vs v1's custom formats), 
   - new ancestor_id/ancestor_id__direct_children (ltree-based "all descendants" vs v1's one-level-only parent_id), 
   - new bbox/within/outside spatial filters, 
   - strict unknown-param rejection (400 + typo suggestions) vs v1's silent ignore. Not ported: searches union queries, hasInstances, org-unit link filters, ids:/refs:/codes: prefixes.
   - see https://github.com/BLSQ/iaso/pull/3306/changes#diff-233d9aea8c3d71bbfb563c2d3df8bdb8f03fa040797f5b908104ed506ca9439eR57-R82
* non-counting pagination (speed up queries)
* a generic cross-endpoint nomenclature contract test (test_filter_nomenclature.py).

## How to test

Explain how to test your PR.

## Print screen / video

## Notes

in the things I noticed we don't have indexes for search by name.
- since most query are icontains : probably need a trigram based indexe to stay performant
- ideally should be doubled by unaccent 

## Doc

### DIFF
--- a/iaso/api/v3/common/fields_parser.py
+++ b/iaso/api/v3/common/fields_parser.py
@@ -1,0 +1,99 @@
+"""Parser for the v3 `fields=` query parameter sub-selector grammar.
+
+Grammar (informal)::
+
+    fields := ws field ws (',' ws field ws)*
+    field  := name ws ['(' ws fields ws ')']
+    name   := [A-Za-z_][A-Za-z0-9_]*
+    ws     := whitespace (optional, anywhere between tokens)
+
+Example::
+
+    >>> parse_fields("id,name,ancestors(id,name,source_ref)")
+    {'id': {}, 'name': {}, 'ancestors': {'id': {}, 'name': {}, 'source_ref': {}}}
+    >>> parse_fields("id, name, ancestors( id, name )")
+    {'id': {}, 'name': {}, 'ancestors': {'id': {}, 'name': {}}}
+
+A field with no sub-selector maps to an empty dict. A relation requested without a sub-selector (e.g.
+plain `ancestors`) also maps to `{}`, meaning "use the default sub-fields for that relation". Whitespace
+around commas/parentheses/names is ignored, so a `fields=` value copied from formatted code or a
+comma-space-separated list (`id, name, ...`) works the same as a tightly-packed one.
+"""
+
+from typing import Dict, Tuple
+
+
+class FieldsParseError(ValueError):
+    """Raised when a `fields=` query parameter value doesn't match the expected grammar."""
+
+
+#: common wrong-bracket typos and the correct replacement, so the error message can point people at the
+#: actual grammar instead of just saying "unexpected character" (e.g. `ancestors[id,name]` -> use `(...)`).
+_BRACKET_TYPOS = {"[": "(", "]": ")", "{": "(", "}": ")"}
+
+
+def _bracket_typo_hint(char: str) -> str:
+    replacement = _BRACKET_TYPOS.get(char)
+    if not replacement:
+        return ""
+    return (
+        f" Did you mean {replacement!r} instead of {char!r}? "
+        "Sub-selectors use parentheses, e.g. fields=id,name,ancestors(id,name)."
+    )
+
+
+def parse_fields(fields_param: str) -> Dict[str, dict]:
+    fields_param = (fields_param or "").strip()
+    if not fields_param:
+        return {}
+    tree, index = _parse_field_list(fields_param, 0)
+    index = _skip_ws(fields_param, index)
+    if index != len(fields_param):
+        char = fields_param[index]
+        raise FieldsParseError(
+            f"Unexpected character {char!r} at position {index} in fields={fields_param!r} "
+            f"(expected ',' or '(' here).{_bracket_typo_hint(char)}"
+        )
+    return tree
+
+
+def _skip_ws(text: str, index: int) -> int:
+    while index < len(text) and text[index].isspace():
+        index += 1
+    return index
+
+
+def _parse_field_list(text: str, index: int) -> Tuple[Dict[str, dict], int]:
+    result: Dict[str, dict] = {}
+    while True:
+        index = _skip_ws(text, index)
+        name, index = _parse_name(text, index)
+        index = _skip_ws(text, index)
+        children: Dict[str, dict] = {}
+        if index < len(text) and text[index] == "(":
+            children, index = _parse_field_list(text, index + 1)
+            index = _skip_ws(text, index)
+            if index >= len(text) or text[index] != ")":
+                got = repr(text[index]) if index < len(text) else "end of string"
+                hint = _bracket_typo_hint(text[index]) if index < len(text) else ""
+                raise FieldsParseError(
+                    f"Expected ')' to close {name!r}(...) but got {got} at position {index} in fields={text!r}.{hint}"
+                )
+            index += 1
+            index = _skip_ws(text, index)
+        result[name] = children
+        if index < len(text) and text[index] == ",":
+            index += 1
+            continue
+        break
+    return result, index
+
+
+def _parse_name(text: str, index: int) -> Tuple[str, int]:
+    start = index
+    while index < len(text) and (text[index].isalnum() or text[index] == "_"):
+        index += 1
+    name = text[start:index]
+    if not name:
+        raise FieldsParseError(f"Expected a field name at position {start} in fields={text!r}")
+    return name, index
--- a/iaso/api/v3/common/filterset.py
+++ b/iaso/api/v3/common/filterset.py
@@ -1,0 +1,44 @@
+from typing import FrozenSet
+
+from django_filters import rest_framework as django_filters
+from rest_framework.exceptions import ValidationError
+
+from .param_validator import build_unsupported_params_error
+
+
+#: pagination/output params every v3 list endpoint accepts regardless of what it's listing - not FilterSet
+#: fields, so each subclass's `extra_allowed_params` needs to include these on top of its own shortcuts
+#: (e.g. `search`, `default_version`). Checked for every v3 FilterSet by
+#: `iaso/tests/api/v3/test_filter_nomenclature.py`, so a new endpoint that forgets one of these fails fast.
+CORE_EXTRA_ALLOWED_PARAMS: FrozenSet[str] = frozenset({"order", "format", "page", "page_size", "with_count", "fields"})
+
+
+class BaseV3FilterSet(django_filters.FilterSet):
+    """FilterSet with a "strict mode".
+
+    django-filter silently ignores query params that don't match a declared filter. v3 endpoints instead
+    reject them with a 400 and `difflib`-based suggestions, to catch typos (`dateFrom` instead of
+    `created_at__gte`) early instead of returning a silently-unfiltered result set.
+
+    Subclasses should set `extra_allowed_params` to the view-level shortcut/output params that are valid
+    but aren't declared filters (e.g. `search`, `order`, `page`, `page_size`, `fields`, ...).
+    """
+
+    #: shortcut/output query params handled outside the FilterSet itself (view-level), still accepted.
+    extra_allowed_params: FrozenSet[str] = frozenset()
+
+    def __init__(self, data=None, *args, **kwargs):
+        if data is not None:
+            self.check_params(data)
+        super().__init__(data, *args, **kwargs)
+
+    @classmethod
+    def known_params(cls):
+        return set(cls.base_filters.keys()) | set(cls.extra_allowed_params)
+
+    @classmethod
+    def check_params(cls, data):
+        known = cls.known_params()
+        unknown = [key for key in data.keys() if key not in known]
+        if unknown:
+            raise ValidationError(build_unsupported_params_error(unknown, known))
--- a/iaso/api/v3/common/pagination.py
+++ b/iaso/api/v3/common/pagination.py
@@ -1,0 +1,93 @@
+from rest_framework.exceptions import NotFound
+from rest_framework.response import Response
+
+from iaso.api.common.pagination import Paginator
+
+
+TRUE_VALUES = ("true", "1", "yes")
+
+
+class V3PagePagination(Paginator):
+    """Page-based pagination for v3 endpoints.
+
+    Subclasses `iaso.api.common.pagination.Paginator` (itself a `PageNumberPagination`) so it satisfies
+    `ReadOnlyModelViewSet`'s `pagination_class` type, but overrides every method: unlike `Paginator`, this
+    does not run a mandatory `COUNT(*)` query on every request - by default `count`/`pages` are `null` and
+    `has_next` is derived by fetching one extra row per page. Pass `?with_count=true` to get an exact
+    `count` (one extra COUNT query).
+    """
+
+    page_size = 100
+    max_page_size = 10_000
+    page_size_query_param = "page_size"
+    results_key = "results"
+
+    def get_results_key(self):
+        view = self.request.parser_context.get("view") if getattr(self, "request", None) else None
+        return getattr(view, "results_key", None) or self.results_key
+
+    def with_count(self, request):
+        return str(request.query_params.get("with_count", "")).lower() in TRUE_VALUES
+
+    def paginate_queryset(self, queryset, request, view=None):
+        self.request = request
+        page_size = self.get_page_size(request)
+        if not page_size:
+            return None
+
+        try:
+            page_number = int(request.query_params.get(self.page_query_param, 1))
+        except (TypeError, ValueError):
+            page_number = 1
+        if page_number < 1:
+            raise NotFound(f"Invalid page number: must be >= 1, got {request.query_params.get(self.page_query_param)}")
+
+        offset = (page_number - 1) * page_size
+        # Fetch one row past the page to know if there's a next page, without a COUNT(*) query.
+        rows = list(queryset[offset : offset + page_size + 1])
+
+        if not rows and page_number > 1:
+            raise NotFound(f"Page {page_number} is out of range (no results at this offset)")
+
+        self.page_number = page_number
+        self.page_size = page_size
+        self.has_next_page = len(rows) > page_size
+        self.count = queryset.count() if self.with_count(request) else None
+        self.num_pages = -(-self.count // page_size) if self.count is not None else None  # ceil division
+
+        return rows[:page_size]
+
+    def get_paginated_response(self, data):
+        return Response(
+            {
+                "count": self.count,
+                self.get_results_key(): data,
+                "has_next": self.has_next_page,
+                "has_previous": self.page_number > 1,
+                "page": self.page_number,
+                "pages": self.num_pages,
+                "page_size": self.page_size,
+            }
+        )
+
+    def get_paginated_response_schema(self, schema):
+        return {
+            "type": "object",
+            "properties": {
+                "count": {
+                    "type": "integer",
+                    "nullable": True,
+                    "description": "Exact total, only computed when `with_count=true` was passed.",
+                },
+                "has_next": {"type": "boolean"},
+                "has_previous": {"type": "boolean"},
+                "page": {"type": "integer", "example": 1},
+                "pages": {
+                    "type": "integer",
+                    "nullable": True,
+                    "description": "Only computed when `with_count=true` was passed.",
+                },
+                "page_size": {"type": "integer", "example": 100},
+                self.get_results_key(): schema,
+            },
+        }
--- a/iaso/api/v3/common/param_validator.py
+++ b/iaso/api/v3/common/param_validator.py
@@ -1,0 +1,30 @@
+"""Strict query-param validation shared by v3 endpoints.
+
+v3 endpoints reject unknown query parameters instead of silently ignoring them (django-filter's
+default behaviour), and suggest the closest known param name(s) using the stdlib `difflib` module -
+no fuzzy-matching dependency needed for typo-level suggestions.
+"""
+
+import difflib
+
+from typing import Dict, Iterable, List
+
+
+def suggest_close_matches(unknown_param: str, known_params: Iterable[str], max_suggestions: int = 3) -> List[str]:
+    """Return up to `max_suggestions` known param names that look like `unknown_param`."""
+    return difflib.get_close_matches(unknown_param, known_params, n=max_suggestions, cutoff=0.5)
+
+
+def build_unsupported_params_error(unknown_params: Iterable[str], known_params: Iterable[str]) -> Dict:
+    """Build the `{"error": ..., "suggestions": ...}` payload for an "unsupported query parameter" 400."""
+    known_params = sorted(known_params)
+    unknown_params = sorted(unknown_params)
+    suggestions = {}
+    for param in unknown_params:
+        matches = suggest_close_matches(param, known_params)
+        if matches:
+            suggestions[param] = matches
+    return {
+        "error": f"Unsupported query parameter(s): {', '.join(unknown_params)}",
+        "suggestions": suggestions,
+    }
--- a/iaso/api/v3/common/spatial_filters.py
+++ b/iaso/api/v3/common/spatial_filters.py
@@ -1,0 +1,100 @@
+"""Spatial filters shared by v3 endpoints.
+
+Core subset only (per product decision): bounding-box intersection (`__bbox`) and containment inside/
+outside another org unit's geometry (`__within_org_unit`/`__outside_org_unit`). Point-in-arbitrary-polygon
+(`__contains`) and radius search (`__near`) are still deferred to a follow-up.
+"""
+
+import django_filters
+
+from django.contrib.gis.geos import Polygon
+from django.core.exceptions import ValidationError as DjangoValidationError
+from rest_framework.exceptions import ValidationError
+
+
+def parse_bbox(value: str) -> Polygon:
+    """Parse a `minx,miny,maxx,maxy` string into a GEOS Polygon (SRID 4326)."""
+    parts = value.split(",")
+    if len(parts) != 4:
+        raise ValidationError(f"Invalid bbox value {value!r}: expected 'minx,miny,maxx,maxy'")
+    try:
+        minx, miny, maxx, maxy = (float(part) for part in parts)
+    except ValueError:
+        raise ValidationError(f"Invalid bbox value {value!r}: all 4 values must be numbers")
+    try:
+        polygon = Polygon.from_bbox((minx, miny, maxx, maxy))
+    except DjangoValidationError as e:
+        raise ValidationError(f"Invalid bbox value {value!r}: {e}")
+    polygon.srid = 4326
+    return polygon
+
+
+def resolve_reference_geometry(org_unit_model, org_unit_id: str, operator_name: str):
+    """Fetch the geometry (`simplified_geom`, falling back to `geom`) of the org unit referenced by a
+    `*__within_org_unit` filter, raising the spec'd 400 error if it has neither."""
+    try:
+        reference = org_unit_model.objects.only("id", "geom", "simplified_geom").get(pk=org_unit_id)
+    except (org_unit_model.DoesNotExist, ValueError, TypeError):
+        raise ValidationError(f"Org unit {org_unit_id!r} does not exist")
+
+    geometry = reference.simplified_geom or reference.geom
+    if geometry is None:
+        raise ValidationError(
+            {
+                "error": f"Org unit {org_unit_id} has no geometry",
+                "detail": (
+                    f"Operator {operator_name!r} requires the referenced org unit to have a geom or simplified_geom."
+                ),
+            }
+        )
+    return geometry
+
+
+class BboxFilter(django_filters.CharFilter):
+    """`<field>__bbox=minx,miny,maxx,maxy` -> keep rows whose geometry intersects the bounding box."""
+
+    def __init__(self, *args, geometry_field: str, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.geometry_field = geometry_field
+
+    def filter(self, qs, value):
+        if value in (None, ""):
+            return qs
+        polygon = parse_bbox(value)
+        return qs.filter(**{f"{self.geometry_field}__intersects": polygon})
+
+
+class _OrgUnitContainmentFilter(django_filters.CharFilter):
+    """Shared base for `within_org_unit`/`outside_org_unit`: both resolve the same reference geometry,
+    they just apply it as a `filter()` vs. an `exclude()`."""
+
+    def __init__(self, *args, geometry_field: str, org_unit_model, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.geometry_field = geometry_field
+        self.org_unit_model = org_unit_model
+
+
+class WithinOrgUnitFilter(_OrgUnitContainmentFilter):
+    """`<field>__within_org_unit=<org unit id>` -> keep rows whose geometry is contained within the
+    referenced org unit's geometry."""
+
+    def filter(self, qs, value):
+        if value in (None, ""):
+            return qs
+        reference_geometry = resolve_reference_geometry(self.org_unit_model, value, self.field_name)
+        return qs.filter(**{f"{self.geometry_field}__within": reference_geometry})
+
+
+class OutsideOrgUnitFilter(_OrgUnitContainmentFilter):
+    """`<field>__outside_org_unit=<org unit id>` -> keep rows that HAVE a `geometry_field` value but it's
+    NOT contained within the referenced org unit's geometry (e.g. a district's org units whose recorded
+    location falls outside the district's own shape - typically a data quality issue). Rows where
+    `geometry_field` is null are excluded rather than treated as "outside": there's nothing to compare."""
+
+    def filter(self, qs, value):
+        if value in (None, ""):
+            return qs
+        reference_geometry = resolve_reference_geometry(self.org_unit_model, value, self.field_name)
+        return qs.filter(**{f"{self.geometry_field}__isnull": False}).exclude(
+            **{f"{self.geometry_field}__within": reference_geometry}
+        )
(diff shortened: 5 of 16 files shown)


iaso/api/v3/common/fields_parser.py
```diff
@@ -0,0 +1,99 @@
+"""Parser for the v3 `fields=` query parameter sub-selector grammar.
```
tdethier, line 1: we should add unit tests for all helpers in this file

iaso/api/v3/common/filterset.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+#: pagination/output params every v3 list endpoint accepts regardless of what it's listing - not FilterSet
+#: fields, so each subclass's `extra_allowed_params` needs to include these on top of its own shortcuts
+#: (e.g. `search`, `default_version`). Checked for every v3 FilterSet by
+#: `iaso/tests/api/v3/test_filter_nomenclature.py`, so a new endpoint that forgets one of these fails fast.
+CORE_EXTRA_ALLOWED_PARAMS: FrozenSet[str] = frozenset({"order", "format", "page", "page_size", "with_count", "fields"})
+
+
+class BaseV3FilterSet(django_filters.FilterSet):
```
tdethier, line 18: i would also add unit tests to check the behavior of this class, with filters that inherit from it and define `extra_allowed_params` or not

iaso/api/v3/common/param_validator.py
```diff
@@ -0,0 +1,30 @@
...
+default behaviour), and suggest the closest known param name(s) using the stdlib `difflib` module -
+no fuzzy-matching dependency needed for typo-level suggestions.
+"""
+
+import difflib
+
+from typing import Dict, Iterable, List
+
+
+def suggest_close_matches(unknown_param: str, known_params: Iterable[str], max_suggestions: int = 3) -> List[str]:
```
tdethier, line 13: very good idea 👍

iaso/api/auth/authentication.py
```diff
@@ -1,6 +1,21 @@
...
+from drf_spectacular.extensions import OpenApiAuthenticationExtension
 from rest_framework.authentication import SessionAuthentication
 
 
 class CsrfExemptSessionAuthentication(SessionAuthentication):
     def enforce_csrf(self, request):
         return  # To not perform the csrf check previously happening
+
+
+class CsrfExemptSessionAuthenticationScheme(OpenApiAuthenticationExtension):
```
tdethier, line 11: not sure why we have this, was it needed for swagger?

iaso/api/v3/common/spatial_filters.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+class _OrgUnitContainmentFilter(django_filters.CharFilter):
+    """Shared base for `within_org_unit`/`outside_org_unit`: both resolve the same reference geometry,
+    they just apply it as a `filter()` vs. an `exclude()`."""
+
+    def __init__(self, *args, geometry_field: str, org_unit_model, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.geometry_field = geometry_field
+        self.org_unit_model = org_unit_model
```
tdethier: not sure why you need to pass the `OrgUnit` model here, if you're checking `geom`  and `simplified_geom`, you already know that it's this model and you don't need to pass it? otherwise it sounds like we have multiple models that store geo data, with the same attribute names, but i don't think we do

iaso/api/v3/common/spatial_filters.py
```diff
@@ -0,0 +1,121 @@
...
+        raise ValidationError(f"Invalid bbox value {value!r}: {e}")
+    polygon.srid = 4326
+    return polygon
+
+
+def resolve_reference_geometry(org_unit_model, org_unit_id: str, operator_name: str):
+    """Fetch the geometry (`simplified_geom`, falling back to `geom`) of the org unit referenced by a
+    `*__within_org_unit` filter, raising the spec'd 400 error if it has neither."""
+    try:
+        reference = org_unit_model.objects.only("id", "geom", "simplified_geom").get(pk=org_unit_id)
```
tdethier: i'm not sure why we don't do `OrgUnit.objects.only(...)` here

iaso/api/v3/common/spatial_filters.py
```diff
@@ -0,0 +1,121 @@
...
+class _BboxFilterBase(django_filters.CharFilter):
+    """Shared base for `bbox`/`outside_bbox`: both parse the same `minx,miny,maxx,maxy` box, they just
+    apply it as a `filter()` (intersects) vs. an `exclude()` (does not intersect)."""
+
+    def __init__(self, *args, geometry_field: str, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.geometry_field = geometry_field
+
+
+class BboxFilter(_BboxFilterBase):
```
tdethier: nitpick: i would maybe call it `WithinBboxFilter` since you also have `Within/OutsideOrgUnit` filters below

mestachs:
nice tool to test the api "validations"

```
TOKEN=$(curl -s -X POST http://localhost:8081/api/token/ \
  -H "Content-Type: application/json" \
  -d '{"username":"testemailstable-2-40-12","password":"testemailstable-2-40-12"}' \
  | jq -r .access)

uvx schemathesis run \
  "http://localhost:8081/swagger/?format=json" \
  --header "Authorization: Bearer $TOKEN" \
  --include-path-regex '^/api/v3/orgunits' \
  -c not_a_server_error \
  -n 300 \
  --report json,junit \
  --report-dir ./schemathesis-report
```

iaso/api/v3/org_units/filters.py
```diff
@@ -0,0 +1,174 @@
...
+    OutsideBboxFilter,
+    OutsideOrgUnitFilter,
+    WithinOrgUnitFilter,
+)
+from iaso.models import OrgUnit, OrgUnitType
+
+
+#: query params handled at the view level (search shortcut, output params) rather than as FilterSet fields:
+#: the shared core (pagination/format/fields, see CORE_EXTRA_ALLOWED_PARAMS) plus org-units-specific ones.
+EXTRA_ALLOWED_PARAMS = CORE_EXTRA_ALLOWED_PARAMS | frozenset(
```
tdethier: maybe this merge should be done at the base level and we could simply list extra params at the children classes level

iaso/api/v3/org_units/filters.py
```diff
@@ -0,0 +1,174 @@
...
+    {
+        "search",
+        "default_version",
+        "roots_for_user",
+        "extra_fields",
+    }
+)
+
+
+class NumberInFilter(django_filters.BaseInFilter, django_filters.NumberFilter):
```
tdethier: we already had this in the codebase,  i would delete the other one, or call the other one in this file instead of repeating it here

iaso/api/v3/common/pagination.py
```diff
@@ -0,0 +1,111 @@
...
+        rows = list(queryset[offset : offset + page_size + 1])
+
+        if not rows and page_number > 1:
+            raise NotFound(f"Page {page_number} is out of range (no results at this offset)")
+
+        self.page_number = page_number
+        self.page_size = page_size
+        self.has_next_page = len(rows) > page_size
+        self.count = queryset.count() if self.with_count(request) else None
+        self.num_pages = -(-self.count // page_size) if self.count is not None else None  # ceil division
```
tdethier, line 77: i hope all of this is thoroughly tested 😅

iaso/api/v3/org_units/filters.py
```diff
@@ -0,0 +1,174 @@
...
+        field_name="org_unit_type__category", choices=OrgUnitType.CATEGORIES
+    )
+    parent__name__icontains = django_filters.CharFilter(field_name="parent__name", lookup_expr="icontains")
+    parent__source_ref = django_filters.CharFilter(field_name="parent__source_ref", lookup_expr="exact")
+
+    # -- dates --
+    created_at__gte = SafeIsoDateTimeFilter(field_name="created_at", lookup_expr="gte")
+    created_at__lte = SafeIsoDateTimeFilter(field_name="created_at", lookup_expr="lte")
+    updated_at__gte = SafeIsoDateTimeFilter(field_name="updated_at", lookup_expr="gte")
+    updated_at__lte = SafeIsoDateTimeFilter(field_name="updated_at", lookup_expr="lte")
```
tdethier: we should probably have the same for `source_created_at`

iaso/api/v3/org_units/filters.py
```diff
@@ -0,0 +1,174 @@
...
+    # `ancestor_id=X` (all descendants, any depth) vs. `ancestor_id__direct_children=X` (one level down
+    # only - a plain `parent_id` equality under the hood, no ltree scan needed for this variant).
+    ancestor_id = django_filters.NumberFilter(method="filter_ancestor_id")
+    ancestor_id__direct_children = django_filters.NumberFilter(field_name="parent_id")
+    depth = django_filters.NumberFilter(field_name="path", lookup_expr="depth")
+
+    # -- spatial (core subset) --
+    geom__bbox = BboxFilter(geometry_field="geom")
+    simplified_geom__bbox = BboxFilter(geometry_field="simplified_geom")
+    location__bbox = BboxFilter(geometry_field="location")
```
tdethier: same comment about using `inside/within`

iaso/api/v3/org_units/serializers.py
```diff
@@ -0,0 +1,673 @@
...
+        "fixed_subfields": list(CREATOR_SUBFIELDS),
+    }
+    fields["version"] = {
+        "default": False,
+        "sub_selector": True,
+        "default_subfields": list(VERSION_DEFAULT_SUBFIELDS),
+        "allowed_subfields": sorted(VERSION_ALLOWED_SUBFIELDS),
+        "nested": {"data_source": {"fixed_subfields": list(DATA_SOURCE_SUBFIELDS)}},
+    }
+    assert set(fields) == ALLOWED_TOP_LEVEL_FIELDS, "build_fields_schema() drifted from ALLOWED_TOP_LEVEL_FIELDS"
```
tdethier: i find it a bit strange to have this `assert` here in the API

iaso/api/v3/org_units/serializers.py
```diff
@@ -0,0 +1,673 @@
...
+            f"Allowed fields: {', '.join(sorted(ALLOWED_TOP_LEVEL_FIELDS))}",
+        )
+    _reject_unknown_subfields(field_tree, "ancestors", ANCESTOR_ALLOWED_SUBFIELDS)
+    _reject_unknown_subfields(field_tree, "parent", PARENT_ALLOWED_SUBFIELDS)
+    _reject_fixed_shape_sub_selector(field_tree, "org_unit_type", ORG_UNIT_TYPE_SUBFIELDS)
+    _reject_fixed_shape_sub_selector(field_tree, "creator", CREATOR_SUBFIELDS)
+
+    version_subfields = _reject_unknown_subfields(field_tree, "version", VERSION_ALLOWED_SUBFIELDS)
+    if version_subfields and version_subfields.get("data_source"):
+        raise bad_request(
```
tdethier: i don't get this check, `data_source` is in `VERSION_ALLOWED_SUBFIELDS`, why would this fail?

iaso/api/v3/org_units/serializers.py
```diff
@@ -0,0 +1,673 @@
...
+
+def serialize_org_units(org_units, field_tree: Optional[Dict[str, dict]]) -> List[dict]:
+    """Build the list of response dicts for a page (or export batch) of `OrgUnit` instances.
+
+    `field_tree` is the parsed `fields=` selector (or `None` for the default field set). Geometry fields,
+    `ancestors`, `parent`, `org_unit_type`, `creator` and `version` are only computed when requested, each
+    in a single batched query for the whole `org_units` collection - no N+1 regardless of page size or
+    tree depth.
+    """
+    org_units = list(org_units)
```
tdethier: `org_units` is a queryset i guess?

iaso/api/v3/org_units/serializers.py
```diff
@@ -0,0 +1,673 @@
...
+
+    geometry_fields = [f for f in scalar_fields if f in GEOMETRY_FIELDS]
+    group_fields = [f for f in scalar_fields if f in GROUP_FIELDS]
+    include_org_unit_type = "org_unit_type" in scalar_fields
+    include_creator = "creator" in scalar_fields
+    plain_fields = [
+        f
+        for f in scalar_fields
+        if f not in GEOMETRY_FIELDS and f not in GROUP_FIELDS and f not in ("org_unit_type", "creator")
+    ]
```
tdethier: nitpick: i would probably loop once over `scalar_fields` and populate all 3 lists with `if` inside the loop, but this works

iaso/api/v3/org_units/serializers.py
```diff
@@ -0,0 +1,673 @@
...
+        path = _field(row, "path")
+        return len(path) if path else None
+    return _field(row, field)
+
+
+def _batch_geojson(org_units: List[OrgUnit], geometry_field: str) -> Dict[int, Optional[dict]]:
+    ids = [unit.id for unit in org_units]
+    if not ids:
+        return {}
+    return _batch_geojson_for_queryset(OrgUnit.objects.filter(id__in=ids), geometry_field)
```
tdethier: so to get geo data, we need to have a queryset, turn it into a list, get ID from every element in the list, then make a new queryset based on IDs? can we get geo data directly from the initial queryset?

iaso/api/v3/org_units/serializers.py
```diff
@@ -0,0 +1,673 @@
...
+    return {feature["id"]: feature["geometry"] for feature in result["features"]}
+
+
+def _batch_group_ids(org_units: List[OrgUnit]) -> Dict[int, List[int]]:
+    """One query for the whole collection instead of one `.groups.values_list(...)` per org unit."""
+    ids = [unit.id for unit in org_units]
+    result: Dict[int, List[int]] = {unit.id: [] for unit in org_units}
+    if not ids:
+        return result
+    pairs = Group.objects.filter(org_units__id__in=ids).values_list("org_units__id", "id")
```
tdethier: i would probably add an `.order_by("id")`

iaso/api/v3/org_units/serializers.py
```diff
@@ -0,0 +1,673 @@
...
+
+def _batch_groups(org_units: List[OrgUnit]) -> Dict[int, List[dict]]:
+    """Same as `_batch_group_ids` but with `{id, name}` objects - one query for the whole collection."""
+    ids = [unit.id for unit in org_units]
+    result: Dict[int, List[dict]] = {unit.id: [] for unit in org_units}
+    if not ids:
+        return result
+    rows = Group.objects.filter(org_units__id__in=ids).values_list("org_units__id", "id", "name")
+    for org_unit_id, group_id, group_name in rows:
+        result.setdefault(org_unit_id, []).append({"id": group_id, "name": group_name})
```
tdethier: so if we want to add another field in groups, there are now 3 places where we need to add the field:
- the serializer
- the white list of fields
- here

iaso/api/v3/org_units/serializers.py
```diff
@@ -0,0 +1,673 @@
...
+        for ancestor in OrgUnit.objects.filter(id__in=all_ancestor_ids).only("id", *subfields):
+            ancestors_by_id[ancestor.id] = {"id": ancestor.id, **{f: getattr(ancestor, f) for f in subfields}}
+
+    result: Dict[int, List[dict]] = {}
+    for unit in org_units:
+        if not unit.path:
+            result[unit.id] = []
+            continue
+        result[unit.id] = [
+            ancestors_by_id[int(ancestor_id)] for ancestor_id in unit.path[:-1] if int(ancestor_id) in ancestors_by_id
```
tdethier: i find this really hard to read

iaso/api/v3/org_units/serializers.py
```diff
@@ -0,0 +1,673 @@
...
+def _batch_versions(version_ids: List[Optional[int]], requested_subfields: Optional[dict]) -> Dict[int, dict]:
+    plain_subfields = _version_plain_subfields(requested_subfields)
+    # `data_source_id` is needed internally to expand `data_source`, even if it wasn't itself requested.
+    fetch_subfields = list({*plain_subfields, "data_source_id"})
+    version_map = _batch_single_fk(version_ids, SourceVersion, fetch_subfields)
+    _attach_data_sources(version_map, requested_subfields)
+    if "data_source_id" not in plain_subfields:
+        for version in version_map.values():
+            version.pop("data_source_id", None)
+    return version_map
```
tdethier: it feels we are manually rebuilding graphql in a REST endpoint, i'm not a fan

iaso/api/v3/org_units/serializers.py
```diff
@@ -0,0 +1,673 @@
...
+        else None
+    )
+    creator_map = (
+        _batch_single_fk_for_queryset(queryset, "creator_id", User, list(CREATOR_SUBFIELDS))
+        if include_creator
+        else None
+    )
+    version_map = _batch_versions_for_queryset(queryset, version_subfields) if version_subfields is not None else None
+
+    def get_value(row, field: str):
```
tdethier: i'm not a fan of these inner functions

iaso/api/v3/org_units/serializers.py
```diff
@@ -0,0 +1,673 @@
...
+            columns.add("parent_id")
+        elif field == "creator":
+            columns.add("creator_id")
+        elif field == "version":
+            columns.add("version_id")
+        else:
+            columns.add(field)
+    if include_ancestors:
+        columns.add("path")
+    return sorted(columns)
```
tdethier: does this handle the column order inside the exported file?

iaso/api/v3/org_units/views.py
```diff
@@ -0,0 +1,666 @@
...
+    "location_geojson",
+    "simplified_geom_geojson",
+    "biggest_polygon_geojson",
+    "groups_exploded",
+    "groups_exploded_code",
+    "groups_json",
+    ":all",
+]
+
+V3_ORG_UNIT_PARAMETERS = [
```
tdethier: can't this be automatically computed from `OrgUnitFilterSetV3`?

iaso/api/v3/org_units/views.py
```diff
@@ -0,0 +1,666 @@
...
+    ),
+]
+
+V3_ORG_UNIT_PARAMETERS_BY_NAME = {parameter.name: parameter for parameter in V3_ORG_UNIT_PARAMETERS}
+
+
+def _flatten_export_cell(value):
+    """Tabular exports (csv/xlsx) can't nest: a list becomes a `;`-separated string, and a `{id, name}`
+    object (one entry of the default `groups` field) is flattened to just the name. Every other
+    non-scalar field (`ancestors(...)`, `org_unit_type`, `parent`) is split into its own columns instead
```
tdethier: why does `groups` behave differently from the others?

iaso/api/v3/org_units/views.py
```diff
@@ -0,0 +1,666 @@
...
+        return queryset
+
+    def _parse_and_validate_fields(self, request):
+        fields_param = request.query_params.get("fields")
+        try:
+            field_tree = parse_fields(fields_param) if fields_param else None
+        except FieldsParseError as e:
+            raise bad_request("Invalid fields= parameter", str(e))
+        validate_field_tree(field_tree)
+        return field_tree
```
tdethier: this should probably be part of a base common view class

iaso/tests/api/v3/test_org_units.py
```diff
@@ -0,0 +1,948 @@
...
+        self.client.force_authenticate(None)
+        response = self.client.get(BASE_URL)
+        self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
+
+    def test_user_from_another_account_sees_only_their_own_account(self):
+        self.client.force_authenticate(self.other_account_user)
+        response = self.client.get(BASE_URL)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        ids = {r["id"] for r in response.json()["results"]}
+        self.assertEqual(ids, {self.marvel_org_unit.id, self.marvel_org_unit_without_geom.id})
```
tdethier: i would use `assertCountEqual` here, to avoid flaky tests due to ordering

iaso/tests/api/v3/test_org_units.py
```diff
@@ -0,0 +1,948 @@
...
+
+    def test_within_org_unit_rejects_org_unit_outside_user_scope(self):
+        # `self.marvel_org_unit` belongs to a different account and DOES have a geom - without the scoped
+        # lookup in `resolve_reference_geometry`, this would succeed (200) using a foreign account's
+        # geometry as the reference shape, leaking that it exists and has a geometry.
+        response = self.client.get(BASE_URL, {"location__within_org_unit": self.marvel_org_unit.id})
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertIn("does not exist", response.json()["error"])
```
tdethier: there's no need to do `response.json()` every time: we have `result = self.assertJSONResponse(response, status.HTTP_XXX)`

iaso/tests/api/v3/test_org_units.py
```diff
@@ -0,0 +1,948 @@
...
+        self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
+
+    def test_user_from_another_account_sees_only_their_own_account(self):
+        self.client.force_authenticate(self.other_account_user)
+        response = self.client.get(BASE_URL)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        ids = {r["id"] for r in response.json()["results"]}
+        self.assertEqual(ids, {self.marvel_org_unit.id, self.marvel_org_unit_without_geom.id})
+
+    def test_within_org_unit_rejects_org_unit_outside_user_scope(self):
```
tdethier: i would move all tests related to filtering to a `test_filters.py` file

iaso/tests/api/v3/test_org_units.py
```diff
@@ -0,0 +1,948 @@
...
+        self.assertEqual(data["fields"]["name"], {"default": True})
+        self.assertEqual(data["fields"]["geom"], {"default": False, "shape": "GeoJSON"})
+        ancestors_schema = data["fields"]["ancestors"]
+        self.assertEqual(ancestors_schema["default"], False)
+        self.assertEqual(ancestors_schema["sub_selector"], True)
+        self.assertIn("source_ref", ancestors_schema["allowed_subfields"])
+
+    # -- FilterSet fields --
+
+    def test_filter_id_exact(self):
```
tdethier: i would still check for HTTP result code, especially now that wrong params throw errors

iaso/tests/api/v3/test_org_units.py
```diff
@@ -0,0 +1,948 @@
...
+        response = self.client.get(BASE_URL, {"geom__bbox": "100,100,200,200"})
+        self.assertEqual(response.json()["results"], [])
+
+    def test_filter_location_bbox_matches(self):
+        response = self.client.get(BASE_URL, {"location__bbox": "0,0,10,10"})
+        self.assertEqual([r["id"] for r in response.json()["results"]], [self.region.id])
+
+    def test_filter_invalid_bbox(self):
+        response = self.client.get(BASE_URL, {"geom__bbox": "not-a-bbox"})
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
```
tdethier: nitpick: i would also check the error message to make sure that the right error is triggered

iaso/tests/api/v3/test_org_units.py
```diff
@@ -0,0 +1,948 @@
...
+                "location__outside_org_unit": self.country.id,
+                "fields": "id,name,latitude,longitude,ancestors(id,name)",
+            },
+        )
+        results = response.json()["results"]
+        self.assertEqual([r["id"] for r in results], [misplaced.id])
+        self.assertEqual(results[0]["latitude"], 50.0)
+        self.assertEqual([a["name"] for a in results[0]["ancestors"]], ["Naboo", "Theed", "Theed District"])
+
+    # -- strict param validation --
```
tdethier: i would also put this in `test_filters.py`? or maybe another test file

iaso/tests/api/v3/test_org_units.py
```diff
@@ -0,0 +1,948 @@
...
+        self.assertIn("source_ref", data["suggestions"]["source_reff"])
+
+    def test_unknown_query_param_without_close_match_has_no_suggestions(self):
+        response = self.client.get(BASE_URL, {"dateFrom": "2020-01-01"})
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        data = response.json()
+        self.assertIn("dateFrom", data["error"])
+        self.assertEqual(data["suggestions"], {})
+
+    # -- fields= selector --
```
tdethier: same comment about splitting

tdethier: i think all of this should be covered by unit tests calling all helpers instead

iaso/tests/api/v3/test_org_units.py
```diff
@@ -0,0 +1,948 @@
...
+    def test_fields_param_geometry_is_opt_in(self):
+        response = self.client.get(
+            BASE_URL,
+            {"fields": "id,geom"},
+        )
+        row = next(r for r in response.json()["results"] if r["id"] == self.country.id)
+        self.assertIsNotNone(row["geom"])
+        self.assertEqual(row["geom"]["type"], "MultiPolygon")
+
+    # -- pagination --
```
tdethier: same comment about splitting the file

iaso/tests/api/v3/test_org_units.py
```diff
@@ -0,0 +1,948 @@
...
+
+    def test_format_typo_is_a_clear_400_not_drfs_bare_404(self):
+        # DRF's own content negotiation reacts to an unrecognized `?format=` value with a bare
+        # `Http404("Not Found")` - confusing for "you typoed a query param value". Caught in
+        # `OrgUnitViewSetV3.initial()` instead.
+        response = self.client.get(BASE_URL, {"format": "parqet"})
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        data = response.json()
+        self.assertIn("parqet", data["error"])
+        self.assertIn("parquet", data["detail"])
```
tdethier: does the `detail` section suggest another spelling? i'm not sure of what i'm looking at in this test

iaso/tests/api/v3/test_org_units.py
```diff
@@ -0,0 +1,948 @@
...
+
+    # -- fields= selector --
+
+    def test_fields_param_flat_subset(self):
+        response = self.client.get(BASE_URL, {"fields": "id,name"})
+        row = response.json()["results"][0]
+        self.assertEqual(set(row.keys()), {"id", "name"})
+
+    def test_fields_param_allows_whitespace_around_commas_and_parens(self):
+        response = self.client.get(BASE_URL, {"fields": "id, name, ancestors( id, name )", "name": "Theed District"})
```
tdethier: in all filtering tests, i would use values from org unit objects instead of hardcoded strings (`"Theed District"`)

iaso/tests/api/v3/test_org_units.py
```diff
@@ -0,0 +1,948 @@
...
+
+
+class OrgUnitV3APITestCase(APITestCase):
+    """Covers `/api/v3/orgunits/`: default fields, FilterSet fields, hierarchy/spatial filters, the
+    `fields=` selector, pagination and permissions. CSV/XLSX exports are smoke-tested here; the parquet
+    export (which needs committed data visible to DuckDB) is covered by `OrgUnitV3ParquetTestCase` below."""
+
+    @classmethod
+    def setUpTestData(cls):
+        cls.star_wars = star_wars = m.Account.objects.create(name="Star Wars")
```
tdethier: you're adding more star wars into the codebase? 😇

iaso/tests/api/v3/test_org_units.py
```diff
@@ -0,0 +1,948 @@
...
+                "ancestors[0].validation_status",
+                "ancestors[1].id",
+                "ancestors[1].name",
+                "ancestors[1].validation_status",
+            ],
+        )
+        row = next(reader)
+        self.assertEqual(row["name"], "Theed District")
+        self.assertEqual(row["ancestors[0].name"], "Naboo")  # root
+        self.assertEqual(row["ancestors[1].name"], "Theed")  # immediate parent
```
tdethier: here as well, i would assert with values from OrgUnit objects from the setup instead of hardcoding values with comments

iaso/tests/api/v3/test_org_units.py
```diff
@@ -0,0 +1,948 @@
...
+        # DRF's own content negotiation reacts to an unrecognized `?format=` value with a bare
+        # `Http404("Not Found")` - confusing for "you typoed a query param value". Caught in
+        # `OrgUnitViewSetV3.initial()` instead.
+        response = self.client.get(BASE_URL, {"format": "parqet"})
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        data = response.json()
+        self.assertIn("parqet", data["error"])
+        self.assertIn("parquet", data["detail"])
+
+    def test_format_csv(self):
```
tdethier: for all csv tests: there's a `self.assertCsvFileResponse()` utility you can use

iaso/tests/api/v3/test_org_units.py
```diff
@@ -0,0 +1,948 @@
...
+        response = self.client.get(BASE_URL, {"format": "csv"})
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response["Content-Type"], "text/csv")
+        content = b"".join(response.streaming_content).decode("utf-8")
+        self.assertIn("Theed", content)
+
+    def test_format_xlsx(self):
```
tdethier: and for all xlsx tests, there's a `self.assertXlsxFileResponse()` utility you can use

iaso/tests/api/v3/test_org_units.py
```diff
@@ -0,0 +1,948 @@
...
+            response = self.client.get(BASE_URL, {"fields": "id,name,depth,org_unit_type"})
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        # the main org unit page-fetch query specifically - not the `org_unit_type` batch query (which
+        # legitimately selects FROM iaso_orgunittype, with an `iaso_orgunit` subquery) or any other one.
+        main_query = next(
+            q["sql"] for q in ctx.captured_queries if q["sql"].strip().startswith('SELECT "iaso_orgunit".')
+        )
+        # `LEFT OUTER JOIN` is what `select_related` would add; the account-scoping subquery
+        # (`filter_for_account`) legitimately uses `INNER JOIN` in its own nested SELECT, which is fine.
+        self.assertNotIn("LEFT OUTER JOIN", main_query)
```
tdethier: are you sure that the first query with `'SELECT "iaso_orgunit".'` will be the right one? are they always in the right order? can't there be another query on this table before the "big" query with select/prefetch related?

iaso/tests/api/v3/test_org_units.py
```diff
@@ -0,0 +1,948 @@
...
+        )
+        self.user = self.create_user_with_profile(username="padme", account=self.star_wars)
+        self.client.force_authenticate(self.user)
+
+    def test_format_parquet(self):
+        response = self.client.get(BASE_URL, {"format": "parquet"})
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assert_parquet_content_type(response)
+        with open("/tmp/v3_orgunits_test.parquet", "wb") as f:
+            write_response_to_file(response, f)
```
tdethier: no assert after writing the file? what's the point of writing the file then?

tdethier (commented):
general comments in any order:
- this is a great starting point, thanks for the huge work 👍 
- i like that the API is aware of all valid params and can refuse unknown ones instead of ignoring them 
- i like that the API can suggest valid options in case of spelling mistakes for instance
- i don't like that adding a new field somewhere is much harder - the same field has to be added in:
  - a serializer (which is ok)
  - the whist list of fields (would it be possible to not make them and generate the white list based off serializer attributes instead?)
  - in `batch_xxx` methods to actually fetch data
  - in `batch_xxx_for_queryset` methods to fetch data for exports
  - in `only_columns_for_field_tree` (i'm not sure why)
  - maybe in other places as well? this sounds very complicated to maintain
- i don't like how csv/xlsx exports are handled: most of the code is more or less duplicated, it's gonna be really hard to maintain + it reuses the same old "get_row" logic with inner functions like old endpoints
- we should have more tests
- the API test file (`test_org_units.py`) should be split in test_views/filters/pagination/... like we usually do
- i didn't check if there were features/fields/... that exist in the current API but are missing in this new version
- we should handle the security alerts that github raised, they make sense
- we should get inputs from other team members and decide on how we go from there - personally, i like the added features, but i think the maintenance cost is too high compared to their added value (at least in the current version); getting rid of some new features, or implementing them differently might be better? i'm not sure

iaso/api/auth/authentication.py
mestachs: it's to be able to test the schema generation : 
iaso/tests/api/v3/test_openapi_schema.py::V3SchemaDocumentationTestCase::test_v3_endpoints_generate_a_clean_openapi_schema

iaso/tests/api/v3/test_org_units.py
mestachs: to be honest I fixed it but I don't like that... it finally feels your "test" is too smart and don't detect anything (like playing with mock, and you don't even realize that you test your mock and not the implementation)

PR #3310, opened 2026-09-07
Console-281: Account usage feature

Phil-V:
Builds on PR #3133 

Move the quota logic to the SaaS plugin, address some review comments and fix some bugs.

### Related JIRA tickets

Console-281

### Related Pull Requests

https://github.com/BLSQ/saas/pull/16 https://github.com/BLSQ/console/pull/162

## Changes

Most of the changes reside in the other (saas plugin) PR. The changes in this PR revolve around bug fixes or making the test suite pass when the Saas signal receivers are active. 

- Moved the quota logic to the SaaS plugin.
- Set the Instance.project field on creation so that it's immediately accessible to post_save signal listeners.
- Fix an error in our task runner when using `transaction.on_commit` hooks. 
- Add a signal for bulk user creation for the SaaS plugin to listen to. 
- Add a skip_if_raw decorator to skip the signals when loading test fixtures.
- Update project uv dependencies.

## How to test

There shouldn't be any changes to any api or functionality but make sure some of the affected features are not accidentally broken:
- Instance creation (setuper or mobile uploads).
- Bulk user creation.


beanstalk_worker/services.py
```diff
@@ -109,7 +109,8 @@ def run_all(self):
 
         if connection.in_atomic_block:
             while connection.run_on_commit:
-                sids, func = connection.run_on_commit.pop(0)
+                item = connection.run_on_commit.pop(0)
+                func = item[1]
```
tdethier: why did you change this?

iaso/tests/api/profiles/test_views/common.py
```diff
@@ -86,8 +87,9 @@ def setUp(self):
             parent=self.org_unit_from_parent_type,
         )
 
+        content_type = ContentType.objects.first()
         self.permission = Permission.objects.create(
-            name="iaso permission", content_type_id=1, codename="iaso_permission"
+            name="iaso permission", content_type=content_type, codename="iaso_permission"
```
tdethier: 👍

iaso/tests/api/profiles/test_views/test_create.py
```diff
@@ -169,9 +169,10 @@ def test_create_profile_with_send_email(self):
 
         with self.captureOnCommitCallbacks(execute=True) as callbacks:
             response = self.client.post(reverse("profiles-list"), data=data, format="json")
+            self.assertEqual(len(mail.outbox), 0)
 
         response_data = self.assertJSONResponse(response, status.HTTP_201_CREATED)
-        self.assertEqual(len(callbacks), 1)
+        self.assertGreaterEqual(len(callbacks), 1)
```
tdethier: did you have to use `assertGreaterEqual()` because signals interfere with tests and multiple emails were sent?

iaso/api/instances/views.py
```diff
@@ -1011,7 +1011,13 @@ def import_data(instances, user, app_id, api_import):
         # it is possible (although it won't happen often) that the instance has already been created by the
         # POST /sync/form_upload/ endpoint.
         file_name = ntpath.basename(instance_data.get("file", None))
-        instance, _ = Instance.objects.get_or_create(file_name=file_name)
+        instance, _ = Instance.objects.get_or_create(
+            file_name=file_name,
+            defaults={
+                "project": project,
+                "uuid": uuid,
+            },
```
tdethier: why adding defaults here if they're added anyway just below?

iaso/api/profiles/views.py
```diff
@@ -273,6 +273,12 @@ def perform_create(self, serializer):
         if send_invite and email:
             transaction.on_commit(lambda: self.send_email_invitation(profile, language))
 
+        # log for quota
```
tdethier: i guess we can remove this block?

iaso/utils/signals.py
```diff
@@ -0,0 +1,39 @@
+from functools import wraps
+
+
+def skip_if_raw(func):
+    """
+    Decorator to skip a signal when raw=True.
+
+    Django passes this parameter when loading raw database records
+    (e.g. during fixture loading), not through usual application logic.
```
tdethier: i would add the link here https://code.djangoproject.com/ticket/5422

tdethier (commented):
since this is a bit difficult to review with all PRs, i'll do a first round of comments and i'll come back later once i've gone through everything

iaso/locale/fr/LC_MESSAGES/django.po
```diff
@@ -1234,6 +1234,22 @@ msgstr "Gestion du compte"
 msgid "Pipeline management"
 msgstr "Gestion des pipelines"
 
+#: iaso/saas/constants.py
```
tdethier: i guess you can regenerate translations, these should probably be gone

beanstalk_worker/services.py
Phil-V: This was a fix for an error when running the iaso test suite with the saas plugin and quota receivers activated. 

That particular line was essentially dormant and broken since at least django 4.2 where the run_on_commit queue holds 3 items instead of 2.
https://github.com/django/django/blob/97aa3b7f08f51669e118f3af5ca91026e39664c3/django/db/backends/base/base.py#L741
Long story short, the saas tests related to this feature were the first to populate the on_commit queue inside of tasks that were covered by tests that use run_all().

iaso/api/instances/views.py
Phil-V: This is so that the project attribute is set on the initial save of the object, when the signal is fired with created=True (the subsequent saves are ignored by the receiver).

iaso/tests/api/profiles/test_views/test_create.py
Phil-V: The plugin adds an extra transaction on_commit hook on profile save which breaks the assertion. I solved it by relaxing the assertion to gte=1 (we could assert == 2 if_saas_plugin_active alternatively). But it doesn't interfere with the email sending.

beanstalk_worker/services.py
tdethier: oh that's very interesting, good catch

PR #3313, opened 2026-09-08
Ia 5400 adapt completeness stats to oug restricted forms

mestachs:

Forms are now possibly assigned via orgunit groups (not only orgunit types) so we need to adapt the completeness query to find the target orgunits not only on their type.

### Related JIRA tickets

IA-5400

## Changes

- split the huge .py in several files to match more the other endpoints
- adapt the target ou query to compute groups and filter on them depending on the forms definition

## How to test

- launch seed command / setuper 
> docker compose run --rm iaso manage seed_test_data --mode=seed --dhis2version=2.40.12
- go in the completeness stats
- check the numbers
- go in the form definition change target group, remove the type
image
 
- check the numbers
image

- remove one of the orgunit from the group
- check the numbers go down by 1

## Print screen / video

image

image

## Notes

## Doc


Bewi (approved):
LGTM

Tried locally and it looks fine with scenario I went through:
- Both org unit types and groups
- Only org unit types
- Only groups
- Nothing at all
- Changing group org units

PR #3320, opened 2026-09-10
[IA-5414] Fix derived From's FormVersion should not be returned to mobile

bmonjoie:

When the mobile application retrieves the `FormVersion`s, some of them are pointing to `Form`s that are "derived" and not returned to the mobile. This results in a Foreign Key exception in the mobile app.

### Related JIRA tickets

IA-5414

## Changes

When an `app_id` is passed, filter out the `derived` `Form`s.

## How to test

1. Request the `FormVersion` using the API:  `/api/formversions/?app_id=...`
2. Mark a form as derived
3. Check that the versions of that form are not returned when calling the same endpoint.

### DIFF
--- a/iaso/tests/api/form_versions/test_form_versions.py
+++ b/iaso/tests/api/form_versions/test_form_versions.py
@@ -493,6 +493,35 @@ class FormsVersionAPITestCase(APITestCase):
         response = self.client.get("/api/formversions/", {APP_ID: self.project.app_id})
         self.assertJSONResponse(response, status.HTTP_200_OK)
 
+    def test_formversions_list_with_app_id_filters_out_version_of_derived_forms(self):
+        """GET /formversions/ with auth for project which requires it: 200"""
+
+        form_derived = m.Form.objects.create(
+            name="Derived",
+            form_id="sample2",
+            period_type="MONTH",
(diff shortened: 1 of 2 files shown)


iaso/tests/api/form_versions/test_form_versions.py
```diff
@@ -493,6 +493,35 @@ def test_formversions_list_with_auth_for_project_requiring_auth(self):
...
+        form_derived_file_mock.name = "test.xml"
+        with open("iaso/tests/fixtures/odk_form_valid_no_settings.xlsx", "rb") as xls_file:
+            form_derived.form_versions.create(
+                file=form_derived_file_mock, xls_file=UploadedFile(xls_file), version_id="2020022401"
+            )
+
+        self.client.force_authenticate(user=self.yoda)
+        response = self.client.get("/api/formversions/", {APP_ID: self.project.app_id})
+        self.assertJSONResponse(response, status.HTTP_200_OK)
+        response_data = response.json()
```
tdethier:
```suggestion
        response_data = self.assertJSONResponse(response, status.HTTP_200_OK)
```

tdethier (approved):
LGTM 👍

PR #3321, opened 2026-09-10
IA-5422 hotfix: mobile reports endpoint ignores app_id

madewulf:

`GET /api/mobile/reports/` accepts an `app_id` parameter but never used it, returning every report on the account instead of only the ones for the requesting project.

### Related JIRA tickets

IA-5422

## Changes

- `MobileReportsViewSet.get_queryset()` now resolves the project from `app_id` (`Project.objects.get_for_user_and_app_id`, same pattern as `StockKeepingUnitMobileViewSet`/`StockRulesVersionMobileViewSet`) and filters `Report` by that project instead of by `project__account`.
- `app_id` stays optional: when omitted, the endpoint falls back to the previous account-wide behaviour instead of rejecting the request, so existing callers that don't pass it aren't broken.

## How to test

`GET /api/mobile/reports/?app_id=<app_id>` should only return reports whose `project.app_id` matches, even when other projects on the same account have their own reports configured. `GET /api/mobile/reports/` without `app_id` should keep returning every report on the account, as before.

## Notes

Hotfix branched from `main` (latest released version, `2026.9.8a`) — contains only this change.

## Doc

N/A

### DIFF
--- a/iaso/api/mobile/reports.py
+++ b/iaso/api/mobile/reports.py
@@ -39,10 +39,13 @@ class MobileReportsViewSet(ModelViewSet):
         return MobileReportSerializer
 
     def get_queryset(self):
-        app_id = AppIdSerializer(data=self.request.query_params).get_app_id(raise_exception=True)
-        project = Project.objects.get_for_user_and_app_id(self.request.user, app_id)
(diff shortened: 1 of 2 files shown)


bmonjoie (approved):
LGTM

PR #3323, opened 2026-09-11
WC2-953: Bangladesh ETL

butofleury:

Intregrating Bangladesh in ETL

### Related JIRA tickets

[WC2-953](https://bluesquare.atlassian.net/browse/WC2-953)

## Changes

Added new country with Children Under 5 and PBWG and allow to run ETL on its data!

## How to test

From coda instance with database having data for Bangladesh, run the command: `uv run manage.py etl_bangladesh`

Then login on django admin and check in [analytics tables](http://localhost:8081/admin/wfp/) and filter on Bangladesh account  for tables:

- [Beneficiary](http://localhost:8081/admin/wfp/beneficiary/?account__id__exact=7)
- [Journey](http://localhost:8081/admin/wfp/journey/?beneficiary__account__id__exact=7)
- [Visits](http://localhost:8081/admin/wfp/visit/?journey__beneficiary__account__id__exact=7)
- [Steps](http://localhost:8081/admin/wfp/step/?visit__journey__beneficiary__account__id__exact=7)
- [Monthly statistics](http://localhost:8081/admin/wfp/monthlystatistics/?account__id__exact=7)

## Print screen / video

## Notes

## Doc

[WC2-953]: https://bluesquare.atlassian.net/browse/WC2-953?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


plugins/wfp/management/commands/bangladesh/Under5.py
```diff
@@ -0,0 +1,92 @@
+import logging
+
+from itertools import groupby
+from operator import itemgetter
+
+from django.core.paginator import Paginator
+
+from iaso.models import *
```
beygorghor: Nit: `Pbwg.py` already uses explicit imports. Could you do the same here?
```suggestion
from iaso.models import Task
from plugins.wfp.common import ETL
from plugins.wfp.models import Beneficiary
```

plugins/wfp/common.py
```diff
@@ -96,11 +100,19 @@
...
         "PBWG_BSFP",
+        "bsfp_child_visit",
+        "bsfp_pbwg_visit",
         "wfp_coda_pbwg_assistance",
         "wfp_coda_pbwg_assistance_followup",
         "ng_pbwg_assistanceassistance_admission_otp",
         "ng_pbwg_assistance",
         "ethiopia_child_assistance_follow_up",
+        # Bangladesh "medical" forms only carry medicine/vaccine items (see extract_assistance),
+        # so they're classified as assistance forms rather than a separate category.
```
beygorghor, line 111: Could you confirm two things about this classification?

1. Journey/visit grouping only keeps groups that contain an anthropometric form. A beneficiary with only these medical submissions will not appear in analytics. Intended?
2. `extract_assistance` already runs on every submission in a visit, so listing them here mainly affects **defaulter** detection (`_check_defaulter` uses the last `ASSISTANCE_FORMS` form’s next-visit date). If these forms have a next-visit date, they will start marking defaulters.
Also worth double-checking that `bsfp_child_visit` / `bsfp_pbwg_visit` are unique across accounts — they land in the global `ADMISSION_FORMS` / `FOLLOWUP_FORMS` / `ASSISTANCE_FORMS` sets used by SSD/NG/Ethiopia too.

beygorghor (changes requested):
`etl_bangladesh` is consistent with Nigeria/Ethiopia, and the runners correctly collect journeys/visits/steps even when `beneficiary` is `None`.

A few things to confirm before running this on coda:

1. **Form field mapping.** The new runners reuse the shared extractors (`actual_birthday__date__` or `age_entry`/`age__int__`/`registration_date`, `gender`/`_gender`, `programme`/`program`, `weight_kgs`, physiology fields). If Bangladesh forms use different keys, entities get skipped (`_is_valid_beneficiary` / `nutrition_programme is None`) and the task still reports success. A small fixture test (one U5 + one PBWG) would make this much safer.

2. **Medical forms are assistance-only**, so medical-only submissions never create a journey/visit (grouping requires an anthropometric form). Intended?

3. **`ASSISTANCE_FORMS` drives defaulter detection**, not `extract_assistance` (that already runs on every submission in a visit). If medical forms carry a next-visit date, they will start marking defaulters.

4. Entity types `bangladesh_under5` and `bangladesh_pbwg` need to exist on the target account — `get_account()` will crash if they don't.

Nit: `Under5.py` uses star imports while `Pbwg.py` already imports `Task` / `Beneficiary` explicitly.

plugins/wfp/common.py
butofleury: @beygorghor 
1. A beneficiary without enrollment(admission) anthropometric form is always skipped in the analytics tables!
2. Yes, `_check_defaulter` looks at the **last** assistance form.  If its scheduled next  visit date (+ next_visit_days) has passed and there is no subsequent anthropometric visit, the entity is considered a defaulter.

Also `bsfp_child_visit` / `bsfp_pbwg_visit` forms are now used only for Bangladesh.

plugins/wfp/common.py
```diff
@@ -239,8 +247,13 @@ def extract_exit_type(data):
     """
     exit_type = None
 
+    discharge_program = data.get("discharge_program") or ""
+    if any(programme in discharge_program for programme in ("TSFP", "OTP", "BSFP")):
```
beygorghor: This runs for every country, not only Bangladesh, and it wins over the transfer / cured / defaulter branches below.
1. `"OTP" in discharge_program` also matches `"OTP - Under 6"`.
2. `exit_type = new_programme` can persist `"TSFP"` / `"OTP"` / `"NONE"`, which are not in `EXIT_TYPES`. `_build_transfer_info` only handles `transfer_to_tsfp` / `transfer_to_otp`, so post-transfer journeys and weight-gain will be wrong.

plugins/wfp/common.py
```diff
@@ -101,8 +101,16 @@
         "ng_pbwg_assistanceassistance_admission_otp",
         "ng_pbwg_assistance",
         "ethiopia_child_assistance_follow_up",
+        "Child Medical Admission_2_u6",
```
beygorghor, line 111: BSFP visit forms are now Bangladesh-scoped via `BANGLADESH_BSFP_FORMS`, but these medical form IDs are still in the global `ASSISTANCE_FORMS`.
Worth doing the same (`BANGLADESH_MEDICAL_FORMS` merged in `assistance_forms`), unless you know these IDs cannot exist on other accounts.

plugins/wfp/locale/fr/LC_MESSAGES/django.po
```diff
@@ -130,6 +134,22 @@ msgstr "Par enfant émacié"
...
+msgid "New admission (MUAC <11.5 cm)"
+msgstr "Nouvelle admission (MUAC <11.5 cm)"
+
+#: plugins/wfp/models.py
+msgid "New admission (WHZ scores <-3SD)"
+msgstr "Nouvelle admission (WHZ scores <-3SD)"
+
+#: plugins/wfp/models.py
+msgid "New admission (Both WHZ and MUAC)"
+msgstr "Nouvelle admission (MUBoth WHZ and MUAC)"
```
beygorghor:
```suggestion
msgstr "Nouvelle admission (MUBoth WHZ et MUAC)"
```

beygorghor (changes requested):
Thanks for the changes !

The new `discharge_program` branch runs for **every** country and short-circuits the existing transfer / cured / defaulter logic. 
Also:
- `"OTP" in discharge_program` is a substring check, so `"OTP - Under 6"` matches.
- If `new_programme` is `"TSFP"` / `"OTP"` / `"NONE"`, that value is stored as `Journey.exit_type`, but those are not in `EXIT_TYPES` (`transfer_to_tsfp`, `cured`, etc.). Transfer chaining (`_build_transfer_info`) and weight-gain then won’t fire.
Please scope this to Bangladesh, or map programme names onto existing exit types, and add a couple of unit tests (Bangladesh + an SSD/NG fixture that already has `discharge_program`).

plugins/wfp/common.py
butofleury: @beygorghor , 
1. for Bangladesh there is no "OTP - Under 6" program for U5. This is only for SSD. Even now OTP has not been yet introduced for Bangladesh. They told us to prepare it in advance, to make it easier when it will be integrated. But according to them, it will be similar to SSD.

2. for this, based on the xls form for [Anthropometric TSFP](http://localhost:8081/dashboard/forms/detail/formId/82) , noticed a new field `referred_to_BSFP` which is a transfer to BSFP! But some points are still unclear for Bangladesh. For instance, [this beneficiary](https://qa.coda.go.wfp.org/dashboard/entities/details/accountId/7/entityId/3400) is 
[discharged from TSFP](https://qa.coda.go.wfp.org/dashboard/forms/submission/accountId/7/instanceId/26136), but nothing show if he/she is transfered to another program or if he/she is cured!
Based on the screenshot when beneficiary is discharged, it shows transfered to OTP, but there is any field in the json that showing it.

image

plugins/wfp/common.py
beygorghor, line 1119: `_process_entity` now treats `transfer_to_bsfp` as a journey-ending transfer, but this helper falls through to `return None` for that exit type.

So the next journey never gets transfer context (`nutrition_programme="BSFP"` + a referred-from admission type). If the following BSFP visit has no `programme` field, the journey is skipped (`nutrition_programme is None`).

Could you add a `transfer_to_bsfp` branch here (or stop passing that exit type into `_build_transfer_info`)?

beygorghor (changes requested):

beygorghor (approved):

PR #3329, opened 2026-09-16
POLIO-2209: Add fIPV campaigns to vaccine management

beygorghor:

As the new campaign category fIPV was added - it’s not appearing on Vaccine Management (Stock, Supply) - which makes sense because only Polio campaigns are configured to appear, which fIPV is a Polio type of vaccine. 

### Related JIRA tickets

POLIO-2209

## Changes
get fIPV campaign too for VRF

## How to test

Make sure you have a fIPV campaign type and create a campaign of this type.
Create a VRF for the country of the campaign, you should be able to select the created camaping and save it

## Print screen / video

Screenshot 2026-09-16 at 16 29 30
Screenshot 2026-09-16 at 16 29 20


Bewi (approved):
LGtM

PR #3331, opened 2026-09-17
IA-5429: Fix missing user group permissions when bulk updating users

Phil-V:

When bulk updating users to assign roles, the user permission groups were not properly updated, leaving the users actual permissions out of sync with their roles.

### Related JIRA tickets

refs: IA-5429

## Changes

- Fix the bulk update code to ensure that the user's built-in django groups stay in sync with our custom `UserRole` model. 
- Add a management command to fix the roles/groups of existing users affected by this issue.

## How to test

You will need a running worker for the bulk update task. 

- Go to the users management page and select one or more users.
- Use the speed dial action to edit them in bulk and add or remove user roles.
- Visit the user details page and make sure the permissions listed on the details page reflect the chosen role. 

## Print screen / video

See Jira ticket

## Notes

Requires running the command after deployment to production. 

## Doc

/


iaso/management/commands/fix_user_role_mismatch.py
```diff
@@ -0,0 +1,112 @@
...
+
+from django.contrib.auth.models import Group
+from django.core.management.base import BaseCommand
+from django.db import transaction
+from django.db.models import Exists, OuterRef
+
+from iaso.models import Profile, UserRole
+
+
+class Command(BaseCommand):
```
tdethier, line 18: please implement a verbosity level so that we can avoid outputs if we want (like in tests) - see [example here](https://github.com/BLSQ/iaso/blob/5830e91f11b17d6a310d88d18ca2c80a7ce29a66/iaso/management/commands/delete_accounts.py#L391), value is [set here](https://github.com/BLSQ/iaso/blob/5830e91f11b17d6a310d88d18ca2c80a7ce29a66/iaso/management/commands/delete_accounts.py#L1277)

iaso/tests/commands/test_fix_user_role_mismatch.py
```diff
@@ -0,0 +1,88 @@
...
+            username="user_2",
+            account=cls.account2,
+            user_roles=[cls.user_role3],
+        )
+
+    def test_fix_adds_missing_groups(self):
+        self.assertEqual(self.user1.groups.count(), 0)
+        self.assertEqual(self.user2.groups.count(), 0)
+
+        management.call_command("fix_user_role_mismatch")
```
tdethier: once verbosity is implemented in the command, you should pass `verbosity=0` in all tests

iaso/management/commands/fix_user_role_mismatch.py
```diff
@@ -0,0 +1,112 @@
...
+                    self.stdout.write(f" - Missing groups: [{group_names}]")
+                    total_groups_added += len(missing_groups)
+
+                if extra_groups and clean_up_groups:
+                    group_names = ", ".join(g.name for g in extra_groups)
+                    self.stdout.write(f" - Extra groups (to be removed): [{group_names}]")
+                    total_groups_removed += len(extra_groups)
+
+                if not dry_run:
+                    with transaction.atomic():
```
tdethier, line 102: i thought we would simply add a single transaction on the whole command like we do in migrations, but since this is triggered manually and should be idempotent, this works 👍

tdethier (approved):
the bulk update now properly grants permissions, nice job 👍 

i'm already approving, but please implement a `verbosity` level on the command (see below) before merging

iaso/management/commands/fix_user_role_mismatch.py
Phil-V: Thanks for the review and examples!

PR #3332, opened 2026-09-18
WC2-967: Impossible to change storage status

beygorghor:

While trying to blacklist or whitelist a storage we got a 400 error

### Related JIRA tickets

WC2-967

## Changes

Fix storage_id parsing

## How to test

Try to change a storage status using the dashboard

## Print screen / video

https://github.com/user-attachments/assets/6099fa04-cd8e-4bd7-b03b-fdffb7069f82


bmonjoie (approved):
LGTM

tdethier (approved):
i find it weird that we had this issue in the first place: i would expect the hex format to simply be computed and displayed in the front, but never sent from or to the backend (only the base64 one).

but the fix works, i managed to change the status of storages 👍

PR #3334, opened 2026-09-18
POLIO-2209: Add IPV vaccine type on VRF

beygorghor:

When creating/editing a VRF, IPV vaccine should be available if the campaign is of type fIPV

### Related JIRA tickets

POLIO-2209

## Changes

- Adding value in Enum + migration
- Python tests
- Adapt front to allow vaccine type 

## How to test

Create a VRF for a campaign of type fIPV, make sure IPV is available in the list of vaccine options.
IPV should not be available if campaign is of polio type

## Print screen / video

Screenshot 2026-09-18 at 15 10 39


plugins/polio/js/src/domains/VaccineModule/SupplyChain/constants.ts
```diff
@@ -37,3 +43,10 @@ export const singleVaccinesList = [
         value: 'bOPV',
     },
 ];
+
+export const vrfVaccineFilterOptions = [
```
Bewi, line 47: I don't know what ipvVaccineOption but I guess it cannot be in the singleVaccinesList which is used in other places ?

plugins/polio/js/src/domains/VaccineModule/SupplyChain/hooks/api/vrf.tsx
```diff
@@ -177,19 +181,34 @@ export const useCampaignDropDowns = ({
...
-                  label: vaccineName.trim(),
-                  value: vaccineName.trim(),
-              }))
-            : singleVaccinesList;
+        const fipvCampaigns = (fIPVData as Campaign[]) ?? [];
+        const isFipvCampaign = Boolean(
+            selectedCampaign &&
+            fipvCampaigns.some(c => c.obr_name === selectedCampaign.obr_name),
+        );
+        const vaccinesFromScopes = selectedCampaign?.single_vaccines
```
Bewi: No need of ternary here:

selectedCampaign?.single_vaccines?.split(',')... ?? []

plugins/polio/js/src/domains/VaccineModule/SupplyChain/hooks/api/vrf.tsx
```diff
@@ -177,19 +181,34 @@ export const useCampaignDropDowns = ({
...
+            ? selectedCampaign.single_vaccines
+                  .split(',')
+                  .map(vaccineName => vaccineName.trim())
+                  .filter(Boolean)
+                  .map(vaccineName => ({
+                      label: vaccineName,
+                      value: vaccineName,
+                  }))
+            : [];
+        let vaccines = vaccinesFromScopes;
```
Bewi: why having another variable here? 
You could use `vaccines` directly when filtering `single_vaccines`.
In your `else` a bit later you can use `vaccines` too.

Bewi: You could also delay the `vaccinesFromScopes` filtering and such only if ìsFipvCampaign = false`

Bewi (approved):
WaE & LGTM 

Only minor technical remarks but non blocking

plugins/polio/js/src/domains/VaccineModule/SupplyChain/constants.ts
beygorghor: no indeed this is the problem, this vaccine should only be visible in VRF

